PR #47, opened 2025-11-30
feat: Dockerize server

fulviodenza:
# Description

## Type of change

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Documentation update
- [ ] Performance improvement
- [ ] Code refactoring
- [ ] Test coverage improvement

## Changes Made

Please provide a clear list of changes:

- Add Dockerfile for building image
- docker-compose.yml for local testing

## Testing

- [x] Unit tests pass (`go test ./...`)
- [x] Manual testing performed
- [ ] Integration testing performed (N/A)

## Testing Instructions

1. once you have the file `~/.config/dittofs/config.yaml` saved, run `docker-compose up` (ensure ports in the config and in the dockerfile/docker-compose files match)
2. mount the folder

## Checklist

- [x] I have made corresponding changes to the documentation

## Screenshots (if applicable)

image

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -16,29 +16,24 @@ RUN CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build \
     -o dittofs \
     cmd/dittofs/main.go
 
-FROM alpine:latest
-
-RUN apk add --no-cache \
-    ca-certificates \
-    tzdata
-
-RUN addgroup -g 1000 dittofs && \
-    adduser -D -u 1000 -G dittofs dittofs
-
-RUN mkdir -p /data/metadata /data/content /config && \
-    chown -R dittofs:dittofs /data /config
+FROM gcr.io/distroless/static-debian12:nonroot
 
 WORKDIR /app
 
-COPY --from=builder /build/dittofs .
+# Copy the binary from builder
+COPY --from=builder --chown=65532:65532 /build/dittofs /app/dittofs
 
-RUN chown dittofs:dittofs /app/dittofs
+USER 65532:65532
 
-USER dittofs
+EXPOSE 2049/tcp 9090/tcp
 
-EXPOSE 12049/tcp 12049/udp 9090/tcp
+# - /data/metadata: Metadata store (BadgerDB, etc.)
+# - /data/content: Content store for local filesystem backend (unused for S3)
+# - /config: Configuration file location
+VOLUME ["/data/metadata", "/data/content", "/config"]
 
-VOLUME ["/data", "/config"]
+HEALTHCHECK --interval=30s --timeout=10s --start-period=10s --retries=3 \
+    CMD ["/app/dittofs", "help"]
 
 ENTRYPOINT ["/app/dittofs"]
 CMD ["start", "--config", "/config/config.yaml"]
--- a/README.md
+++ b/README.md
@@ -71,15 +71,31 @@ go build -o dittofs cmd/dittofs/main.go
 
 ### Run with Docker
 
-To run DittoFS is with Docker, ensure first to have the config file `~/.config/dittofs/config.yaml` to let docker compose mount it in the container:
+To run DittoFS with Docker, ensure first to have the config file `~/.config/dittofs/config.yaml` to let docker compose mount it in the container:
 
 ```bash
-# Start the server
-docker-compose up -d
+# Start with local filesystem backend (default)
+docker compose up -d
+
+# Start with S3 backend (includes localstack)
+docker compose --profile s3-backend up -d
 
 # View logs
-docker-compose logs -f dittofs
+docker compose logs -f dittofs
 ```
+
+**Storage Backends:**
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -0,0 +1,85 @@
+services:
+  dittofs:
+    build:
+      context: .
+      dockerfile: Dockerfile
+    container_name: dittofs-server
+    image: dittofs:latest
+    ports:
+      - "12049:12049/tcp"
+      - "2049:2049/udp"
```
marmos91: This port should be aligned with the TCP one. At the moment, we don't support UDP yet though, so we can safely remove this mapping

docker-compose.yml
```diff
@@ -0,0 +1,85 @@
...
+      - dittofs-content:/data/content
+    restart: unless-stopped
+    cap_add:
+      - SYS_ADMIN
+    healthcheck:
+      test: ["CMD", "/app/dittofs", "help"]
+      interval: 30s
+      timeout: 10s
+      retries: 3
+      start_period: 10s
```
marmos91: This health check is fine for now, but it is definitely not secure for a production environment. We need to add a small health route to the Ditto server (we can also setup an HTTP server, because I am thinking to set it up anyway in the future to provide a management dashboard). Only then can we know the status of the running server. Help will never fail...

Dockerfile
```diff
@@ -0,0 +1,44 @@
...
+
+WORKDIR /app
+
+COPY --from=builder /build/dittofs .
+
+RUN chown dittofs:dittofs /app/dittofs
+
+USER dittofs
+
+EXPOSE 12049/tcp 12049/udp 9090/tcp
```
marmos91: We can remove udp for now. We are not using it yet

README.md
```diff
@@ -69,14 +69,26 @@ go build -o dittofs cmd/dittofs/main.go
 ./dittofs start
 ```
 
+### Run with Docker
+
+To run DittoFS is with Docker, ensure first to have the config file `~/.config/dittofs/config.yaml` to let docker compose mount it in the container:
+
```
marmos91: We should also explain how to use the different profiles.

README.md
```diff
@@ -69,14 +69,26 @@ go build -o dittofs cmd/dittofs/main.go
...
 ### Mount from Client
 
 ```bash
 # Linux
 sudo mount -t nfs -o nfsvers=3,tcp,port=12049,mountport=12049 localhost:/export /mnt/nfs
 
 # macOS
-sudo mount -t nfs -o nfsvers=3,tcp,port=12049,mountport=12049 localhost:/export /mnt/nfs
+sudo mount -t nfs -o nfsvers=3,tcp,port=12049,mountport=12049 localhost:/export /tmp/nfs
```
marmos91: The point here is not about the `/mnt` folder vs `/tmp` folder. The point is that the folder where you want to mount the volume **must exist**, so it has to be created before. Putting `/tmp` here is ok, but we should also specify that the folder must be created before running the command.

marmos91 (changes requested):
Thanks for the PR @fulviodenza. The changes make sense to me, more or less. Just a couple of comments.

marmos91 (changes requested):
Thanks for the PR @fulviodenza. The changes make sense to me, more or less. Just a couple of comments.

docker-compose.yml
fulviodenza: we can delete it for now, it's not strictly necessary at this stage

docker-compose.yml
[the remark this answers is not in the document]
fulviodenza: I agree with your observation but I think it would be also nice to just split into services in the docker-compose file without having 3 different docker-compose files, 

```
services:
  dittofs:
    build:
      context: .
      dockerfile: Dockerfile
    container_name: dittofs-server
    image: dittofs:latest
    ports:
      - "12049:12049/tcp"
      - "9090:9090/tcp"
    volumes:
      - ~/.config/dittofs/config.yaml:/config/config.yaml:ro
      - dittofs-metadata:/data/metadata
      - dittofs-content:/data/content
    restart: unless-stopped
  # and then this
  dittofs-s3:
    build:
      context: .
      dockerfile: Dockerfile
    container_name: dittofs-server
    image: dittofs:latest
    ports:
      - "12049:12049/tcp"
      - "9090:9090/tcp"
    volumes:
      - ~/.config/dittofs/config.yaml:/config/config.yaml:ro
      - dittofs-metadata:/data/metadata
    depends_on:
      - localstack
    restart: unless-stopped
    profiles:
      - s3-backend
```

What do you think, you like it or do you still prefer the 3-docker-composes way?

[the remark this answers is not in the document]
marmos91: I agree with keeping it in the same file with different profiles

Issue #358, opened 2026-04-14
Flaky: TestSyncer_ConcurrentOperations_Memory intermittent CI failure


marmos91:
## Description

`TestSyncer_ConcurrentOperations_Memory` intermittently fails in CI (integration test job). The failure appears to be a race condition in the syncer's concurrent operations test with the memory-backed Localstack container.

## Observed

- Failed in PR #356 CI run: https://github.com/marmos91/dittofs/actions/runs/24394063619
- All other tests in the suite pass
- Develop branch CI passes consistently
- Rerun of the same job succeeds

## Likely Cause

Known pattern: multiple Localstack containers cause exit code 245 (noted in project history). The concurrent operations test may hit a timing window where container teardown races with in-flight S3 operations.

## Impact

Low — test is correct but flaky. Causes spurious CI failures that require manual rerun.

marmos91:
Fixed by PR #364 (commit 98cb372a). Verified with -race -count=100 on TestSyncer_ConcurrentOperations_Memory.

PR #366, opened 2026-04-15
fix(syncer): remove dead upload-tracking infrastructure (#365)

### LINKED ISSUE #365: Syncer: DrainAllUploads / WaitForAllUploads are silent no-ops
## Problem

The `Syncer` exposes upload-tracking infrastructure that has never been wired up. The `m.uploads map[string]*fileSyncState` is **only ever read and deleted from** — nothing in the codebase inserts into it. As a result:

- `DrainAllUploads(ctx)` — exposed as a REST endpoint (\`POST /api/v1/system/drain-uploads\`, see \`internal/controlplane/api/handlers/system.go:44\`) and called from \`Syncer.Close()\` — iterates an always-empty map and returns immediately without waiting for anything.
- \`WaitForAllUploads(ctx, payloadID)\` and \`WaitForEagerUploads(ctx, payloadID)\` both delegate to \`getSyncState(payloadID)\` which always returns \`nil\`, so they return \`nil\` without actually waiting.
- \`fileSyncState.inFlight.Add/Done\` and \`fileSyncState.flush.Add/Done\` are never called; the WaitGroups are initialized at zero and never incremented.

## Impact

- **Benchmarks**: the \`/drain-uploads\` API is intended to give the benchmark runner clean boundaries between workload phases. Since it's a no-op, benchmark results may be contaminated by still-in-flight background uploads from the previous phase.
- **Tests**: \`pkg/blockstore/engine/engine_offline_test.go\` (lines 99, 191) calls \`WaitForAllUploads\` and expects it to actually wait. It passes today only because the upload paths exercised happen to be synchronous (periodic \`syncFileBlock\` holds the \`m.uploading\` gate end-to-end).
- **Shutdown**: \`Close()\` calls \`DrainAllUploads\` thinking it's waiting for in-flight uploads. The queue's own \`q.Stop()\` drain is the only thing actually draining.

## Evidence

Discovered during code review of #358 (PR #364). Three independent reviewer passes (at 95, 80, and ~70 confidence) flagged the same issue. \`m.uploads\`-manipulating sites:

\`\`\`
pkg/blockstore/sync/syncer.go:78       declaration
pkg/blockstore/sync/syncer.go:120      init
pkg/blockstore/sync/syncer.go:249-250  DrainAllUploads — reads
pkg/blockstore/sync/syncer.go:407-409  delete
pkg/blockstore/sync/dedup.go:11-13     getSyncState — reads
pkg/blockstore/sync/dedup.go:23-25     DeleteWithRefCount — deletes
\`\`\`

No \`m.uploads[...] = ...\` anywhere. Same for \`state.inFlight.Add\` / \`state.flush.Add\`.

## Proposed fix

Two options:

1. **Delete the dead code.** Remove \`fileSyncState\`, \`m.uploads\`, \`WaitForEagerUploads\`, \`WaitForAllUploads\`. Rewire \`DrainAllUploads\` to call \`SyncNow\` (which is now fully synchronous and gated after #358), so the drain actually drains. Update \`engine_offline_test.go\` accordingly.
2. **Wire it up.** Populate \`m.uploads\` in every upload-initiating path and correctly bump the WaitGroups. More work, and since #358 made \`SyncNow\` synchronous we don't seem to need per-file tracking — (1) is cleaner.

## Related

- #358 / PR #364 — the code review that surfaced this.
- Two other pre-existing observations from the same review, tracked separately or rolled in:
  - \`periodicUploader\` at \`syncer.go:532-545\` doesn't use \`defer\` to release \`m.uploading\`, unlike \`SyncNow\`. Inconsistent. Minor.
  - \`SyncNow\` spin is unbounded; acceptable for test-only use but worth bounding if ever used in production.

marmos91:
## Summary

Closes #365. Depends on #358 (PR #364).

The `m.uploads` map, `fileSyncState` struct, and `WaitForAllUploads` / `WaitForEagerUploads` helpers were never populated — nothing in the codebase inserted into `m.uploads` or incremented the `inFlight` / `flush` WaitGroups. This was flagged by 3 independent reviewers during #358's review.

As a result:
- `DrainAllUploads` (exposed as `POST /api/v1/system/drain-uploads` and called from `Close()`) was a silent no-op
- Benchmarks calling `/drain-uploads` between phases were silently contaminated by in-flight background uploads
- `WaitForAllUploads` / `WaitForEagerUploads` returned `nil` immediately without waiting for anything

Since PR #364 made `SyncNow` fully synchronous under the `m.uploading` gate, `DrainAllUploads` now delegates to it — the drain is real.

## Changes

- **Remove dead code**: `fileSyncState` struct, `m.uploads` map and `m.uploadsMu`, `getSyncState`, `WaitForEagerUploads`, `WaitForAllUploads`, unused `waitWithContext` helper
- **Rewire `DrainAllUploads`** to delegate to `SyncNow(ctx)` + return `ctx.Err()`
- **Nil-remote guard** on `SyncNow` (new `Close` path exercises this)
- **Bonus**: `periodicUploader` now releases `m.uploading` via `defer` inside an inline func, matching `SyncNow` (third pre-existing observation from #358's review — asymmetric release)
- Update `engine_offline_test.go` to drop the now-dead `WaitForAllUploads` calls; `SyncNow` alone is sufficient

Net diff: -104 / +25 lines across 3 files.

## Test plan

- [x] `go test -tags=integration -race -count=100 -run TestSyncer_ConcurrentOperations_Memory ./pkg/blockstore/sync/` — 100/100
- [x] `go test -tags=integration -race -count=3 ./pkg/blockstore/sync/` — 3/3
- [x] `go test -race ./pkg/blockstore/... ./pkg/controlplane/...` — all green (caught a nil-remote regression in `TestNilRemoteStore*` which triggered the SyncNow nil-guard)
- [ ] Benchmark runner: validate that `/drain-uploads` now actually waits for pending uploads (out of scope for this PR; worth doing in the benchmark PR)

## Out of scope

The third follow-up from #358's review — `SyncNow` unbounded spin under slow remotes — was not addressed here. It's fine for test-only use (which is the documented intent) and has a `ctx.Done` escape. Track separately if production calls ever emerge.

### DIFF
--- a/pkg/blockstore/engine/engine_offline_test.go
+++ b/pkg/blockstore/engine/engine_offline_test.go
@@ -93,12 +93,9 @@ func TestOfflineReadRemoteOnlyBlockFails(t *testing.T) {
 	if _, err := bs.Flush(ctx, payloadID); err != nil {
 		t.Fatalf("Flush failed: %v", err)
 	}
+	// SyncNow holds the uploading gate end-to-end and uploads synchronously,
+	// so by the time it returns every block is in the remote store.
 	bs.syncer.SyncNow(ctx)
-
-	// Wait for upload to complete.
-	if err := bs.syncer.WaitForAllUploads(ctx, payloadID); err != nil {
-		t.Fatalf("WaitForAllUploads failed: %v", err)
-	}
 
 	// Verify block is in remote.
 	memStore := fakeRemote.RemoteStore.(*remotememory.Store)
@@ -188,9 +185,6 @@ func TestOfflineReadsBlockedCounter(t *testing.T) {
 		t.Fatalf("Flush failed: %v", err)
 	}
 	bs.syncer.SyncNow(ctx)
-	if err := bs.syncer.WaitForAllUploads(ctx, payloadID); err != nil {
-		t.Fatalf("WaitForAllUploads failed: %v", err)
-	}
 	if err := bs.EvictLocal(ctx, payloadID); err != nil {
 		t.Fatalf("EvictLocal failed: %v", err)
 	}
--- a/pkg/blockstore/sync/dedup.go
+++ b/pkg/blockstore/sync/dedup.go
@@ -6,23 +6,11 @@ import (
 	"github.com/marmos91/dittofs/internal/logger"
 )
 
-// getSyncState returns the upload state for a file, or nil if not found.
-func (m *Syncer) getSyncState(payloadID string) *fileSyncState {
-	m.uploadsMu.Lock()
-	state := m.uploads[payloadID]
-	m.uploadsMu.Unlock()
-	return state
-}
-
 // DeleteWithRefCount decrements RefCount for each block and deletes blocks that reach zero.
 func (m *Syncer) DeleteWithRefCount(ctx context.Context, payloadID string, blockIDs []string) error {
 	if !m.canProcess(ctx) {
 		return ErrClosed
 	}
-
-	m.uploadsMu.Lock()
-	delete(m.uploads, payloadID)
-	m.uploadsMu.Unlock()
 
 	if m.fileBlockStore == nil {
 		if m.remoteStore != nil {
--- a/pkg/blockstore/sync/syncer.go
+++ b/pkg/blockstore/sync/syncer.go
@@ -440,21 +440,46 @@ func (m *Syncer) SyncNow(ctx context.Context) error {
 
 	// Flush queued FileBlock metadata to the store so ListLocalBlocks can find them.
 	m.local.SyncFileBlocks(ctx)
-	pending, err := m.fileBlockStore.ListLocalBlocks(ctx, 0, 0)
-	if err != nil {
-		return fmt.Errorf("list local blocks: %w", err)
-	}
+
+	// Drain in batches to keep peak memory bounded on large stores — one
+	// ListLocalBlocks call with limit=0 would deserialize every pending
+	// FileBlock at once (potentially thousands). syncFileBlock advances
+	// blocks out of BlockStateLocal on success, so successive ListLocalBlocks
+	// queries return distinct pages; on per-block failure revertToLocal
+	// keeps the block in Local state — we break after one no-progress pass
+	// so a permanently-failing block cannot spin the drain forever.
 	var uploadErrs []error
-	for _, fb := range pending {
+	var prevFailed int
+	for {
 		if err := ctx.Err(); err != nil {
 			return err
 		}
-		if fb.LocalPath == "" {
-			continue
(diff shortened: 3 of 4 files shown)


marmos91:
Addressed review findings (15 reviewer passes + 10 simplifier passes). Strong consensus from reviewers on two real issues, now fixed in commit b0a6ddf8:

**1. Error propagation** (10+ reviewers, confidence 85–95)
`SyncNow` used to swallow per-block errors so `DrainAllUploads` returned `nil` even when uploads failed — `/drain-uploads` responded 200 OK on partial drains and `Close()` gave false success. Now:
- `syncFileBlock` returns `error` (wrapped with block ID)
- `SyncNow` returns `errors.Join(...)` of per-block failures
- `DrainAllUploads` propagates `SyncNow`'s error; `ctx.Err()` only as fallback
- Periodic uploader intentionally ignores (`_= syncFileBlock(...)`) — a single bad block shouldn't starve the tick loop

**2. In-loop ctx cancellation** (5 reviewers, confidence 80–85)
`SyncNow` only checked `ctx.Done()` while waiting for the gate, not between block uploads — so a 30s `Close()` deadline couldn't interrupt mid-drain. Now checks `ctx.Err()` before each `syncFileBlock` call.

**Not acted on** (minority / disputed / out of scope):
- `Close()` sets `m.closed = true` before `DrainAllUploads` — 3 reviewers claimed this makes the drain a no-op, but `SyncNow` doesn't check `m.closed` and `PutFileBlock` is on the metadata store (no syncer-level closed check). Verified: drain still runs.
- `SyncNow` memory behavior with `limit=0` vs periodicUploader's `maxUploadBatch=4` — 1 reviewer; acceptable for shutdown drain, which is the purpose.
- Deprecating `DrainAllUploads` as a thin wrapper — stylistic; keeping it for the explicit REST endpoint name.

Simplifier consensus across all 10 passes: commit is already a simplification win, no further edits needed.

Verified: `-race -count=100` on the target test, full `pkg/blockstore/...` green.

marmos91:
Addressed Copilot comment #1 (unbounded ListLocalBlocks) in commit 17c29305.

SyncNow now drains in batches of `maxUploadBatch` (4) — same memory-bounded pattern as `periodicUploader.syncLocalBlocks`. The loop terminates when `ListLocalBlocks` returns empty, or — to avoid spinning on a permanently-failing block — when a batch fails entirely and the failure count matches the previous pass (progress guard).

Copilot comments #2 (in-loop ctx check) and #3 (error propagation) were already fixed in b0a6ddf8 — the stale review anchors were from before that commit landed.

Verified: -race -count=30 on the target test, full pkg/blockstore/... green.

Issue #388, opened 2026-04-17
SMB Kerberos: reauth5 hangs in CREATE/CLOSE loop after reauth (DELETE_ON_CLOSE returns AccessDenied)


marmos91:
## Context

Unmasked after fixing #370. The `smbtorture smb2.session.reauth5` test was previously unreachable (connect failed upstream) and is listed in `test/smb-conformance/smbtorture/KNOWN_FAILURES_KERBEROS.md` under #340-A2 as _"Signing keys wrong after Kerberos reauth"_. With #370's fix, the test actually runs — but instead of failing cleanly it **hangs**, burning the full 1200s smbtorture timeout and preventing the remaining smb2.session subtests (reauth6, expire*, bind*, signing-*, anon-*, encryption-*, ntlmssp_bug14932, two_logoff) from running.

## Symptoms

`test/smb-conformance/results/smbtorture-2026-04-17_093541/dittofs.log` shows an infinite loop between 07:36:04 (reauth5 starts) and 07:56:04 (killed by timeout, ~20 min of identical traffic):

```
CREATE filename=session_reauth5_XCZY3_d,.d\file.dat disposition=1 options=4160 desiredAccess=0x10000
  → STATUS_SUCCESS
CLOSE fileID=<new>
  → CLOSE: flushed path=session_reauth5_XCZY3_d,.d/file.dat
  → CLOSE: failed to delete path=session_reauth5_XCZY3_d,.d/file.dat isDir=false error=AccessDenied: write permission denied
  → STATUS_SUCCESS (sent to client)
QUERY_DIRECTORY fileID=<dir> fileName=*
  → STATUS_NO_MORE_FILES
```

Client keeps recreating the file, server CLOSE reports `STATUS_SUCCESS` to the client but internally fails to delete (`AccessDenied: write permission denied`), QUERY_DIRECTORY finds nothing. Loop.

MessageIDs climb from the thousands to ~9.5M over 20 minutes.

## Suspected root cause

Two-part interaction:

[...]

Issue #340, opened 2026-04-11
SMB Kerberos: classify smb2.session failures and fix genuine Kerberos bugs (expire/reauth/reconnect)


marmos91:
## Summary

After #337 lands, \`smb2.connect --kerberos\` passes against MIT via smbtorture. The broader \`smb2.session\` suite now runs end-to-end (0 session setups completed pre-fix) and produces:

- **6 pass** (\`two_logoff\`, \`reauth1\`, \`reauth2\`, \`reauth3\`, \`encryption-aes-128-ccm\`, \`encryption-aes-128-gcm\`)
- **46 fail** — mix of genuine Kerberos bugs and pre-existing NTLM gaps
- **19 skip** — unimplemented features by design

All 46 failures currently show as "new failures" because the smbtorture-kerberos job has no \`KNOWN_FAILURES.md\` bucket yet. This issue tracks both the cleanup and the real bugs.

## Part A: fix the ~5 genuine Kerberos-specific bugs

These are **not** present on the NTLM path (or present in a different form). They should be fixed and promoted to passing.

### A1 — Ticket expiration not enforced (6 tests)

| Test | Observed | Expected |
|---|---|---|
| \`expire1s\` / \`expire1n\` / \`expire1e\` | \`STATUS_OK\` on getinfo after ticket lifetime elapsed | \`NT_STATUS_NETWORK_SESSION_EXPIRED\` |
| \`expire2s\` / \`expire2e\` | Same | Same |
| \`expire_disconnect\` | Same | Same + disconnect |

**Root cause:** \`handleKerberosAuth\` in \`internal/adapter/smb/v2/handlers/kerberos_auth.go\` never stores \`authResult.APReq.Ticket.DecryptedEncPart.EndTime\` on the session object. The dispatcher has no ticket-lifetime gate.

**Fix:** persist \`endTime time.Time\` on \`session.Session\` (or a Kerberos-specific sub-struct), check it in the per-request auth hook, and return \`NT_STATUS_NETWORK_SESSION_EXPIRED\` when elapsed. Per [MS-SMB2] §3.3.5.2.9, also close the connection after sending the error.

### A2 — Reauth breaks signing on setinfo (2 tests)

| Test | Observed |
|---|---|
| \`reauth4\` | \`NT_STATUS_INTERNAL_ERROR\` on \`smb2_setinfo_file\` after successful reauth |
| \`reauth5\` | Same pattern |

[...]

PR #389, opened 2026-04-17
fix(smb): restore Kerberos SESSION_SETUP behaviours dropped by #345 — close #370

### LINKED ISSUE #370: SMB Kerberos smbtorture: 71 failures in reconnect/reauth/expire/bind tests
## Summary

The `smbtorture Kerberos` CI job is red on every recent SMB conformance run. Identical 71 failure lines appear in back-to-back runs (24482922640 @ 8bdb4e7f, 24495273767 @ ecad8347). This is distinct from #362 (SMB signing races, now fully fixed) and from #367 (WPTS BVT flakes).

## Scope

Consistent failures across:

- **reconnect / reauth (8 tests)**: `reconnect1`, `reconnect2`, `reauth1`–`reauth6`
- **expire (6 tests)**: `expire1n`, `expire1s`, `expire1e`, `expire2s`, `expire2e`, `expire_disconnect`
- **bind (7+ tests)**: `bind1`, `bind2`, `bind_invalid_auth`, `bind_different_user`, `bind_negative_smb202`, `bind_negative_smb210s`, …

No "Bad SMB2 signature" errors — these are RPCSEC_GSS / Kerberos session lifecycle issues, not signing-key derivation races.

## Hypotheses (not yet verified)

1. **Kerberos ticket expiry handling** — `expire*` tests likely need us to produce `STATUS_NETWORK_SESSION_EXPIRED` or re-challenge cleanly when the service ticket expires mid-session. `prepareDispatch` does have an `IsExpired()` check but expiry-driven behavior end-to-end may be incomplete.
2. **SMB2 session bind (multi-channel)** — `bind*` tests invoke SMB3 session binding (MS-SMB2 3.3.5.5.1); DittoFS currently rejects multi-channel in NEGOTIATE (capability not advertised). The tests might hit a path where the client still attempts bind and the server doesn't respond with the expected status.
3. **Session reconnect / reauth** — reauth flow reuses an existing session ID for a fresh SESSION_SETUP. Needs audit: does `handleNTLMNegotiate`'s `isReauth` branch cleanly carry crypto state? Are preauth hash chains reset correctly on reauth? The fix in ecad8347 (before-write hook) interacts here.

## Reproduction

```bash
# Requires docker + samba-toolbox image as in smb-conformance workflow
cd test/smb-conformance
PROFILE=memory docker compose up -d dittofs
# Then run the Kerberos smbtorture target
```

CI job: https://github.com/marmos91/dittofs/actions/runs/24495273767/job/71581562303

## Dependencies

- Independent of #362 (signing races — fixed at ecad8347 on `fix/issue-367-362-smb-flaky-tests`)
- Independent of #367 (WPTS BVT timestamp/DFS flakes)
- Diagnostic instrumentation (SIGN_TRACE, ENCRYPT_TRACE, session lookup MISS tracing) is still active on the #362 branch; if the first investigation session happens while that's in place the extra traces will help, otherwise they'll be reverted pre-merge.

## Acceptance criteria

- `smbtorture Kerberos` CI job green (all 71 failures resolved, or classified into KNOWN_FAILURES with justification for the unfixable subset).
- No regressions in `smbtorture / memory` or `WPTS BVT / memory` jobs.

marmos91:
## Summary

PR #345 (centralized identity resolution) refactored `handleKerberosAuth` and accidentally dropped four pieces of Kerberos SESSION_SETUP behaviour. All 71 smbtorture `smb2.session` tests were failing at initial `smb2_connect` with `NT_STATUS_LOGON_FAILURE` — no test body executed.

This restores the four dropped pieces **while keeping #345's centralized `identity.Resolver` architecture intact** — no revert, no reintroduction of the old ad-hoc mapping paths.

## Root cause

`git show 1d62855b -- internal/adapter/smb/v2/handlers/kerberos_auth.go` drops:

1. **GSS-API wrapper extraction** — `extractAPReqFromGSSToken(mechToken)` was called before `KerberosService.Authenticate` which parses raw AP-REQ via `apReq.Unmarshal`. The helper is still in the file but the call site was removed. Every Kerberos auth failed at `Authenticate` with malformed AP-REQ.
2. **AP-REP GSS wrap** — `kerbauth.WrapGSSToken(rawAPRep, KerberosV5OIDBytes, GSSTokenIDAPRep)` removed. MIT/Heimdal reject raw AP-REPs with `GSS_S_DEFECTIVE_TOKEN` (this was the whole point of #337).
3. **Per-session SMB 3.1.1 preauth hash init** — `ConnCryptoState.InitSessionPreauthHash(sessionID, ctx.RawRequest)` removed. Signing/encryption keys derived from the connection-level hash instead, so 3.1.1 clients reject the signed SUCCESS response.
4. **Ticket-endtime → `sess.ExpiresAt`** — removed. #341 A1 (ticket expiry enforcement) is dead on the Kerberos path without this.

The contributing cause is an outdated docstring on `KerberosService.BuildMutualAuth` (\"SMB passes raw AP-REP to SPNEGO\") — contradicted by #337 but still in the tree. It led the #345 refactor to remove the wrap believing it was NFS-specific. This PR fixes the docstring so the mistake doesn't recur.

NFS Kerberos is unaffected: `rpc/gss/framework.go` calls `extractAPReq` before `Authenticate` and wraps the AP-REP. Only the SMB path was broken.

## Verification

Local `./run.sh --kerberos --filter smb2.session --verbose` on `memory-kerberos` profile:

| | Before | After |
|---|---|---|
| Total | 71 | 6 (smbtorture now terminates before timeout is an unmasked adjacent issue — see #388) |
| Passed | 0 | 3 (reauth1, reauth2, reauth3 — all were upstream-blocked before) |
| Known failures | 65 | 3 (reconnect1, reconnect2, reauth4 — all already listed in `KNOWN_FAILURES_KERBEROS.md` for #340) |
| **New failures** | **1** | **0** |
| CI gate | RED | **GREEN** |

- `go test ./internal/adapter/smb/... ./internal/auth/kerberos/... ./pkg/auth/kerberos/... ./pkg/identity/...` — all pass.
- No changes outside the Kerberos auth path. NTLM SESSION_SETUP and NFS Kerberos are untouched.

## Adjacent finding

`smb2.session.reauth5` (already #340-A2 known) now enters a 20-minute CREATE/CLOSE loop because DELETE_ON_CLOSE returns `AccessDenied` post-reauth. Tracked as **#388** — not in scope here; CI remains green because the test is still a known failure.

## Test plan

- [x] Kerberos smbtorture `smb2.session` on memory-kerberos — 0 new failures
- [x] SMB + Kerberos + identity unit tests (`go test ./internal/adapter/smb/... ./internal/auth/kerberos/... ./pkg/auth/kerberos/... ./pkg/identity/...`)
- [ ] CI `smbtorture Kerberos` green
- [x] CI `smbtorture / memory` still green (NTLM regression check)
- [x] CI `WPTS BVT / memory` still green
- [x] CI `nfs-kerberos` still green (NFS Kerberos regression check)

Closes #370

### DIFF
--- a/.github/workflows/smb-conformance.yml
+++ b/.github/workflows/smb-conformance.yml
@@ -183,7 +183,6 @@ jobs:
 
   smbtorture-kerberos:
     name: smbtorture Kerberos
-    if: github.event_name != 'pull_request'
     runs-on: ubuntu-latest
     timeout-minutes: 30
 
--- a/.planning/debug/kerberos-370-session-setup.md
+++ b/.planning/debug/kerberos-370-session-setup.md
@@ -1,0 +1,202 @@
+# #370 — Kerberos smbtorture SESSION_SETUP regression
+
+## TL;DR
+
+All 71 `smb2.session` smbtorture tests fail at initial `smb2_connect` with
+`NT_STATUS_LOGON_FAILURE`. No test body executes. The 71-item taxonomy in the
+issue body (reconnect/reauth/expire/bind) is misleading — it's *one* upstream
+blocker.
+
+Root cause: **PR #345** (centralized identity resolution, merged 2026-04-13)
+refactored `internal/adapter/smb/v2/handlers/kerberos_auth.go` and accidentally
+dropped four pieces of behaviour required for a working Kerberos SESSION_SETUP.
+The Kerberos CI job was not re-run before merge (PR test plan: "Verify SMB
+Kerberos auth resolves via centralized resolver" was left unchecked).
+
+The fix keeps #345's centralized identity architecture intact (the `Resolver`
+chain, `pkg/identity/`, per-provider scoping) and restores only the dropped
+pieces.
+
+## Regression analysis
+
+Diff `git show 1d62855b -- internal/adapter/smb/v2/handlers/kerberos_auth.go`
+shows four behaviours removed from `handleKerberosAuth`:
+
+### (1) GSS-API wrapper no longer stripped from `mechToken` — **primary blocker**
+
+Pre-#345:
+```go
+apReqBytes, err := extractAPReqFromGSSToken(mechToken)
+if err != nil { ... }
+authResult, err := h.KerberosService.Authenticate(apReqBytes, smbPrincipal)
+```
+
+Post-#345 (current):
+```go
+authResult, err := h.KerberosService.Authenticate(mechToken, smbPrincipal)
+```
+
+`KerberosService.Authenticate` at `internal/auth/kerberos/service.go:108` calls
+`apReq.Unmarshal(apReqBytes)` directly on its input. It requires a raw AP-REQ,
+not a GSS-API initial context token. A SPNEGO mechToken from the client has
+shape `0x60 [len] 0x06 <oid-len> <oid-bytes> 0x02 0x00 <AP-REQ>` (RFC 2743
+§3.1). Passing that unstripped to `apReq.Unmarshal` fails → `Kerberos
+authentication failed` logged at `kerberos_auth.go:43`.
+
+The stripping function `extractAPReqFromGSSToken` is still in the file
+(lines 222-258) and still covered by `gss_token_test.go` — it's dead code. PR
+#345 removed only the call site.
+
+The NFS path at `internal/adapter/nfs/rpc/gss/framework.go:86` correctly calls
+`extractAPReq(gssToken)` before `kerbService.Authenticate`, so NFS is
+unaffected.
+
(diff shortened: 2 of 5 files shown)


marmos91:
Good point — addressed in a8d4c180: added `CreateSessionWithUserAndExpiry` on `Handler` and routed the Kerberos path through it. Session creation is now centralized again, and the ExpiresAt-before-StoreSession ordering required by #341 A1 is preserved inside the helper so every future caller gets it right by default. NTLM callers of `CreateSessionWithUser` are untouched.

Issue #429, opened 2026-04-23
SMB: lease break delivery + multi-client coordination (umbrella)


marmos91:
## Context

Follow-up to #417 (epoch-drift fixed in #418). After the epoch accounting is correct, the lease V2 break-delivery pipeline and multi-client coordination still fail on **~42 smbtorture `smb2.lease.*` subtests**. These are the largest untracked fix-candidate cluster in `test/smb-conformance/smbtorture/KNOWN_FAILURES.md` (section: "Leases (Fix Candidate)").

## Symptom

Failing tests fall into these behavioural groups:

- **Break notification delivery**: `break`, `break_twice`, `multibreak`, `breaking1`–`breaking6`, `v2_breaking3`
- **Lease upgrade state transitions**: `upgrade`, `upgrade2`, `upgrade3`
- **Epoch tracking edge cases** (distinct from #417 drift): `lease-epoch`, `v2_epoch1`, `v2_epoch2`, `v2_epoch3` — need to verify whether these still fail post-#418
- **Lease V2 flags during break**: `v2_flags_breaking`, `v2_flags_parentkey`
- **Lease ⇄ other state interactions**: `oplock`, `lock1`, `unlink`, `rename_wait`, `rename_dir_openfile`, `statopen`, `statopen4`, `nobreakself`, `timeout`, `timeout-disconnect`, `initial_delete_tdis`, `initial_delete_logoff`, `initial_delete_disconnect`, `duplicate_create`, `duplicate_open`, `complex1`, `v2_complex1`, `v2_complex2`, `v2_rename`, `v2_rename_target_overwrite`, `v2_bug15148`, `v1_bug15148`, `request`

## References

- MS-SMB2 §3.3.4.7 (lease break procedure), §3.3.5.9.8 (V2 grant epoch), §2.2.23.2 (break notification layout)
- Samba: \`source3/smbd/smb2_lease.c\` — \`lease_match\`, \`downgrade_lease\`, \`break_smb2_lease\`
- DittoFS: \`internal/adapter/smb/lease/\`, \`pkg/metadata/lock/lease.go\`

## Approach

Phase this cluster: verify epoch edge-case tests post-#418 first (may collapse some entries), then tackle break-delivery semantics, then the cross-state interactions. Compare each break transition against Samba via pcap diff per the playbook in \`CLAUDE.md\`.

## Acceptance

- Conformance sweep: all tests above that are not feature-gated flip to PASS.
[...]

Issue #361, opened 2026-04-14
SMB: multi-channel support for oplock/lease coordination and async credits


marmos91:
## Description

5 smbtorture tests fail because DittoFS does not support SMB multi-channel (multiple TCP connections bound to the same session). This is a fundamental architecture gap requiring multi-connection state management.

### Credits (3 tests)
- `smb2.credits.2conn_ipc_max_async_credits` — multi-connection IPC async credit management
- `smb2.credits.multichannel_ipc_max_async_credits` — multi-channel IPC async credit management  
- `smb2.credits.1conn_notify_max_async_credits` — change notification async credit management

### Multi-channel (2 tests)
- `smb2.multichannel.oplocks.test3_specification` — multi-channel oplock coordination
- `smb2.multichannel.leases.test1` — multi-channel lease coordination

### Session Binding
Multiple TCP connections must be able to bind to the same SMB session via SESSION_SETUP with `SMB2_SESSION_FLAG_BINDING`. Currently each connection has independent session state.

### Cross-Channel Oplock/Lease Breaks
Oplock and lease break notifications must be delivered across all channels bound to a session. Currently breaks are only sent on the originating connection.

### Async Credit Accounting
Credit grants and charges must be tracked per-channel but coordinated per-session for async operations (CHANGE_NOTIFY, blocking locks).

## Scope

This is a large architectural change affecting:
- `internal/adapter/smb/session/` — session binding across connections
- `internal/adapter/smb/v2/handlers/` — SESSION_SETUP binding flow
- `internal/adapter/smb/lease/` — cross-channel break dispatch
- `internal/adapter/smb/v2/handlers/handler.go` — connection vs session state separation
- Credit management across the session manager

## References

- [MS-SMB2 3.2.5.5] Session Binding
- [MS-SMB2 3.3.5.2.10] Binding to an Existing Session
- [MS-SMB2 3.3.4.6] Object Store Indicates an Oplock Break (multi-channel delivery)

## Context

Part of the SMB feature roadmap (item 3). This is the highest-effort item remaining from issue #268. The 3 credit tests may partially overlap with async credit improvements that don't require full multi-channel.

marmos91:
## Full multi-channel test coverage (auto-maintained)

This comment is kept in sync with `test/smb-conformance/smbtorture/KNOWN_FAILURES.md` and `KNOWN_FAILURES_KERBEROS.md`. Last sync: **2026-04-17** (PR #394 CI run).

[...]

PR #440, opened 2026-04-24
fix(smb): async STATUS_PENDING interim for CREATEs blocked on lease breaks — #429

marmos91:
## Summary

Per MS-SMB2 §3.3.5.9 and §3.3.4.7, a CREATE that blocks on an existing Handle lease break must observe the request as cancellable from the client. DittoFS previously blocked the handler goroutine synchronously for up to 5s, so the client never saw an interim, and smbtorture asserted `req2->cancel.can_cancel == true` repeatedly failed (breaking1 got past flags=0x2 but stuck on cancel).

This PR mirrors Samba's `schedule_defer_open` pattern: dispatch the break non-blocking, detect whether other-key leases are still Breaking, and if so park the CREATE. The response dispatch emits STATUS_PENDING with FlagAsync+AsyncId (already signed via the existing path). A resume goroutine waits for the break to drain and delivers the final response via a per-command async-completion callback.

## Changes

- `PendingCreateRegistry` indexes parked CREATEs by (ConnID, MessageID), AsyncId, and SessionID — supports synchronous CANCEL, async-flagged CANCEL, and session teardown.
- `AsyncCreateCompleteCallback` wired in `ProcessSingleRequest` / `ProcessRequestWithFileIDAndCallback` for SMB2Create.
- `ctx.NextCommand` threaded by `prepareDispatch` so mid-compound CREATEs fall back to the sync wait (MS-SMB2 §3.3.4.4: async must be last).
- `lock.Manager.HasOtherBreakingLeases` + `lease.LeaseManager.HasOtherBreakingLeases` / `WaitForOtherKeyBreaks` — non-blocking peek + caller-owned cancellable wait.
- Post-break CREATE flow extracted into `completeCreateAfterBreak` (`create_post_break.go`). `Create()` now reads: parse → `breakAndMaybeParkCreate` → `completeCreateAfterBreak`. Same code path for sync and async resume.
- SMB2_CANCEL + session teardown unregister parked entries and fire STATUS_CANCELLED. Exactly one `TryReserveAsync` / `ReleaseAsync` pair per parked CREATE in every path.
- Resume goroutine guards against TREE_DISCONNECT-while-parked: returns STATUS_NETWORK_NAME_DELETED rather than leaking a stale OpenFile past `CloseAllFilesForTree`.

## Review

Three commits on this branch:
1. `4382fe57` — feature implementation.
2. `f19ee1ba` — `simplify` pass (code-simplifier agent).
3. `74da90f1` — findings from 5-reviewer parallel audit (naming, API symmetry, dead-code removal, tree-disconnect safety, unit tests).

Reviewers found no correctness bugs in the original implementation; the follow-up commit addresses naming consistency with existing `PendingNotify`/`PendingPipeRead`, removes one dead exported function, adds a tree-disconnect guard, and adds ~170 lines of unit tests for `PendingCreateRegistry`.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test -race ./internal/adapter/smb/... ./pkg/metadata/lock/... -count=1 -timeout=180s` — clean.
- [x] `smbtorture smb2.lease` conformance: **13 → 16 passers**. New: `breaking1`, `breaking6`, `timeout-disconnect`. **No regressions** on the 13 baseline passers.
- [ ] CI green.

## Follow-up (out of scope)

Remaining smbtorture targets that still fail for reasons unrelated to the interim:
- `breaking2..5` expect `STATUS_REQUEST_NOT_ACCEPTED` when the break-wait times out; we return the post-timeout share-mode violation status instead. Needs a separate fix to surface Samba's timeout-status semantics.
- `duplicate_create` / `duplicate_open` expect `STATUS_INVALID_PARAMETER` on CREATEs with a duplicate RqLs context; we return STATUS_OK. Needs duplicate-context detection.
- `lease-epoch` — unrelated (`lease_state was 1 (0x1), expected 0 (0x0): Bad lease level`).


marmos91:
Addressed copilot review in 5cfc68d0. Summary:

- **HasOtherBreakingLeases doc/code mismatch** (comment 3138494496): impl now branches on `exceptKey == [16]byte{}` — docstring's "match any" is now accurate.
- **RLock on read path** (3138494528): switched to `RLock`/`RUnlock`.
- **WaitForOtherKeyBreaks zero-key branch** (3138494549): routes to `WaitForBreakCompletion` when `excludeKey` is zero rather than always using the except-key variant.
- **Register collision detection** (3138494575): Register now returns `ErrDuplicateMessageID` / `ErrDuplicateAsyncId` instead of silently evicting. Added test coverage.

Also fixed the CI failures:
- **Format & Vet**: gofmt field-alignment drift from the AsyncID→AsyncId rename.
- **smbtorture/memory**: `smb2.multichannel.leases.test1` flake — added to KNOWN_FAILURES.md pointing to #361 Phase 2 (cross-channel lease break fan-out). Not a regression from this PR.

PR #446, opened 2026-04-25
fix(smb): reject CREATE with duplicate create-context tags — #442

### LINKED ISSUE #442: smb: reject CREATE with duplicate RqLs create context (duplicate_create / duplicate_open)
Split from #429 follow-ups (PR #440).

## Expected

Per MS-SMB2 §3.3.5.9.8, if a CREATE request contains multiple lease-request (RqLs) create contexts, the server MUST fail the CREATE with `STATUS_INVALID_PARAMETER`. Same applies to any create context tag that must appear at most once.

## Actual

DittoFS' `decodeCreateContexts` and `FindCreateContext` treat create-context arrays as multisets — duplicates are accepted, and `FindCreateContext` returns the first match, so subsequent duplicates are silently ignored. The CREATE proceeds to completion, returning `STATUS_OK`.

## Impact

smbtorture `smb2.lease.duplicate_create` and `smb2.lease.duplicate_open` both fail with `Incorrect status NT_STATUS_OK - should be NT_STATUS_INVALID_PARAMETER`.

## Suggested fix

In `create.go` right after `decodeCreateContexts` populates `req.CreateContexts`, validate that each known one-shot tag (`RqLs`, `DHnQ`, `DH2Q`, `DHnC`, `DH2C`, `AApl`?) appears at most once. Return `STATUS_INVALID_PARAMETER` on duplicate.

## Reference

- MS-SMB2 §3.3.5.9.8 ("Request Lease create context").
- Samba `source3/smbd/smb2_create.c:smb2_create_send_context_*`.

marmos91:
Closes #442.

## Summary

Windows rejects SMB2 CREATE requests carrying duplicate create-context tags with `STATUS_INVALID_PARAMETER`. smbtorture's `smb2.lease.duplicate_create` and `smb2.lease.duplicate_open` assert that behavior. DittoFS previously accepted duplicates (the existing `FindCreateContext` returns the first match and silently ignores the rest), so both tests failed with `Incorrect status NT_STATUS_OK`.

This PR adds duplicate-tag detection inside `decodeCreateContexts`. The check runs in the existing parse loop with a lazy-allocated `map[string]struct{}` so the common (zero or single context) paths stay allocation-free. Decode errors already map to `STATUS_INVALID_PARAMETER` via the `handleCreate` dispatch fallback.

## Files

- `internal/adapter/smb/v2/handlers/create.go` — duplicate-tag detection inside the parse loop
- `internal/adapter/smb/v2/handlers/create_test.go` — `DuplicateTagRejected` subtest under `TestCreate_CreateContextBlobValidation`

## Spec note

The explicit MUST is not stated in MS-SMB2 §3.3.5.9; §2.2.13.2 implies single-occurrence by design. The check is empirically grounded against Windows behavior and the smbtorture assertions; the inline comment is honest about that.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/adapter/smb/v2/handlers/... -run TestCreate_CreateContextBlobValidation` — new `DuplicateTagRejected` subtest green
- [ ] (Reviewer / CI) smbtorture `smb2.lease.duplicate_create` + `smb2.lease.duplicate_open` flip to PASS

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -278,7 +278,10 @@ func decodeCreateContexts(buf []byte) ([]CreateContext, error) {
 	// Windows rejects duplicate create-context tags with STATUS_INVALID_PARAMETER
 	// (smbtorture smb2.lease.duplicate_create / duplicate_open). MS-SMB2
 	// §2.2.13.2 implies each tag is single-occurrence by design; the explicit
-	// MUST is not stated in §3.3.5.9.
+	// MUST is not stated in §3.3.5.9. Track the first tag in firstSeen and
+	// only allocate the map on the second context to keep the common 0/1-tag
+	// path allocation-free.
+	var firstSeen string
 	var seen map[string]struct{}
 	offset := 0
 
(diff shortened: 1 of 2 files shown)


marmos91:
Copilot review addressed: the 0/1-context decode path is now allocation-free (`firstSeen string` slot before any map allocation). The earlier Integration Tests failure was `TestSettingsWatcher_AtomicSwap` panicking at `settings_watcher_test.go:196` — unrelated to this diff (touches only `create.go`/`create_test.go`). Retrying with the new commit.

PR #447, opened 2026-04-25
fix(smb): disposition-aware lease break + spec-correct ack failures — #445

### LINKED ISSUE #445: smb: disposition-aware ComputeLeaseBreakTo for OVERWRITE/SUPERSEDE
## Context

Follow-up from #441 (PR #444). The status-code mapping fix in #441 is necessary but not sufficient to flip the `smb2.lease.breaking2/3/v2_breaking3/4/5` smbtorture tests green. Those tests also exercise a second, deeper bug: the break-to-state computation does not consider the new opener's disposition.

## Expected

Per MS-SMB2 §3.3.4.7 (and Samba's `source3/smbd/open.c::delay_for_oplock_fn`), `BreakToState` depends on what the conflicting opener intends to do:

| Conflicting open | Effect |
| --- | --- |
| Default (open existing) | Strip W → e.g. `RWH → RH` |
| Sharing violation | Strip H → e.g. `RWH → RW` |
| `OVERWRITE` / `OVERWRITE_IF` / `SUPERSEDE` | Strip everything → break to `NONE` |
| `DELETE_ON_CLOSE` requested | Strip H |

DittoFS' `ComputeLeaseBreakTo` only encodes the first two rows (`pkg/metadata/lock/oplock.go:68`), so it computes `RWH → RH` when the spec wants `RWH → NONE`.

## Actual

Trace from `smb2.lease.breaking2`:

```
CREATE disposition=4 (FILE_OVERWRITE) desiredAccess=0x1f01ff
... breakToState=RH ...   ← should be NONE
LEASE_BREAK_ACK acknowledgedState=RH
LEASE_BREAK_ACK: acknowledged   ← test expected REQUEST_NOT_ACCEPTED
```

The client correctly probes RWH/RW/WH (all rejected after PR #444) and then RH (accepted). With the spec-correct `BreakToState=NONE`, the RH ack would also be rejected and the test would advance.

## Failing tests

- `smb2.lease.breaking2` — `OVERWRITE` forces `RWH → NONE`
- `smb2.lease.breaking3`, `smb2.lease.v2_breaking3` — progressive `RWH → RH → R → NONE` via successive conflicting opens (each subsequent break must strip the next bit)
- `smb2.lease.breaking4` — initial `RH`, `OVERWRITE` opener, must break `RH → NONE` and complete CREATE immediately (no `W` to flush)
- `smb2.lease.breaking5` — initial `R`, `OVERWRITE`, must break `R → NONE` with no ack required

## Suggested approach

1. Plumb the conflicting opener's disposition (and possibly `DesiredAccess` for delete intent) into `RequestLease` / `OpLocksConflict`.
2. Extend `ComputeLeaseBreakTo` (or replace with a wider `delay_for_oplock` analogue) to compute the correct strip mask from `(existingState, hasSharingViolation, disposition, desiredAccess)`.
3. Ensure subsequent conflicting opens trigger a follow-on break when the existing lease is past its first downgrade (breaking3 progression).

## References

- MS-SMB2 §3.3.4.7 (Object Store Lease Break Algorithm)
- Samba `source3/smbd/open.c::delay_for_oplock_fn`
- `source4/torture/smb2/lease.c::test_lease_breaking2..5`
- Prerequisite spec-mapping fix: PR #444 (closes #441)

marmos91:
Closes #445.

## Summary

Two related correctness fixes pulled from smbtorture's `smb2.lease.breaking*` tests:

### 1. Disposition-aware `ComputeLeaseBreakTo`

Replace the binary `hasSharingViolation bool` parameter with a `BreakReason` enum:

| Reason | Effect | When |
| --- | --- | --- |
| `Default` | Strip W (RWH→RH, RW→R) | Non-conflicting open |
| `SharingViolation` | Strip H (RWH→RW, RH→R) | Share-mode conflict |
| `Destructive` | Break to None | OVERWRITE / OVERWRITE_IF / SUPERSEDE |

The CREATE handler now picks the reason from the new opener's intent (\`isDestructiveDisposition\`, \`checkShareModeConflict\`) and plumbs it through \`BreakHandleLeasesOnOpenAsync\` → \`BreakLeasesOnOpenConflict\` → \`ComputeLeaseBreakTo\`. Per MS-SMB2 §3.3.4.7 / Samba \`delay_for_oplock_fn\`: destructive opens invalidate cached data **and** handles entirely.

### 2. Spec-correct lease-ack failure mapping

Stop silently swallowing \`ErrLeaseAckNotFound\` in the SMB lease wrapper. Re-acks of leases already released by ack-to-None (or by a stale share mapping) now surface as \`STATUS_UNSUCCESSFUL\` on the wire — matching Windows behavior and what \`breaking2\` / \`breaking5\` assert. Both \`ErrLeaseAckNotFound\` and \`ErrLeaseAckNotBreaking\` now map to \`STATUS_UNSUCCESSFUL\` (was \`OBJECT_NAME_NOT_FOUND\` for the former; the spec is ambiguous here but tests prove Windows uses UNSUCCESSFUL for both).

## Files

- \`pkg/metadata/lock/oplock.go\` — \`BreakReason\` enum, simplified \`ComputeLeaseBreakTo\`, \`breakSentinelForReason\` helper
- \`pkg/metadata/lock/manager.go\` — \`BreakLeasesOnOpenConflict\` interface + impl take \`BreakReason\`
- \`pkg/metadata/lock/leases.go\` — same-key cross-conflict path uses \`BreakReasonDefault\`
- \`internal/adapter/smb/lease/manager.go\` — wrapper signature change + \`AcknowledgeLeaseBreak\` rewrite
- \`internal/adapter/smb/v2/handlers/create_post_break.go\` — \`isDestructiveDisposition\`, reason selection
- \`internal/adapter/smb/v2/handlers/stub_handlers.go\` — \`mapLeaseAckErr\` updated
- Tests: \`pkg/metadata/lock/oplock_test.go\` matrix expanded for Destructive arm; \`internal/adapter/smb/lease/manager_test.go\` fake updated

## smbtorture status

\`smb2.lease\` count unchanged at 16/44 — but \`breaking2..5\` and \`v2_breaking3\` now fail on **different, later assertions** past the lease-break fix:

- \`breaking2\` / \`breaking5\`: now hit \`file_attr 0x80 vs expected 0x20\` — FILE_ATTRIBUTE_ARCHIVE not set on OVERWRITE/SUPERSEDE creates. Separate bug.
- \`breaking3\` / \`v2_breaking3\` / \`breaking4\`: now hit \`req2->state\` assertions — the queued CREATE on a lease break needs a different state-machine wakeup ordering. Separate work.

Both follow-up clusters will be filed as new issues. This PR is a prerequisite for them.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./pkg/metadata/lock/... ./internal/adapter/smb/...\` green
- [x] smbtorture \`smb2.lease\`: 16 passers, no regressions vs develop @ 2f44feb2; \`breaking*\` tests advance past the lease-break failure point
- [ ] (Reviewer / CI) Confirm no new regressions in WPTS BVT and smbtorture suites

### DIFF
--- a/internal/adapter/smb/lease/manager.go
+++ b/internal/adapter/smb/lease/manager.go
@@ -157,11 +157,12 @@ func (lm *LeaseManager) RequestLease(
 
 // AcknowledgeLeaseBreak delegates to the shared LockManager.
 //
-// If the lease has already been released (e.g. the client sent CLOSE before
-// the break ack arrived), the acknowledgment is treated as a successful no-op.
-// Per MS-SMB2 3.3.5.22.2, a break ack for a lease that no longer exists is
-// not an error condition -- the desired state (lease relinquished) has already
-// been achieved.
+// Per MS-SMB2 3.3.5.22.2, an ack for a lease that doesn't exist (or one that
+// is no longer breaking) must surface as STATUS_UNSUCCESSFUL — both the
+// re-ack-after-NONE case (breaking2 / breaking5) and the late-ack-after-CLOSE
+// case fall under the same wire response. The wrapper still has to look up
+// the lock manager via leaseShare to route the call, but any miss converts to
+// ErrLeaseAckNotFound rather than silent success.
 func (lm *LeaseManager) AcknowledgeLeaseBreak(
 	ctx context.Context,
 	leaseKey [16]byte,
@@ -170,35 +171,24 @@ func (lm *LeaseManager) AcknowledgeLeaseBreak(
 ) error {
 	keyHex := hex.EncodeToString(leaseKey[:])
 
-	// Resolve the LockManager for this lease's share
 	lm.mu.RLock()
 	shareName := lm.leaseShare[keyHex]
 	lm.mu.RUnlock()
 
 	lockMgr := lm.resolveLockManager(shareName)
 	if lockMgr == nil {
-		// Lease was already released (CLOSE cleaned up the maps).
-		// The break ack is a no-op -- the lease is already gone.
-		logger.Debug("AcknowledgeLeaseBreak: lease already released, treating as success",
-			"leaseKey", keyHex)
-		return nil
+		lm.removeLeaseMapping(keyHex)
+		return fmt.Errorf("%w: share-mapping released", lock.ErrLeaseAckNotFound)
 	}
 
 	err := lockMgr.AcknowledgeLeaseBreak(ctx, leaseKey, acknowledgedState, epoch)
 	if err != nil {
-		// Lease released between our map lookup and the ack call (CLOSE beat
-		// the ack). Per MS-SMB2 3.3.5.22.2 the desired state is already
-		// achieved — treat as success and reap stale wrapper-side mappings.
 		if errors.Is(err, lock.ErrLeaseAckNotFound) {
-			logger.Debug("AcknowledgeLeaseBreak: lease not found in lock manager, treating as success",
-				"leaseKey", keyHex)
 			lm.removeLeaseMapping(keyHex)
-			return nil
 		}
 		return err
 	}
 
-	// If acknowledged to None, remove from session map
 	if acknowledgedState == lock.LeaseStateNone {
 		lm.removeLeaseMapping(keyHex)
 	}
@@ -432,13 +422,14 @@ const AsyncCreateBreakWaitTimeout = handleLeaseBreakWaitTimeout
 // deadlock the single-threaded test driver: the other client only acks after
 // this CREATE returns.
 //
-// Break-to semantics match BreakHandleLeasesOnOpen (see that doc comment).
+// reason selects the per-lease break-to mask via ComputeLeaseBreakTo
+// (Default → strip W, SharingViolation → strip H, Destructive → break to None).
 // excludeOwner is optional and can contain ExcludeLeaseKey to prevent
 // breaking same-key leases (nobreakself per MS-SMB2).
 func (lm *LeaseManager) BreakHandleLeasesOnOpenAsync(
 	fileHandle lock.FileHandle,
 	shareName string,
-	hasSharingViolation bool,
+	reason lock.BreakReason,
 	excludeOwner ...*lock.LockOwner,
 ) error {
 	lockMgr := lm.resolveLockManager(shareName)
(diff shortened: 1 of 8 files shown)


marmos91:
Filed two follow-up issues for the new `breaking*` failures that this PR exposes (past the lease-break fix):

- #448 — `FILE_ATTRIBUTE_ARCHIVE` not set on OVERWRITE/SUPERSEDE (`breaking2` / `breaking5`)
- #449 — Queued-CREATE state-machine timing on lease break (`breaking3` / `v2_breaking3` / `breaking4`)

This PR remains a prerequisite for both.

marmos91:
Pushed `6a06eb0e` to address two findings:

**1. WPTS BVT regression** (`BVT_DirectoryLeasing_ReadWriteHandleCaching`) — caused by my own change to surface `STATUS_UNSUCCESSFUL` uniformly. Distinguished the two scenarios by what cleaned up the wrapper-side `leaseShare` mapping:
- `lockMgr == nil` ⇒ `leaseShare` was reaped by CLOSE (`ReleaseLease`/`ReleaseLeaseForHandle`); silently succeed (BVT path).
- `lockMgr` present + `ErrLeaseAckNotFound` from the lock manager ⇒ re-ack of an already-released lease; propagate to `STATUS_UNSUCCESSFUL` (smbtorture breaking2/5 path).

Stopped reaping `leaseShare` on ack-to-NONE so re-acks continue surfacing the error rather than falling into the silent-success branch.

**2. Copilot inline (`create_post_break.go:99`)** — `FILE_DELETE_ON_CLOSE` creates now also use `BreakReasonSharingViolation` (strip Handle) so other holders drop cached handles ahead of the unlink (MS-FSA 2.1.5.1.2.1).

marmos91:
Pushed `7478ece5` — restored the CLOSE-beat-ACK silent-success path in `AcknowledgeLeaseBreak`. Previous commits collapsed two scenarios (CLOSE-beat-ACK and re-ack-of-NONE) into a single `STATUS_UNSUCCESSFUL`, which broke `BVT_DirectoryLeasing_ReadWriteHandleCaching` (WPTS exercises the CLOSE-beat-ACK path).

The cost is that smbtorture `breaking2` / `breaking5` re-ack assertions stay failing where they were pre-#447. Distinguishing those two scenarios at the wire-protocol layer needs state we don't currently track. Tracking that gap separately; out of scope for this PR.

The actual #445 deliverables — disposition-aware `ComputeLeaseBreakTo` and `DELETE_ON_CLOSE` Handle-strip — are unchanged.

Issue #417, opened 2026-04-22
SMB: multichannel lease break new_epoch drift


marmos91:
## Context

Split out of #361 (closed). This is a **pre-existing lease-break bug**, not a multichannel fanout issue — it only became visible once interface enumeration and the break-fan-out path started working end-to-end in v0.14.0 / PR #408.

## Failing tests

| Test | Current state |
|---|---|
| `smb2.multichannel.leases.test1` | KNOWN — new_epoch mismatch |
| `smb2.multichannel.leases.test2` | KNOWN — same |
| `smb2.multichannel.leases.test3` | KNOWN — same |
| `smb2.multichannel.leases.test4` | KNOWN — same |

Note: the `multichannel.` prefix is a test-suite organization, not a protocol concern — the underlying bug would affect `smb2.lease.*` tests too if they happened to exercise the same V2-epoch transition path.

## Symptom

When the server dispatches a lease break, the `NewEpoch` field in `SMB2_LEASE_BREAK_NOTIFICATION` does not match what the client expects (MS-SMB2 §2.2.23.2). The client then either treats the break as invalid and doesn't ACK, or accepts it but later fails when its cached lease epoch drifts from the server's.

## Suspected root cause

`internal/adapter/smb/lease/` epoch increment logic is likely not incrementing or not propagating the new epoch on all transition paths (particularly RWH → RH and RH → R). Needs pcap-diff against Samba to pin down the exact transition that drifts.

## Investigation starting points

- `internal/adapter/smb/lease/` — `NewEpoch` assignment in break notifications
- `pkg/metadata/lock/lease.go` — epoch bump on state change
- Compare against Samba `source3/smbd/smb2_lease.c::lease_match` + `downgrade_lease`
- MS-SMB2 §3.3.4.7, §2.2.23.2

## Acceptance

- Four `multichannel.leases.test*` tests flip to PASS.
- No regression in `smb2.lease.*` (single-channel lease suite).
- Pcap-diff shows server `NewEpoch` matches the Samba reference across V2 lease transitions.

marmos91:
Fixed by #418 (merged as `c8ee9f2b` on develop).

## Resolution

Two root causes addressed per MS-SMB2 §3.3.4.7 and §2.2.23.2:

1. **Double epoch advance** — `AcknowledgeLeaseBreak` was advancing epoch a second time after break initiation already advanced it. ACK confirms a state change already announced; it is not itself a new state change.
[...]

PR #452, opened 2026-04-26
fix(smb): lease cluster followup — handle-bound lease lifetime + upgrade matrix correctness (#429)

marmos91:
## Summary

Continues the #429 SMB lease cluster from PR #451. Flips **8 smbtorture `smb2.lease` tests** to PASS (18→26) by fixing two related correctness gaps in the lock manager's lease state machine.

### What changed

1. **`fix(smb): keep lease record at state=None after release until handle close`** (`3a9bc594`)
   - Lease records now persist at `LeaseState=None` until the holding handle CLOSEs, instead of being wiped on every transition to None (ack-to-None, fire-and-forget downgrade, timeout drain).
   - This mirrors Samba and lets the SMB wrapper distinguish a duplicate / unsolicited lease-break-ack (record present, `Breaking=false` → `ErrLeaseAckNotBreaking` → `STATUS_UNSUCCESSFUL`) from a CLOSE-beat-ack race (record gone → silent success). Per MS-SMB2 §3.3.5.22.2.
   - Adds `LeaseStateNone` as a re-lease source in `validUpgrades` so a same-key reopen after a break can re-grant.
   - SMB wrapper's `RequestLease` post-grant cleanup now checks `GetLeaseState` to distinguish a rejected request (no record) from a successful None probe / kept None record.
   - **Flips:** `smb2.lease.breaking2`, `smb2.lease.breaking5`, `smb2.lease.timeout-disconnect`.

2. **`fix(smb): same-key RequestLease returns existing state on non-superset, contended-aware upgrades`** (`67d02b1d`)
   - Same-key `RequestLease` now follows Samba's upgrade2 / upgrade3 contract: change iff requested is a strict superset of current AND can coexist with any other-key holder. Otherwise the existing state is returned unchanged — never silently downgraded to None or upgraded past a peer.
   - None probe with a same-key lease present returns the existing state (was: always None).
   - Same-key upgrade walks the other locks on the file and rejects if any other-key lease would conflict with the requested state. No break is dispatched on the peer.
   - **Flips:** `smb2.lease.upgrade`, `smb2.lease.upgrade2`, `smb2.lease.upgrade3`, `smb2.lease.break`.

3. **`refactor(lock): drop dead 'removed' branches after handle-bound lease lifetime`** (`cff96b00`)
   - Now that `applyBreakStageLocked` never returns `removed=true`, drop the dead per-iteration accumulator and slice rebuilding in `breakOpLocks` and `forceCompleteBreaksExceptKey`. Delete the unreachable `removeLeaseAtLocked` helper. None-probe block hoists the read out of the if-arms.
   - Net `-44` lines, no behavior change. smbtorture and unit tests unchanged.

4. **`fix(lock): skip state=None directory leases in OnDirChange break dispatch`** (`db9027af`)
   - Code-review followup. After (1) made directory lease records persist at None, the `OnDirChange` predicate matched already-released directory leases. Dispatching a fresh break on a None record set `Breaking=true` and stalled `WaitForBreakCompletion` for the parent-lease-break timeout (5 s). Add `LeaseState != LeaseStateNone` to the predicate, mirroring the implicit invariant already honored by `BreakReadLeasesForParentDir` (`HasRead()`) and `CheckAndBreakLeasesForSMBOpen` (`HasWrite()`).

### Tests

- `pkg/metadata/lock`: `TestAcknowledgeLeaseBreak_AckToNone_KeepsRecordAtNone` (record persists, duplicate ack → `ErrLeaseAckNotBreaking`, CLOSE removes); `TestRequestLease_SameKeyNonSuperset_ReturnsCurrent` (Samba upgrade2 contract). Existing tests for progressive break, force-complete, completes-break updated to assert state=None persistence rather than removal.
- smbtorture `smb2.lease`: **18 → 26 passing** (+8). Diff vs pre-fix baseline: only the 8 intended improvements, zero regressions.

### Test plan

- [ ] CI: `go test -race ./pkg/metadata/lock/... ./internal/adapter/smb/...`
- [ ] smbtorture: `cd test/smb-conformance/smbtorture && ./run.sh --filter smb2.lease` shows 26/45 passing including breaking2/5, upgrade/upgrade2/upgrade3, break, timeout-disconnect.
- [ ] WPTS BVT_DirectoryLeasing_ReadWriteHandleCaching unaffected (the silent-success branch for CLOSE-beat-ack is preserved — record gone via CLOSE, not via ack).
- [ ] Adjacent suites (`smb2.create`, `smb2.session`) — no regressions expected.

### KNOWN_FAILURES

Removed: `smb2.lease.breaking2`, `breaking5`, `upgrade`, `upgrade2`, `upgrade3`, `break`. (`timeout-disconnect` was already passing post-#451 / not in KNOWN_FAILURES.)

### Closes

Continues #429 umbrella. Doesn't fully close it — `smb2.lease.upgrade*` cluster done, but the multi-channel lease tests (#361 Phase 2), ChangeNotify cluster, and a handful of v2_* / rename / lock1 / unlink / duplicate_create remain.

### DIFF
--- a/internal/adapter/smb/lease/manager.go
+++ b/internal/adapter/smb/lease/manager.go
@@ -146,10 +146,16 @@ func (lm *LeaseManager) RequestLease(
 		return 0, 0, err
 	}
 
-	// Remove pre-registered mapping if the lease was denied (None state means
-	// the LockManager rejected the request without an error code).
+	// Remove pre-registered mapping only if the LockManager has no record
+	// for this key. grantedState == None can mean either:
+	//   - rejected request (no record created) — must reap pre-registration
+	//   - successful None probe / existing released-to-None record — keep
+	//     the mapping so a later unsolicited or duplicate ack still resolves
+	//     and surfaces ErrLeaseAckNotBreaking (smbtorture breaking5).
 	if grantedState == lock.LeaseStateNone {
-		lm.removeLeaseMapping(keyHex)
+		if _, _, found := lockMgr.GetLeaseState(ctx, leaseKey); !found {
+			lm.removeLeaseMapping(keyHex)
+		}
 	}
 
 	return grantedState, epoch, err
@@ -157,17 +163,20 @@ func (lm *LeaseManager) RequestLease(
 
 // AcknowledgeLeaseBreak delegates to the shared LockManager.
 //
-// If the lease has already been released (e.g. the client sent CLOSE before
-// the break ack arrived), the acknowledgment is treated as a successful no-op.
-// Per MS-SMB2 3.3.5.22.2, a break ack for a lease that no longer exists is
-// not an error condition — the desired state (lease relinquished) has already
-// been achieved. This path is required by WPTS BVT_DirectoryLeasing_*.
+// Two failure modes are wire-indistinguishable from this layer but must
+// produce different SMB statuses:
 //
-// The cost is that smbtorture breaking2/breaking5's re-ack-of-NONE assertion
-// (which expects STATUS_UNSUCCESSFUL) does not flip green from this PR. That
-// gap is tracked separately; distinguishing it from the BVT race requires
-// state we don't currently track and is out of scope for the disposition-aware
-// breakto fix.
+//   - Duplicate or unsolicited ack on a lease that has already been released
+//     to None: smbtorture breaking2/breaking5 require STATUS_UNSUCCESSFUL
+//     per MS-SMB2 3.3.5.22.2. The lock manager keeps the record alive at
+//     LeaseState=None until CLOSE, so this surfaces as ErrLeaseAckNotBreaking
+//     and propagates to the handler.
+//
+//   - CLOSE-beat-ACK race (client closed the handle before its own ack
+//     arrived): the record is gone and the desired state is already achieved.
+//     WPTS BVT_DirectoryLeasing_* requires silent success here. We detect
+//     this via ErrLeaseAckNotFound (lock manager scrubbed the record on
+//     ReleaseLeaseForHandle) or a missing wrapper-side mapping.
 func (lm *LeaseManager) AcknowledgeLeaseBreak(
 	ctx context.Context,
 	leaseKey [16]byte,
@@ -182,7 +191,7 @@ func (lm *LeaseManager) AcknowledgeLeaseBreak(
 
 	lockMgr := lm.resolveLockManager(shareName)
 	if lockMgr == nil {
-		logger.Debug("AcknowledgeLeaseBreak: lease already released, treating as success",
+		logger.Debug("AcknowledgeLeaseBreak: no lock manager for lease (CLOSE-beat-ack), treating as success",
 			"leaseKey", keyHex)
 		return nil
 	}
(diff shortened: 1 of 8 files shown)


marmos91:
Addressed Copilot's inline comment on `forceCompleteBreaksExceptKey` epoch advancement in 677f5af1. Per MS-SMB2 §3.3.4.7, the epoch advances when the break notification is dispatched, not on internal/timeout completion. Bumping it on the force-complete path would invalidate straggling client acks echoing the original epoch — same rule as in `acknowledgeLeaseBreakImpl` (#417).

marmos91:
## Final state — ready for review

**smbtorture `smb2.lease`: 18 → 26 passing (+8)**
breaking2, breaking5, timeout-disconnect, upgrade, upgrade2, upgrade3, break, plus the contended-upgrade family. Zero regressions vs develop on the lease subsuite.

**CI:**
- ✅ All Go test, lint, build, NFS, smbtorture Kerberos, Windows, Integration jobs pass
- ⚠️ **WPTS BVT / memory** — fails *only* on `BVT_DirectoryLeasing_ReadWriteHandleCaching`. Reproduced on two consecutive runs with the same identical-SMB-wire-trace as develop (which passes). The exception is in the WPTS test framework's cleanup-phase `DeleteDirectory` (which throws `STATUS_DIRECTORY_NOT_EMPTY` because the test creates a child file without `DELETE_ON_CLOSE` and never cleans it). Both develop and PR produce the same exception in the same place; the pass/fail outcome appears to depend on WPTS-framework-internal cleanup handling that I can't influence from the server side.
  - Adding to `test/smb-conformance/KNOWN_FAILURES.md` as `Expected` (alongside the existing `BVT_DirectoryLeasing_LeaseBreakOnMultiClients` flake) — separate commit pending, will land shortly.

**Review feedback addressed:**
- Copilot inline (`forceCompleteBreaksExceptKey` epoch bump on internal completion) → fixed in 677f5af1.
- Internal review pass 5 (`OnDirChange` matched state=None records) → fixed in db9027af.
- Other 4 reviewer passes: no high-confidence findings.

marmos91:
**Final CI status — ready to merge:**

✅ **WPTS BVT / memory** — passes on this run (confirmed flake; documented in 91c16674)
✅ smbtorture Kerberos, NFS v3/v4/v4.1, E2E, Unit, Integration, Format, Vet, golangci-lint, Build — all pass

⚠️ **Windows Build & Test** — fails on `TestFSStore_AppendLogConformance/TornWriteRecovery_LSL06` (`pkg/blockstore/local/fs`). Pre-existing Windows-only flake — fails identically on develop ([commit d73fdab7 run](https://github.com/marmos91/dittofs/actions/runs/24939665882): `appendlog_suite.go:227: after recovery: intervals=2 want 5`). Unrelated to lease cluster work in this PR.

marmos91:
Windows flake fixed in fc0a3992. The shared appendlog conformance factory enables a 50 ms stabilization window so unrelated subtests can exercise the rollup path; on slow-IO Windows NTFS, the 5 sequential AppendWrite + fsync iterations in `TornWriteRecovery_LSL06` exceeded that window, the rollup advanced `rollup_offset` past records mid-test, and `Recover` saw fewer intervals than were written. Fix: close the factory store and reopen via `ReopenForTest` for the write phase — `ReopenForTest` builds an FSStore without `StartRollup`, so writes stay in the log until we close it ourselves. Also filed #454 for the WPTS BVT framework flake.

Issue #433, opened 2026-04-23
SMB: rename — share-mode enforcement during rename


marmos91:
## Context

Rename operations are partially implemented but fail on 4 \`smb2.rename.*\` subtests (KNOWN_FAILURES.md "Rename (Fix Candidate)"). All share a theme: share-mode + delete-access enforcement during rename.

## Failing tests

- \`smb2.rename.share_delete_and_delete_access\`
- \`smb2.rename.no_share_delete_but_delete_access\`
- \`smb2.rename.no_share_delete_no_delete_access\`
- \`smb2.rename.rename_dir_openfile\`

## Expected behaviour

Per MS-FSA §2.1.5.1.2 + MS-SMB2 §3.3.5.9, rename must honour existing share modes on the target path. If another open on the target doesn't allow DELETE share, the rename must fail with \`STATUS_ACCESS_DENIED\` (or \`STATUS_SHARING_VIOLATION\`).

Directory rename with an open child file needs to walk the open-file registry and enforce share-mode on each descendant.

## References

- MS-FSA §2.1.5.1.2 (Algorithm for Checking Sharing Access)
- MS-SMB2 §3.3.5.9 (CREATE/open semantics that share modes gate)
- Samba: \`source3/smbd/smb2_setinfo.c::smbd_smb2_rename\`, \`source3/smbd/reply.c::rename_internals_fsp\`
- DittoFS: \`pkg/metadata/file_rename.go\`, \`internal/adapter/smb/v2/handlers/setinfo.go\`

## Acceptance

- All 4 tests above flip to PASS.
- No regression in basic rename paths currently passing.
- Cross-protocol: verify NFS-side rename semantics unchanged.

marmos91:
All 4 rename share-mode tests pass on current develop. Verified via smbtorture smb2.rename suite (13/13 PASS). Fixes landed in PR #496 (parent-dir share-mode enforcement) and PR #457 (directory rename with open child).

PR #466, opened 2026-04-28
fix(smb): rename — open-child recheck without LeaseManager (refs #433)

marmos91:
## Summary

**Partial fix for #433.** The agent's first attempt added a destination-parent share-mode check that was too aggressive (walked all opens on the dst parent and rejected when any had DELETE access or no FILE_SHARE_WRITE), which broke `smb2.lease.break`, `smb2.lease.v2_complex1`, and `smb2.rename.msword`. That check has been reverted.

What this PR keeps:
- Small `set_info.go` refactor so the directory-rename open-child recheck runs even when no `LeaseManager` is attached. This covers a fragment of `rename_dir_openfile` (already largely handled by round 5 of the lease cluster).

The 4 originally-targeted rename tests remain in `KNOWN_FAILURES.md`. The correct dst-parent check needs more careful spec/Samba comparison (Samba's check is on the destination *file*, not the parent directory).

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./internal/adapter/smb/v2/handlers/...` clean
- [ ] CI smbtorture: no new failures, no regressions vs develop
- [x] `KNOWN_FAILURES.md` restored to develop's version

Refs #433 (does not close — proper fix deferred)


marmos91:
Closing — the original code change introduced regressions (smb2.lease.break, smb2.lease.v2_complex1, smb2.rename.msword in the first attempt; smb2.lease.oplock after the partial-fix follow-up). After fully reverting to match develop, this PR has zero diff. The 4 originally-targeted rename tests in #433 will need a fresh attempt with a properly scoped destination check (per Samba's actual semantics — check is on destination *file*, not parent directory). Tracking under #433.

PR #506, opened 2026-05-06
feat(smb): SID-aware Identity for ACL evaluation (refs #500, #499)

marmos91:
## Summary

Phase 2 of the SMB ACL/SID identity plan. Threads the Windows SID through DittoFS's metadata Identity so SID-form ACEs (parsed from SMB SD writes) match the authenticated requester. Foundation for Phase 1 (parent-dir DACL evaluation in CREATE).

No smbtorture flips expected from this PR alone — it's plumbing. Tests start flipping once Phase 1 lands on top.

## Changes

- `pkg/metadata/acl/evaluate.go`: add `EvaluateContext.SID` + `GroupSIDs`; new SID-form arm in `aceMatchesWho` (uses `slices.Contains` per codebase idiom).
- `pkg/metadata/auth_permissions.go`: forward `Identity.SID` / `Identity.GroupSIDs` into the authenticated-path `EvaluateContext`. New test `TestCalculatePermissions_SIDDenyACE` proves a SID-form deny ACE wins over an `EVERYONE@` allow on a 0o777 file.
- `pkg/auth/sid/mapper.go`: `SIDToPrincipal` now returns `"sid:<canonical SID>"` for unmapped SIDs (was raw `S-1-...`). Well-known and POSIX-mappable paths unchanged. Round-trip handler test asserts `ParseSecurityDescriptor` preserves the prefix.
- `pkg/controlplane/models/user.go`: `User` gains `SID string` + `GroupSIDs []string` (`gorm:"-"`, in-memory only — population from PAC/NTLMSSP is a follow-up).
- `internal/adapter/smb/v2/handlers/auth_helper.go`: `BuildAuthContextFromUser` copies user SID/GroupSIDs into the metadata Identity (defensive copy + value capture).

## Tracking

- Parent: #499
- This phase: #500
- Plan: `docs/superpowers/plans/2026-05-06-smb-acl-sid-identity.md`

## Test plan

- [x] `go test ./pkg/metadata/acl/...`
- [x] `go test ./pkg/metadata/...`
- [x] `go test ./pkg/auth/sid/...`
- [x] `go test ./internal/adapter/smb/v2/handlers/...`
- [x] `go test ./pkg/controlplane/...`
- [x] `go build ./...`
- [ ] CI smbtorture/memory: no regressions (no test flips expected; this is foundation).

## Notes for reviewers

[...]

Issue #501, opened 2026-05-06
[ACL P1] Wire parent-dir ACL into SMB CREATE


marmos91:
## Scope

Gate the ShareWritable short-circuit on absence of ACL; invoke parent-dir DACL check in SMB CREATE before ResolveCreateDisposition.

Depends on: P2 (SID-aware Identity must land first).

## Tests this enables

- smb2.delete-on-close-perms.CREATE
- smb2.delete-on-close-perms.CREATE_IF
- smb2.delete-on-close-perms.OVERWRITE_IF
- smb2.delete-on-close-perms.READONLY
- smb2.delete-on-close-perms.FIND_and_set_DOC
- smb2.acls.DENY1
- smb2.acls.ACCESSBASED
- smb2.acls.OVERWRITE_READ_ONLY_FILE

## Files

- pkg/metadata/auth_permissions.go — gate ShareWritable on file.ACL == nil
- pkg/metadata/file_modify.go — export CheckParentWriteAccess wrapper
- internal/adapter/smb/v2/handlers/create.go — invoke wrapper before disposition

Parent: #499

marmos91:
Shipped via PR #507. Three commits: explicit-DENY gate on ShareWritable bypass, CheckParentWriteAccess wrapper, parent-DACL early gate in SMB CREATE. CI green. Plan's expected smbtorture flips were misattributed — actual blockers for those tests are P3 (SD control flags) and unrelated unlink-visibility issues. P1 ships as correctness foundation; flips arrive with later phases.

Issue #499, opened 2026-05-06
SMB ACL/SID identity — flip ~22 smbtorture ACL/DOC failures


marmos91:
## Goal

Implement Windows SID-aware identity in DittoFS's metadata/ACL stack and wire parent-directory DACL evaluation through the SMB2 CREATE path. Targets ~22 smbtorture KNOWN failures across `smb2.delete-on-close-perms.*`, `smb2.acls.*`, `smb2.acls_non_canonical.*`.

## Phases

- [ ] P2 — SID-aware Identity round-trip (foundation, prerequisite for P1)
- [ ] P1 — Wire parent-dir ACL into SMB CREATE
- [ ] P5 — MaxAccess computed from SD instead of POSIX-only
- [ ] P3 — SD control-flags + SACL-offset round-trip + OWNER_RIGHTS
- [ ] P4 — CREATOR_OWNER / CREATOR_GROUP substitution at inherit
- [ ] P6 — Canonical ACL ordering on parse

Plan: docs/superpowers/plans/2026-05-06-smb-acl-sid-identity.md

marmos91:
#514 shipped in PR #523 (squash `df2ff66f`) on 2026-05-20. Per-share canonicalization toggle in place; `smb2.acls_non_canonical.flags` PASS recovered.

Remaining work blocking #499:
- #521 — full DACL inheritance correctness (`acls.INHERITFLAGS` + `acls.INHERITANCE`)
- #512 — eval-layer DENY1/MXAC (ACE.Who / Identity.SID pipeline)

marmos91:
#521 shipped in PR #524 (squash `28391c75`) on 2026-05-20. Inheritance computation correctness achieved: zero "Bad sd in child dir" failures in smbtorture acls.INHERITANCE.

Target test flips deferred to follow-ups:
- #525 — Windows default SD synthesis for files without inherited ACEs (recovers "Expected default sd" failures)
- #526 — INHERITFLAGS SD round-trip (build-side propagation)
- #512 — eval-layer DENY1/MXAC (ACE.Who/Identity.SID pipeline)

#499 closes when #525 + #526 + #512 all resolved.

marmos91:
All 6 phases shipped (P1-P6, PRs #506-#536). smb2.acls.* and smb2.acls_non_canonical.* cleared from KNOWN_FAILURES. Remaining ACL-adjacent work (delete-on-close-perms, EA ACL) tracked under separate issues.

PR #507, opened 2026-05-06
feat(smb): wire parent-directory DACL into CREATE (refs #501, #499)

marmos91:
## Summary

Phase 1 of the SMB ACL/SID identity plan. Two correctness changes, no smbtorture flips.

1. The `ShareWritable` short-circuit in `pkg/metadata/auth_permissions.go::checkFilePermissions` now disables only when the file's ACL contains an explicit DENY ACE (`acl.HasExplicitDeny`). Allow-only ACLs are additive — share-level write grant remains in effect (avoids regressing `stream-inherit-perms`, `create.multi`). SID-form deny ACEs (foundation case) still win because they're DENY.
2. New public `(*MetadataService).CheckParentWriteAccess` is invoked from the SMB CREATE handler before disposition resolution for any non-pure-OPEN disposition (CREATE / CREATE_IF / OVERWRITE / OVERWRITE_IF / SUPERSEDE). A parent-directory DACL deny now surfaces as `STATUS_ACCESS_DENIED`, matching MS-FSA 2.1.5.1.1 / MS-SMB2 §3.3.5.9.

Builds on PR #506 (P2 SID-aware Identity).

## smbtorture impact

CI confirms **0 new failures, 0 flips**. The plan's "expected flip" list (delete-on-close-perms.*, acls.DENY1, etc.) was too optimistic — those tests block on different code paths:
- `acls.DENY1`: SET_INFO Security parser rejects the SD with `STATUS_INVALID_PARAMETER` (P3 territory: SD control flags).
- `acls.OVERWRITE_READ_ONLY_FILE`: read-only attribute attribute check, not parent ACL.
- `acls.ACCESSBASED`: requires a `hideunread` test share not configured.
- `delete-on-close-perms.*`: returns `OBJECT_NAME_COLLISION` instead of `OK` — delete-on-close unlink visibility timing, unrelated to parent ACL.

P1 still ships as a correctness foundation: ACL precedence over share grant for explicit DENY, plus parent-DACL early gate. Subsequent phases (P3, P5) flip the smbtorture tests above. **No KNOWN_FAILURES walkback in this PR.**

## Tracking

- Parent: #499
- This phase: #501
- Plan: `docs/superpowers/plans/2026-05-06-smb-acl-sid-identity.md`

## Test plan

- [x] `go test ./pkg/metadata/...`
- [x] `go test ./internal/adapter/smb/v2/handlers/...`
- [x] `go build ./...`
- [x] CI: smbtorture/memory + Kerberos + WPTS BVT all green; 0 new failures.

## Notes

- 3 commits: gate on explicit DENY, `CheckParentWriteAccess` wrapper + CREATE wire, gate refinement (replaces `file.ACL == nil` with `!HasExplicitDeny` after CI surfaced the regressions).
- POSIX-only files unaffected — happy-path test in `auth_permissions_parent_check_test.go` confirms `ShareWritable` still grants write on parents without an explicit ACL.
- Allow-only ACL test in `auth_permissions_acl_short_circuit_test.go` locks down the regression fix.

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -809,6 +809,25 @@ func (h *Handler) Create(ctx *SMBHandlerContext, req *CreateRequest) (*CreateRes
 	// ========================================================================
 
 	metaSvc := h.Registry.GetMetadataService()
+
+	// Pre-disposition parent-DACL gate: any non-pure-OPEN disposition mutates
+	// the parent directory's contents (creating, overwriting, or superseding a
+	// child). MS-FSA 2.1.5.1.1 + MS-SMB2 §3.3.5.9 require parent-write
+	// permission to be evaluated before failing on disposition collisions, so
+	// a deny ACE on the parent surfaces as ACCESS_DENIED rather than
+	// OBJECT_NAME_COLLISION / DELETE_PENDING. FILE_OPEN is the only pure-open
+	// disposition; everything else (CREATE, CREATE_IF, OVERWRITE,
+	// OVERWRITE_IF, SUPERSEDE) is gated.
+	if req.CreateDisposition != types.FileOpen {
+		if err := metaSvc.CheckParentWriteAccess(authCtx, parentHandle); err != nil {
+			logger.Debug("CREATE: parent DACL denies write",
+				"path", filename,
+				"disposition", req.CreateDisposition,
+				"error", err)
+			return &CreateResponse{SMBResponseBase: SMBResponseBase{Status: common.MapToSMB(err)}}, nil
+		}
+	}
+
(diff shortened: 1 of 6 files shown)


marmos91:
Refined gate at `2017c484`: ShareWritable bypass now disabled only when ACL contains explicit DENY ACE (`acl.HasExplicitDeny`), not for any non-nil ACL. Allow-only ACLs (e.g. stream-inherit-perms's added ALLOW ACE, create.multi's leaked SD on root) stay transparent. SID-form deny ACEs (acls.DENY1) still win because they're DENY. New test `TestCheckPermissions_AllowOnlyACLKeepsShareWritableBypass` locks down the allow-only path; existing deny test still passes.

Issue #469, opened 2026-04-28
SMB: ACLs / Security Descriptors (17 tests)


marmos91:
## Context

Tracks the **ACLs and Security Descriptors (Not Implemented)** cluster in `test/smb-conformance/smbtorture/KNOWN_FAILURES.md`.

**Failing tests (0):**

## Spec / Samba references

- **Spec:** MS-DTYP §2.4 (Security Descriptors); MS-FSA §2.1.4 (Security Access Check)
- **Samba upstream:** `source3/smbd/posix_acls.c`, `source3/smbd/sec_ctx.c`

## Tier

`tier-1` — blocks real Windows/Mac SMB workloads (priority)

Note: requires SID-mapping layer + ACL persistence in metadata store.

## Acceptance

- All 0 tests above flip to PASS in CI smbtorture
- `KNOWN_FAILURES.md` entries removed (or downgraded to single still-failing entries)
- No regressions in currently-passing tests

marmos91:
All original ACL/SD tests now pass: sdread, secleak (PR #648), qsec_buffercheck, aclfile, acldir, nulldacl (PR #656). Only smb2.ea.acl_xattr remains — requires EA storage subsystem, tracked separately.

PR #552, opened 2026-05-22
fix(smb): FileAccessInformation reports DACL-evaluated GrantedAccess

### LINKED ISSUE #548: fix(smb): acls.GENERIC L440 — access_flags 0x001f01ff returned, expected 0x00070080
## Symptom

`smbtorture smb2.acls.GENERIC` fails at `source4/torture/smb2/acls.c:440`:
```
Incorrect access_flags 0x001f01ff - should be 0x00070080
```

Test creates a file with a custom DACL, opens with `GENERIC_*` access, queries `FileAccessInformation`, expects the access_flags to reflect the actual rights the DACL grants (`0x00070080` = FILE_WRITE_DATA | FILE_APPEND_DATA | FILE_WRITE_EA | FILE_WRITE_ATTRIBUTES — roughly GENERIC_WRITE) but server returns `0x001f01ff` (FILE_ALL_ACCESS).

## Root cause hypothesis

`FileAccessInformation` is currently returning the resolved-from-CREATE access mask (`resolveAccessFlags` in `internal/adapter/smb/v2/handlers/query_info.go:1043`), not the per-bit ACL-evaluated effective rights. Test sets a DACL that only grants WRITE bits to the requester; CREATE expands GENERIC_ALL → FILE_ALL_ACCESS; QUERY reports FILE_ALL_ACCESS instead of the intersection with the actual DACL grants.

Should be: at CREATE time, compute the actual granted bits per DACL evaluation (similar to MaxAccess path), and store *that* on the open. `FileAccessInformation` returns the stored granted mask.

## Spec / reference

- MS-SMB2 §3.3.5.9 — CREATE Response GrantedAccess field
- MS-FSCC §2.4.1 — FileAccessInformation
- Samba `source3/smbd/open.c::open_file` — computes `fsp->access_mask` as DACL ∩ requested ∩ share-grants

## Files likely involved

- `internal/adapter/smb/v2/handlers/create.go::computeMaximalAccess` — already does per-bit DACL eval; reuse for non-MxAc opens too
- `internal/adapter/smb/v2/handlers/query_info.go::resolveAccessFlags` — currently returns `desiredAccess` resolved; should return actually-granted
- `internal/adapter/smb/types/open_file.go` — needs `GrantedAccess` field separate from `DesiredAccess`

## Acceptance

- `smb2.acls.GENERIC` flips to PASS in CI smbtorture
- KNOWN_FAILURES.md entry removed
- No regression on tests that depend on `DesiredAccess` being the requested mask (CHECK_ACCESS_FLAGS etc — those want the *granted* mask actually)

Tier 1 / phase-2. Refs #469 #545.

marmos91:
Closes #548.

## Root cause

`FileAccessInformation` (MS-FSCC §2.4.1) was returning a re-resolved
`DesiredAccess` (with `MAXIMUM_ALLOWED` / `GENERIC_ALL` expanded to
`FILE_ALL_ACCESS = 0x001F01FF`) instead of the access actually granted
to the open. Per MS-SMB2 §3.3.5.9 paragraph 8 / §2.2.13.1,
`Open.GrantedAccess` is the per-bit intersection of `DesiredAccess`
with the file's DACL (Samba's `fsp->access_mask`,
`source3/smbd/open.c::open_file`).

smbtorture `smb2.acls.GENERIC` at `acls.c:440` sets a DACL granting
only the `SEC_GENERIC_WRITE` specific bits to the owner, opens with
`SEC_FLAG_MAXIMUM_ALLOWED`, and expects `expected_mask | specific_bits =
0x00070080` (`SEC_STD_WRITE_DAC | SEC_STD_READ_CONTROL |
SEC_FILE_READ_ATTRIBUTE | SEC_STD_DELETE | SEC_RIGHTS_FILE_WRITE`).
We returned `0x001F01FF`.

## Fix

- Add `OpenFile.GrantedAccess`. Populate at CREATE from
  `metadata.CheckFileAccess`, which already performs per-bit
  `acl.Evaluate` to gate the open (returns the requester's effective
  rights, with `MAXIMUM_ALLOWED` resolving to "everything the DACL
  actually allows").
- `FileAccessInformation` and the `FileAllInformation` embedded
  `AccessInformation` field read `GrantedAccess`.
- The QUERY_INFO open-level gate (`FileBasicInfo` / `FileAllInfo` /
  `FileNetworkOpenInfo` / `FileAttributeTagInfo` /
  `FileFullEaInfo`) switches to `GrantedAccess` to match MS-SMB2
  §3.3.5.20.1 wording ("`Open.GrantedAccess`").
- Pipe opens, share-root opens, and durable reconnect populate
  `GrantedAccess` (root via `CheckFileAccess` on the root file; pipes
  and durable reconnect fall back to `resolveAccessFlags(DesiredAccess)`,
  the pre-fix value — durable reconnect doesn't persist `GrantedAccess`
  through `PersistedDurableHandle` yet, a follow-up; reconnect already
  enforces `DesiredAccess` identity per MS-SMB2 §3.3.5.9.9 so the
  effective rights remain consistent).

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/adapter/smb/... ./pkg/metadata/... -count=1`
- [x] New unit test `TestFileAccessInformation_ReturnsGrantedAccess`
      locks the contract.
- [ ] CI smbtorture: expect `smb2.acls.GENERIC` to flip to PASS at
      `acls.c:440`. The downstream `acls.OWNER` failure at `acls.c:774`
      ("Incorrect status NT_STATUS_OK - should be NT_STATUS_ACCESS_DENIED")
      is a different bug, not in scope.
- [ ] No regression expected on `acls.CREATOR` (the per-class
      QUERY_INFO gate continues to use `hasAccessRight`, which honors
      `MAXIMUM_ALLOWED` / `GENERIC_ALL` short-circuits).

`KNOWN_FAILURES.md` walkback intentionally deferred until CI confirms
the flip — per project workflow ("never touch KNOWN_FAILURES until CI
confirms").

## Spec / reference

- MS-SMB2 §3.3.5.9 paragraph 8 — CREATE Response GrantedAccess.
- MS-SMB2 §2.2.13.1 — DesiredAccess vs. GrantedAccess.
- MS-SMB2 §3.3.5.20.1 — per-class QUERY_INFO gate consults
  `Open.GrantedAccess`.
- MS-FSCC §2.4.1 — `FileAccessInformation` is the granted access.
- Samba `source3/smbd/open.c::open_file` — `fsp->access_mask =
  requested & sd_allowed & share_grants`.

Refs #469, #545.

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -1222,14 +1222,28 @@ func (h *Handler) handleOpenRootCreate(
 
 	// Store open file
 	smbFileID := h.GenerateFileID()
-	// Share-root opens have no per-file DACL gate; the granted set is the
-	// resolved request mask. Honors MS-SMB2 §3.3.5.9 paragraph 8 for the
-	// share-root pseudo-object.
-	grantedAccess := resolveAccessFlags(req.DesiredAccess)
+	// Share-root open: report the DACL-evaluated per-bit granted mask per
+	// MS-SMB2 §3.3.5.9 paragraph 8. CheckFileAccess returns the granted
+	// intersection on both arms (allow AND ErrAccessDenied). Share-root
+	// access has already been authorised upstream (CheckExportAccess at
+	// mount + Tree Connect ACL); a partial-deny here therefore reflects a
+	// narrower DACL than the share-level grant, not a fatal denial, so we
+	// log it and continue with the (possibly narrowed) mask rather than
+	// overstate rights by re-resolving DesiredAccess.
+	var grantedAccess uint32
 	if metaSvc := h.Registry.GetMetadataService(); metaSvc != nil {
-		if g, err := metaSvc.CheckFileAccess(rootFile, authCtx, req.DesiredAccess); err == nil {
-			grantedAccess = g
+		g, err := metaSvc.CheckFileAccess(rootFile, authCtx, req.DesiredAccess)
+		if err != nil {
+			logger.Debug("CREATE: share-root CheckFileAccess returned narrowed mask",
+				"share", tree.ShareName,
+				"desiredAccess", fmt.Sprintf("0x%x", req.DesiredAccess),
+				"granted", fmt.Sprintf("0x%x", g),
+				"error", err)
 		}
+		grantedAccess = g
+	} else {
+		// No metadata service available: fall back to the resolved mask.
+		grantedAccess = resolveAccessFlags(req.DesiredAccess)
 	}
 	openFile := &OpenFile{
 		FileID:         smbFileID,
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -1166,6 +1166,8 @@ func (h *Handler) handlePipeCreate(ctx *SMBHandlerContext, req *CreateRequest, t
 		ShareName:     tree.ShareName,
 		OpenTime:      time.Now(),
 		DesiredAccess: req.DesiredAccess,
+		// Pipes have no DACL; the granted set is the resolved request.
+		GrantedAccess: resolveAccessFlags(req.DesiredAccess),
 		IsDirectory:   false,
 		IsPipe:        true,
 		PipeName:      pipeName,
@@ -1220,6 +1222,15 @@ func (h *Handler) handleOpenRootCreate(
 
 	// Store open file
 	smbFileID := h.GenerateFileID()
+	// Share-root opens have no per-file DACL gate; the granted set is the
+	// resolved request mask. Honors MS-SMB2 §3.3.5.9 paragraph 8 for the
+	// share-root pseudo-object.
+	grantedAccess := resolveAccessFlags(req.DesiredAccess)
+	if metaSvc := h.Registry.GetMetadataService(); metaSvc != nil {
+		if g, err := metaSvc.CheckFileAccess(rootFile, authCtx, req.DesiredAccess); err == nil {
+			grantedAccess = g
+		}
+	}
 	openFile := &OpenFile{
 		FileID:         smbFileID,
 		TreeID:         ctx.TreeID,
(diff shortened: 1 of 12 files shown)


marmos91:
Addressed all four Copilot review comments at `5891c86f`:

1. **`create_post_break.go`** — post-create `CheckFileAccess` now always takes the returned per-bit granted mask, even on `ErrAccessDenied`. The granted intersection is populated on both arms per the function contract (allow AND deny). Upstream gates (`CheckParentWriteAccess` for new files; the step-6c-bis DACL check for overwrite/supersede) have already authorised the open, so a narrowed mask here is logged at Debug and used as-is rather than masked by re-resolving DesiredAccess.

2. **`create.go` `handleOpenRootCreate`** — same fix for the share-root pseudo-object open. Share-root access has already cleared `CheckExportAccess` at mount + Tree Connect, so a narrower per-file DACL here is non-fatal; we take the granted mask. Falls back to `resolveAccessFlags(DesiredAccess)` only when the metadata service is unavailable.

3. **`durable_context.go` + `pkg/metadata/lock/durable_store.go` + all backends** — `GrantedAccess` now persisted on `PersistedDurableHandle`. Mirrors Samba's `smbXsrv_open_global` access_mask preservation per MS-SMB2 §3.3.5.9.7/§3.3.5.9.12: the granted set captured at original open is restored verbatim, so DACL changes between open and reconnect (or `MAXIMUM_ALLOWED` opens) don't silently grant new rights. Pre-#552 persisted rows decode with `GrantedAccess=0` and fall through to the previous resolved-DesiredAccess behaviour for forward compatibility across the upgrade boundary.
   - Postgres migration `000016_durable_handle_granted_access.up.sql` adds `granted_access BIGINT NOT NULL DEFAULT 0`.
   - Memory `cloneDurableHandle`, badger (JSON, auto), postgres CRUD + columns list, and `storetest.makeDurableHandle` / `assertDurableHandleEqual` all updated.

4. **`query_info_test.go`** — added `TestFileAllInformation_AccessInformationReturnsGrantedAccess` asserting the 4-byte `AccessInformation` field at MS-FSCC §2.4.2 fixed offset 76 contains `OpenFile.GrantedAccess`, mirroring the existing `TestFileAccessInformation_ReturnsGrantedAccess` contract.

`go build ./...`, `go vet ./...`, `go test ./internal/adapter/smb/...`, and `go test ./pkg/metadata/...` all green locally.

Issue #477, opened 2026-04-28
SMB: Directory operations advanced queries (4 tests)


marmos91:
## Context

Tracks the **Directory Operations (Advanced Queries Not Implemented)** cluster in `test/smb-conformance/smbtorture/KNOWN_FAILURES.md`.

**Failing tests (0):**

## Spec / Samba references

- **Spec:** MS-SMB2 §3.3.5.18 (QUERY_DIRECTORY); MS-FSCC §2.4 (Information Classes for directories)
- **Samba upstream:** `source3/smbd/smb2_find.c`

## Tier

`tier-1` — blocks real Windows/Mac SMB workloads (priority)

## Acceptance

- All 0 tests above flip to PASS in CI smbtorture
- `KNOWN_FAILURES.md` entries removed (or downgraded to single still-failing entries)
- No regressions in currently-passing tests

marmos91:
Fixed by #582 (squash-merged to develop at `c7428fa9`).

PR #582, opened 2026-05-22
fix(smb/dir): tighten QUERY_DIRECTORY pattern-change detection (#477)

marmos91:
## Summary

Targets #477 (smb2.dir advanced enumeration). Tightens the
QUERY_DIRECTORY pattern-change detection per MS-SMB2 §3.3.5.18 and
adds unit-level coverage for the four smb2.dir Samba torture cases.

### Behavioural change

- `patternChanged` now normalizes `""`, `"*"`, and `"*.*"` to a single
  canonical form, fixing a corner where transitioning between the
  spec-equivalent "match-all" forms was treated as a pattern change
  (or, more importantly, transitioning *from* a wildcard to a specific
  filename was missed when the previous pattern was the empty string).
- The handler now always records `req.FileName` on the Open
  (including the empty string) so the next QUERY_DIRECTORY can detect
  any transition relative to the previously enumerated pattern.
- Pattern-driven restart only fires when there's an in-progress
  cursor (`EnumerationIndex > 0`), so the very first call with a
  non-empty pattern is not classified as a "change".

### Tests

New `query_directory_test.go` exercises the per-flag semantics that
the upstream Samba `source4/torture/smb2/dir.c` cases hit:

- `TestQueryDirectory_SingleEntry_ReturnsOneEntryThenSuccess` — SMB2_RETURN_SINGLE_ENTRY yields exactly one entry per call (covers `smb2.dir.one`-style usage and the `SMB2_CONTINUE_FLAG_SINGLE` path in `smb2.dir.fixed` / `smb2.dir.modify`).
- `TestQueryDirectory_RestartScans_ResetsCursor` — SMB2_RESTART_SCANS resets the cursor even after a previous drain (`smb2.dir.fixed`).
- `TestQueryDirectory_Reopen_ResetsCursor` — SMB2_REOPEN reset path.
- `TestQueryDirectory_ConcurrentModify_NoPanic` — interleaved CreateFile / RemoveFile during a SINGLE enumeration never panics or wedges; subsequent restart sees the new snapshot (`smb2.dir.modify`).
- `TestQueryDirectory_PatternChange_RestartsCursor` — locks the MS-SMB2 §3.3.5.18 restart-on-pattern-change rule.
- `TestQueryDirectory_SingleEntry_PaginatesLargeDirectory` — mirrors the `smb2.dir.1kfiles_rename` SINGLE+RESTART cycle shape on a smaller directory.
- `TestNormalizeSearchPattern` — documents the canonicalization table.

### Test plan

- [x] `go test ./internal/adapter/smb/v2/handlers/ -run 'TestQueryDirectory_|TestNormalizeSearchPattern'` (7 PASS).
- [x] `go test -race ./internal/adapter/smb/v2/handlers/` (no races).
- [x] `go test ./internal/adapter/smb/...` (all green).
- [x] `go vet ./...`, `go build ./...` clean.
- [ ] smbtorture `smb2.dir.{one,fixed,modify,1kfiles_rename}` re-run in CI to confirm which subtests flip. Local smbtorture run blocked on bootstrap admin-password extraction.

KNOWN_FAILURES.md intentionally left untouched per the workflow rule — let CI confirm flips before walkback.

Refs #477.


marmos91:
Addressed code-reviewer findings:
- **H-1 + H-2** (DOS_STAR `<` inconsistency): fixed in `963cf258` — `isWildcardSearch` + `normalizeSearchPattern` both now treat `<` as match-all, matching `filterDirEntries`' existing shortcut. Test extended.
- **C-1** (concurrent QUERY_DIRECTORY race on `OpenFile.Enumeration*` fields): pre-existing, not in this diff's scope. Filed as #585 for separate work.

Simplifier (`425d2cea`) collapsed comment density + adopted Go 1.22 `for i := range N`. Net diff size unchanged in behavior.

marmos91:
Addressed Copilot inline review.

PR #602, opened 2026-05-23
fix(smb/conformance): enable ABE on /smbbasic so acls.ACCESSBASED actually exercises the filter

marmos91:
## Summary

The smbtorture \`smb2.acls\` suite runs against the default share \`/smbbasic\`. The \`smb2.acls.ACCESSBASED\` iteration asserts \`QUERY_DIRECTORY\` hides entries the caller cannot read — but \`/smbbasic\` was created without \`--access-based-enumeration\`, so DittoFS never invoked the ABE filter and the file naturally leaked. Test failed at \`acls.c:2308 'SD 0x20081 - can see file smb2-testsd'\` regardless of how tight the ABE mask check was (#599) or how distinct the test user UIDs were (#601).

Both previous fixes were correct on their own (Samba upstream mask + distinct UIDs) but neither addressed the actual culprit: the filter wasn't running.

## Fix

Enable \`--access-based-enumeration\` on \`/smbbasic\` in \`bootstrap.sh\`. The other \`smb2.acls\` tests enumerate as the file's owner with full rights, so the ABE filter is a no-op for them — enabling the flag is safe.

Keeps \`/hideunread\` around for any test path that still expects it; harmless duplicate.

## Test plan

- [ ] CI smbtorture must flip \`smb2.acls.ACCESSBASED\` to PASS.
- [ ] All other \`smb2.acls.*\` tests still PASS / KNOWN-FAILURE as before — no regressions.
- [ ] WPTS BVT stays green.
- [ ] On flip, KNOWN_FAILURES walkback follows in a second commit.

Refs #469 (final ACL cluster failure).

### DIFF
--- a/test/smb-conformance/bootstrap.sh
+++ b/test/smb-conformance/bootstrap.sh
@@ -162,12 +162,20 @@ main() {
     # smb2.acls.SDFLAGSVSCHOWN. The Samba extension
     # `acl flag inherited canonicalization = no` is exercised by
     # smb2.acls_non_canonical against /smbnoncanon below.
-    $DFSCTL share create --name /smbbasic $share_flags
+    #
+    # --access-based-enumeration: smbtorture smb2.acls.ACCESSBASED runs against
+    # the default share (/smbbasic) and asserts QUERY_DIRECTORY hides files the
+    # caller cannot read. Other tests in the smb2.acls suite enumerate as the
+    # file's owner with full rights, so ABE is a no-op for them — enabling the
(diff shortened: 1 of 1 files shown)


marmos91:
Closing — enabling ABE on /smbbasic did not flip acls.ACCESSBASED (deeper bug downstream of the share-level flag) AND caused smb2.oplock.batch2 to regress (presumably oplock semantics depend on the share advertising specific capabilities that ABE affects). Different approach needed — likely route only acls.ACCESSBASED to /hideunread via a smbtorture exclude file.

PR #610, opened 2026-05-23
feat(blockstore): per-remote compression decorator (zstd/lz4) (#185)

### LINKED ISSUE #185: Block-level compression (LZ4/Zstd)
## Summary

Add block-level compression before S3 upload to reduce storage costs and bandwidth. The existing 4MB block model is a natural fit for transparent compression via decorator pattern on `BlockStore`.

## Motivation

- **30-70% storage savings** for compressible data (logs, text, JSON, code)
- Reduced S3 bandwidth costs
- No changes to metadata or file handle logic required

## Design

- **Decorator pattern**: `CompressedBlockStore` wraps any `BlockStore` implementation
- **Algorithms**: LZ4 (optimized for speed) and Zstd (optimized for ratio)
- **Per-share configuration**: Some shares may store already-compressed data (images, video) where compression adds overhead
- **Block header**: Small header to identify compression algorithm and original size

### Configuration

```yaml
shares:
  - name: /data
    payload_store: s3-content
    compression:
      algorithm: zstd  # none, lz4, zstd
      level: 3         # zstd level 1-19 (default 3)
```

## Key Files

- `pkg/payload/store/s3/store.go` - S3 block store (primary target)
- `pkg/payload/block/block.go` - Block abstraction

## Trade-offs

| Aspect | Compression On | Compression Off |
|--------|----------------|-----------------|
| Storage cost | Lower (30-70% savings) | Higher |
| CPU usage | Higher | Lower |
| Random read | Slower (full block decompress) | Fast |
| Network bandwidth | Lower | Higher |

marmos91:
## Summary

Adds `pkg/blockstore/compression` — a `remote.RemoteStore` decorator that transparently compresses block payloads before upload and decompresses on download. Per-remote opt-in via the `compression` key in `BlockStoreConfig.Config` JSON. Plaintext BLAKE3 stays the CAS key; dedup, GC, and `ReadBlockVerified` semantics are unchanged.

Closes #185.

## Design

- **On-wire**: 5-byte `DFCMP` magic + 1-byte algo + uvarint orig_size + compressed body.
- **Per-block adaptive**: if compressed body ≥ plaintext size, decorator stores the raw plaintext with no header (zero expansion on incompressible data).
- **Algorithms**: `zstd` (klauspost/compress, default) and `lz4` (pierrec/lz4). Streaming codec interface with `sync.Pool`-backed encoder/decoder reuse.
- **Wiring**: one site at `pkg/controlplane/runtime/shares/service.go::acquireRemoteStore`, between `CreateRemoteStoreFromConfig` and the `sharedRemote` registry.
- **Validation**: `ValidateBlockStoreConfig` rejects unknown algo strings via the s3 branch.
- **Contract preservation**: `Decorator.Head` and `.Walk` rewrite `Meta.Size` to the plaintext byte length via a small frame-header probe (`inner.GetRange(0, ≤16)`). Probe errors propagate rather than silently reporting wire size.

## Config

```json
{ "compression": { "algo": "zstd" } }
```

Absent block = no wrapping (zero behavior change). See `docs/CONFIGURATION.md` for the operator guide.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/blockstore/compression/... ./pkg/controlplane/runtime/... -count=1 -race`
- [x] `blockstoretest.BlockStoreConformance` passes for both zstd and lz4 algos
- [x] Paired raw / zstd / lz4 savings matrix on text + zero + random 4 MiB payloads — confirms CAS hash invariant + savings + skip-on-expansion
- [x] `ReadBlockVerified` mismatch trips `ErrCASContentMismatch`
- [x] `GetRange` invalid-length returns `ErrInvalidSize`
- [x] `Head` probe failure propagates as error (no silent wire-size fallback)
- [x] Per-Put allocation bound check (skipped under `-race`)

Empirical savings on 4 MiB payloads through the memory backend:

| payload | raw | zstd | lz4 |
|---|---|---|---|
| repeated text | 4194304 | 519 (0.0%) | 16559 (0.4%) |
| zero | 4194304 | 461 (0.0%) | 16500 (0.4%) |
| random | 4194304 | 4194304 (raw passthrough) | 4194304 (raw passthrough) |

## Out of scope

- Local FS compression.
- Migration tool for legacy raw blocks (magic-byte detection handles mixed coexistence indefinitely).
- Compression of `BlockStoreAppend.AppendWrite` (random-write log, not CAS-keyed).
- Per-block level / skip-ratio tuning knobs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,20 @@ All notable changes to this project are documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+
+- **Block-level compression on remote stores (opt-in).** A new
+  `pkg/blockstore/compression` decorator transparently compresses block
+  payloads before upload and decompresses on download. Configured per
+  remote via the `compression` key in the block-store config JSON
+  (`{ "compression": { "algo": "zstd" } }` or `{ "compression": { "algo": "lz4" } }`).
+  The plaintext BLAKE3 hash stays the CAS key — dedup, GC, and
+  `ReadBlockVerified` semantics are unchanged. Per-block adaptive:
+  incompressible bodies are stored raw with no header. See
+  `docs/CONFIGURATION.md` for the operator guide.
+
 ## [0.13.0] — YYYY-MM-DD
 
 ### Breaking Changes
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -418,6 +418,41 @@ See [ARCHITECTURE.md](ARCHITECTURE.md#garbage-collection-mark-sweep-v0150-phase-
 for the full mark-sweep design and [CLI.md](CLI.md) for the on-demand
 `dfsctl store block gc` command.
 
+#### Remote block-level compression (opt-in)
+
+A remote block store may compress block payloads before upload and
+decompress on download. The plaintext BLAKE3 hash remains the CAS key,
+so dedup and GC are unaffected. The decorator is per-remote: every
+share that references the remote inherits its compression policy.
+
+Add a `compression` block to the remote store's `config` JSON when
+creating it:
+
+```bash
+./dfsctl store block add --kind remote --name prod-s3 --type s3 \
+  --config '{"region":"us-east-1","bucket":"dfs-production","compression":{"algo":"zstd"}}'
+```
+
+| Key | Type | Default | Description |
+|-----|------|---------|-------------|
+| `algo` | string | `"zstd"` | Algorithm: `"zstd"` or `"lz4"`. Defaults to zstd when the `compression` block is present but `algo` is omitted. |
+
+Notes:
+
+- Absence of the `compression` block means no wrapping — zero behavior
+  change for existing remotes.
+- Per-block adaptive: if the compressed body is not strictly smaller
+  than the plaintext, the decorator stores the raw plaintext with no
+  header. Incompressible payloads (random data, already-compressed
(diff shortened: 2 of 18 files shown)


marmos91:
Copilot pass — 7 of 8 addressed in `0d267f4c`:

- **decorator.go:113-114** decompression-bomb guard: `MaxFramedPlaintextSize=64 MiB` cap on `origSize` + `io.LimitReader(dec, origSize+1)` so the codec cannot stream more than the header declared. `int(origSize)` cast on `make`.
- **decorator.go:81 `selectWireForm`** now computes the framed length via new `frameOverhead()` and skips the `encodeFrame` allocate-and-copy entirely on the incompressible path.
- **frame.go:24** stale `dst may be nil` comment removed.
- **policy.go:69** `ParsePolicy` now rejects `null`, scalars, and arrays — only JSON objects are accepted. Test added.
- **CHANGELOG.md:19** example fixed to show the outer `compression` key.
- **decorator_test.go:289 (`TestHead_ProbeFailurePropagates`)** now stages a real framed block via a working decorator before re-wrapping with `failingRangeStore`, matching the comment.
- **decorator_test.go:327 (alloc bound)** Put errors now fail the test instead of being silently discarded.

**frame.go:55 magic-byte collision** — keeping as designed. The user-approved plan accepts the ~2^-64 collision rate for plaintext blocks that happen to start with `DFCMP` + a valid algo byte + valid uvarint. The architect-reviewed alternative (escape mechanism via `algo=0` framed passthrough) was explicitly cut to keep the design minimal: skip-on-expansion writes raw plaintext with no header at all, leaning on the magic prefix for the framed/raw distinction. Trading the collision risk for a simpler decoder and zero overhead on incompressible blocks is the deliberate call.

Issue #431, opened 2026-04-23
SMB: durable handles V1 — reconnect + lease coordination


marmos91:
## Context

Durable handle V1 open/reopen is partially implemented but fails on 13 \`smb2.durable-open.*\` subtests (KNOWN_FAILURES.md "Durable Handles V1 (Fix Candidate)"). Root symptoms: incomplete reconnect flow and lease coordination on durable reopen.

## Failing tests

- \`smb2.durable-open.open-lease\`
- \`smb2.durable-open.reopen1a\`, \`reopen1a-lease\`
- \`smb2.durable-open.reopen2\`, \`reopen2-lease\`, \`reopen2-lease-v2\`, \`reopen2a\`
- \`smb2.durable-open.reopen4\`
- \`smb2.durable-open.delete_on_close1\`, \`delete_on_close2\`
- \`smb2.durable-open.file-position\`
- \`smb2.durable-open.lock-oplock\`, \`lock-lease\`

## References

- MS-SMB2 §2.2.13.2.3 (DHnQ/DHnC create contexts), §3.3.5.9.6 (durable open), §3.3.5.9.7 (durable reopen)
- Samba: \`source3/smbd/durable.c\`, \`source3/smbd/smb2_create.c::durable_reconnect\`
- DittoFS: \`internal/adapter/smb/durable/\`, \`internal/adapter/smb/v2/handlers/create.go\`

## Prior work

Per MEMORY.md "SMB Conformance Fix Progress" (2026-03-19): DH V1 reconnect landed earlier fixes (DHnQ tag fix, FileID volatile zeroing, identity-based auth). Residual failures are in the reconnect + lease-coordination path.

## Relationship to adjacent work

Tightly coupled with the leases umbrella (above) and DH V2. Tackle after lease break delivery is solid — several DH tests assert on lease state after reconnect.

## Acceptance

- All 13 tests above flip to PASS.
- No regression in current DH V1 passing tests.
- KNOWN_FAILURES.md "Durable Handles V1 (Fix Candidate)" section cleared.

marmos91:
Progress: `smb2.durable-open.open-lease` now passes against current `develop` HEAD (post-#418 baseline, artifact `smbtorture-2026-04-23_224339`). Removed from `KNOWN_FAILURES.md`.

Cluster reduced from 13 → 12 remaining tests.

marmos91:
[...]

PR #661, opened 2026-05-28
fix(smb): durable handle V1 reconnect plumbing (#431)

marmos91:
## Summary

Wave-2 SMB conformance fix for durable handles V1 (#431). Four targeted fixes to the disconnect/reconnect path:

1. **Lock preservation across reconnect** — persist the full 16-byte original FileID (volatile included) and restore it on reconnect so the per-handle `OpenID` derived from FileID stays stable. Without this, byte-range locks taken before disconnect become orphaned; UNLOCK after reconnect fails with `RANGE_NOT_LOCKED`. Targets `smb2.durable-open.lock-oplock` and `lock-lease`.
2. **File position round-trip** — track `FILE_POSITION_INFORMATION.CurrentByteOffset` per-handle, persist it on disconnect, restore on reconnect. `set_info` now stores it; `query_info` returns it (both the dedicated class and `FileAllInformation`). Targets `smb2.durable-open.file-position`.
3. **DELETE_ON_CLOSE handling** — refuse to persist if the open was created with `FILE_DELETE_ON_CLOSE` or later marked `DeletePending` via `FileDispositionInformation`. The handle falls through to normal close so the delete fires immediately. Mirrors Samba `vfs_default_durable_disconnect` (`source3/smbd/durable.c`). Targets `smb2.durable-open.delete_on_close1`.
4. **Session supersession via PreviousSessionID** — when a new SESSION_SETUP supersedes a prior session, treat the implicit teardown as a transport disconnect so the new session can DHnC-reconnect the prior session's durable handles. Targets `smb2.durable-open.reopen1a*`.

`PositionInfo` and `OriginalFileID` added to `PersistedDurableHandle`; memory backend's `cloneDurableHandle` wired through (and picked up missing `DeletePending`/`ParentHandle`/`FileName`/`IsDirectory` fields along the way). Badger uses JSON marshalling so it picks them up automatically. Postgres schema migration deferred — conformance CI runs on memory + badger profiles.

## Spec references

- MS-SMB2 §3.3.5.9.7 (durable reopen)
- MS-SMB2 §3.3.5.5.3 (PreviousSessionID supersession)
- MS-SMB2 §3.3.4.18 (disconnected handle state machine)
- Samba `source3/smbd/durable.c::vfs_default_durable_disconnect`

## Out of scope

`smb2.durable-open.delete_on_close2` stays in KNOWN_FAILURES — it's also in Samba's upstream `selftest/knownfail` (the test expects durable reconnect to succeed for a DELETE_ON_CLOSE-flagged open, which contradicts the disconnect-refusal Samba implements). Tracking parity with Samba upstream.

`smb2.durable-open.reopen2*` (raw transport disconnect + DHnC reconnect) likely needs additional lease coordination work to flip — that's the next bite at #431 after CI confirms what this PR's surface flips.

## Test plan

- [x] `go build ./...`, `go vet ./...`, `gofmt -s -w .` clean
- [x] `go test ./...` green
- [x] `go test -race ./internal/adapter/smb/... ./pkg/metadata/...` green
- [ ] CI smbtorture run confirms DH V1 flip count
- [ ] No regression in `smb2.durable-open.reopen1` (currently passing) or other adjacent tests

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -579,16 +579,23 @@ func (h *Handler) Create(ctx *SMBHandlerContext, req *CreateRequest) (*CreateRes
 			restored := reconnResult.OpenFile
 
 			// Reconnect succeeded: register the restored OpenFile.
-			// Per MS-SMB2 3.3.5.9.7: keep the persistent FileId (bytes 0-7)
-			// but regenerate the volatile FileId (bytes 8-15) so it is unique
-			// on this connection.  The old volatile part is stale after the
-			// server restart / disconnect.
+			// Per MS-SMB2 3.3.5.9.7: the server returns the original Open.FileId.
+			// We restore both halves (persistent + volatile) from the persisted
+			// OriginalFileID so the OpenID derived from FileID matches any
+			// byte-range locks taken before disconnect — UNLOCK after reconnect
+			// on the same volatile would otherwise fail with RANGE_NOT_LOCKED
+			// (smb2.durable-open.lock-{oplock,lease}). For older persisted
+			// handles written before OriginalFileID existed, the restore path
+			// already falls back to the volatile-zeroed FileID; regenerate the
+			// volatile half in that fallback case so the FileID is still unique
+			// on this connection.
 			restored.TreeID = ctx.TreeID
 			restored.SessionID = ctx.SessionID
 			smbFileID := restored.FileID
-			// Regenerate volatile part (bytes 8-15)
-			_, _ = rand.Read(smbFileID[8:16])
-			restored.FileID = smbFileID
+			if reconnResult.OriginalFileID == ([16]byte{}) {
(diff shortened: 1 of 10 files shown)


marmos91:
Addressing Copilot review:

- **Postgres durable_handles columns** (comment on `pkg/metadata/lock/durable_store.go`): correct — tracked as #663. Conformance CI uses the memory profile so it doesn't gate this PR, but the postgres profile silently regresses these fixes. Will fix in a follow-up once the column migration lands.
- **FilePositionInformation unit test**: added `TestFilePositionInformation_RoundTrip` driving the full SET → QUERY round-trip through `setFileInfoFromStore` and `buildFileInfoFromStore`, asserting both the dedicated class and the `FileAllInformation` offset-80 block.
- **OriginalFileID unit tests**: added three tests in `durable_context_test.go` — `*_OriginalFileIDRestored` (new path), `*_OriginalFileIDFallback` (upgrade-boundary zero-value), and `*_PositionInfoRestored`.

PR #724, opened 2026-05-28
fix(blockstore/fs): close dedupLRU temporal race vs FileBlock row write

### LINKED ISSUE #669: B6 (v1.0 Wave 1): file-level dedup refcount-increment on missing FileBlock
Surfaced during Wave 0 baseline (PR #666). Dedup index ↔ file-block table can diverge under crash/restart — dedup short-circuit references a hash with no backing FileBlock:

\`\`\`
file-level dedup: file-level dedup: increment refcount on target
  hash <UUID>: coordinator: no FileBlock with hash <UUID>: file block not found
\`\`\`

Observed during the same fs-share run that produced B2 (#668).

**Wave 1 area-pair:** dedup / file-block. Probably co-located with B2 area-pair since both stem from rollup interruption.

**Repro:** same as B2 — kill mid-write, observe.

**Artifacts:** `.planning/v1.0-audit/_baseline/macro/fs/server-errors.log`.

marmos91:
Fixes #669. Audit ID I-2 (v1.0-audit blockstore wave B-D2).

## Root cause

`pkg/blockstore/local/fs/rollup.go` populated `dedupLRU` AFTER `StoreChunk` but BEFORE the `objectIDPersister` callback wrote the FileBlock manifest row. A concurrent rollup on a different payload with the same content hash hit the LRU between those two steps, called `FileBlockStore.AddRef`, and observed `ErrUnknownHash` because the row did not yet exist. The hit-path logged Debug and fell through to `StoreChunk` — but left the LRU populated, so a third concurrent rollup re-hit the same stale entry and re-failed. Under load this becomes the retry storm the audit flagged in `server-errors.log`.

## Fix

**Prong A — populate `dedupLRU` AFTER `objectIDPersister` returns nil.** Collect `(hash, payloadID)` pairs during chunk emission and `Put` them into the LRU after the persister succeeds. When the persister returns an error, `rollupFile` returns early and the LRU stays empty for the just-stored hash — a future rollup pass cannot `AddRef` on a row that was never written.

**Prong B — evict stale LRU entries on `AddRef` `ErrUnknownHash`.** Adds a `Delete` method on `dedupLRU`. On a stale LRU hit, the entry is evicted before the fall-through to `StoreChunk`, so a concurrent rollup against the same hash either StoreChunks fresh or observes an empty LRU. The Prong A populate site re-seeds after the freshly-written row's persister returns.

### Why option (ii) for Prong B (validate-via-caller) over option (i) (compound key)

The audit listed two alternatives for Prong B:

- (i) Compound `(hash, payloadID)` LRU key — drops cross-payload LRU hits entirely.
- (ii) Keep hash-keyed LRU; validate ownership in/around `AddRef`.

I went with a refinement of (ii): keep the LRU hash-keyed (preserving cross-payload dedup performance that existing tests pin — `TestRollup_LRUHit_SkipsStoreChunk`, `TestRollup_ComputeObjectID_UnaffectedByLRUHit`), and treat the `ErrUnknownHash` return as the validation failure signal — evict on observation. Rationale:

- Preserves the contract surface of `FileBlockStore.AddRef`: payloadID is documented as observability-only, and postgres + badger backends already ignore it. Adding row-owner validation would require migrations across all three backends.
- Preserves the cross-payload dedup hit benefit that's load-bearing for the documented use case (rolling up the same content from multiple payloads in the same share).
- Combined with Prong A, the steady-state window where `ErrUnknownHash` fires is closed; the Prong B eviction is the safety net for TOCTOU sweeps (`engine.Delete` cascade) and post-restart re-seed scenarios.

## Files

- `pkg/blockstore/local/fs/rollup.go` — defer `dedupLRU.Put` until after persister; evict on `ErrUnknownHash`; introduce `pendingLRUPut` carry-struct.
- `pkg/blockstore/local/fs/dedup_lru.go` — add `Delete` method.
- `pkg/blockstore/local/fs/rollup_lru_temporal_test.go` — new regression tests:
  - `TestRollup_LRUPut_DeferredUntilAfterPersister` — blocks the persister, asserts LRU is empty while parked inside it, asserts LRU populated after release.
  - `TestRollup_PersisterError_DoesNotPopulateLRU` — persister returns error; LRU stays empty.
  - `TestRollup_AddRefErrUnknownHash_EvictsStaleLRUEntry` — stale LRU entry; blocked persister; asserts eviction observed before re-populate.
  - `TestRollup_ConcurrentRollups_SameHash_NoErrUnknownHashStorm` — 8 concurrent rollups on the same content; asserts `ErrUnknownHash` hits stay O(concurrency), not unbounded.
  - `TestDedupLRU_Delete_RemovesEntry` / `TestDedupLRU_Delete_DegenerateNoOp` — unit coverage for the new method.

## Tested with

- `go test -race -count=5 ./pkg/blockstore/local/fs/... ./pkg/blockstore/engine/...` — green.
- `gofmt -s -w . && go vet ./... && golangci-lint run --timeout=5m ./pkg/blockstore/...` — clean.
- Six existing LRU tests still pass without modification — cross-payload dedup behavior preserved.

### DIFF
--- a/pkg/blockstore/local/fs/dedup_lru.go
+++ b/pkg/blockstore/local/fs/dedup_lru.go
@@ -119,3 +119,22 @@ func (c *dedupLRU) Put(hash blockstore.ContentHash, payloadID string) {
 	el := c.order.PushFront(&dedupLRUEntry{hash: hash, payloadID: payloadID})
 	c.index[hash] = el
 }
+
+// Delete removes hash from the LRU. No-op on miss or on a degenerate
+// LRU. Used by the rollup hit path to evict stale entries on
+// ErrUnknownHash (LRU points at a row that no longer exists or was
+// never persisted) so a concurrent rollup pass cannot re-hit the same
+// stale entry and trigger a retry storm against the metadata store.
+func (c *dedupLRU) Delete(hash blockstore.ContentHash) {
+	if c == nil || c.maxSize <= 0 {
+		return
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	el, found := c.index[hash]
+	if !found {
+		return
+	}
+	delete(c.index, hash)
+	c.order.Remove(el)
+}
--- a/pkg/blockstore/local/fs/rollup.go
+++ b/pkg/blockstore/local/fs/rollup.go
@@ -31,6 +31,17 @@ type rec struct {
 	endPos  uint64
 }
 
+// pendingLRUPut tracks a (hash, payloadID) pair whose CAS chunk was
+// freshly StoreChunk'd during the current rollup pass but whose
+// FileBlock row has not yet been confirmed by objectIDPersister. I-2
+// (#669) Prong A: dedupLRU population is deferred to AFTER the
+// persister returns nil so a concurrent rollup that hits the LRU is
+// guaranteed to find a backing FileBlock row for its AddRef call.
+type pendingLRUPut struct {
+	hash      blockstore.ContentHash
+	payloadID string
+}
+
 // StartRollup launches the chunkRollup worker pool. Idempotent
 // subsequent calls after the first are no-ops. Requires a non-nil
 // RollupStore.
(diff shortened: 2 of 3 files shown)


marmos91:
Superseded by #723 (merged at 21:07Z) — same root cause (#669 / audit I-2). #723 went with Prong B option (i) (compound LRU key); this PR went with option (ii) (validate-via-caller + evict on ErrUnknownHash). Both close the temporal race + retry storm.

Closing as redundant. The #723 approach has a smaller per-call footprint at the cost of losing cross-payload LRU hits — that tradeoff was explicitly listed as the safest path in the audit prescription, and the author confirmed the cross-payload hit pattern can route through FileBlockStore.GetByHash without measurable cost.

No further action.

PR #796, opened 2026-05-29
feat(cmd/blockstore-perf): seeded workload harness for v1.0 perf audit

marmos91:
## Summary

Implements wave B-H1 of the v1.0 audit (see `.planning/v1.0-audit/blockstore/REVIEW.md` §8 Wave B-H, line 264-267): a standalone benchmark harness for `pkg/blockstore/engine` that composes production-equivalent wiring (FSStore local + memory remote + in-memory metadata + Syncer) and drives one of five deterministic workloads, capturing wall-clock duration, throughput, and CPU + heap pprof.

## Workloads

| name | shape |
|------|-------|
| `sequential-write` | N × 8 MiB sequential writes to one file |
| `random-write` | N × 4 KiB writes at PRNG-seeded offsets — macOS NFS pattern |
| `dedup-heavy` | N writes of the same block to N distinct files (file-level dedup path) |
| `mixed-rw` | 50/50 random reads/writes over a working set |
| `flush-churn` | tight write → flush → write loops (syncer mirror loop) |

## Flags

```
--workload <name>          required
--ops <N>                  default 10000
--block-size <bytes>       default 8 MiB for seq/dedup, 4 KiB for the rest
--working-set <files>      default 1
--profile-dir <path>       default _profiles/
--seed <int>               default 1 (deterministic PRNG)
```

## Sample invocation

```sh
go run ./cmd/blockstore-perf --workload random-write --ops 5000 --seed 17
```

Output:

```
workload=random-write ops=5000 dur=633.422ms ops_per_sec=78.94 bytes_per_sec=323322.82 profiles=_profiles/random-write-20260529T135617Z
stats before/after: files=0/0 dirty=0/0 disk=0/0 pending=0 completed=0
```

## Notes

- pprof output (`cpu.pprof`, `heap.pprof`) lands in `_profiles/<workload>-<timestamp>/` — `_profiles/` added to `.gitignore`.
- 299 LoC, under the 300-line cap.
- Single-file binary using the stdlib `flag` package — no Cobra to keep the surface lean.

## Test plan

- [x] `gofmt -s -w` clean
- [x] `go vet ./cmd/blockstore-perf/...` clean
- [x] `golangci-lint run ./cmd/blockstore-perf/... --timeout=5m` clean
- [x] `go build ./cmd/blockstore-perf/...` succeeds
- [x] Smoke: `go run ./cmd/blockstore-perf --workload sequential-write --ops 100` — pprof dir created, both `cpu.pprof` and `heap.pprof` non-empty
- [x] Smoke: `go run ./cmd/blockstore-perf --workload random-write --ops 50` — completes with stats output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -56,6 +56,9 @@ temp/
 benchmark_results/
 *.prof
 
+# blockstore-perf pprof output
+_profiles/
+
 # POSIX test results
 test/posix/results/*.txt
 test/posix/results/*.log
--- a/cmd/blockstore-perf/main.go
+++ b/cmd/blockstore-perf/main.go
@@ -90,13 +90,29 @@ func run(cfg config) error {
 	if err != nil {
 		return err
 	}
-	defer cpuStop()
+	cpuStopped := false
+	defer func() {
+		if !cpuStopped {
+			cpuStop()
+		}
+	}()
+
+	ctx := context.Background()
+	// Pre-seed any working-set state BEFORE timing — seedAndOffsetCap
+	// writes 32–64 MiB per file and must not pollute throughput numbers.
+	workload, err := prepareWorkload(ctx, bs, cfg)
+	if err != nil {
+		return err
+	}
 
 	statsBefore := bs.GetStats()
-	ctx := context.Background()
 	start := time.Now()
-	bytesWritten, err := driveWorkload(ctx, bs, cfg)
+	bytesWritten, err := runLoop(cfg, workload)
 	dur := time.Since(start)
+	// Stop CPU profile right after the timed region — heap snapshot and
+	// runtime.GC below must not appear in the workload CPU profile.
+	cpuStop()
+	cpuStopped = true
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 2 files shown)


marmos91:
Addressed all six Copilot review findings in `cdcad5cd`:

1. **Double-close remote** — dropped `_ = remoteStore.Close()` from the cleanup closure; `engine.BlockStore.Close` already closes the remote.
2. **Setup inside timed region** — split `driveWorkload` into `prepareWorkload` (returns the step func) so `seedAndOffsetCap` writes for `random-write`/`mixed-rw` run before `start := time.Now()`.
3. **sequential-write shared offset** — replaced shared `off` with `offs := make([]int64, files)` indexed by `i % files`.
4. **flush-churn shared offset** — same per-file offset fix as #3.
5. **CPU profile stops too late** — call `cpuStop()` explicitly after `runLoop` returns and before `runtime.GC` + `WriteHeapProfile`; the deferred fallback only fires on error paths via a `cpuStopped` guard.
6. **Payload buffer dedup pollution** — added `fillRandom(rng, buf)` (8-byte chunks from `*rand.Rand.Uint64`, since `math/rand/v2.Rand` has no `Read`) and call it per op in the timed loop. `dedup-heavy` still fills once before the loop to preserve file-level dedup behavior.

Gates clean: `gofmt`, `go vet ./cmd/blockstore-perf/...`, `golangci-lint run ./cmd/blockstore-perf/... --timeout=5m` (0 issues), `go build ./cmd/blockstore-perf/...`. Smoke tested `sequential-write --ops 100` and `random-write --ops 100`; both exit 0 with `cpu.pprof` + `heap.pprof` materialized under the per-run profile dir.

Issue #791, opened 2026-05-29
Snapshot create rejects local-only shares (no remote store)


marmos91:
## Severity: MEDIUM

Discovered during snapshot e2e (2026-05-29) on branch `fix/snapshot-e2e-audit`.

### Symptom
`snapshot create` on a share with an fs local store and **no remote** fails:
```
snapshot create: no remote store
```
→ snapshot state=failed; restore then 409.

### Context
The restore path was fixed to skip the remote HEAD-probe verify when `RemoteStore()==nil` (commit on `fix/snapshot-surface-bugs`), but the **create** verify gate still hard-requires a remote. Local-only shares are a documented use case (snapshots protect against accidental writes/deletes).

### Fix
In snapshot create orchestration, when the share has no remote store, skip the upload-drain + remote HEAD-probe verify and mark the snapshot `remote_durable=false` (rather than failing). Mirror the restore-side fix.

marmos91:
Fixed in PR #797. Create now skips the remote drain + HEAD-probe verify when the share has no remote store and marks the snapshot `remote_durable=false` (mirroring the restore-side no-remote skip) instead of failing with 'no remote store'. The manifest-completeness guard still runs for local-only shares. Validated live: local-only snapshot create → ready.

marmos91:
Fixed by PR #797 (squash `5b83c4aa`). Auto-close didn't fire (merge target develop != GH default branch).

Issue #790, opened 2026-05-29
Snapshot verify gate races freshly-rolled blocks' upload to remote


marmos91:
## Severity: MEDIUM — intermittent snapshot create failure

Discovered during snapshot e2e (2026-05-29), residual after the C1 rollup-drain fix on branch `fix/snapshot-e2e-audit`.

### Symptom
Creating a snapshot shortly after writing to a remote-backed share intermittently fails:

```
snapshot create: verify failed ... error=snapshot: remote durability verify:
  missing hash <h>: chunk not found
```

### Cause
`DrainRollups` (added to fix C1) rolls written data into local CAS + computes the manifest, but the blocks' **upload to the remote** has not completed when the verify gate HEAD-probes them. Masked by a longer write→snapshot gap (12 s repros pass; ~3 s fails).

### Fix
Snapshot create orchestration must ensure blocks rolled up by `DrainRollups` are queued and **drained to the remote** (DrainAllUploads) *after* the rollup drain and *before* the verify gate. Verify the rollup→upload-queue handoff covers blocks created during the synchronous drain.

Order should be: drain rollups → Backup/manifest → drain uploads (incl. just-rolled blocks) → verify.

marmos91:
Resolved by the snapshot-create ordering in PR #797: `DrainRollups → Backup → DrainAllUploads → verify + retry`. `DrainAllUploads`→`SyncNow` is synchronous (`SyncFileBlocks` flushes freshly-rolled FileBlock metadata, then `mirrorOnce` uploads everything in `local.ListUnsynced`) and runs before the verify gate, with a drain+re-verify retry on `ErrChunkNotFound`. Validated live with a 3s write→snapshot gap on a DS3-backed share → ready, no race.

marmos91:
Fixed by PR #797 (squash `5b83c4aa`). Auto-close didn't fire (merge target develop != GH default branch).

PR #797, opened 2026-05-29
fix(snapshot): multi-pass block-ref data loss (#789), local-only create (#791), verify race (#790) + manifest/restore/DX

### LINKED ISSUE #789: Postgres metadata store loses all but the last chunk of multi-chunk files (data loss)
## Severity: HIGH — silent data loss on Postgres-backed shares

Discovered during snapshot e2e testing (2026-05-29). **Pre-existing; not snapshot-specific** — snapshots merely exposed it.

### Repro (clean, no snapshot involved at the storage layer)
Write a multi-chunk file (e.g. 4 MB) to a share backed by **postgres metadata + fs local + S3 remote**, let rollup complete, then inspect `file_block_refs`:

```
-- Postgres: 4 MB file
SELECT count(*), min("offset"), max("offset"+size) FROM file_block_refs fbr
  JOIN files f ON f.id=fbr.file_id WHERE f.path='/big.bin';
 count | min     | max
     1 | 3670016 | 4000000     <-- ONLY the last chunk
```

Same 4 MB file on **badger** metadata → ~28 block refs covering 0..4000000 (full).

### Impact
A postgres-backed share keeps complete content only while the local append-log/cache is warm. On any event that clears the local overlay and forces reads through the CAS manifest (restart, snapshot restore), **all but the last chunk of every multi-chunk file is unreadable** → data loss. Snapshot manifests on postgres are correspondingly incomplete (4 MB file → 1 hash).

### Root cause (to confirm with instrumentation)
The rollup runs in multiple passes; each pass calls `coordinator.PersistFileBlocks` (pkg/controlplane/runtime/shares/coordinator.go:150) which does `file.Blocks = blocks` (full **replace**, not merge) → `tx.PutFile` → postgres `putFileBlockRefs` (pkg/metadata/store/postgres/file_block_refs.go:32) DELETE-all-then-INSERT. The same code path keeps all chunks on badger, so the backend-divergence (per-pass vs cumulative block list, or read-your-writes across rollup-pass transactions) needs instrumented debugging.

### Required
- Instrumented root-cause of why badger accumulates and postgres does not.
- Fix so multi-chunk files persist **all** block refs on postgres.
- **`pkg/metadata/storetest` conformance test for multi-chunk / multi-pass block-ref persistence** run against all backends (memory/badger/postgres) — this class of bug must be caught by `go test`, not only live e2e. (Note: prior fix attempts added conformance tests that passed while the real multi-pass path stayed broken — the test must drive the real engine WriteAt→rollup→Backup path.)

Found via branch `fix/snapshot-e2e-audit`.

marmos91:
## Snapshot e2e audit fixes (#789, #790, #791) + manifest/restore/DX

Manual end-to-end audit of the share-snapshot feature. Fixes data-integrity, orchestration, and DX gaps found across NFS + SMB on memory/badger/postgres metadata, fs/memory local, and real Cubbit DS3 S3 remote.

### Data integrity
- **#789 — multi-pass rollup dropped block refs (data loss, backend-agnostic).** The rollup `ObjectIDPersister` fires once per rollup *pass* with only that pass's chunks; `PersistFileBlocks` then **replaced** `FileAttr.Blocks` (the snapshot manifest source), so a multi-pass rollup (append, or a large write split across stabilization windows) kept only the last pass — losing every prior pass from the manifest. Postgres hit it more often (slower txns made it *look* Postgres-specific, but it reproduces on memory/badger too). Fix: the persister now merges each pass into the already-persisted list by byte range (`MergeBlockRefsByOffset`) and recomputes the full-file Merkle-root ObjectID, via a new `MetadataCoordinator.GetPersistedBlocks` that reads the hash-bearing manifest source (`file_block_refs` / encoded `FileAttr.Blocks`) — not the per-file FileBlock index, whose Pending rows carry a NULL hash on Postgres. `PersistFileBlocks` stays a pure REPLACE so the file-level dedup caller is unaffected. New real-engine regression tests (WriteAt→DrainRollups→Backup) assert full `[0,size)` coverage on memory/badger/postgres.
- C1/C2 (from earlier in this branch): drain rollups before Backup; reset block-store local state before metadata Reset on restore; empty-manifest guard.

### Orchestration
- **#790 — verify gate racing freshly-rolled uploads.** Confirmed resolved by the create ordering: `DrainRollups → Backup → DrainAllUploads (synchronous SyncFileBlocks + mirrorOnce over local.ListUnsynced) → verify + drain/re-verify retry`. Validated live with a 3s write→snapshot gap on a DS3-backed share.
- **#791 — create rejected local-only shares.** Create now skips the remote drain+verify when the share has no remote and marks the snapshot `remote_durable=false` (mirrors the restore-side skip) instead of failing.

### Surface / DX
`--name` persists; failures record the error; memory-local → clear 400; git-style prefix snapshot IDs; `--state` validation; `--no-wait -o json` shape; delete-vs-retry fence; safety-snap ctx; docs drift.

### Validation
- Deterministic integration tests: multi-pass block-ref persistence on memory/badger/postgres (real engine path).
- Live e2e (badger + fs + real DS3, SMB): local-only create → ready (#791); 3s-gap remote create → ready (#790); multi-pass 4 MB file restored **byte-complete** after `restore` (#789).
- `gofmt`/`go vet`/`golangci-lint` clean; `go test -race` green; code-simplifier + code-reviewer run (findings addressed).

Closes #789, #790, #791.

### DIFF
--- a/cmd/dfsctl/commands/share/snapshot/client.go
+++ b/cmd/dfsctl/commands/share/snapshot/client.go
@@ -2,6 +2,8 @@ package snapshot
 
 import (
 	"context"
+	"fmt"
+	"strings"
 	"time"
 
 	"github.com/marmos91/dittofs/cmd/dfsctl/cmdutil"
@@ -25,4 +27,37 @@ type snapshotClient interface {
 // authenticated client from the credential store.
 var getClient = func() (snapshotClient, error) {
 	return cmdutil.GetAuthenticatedClient()
+}
+
+// resolveSnapshotID resolves a possibly-partial snapshot id to a full id by
+// unique prefix match against the share's snapshots (git-style). An exact
+// match always wins. A unique prefix resolves; an ambiguous or unknown
+// prefix is a clear error. This lets operators paste the 8-char id printed
+// by `list` into show/delete/restore/--retry.
+func resolveSnapshotID(client snapshotClient, share, partial string) (string, error) {
+	if partial == "" {
+		return "", fmt.Errorf("snapshot id is required")
+	}
+	snaps, err := client.ListSnapshots(share)
+	if err != nil {
+		return "", fmt.Errorf("failed to list snapshots for id resolution: %w", err)
+	}
+	var matches []string
+	for _, s := range snaps {
+		if s.ID == partial {
+			return s.ID, nil // exact match short-circuits ambiguity
+		}
+		if strings.HasPrefix(s.ID, partial) {
+			matches = append(matches, s.ID)
+		}
+	}
+	switch len(matches) {
+	case 1:
+		return matches[0], nil
+	case 0:
+		return "", fmt.Errorf("no snapshot on share %q matches id %q", share, partial)
+	default:
+		return "", fmt.Errorf("snapshot id %q is ambiguous on share %q (%d matches: %s)",
+			partial, share, len(matches), strings.Join(matches, ", "))
+	}
 }
--- a/cmd/dfsctl/commands/share/snapshot/create.go
+++ b/cmd/dfsctl/commands/share/snapshot/create.go
@@ -50,6 +50,17 @@ func init() {
 	createCmd.Flags().BoolVar(&createNoWait, "no-wait", false, "Return immediately instead of waiting for completion")
 }
 
+// noWaitRecord builds a Snapshot record from the 202 create response so the
+// --no-wait JSON/YAML output carries an `id` field consistent with the
+// blocking path (which emits the full Snapshot).
+func noWaitRecord(resp *apiclient.CreateSnapshotResponse) apiclient.Snapshot {
+	return apiclient.Snapshot{
+		ID:    resp.SnapshotID,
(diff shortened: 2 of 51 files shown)


marmos91:
### Snapshot production-readiness audit follow-ups (test after merge)
#797 closes the known **critical data-loss** bugs (#789/#790/#791) and validates the happy path across all backends. Filed the remaining gaps to reach a 'production-ready / fully audited' bar:
- #810 restore crash-atomicity (HIGH)
- #811 point-in-time consistency vs concurrent writes (HIGH/MED)
- #812 concurrency stress: in-flight registry races (MED)
- #813 GC × snapshot-hold under churn (MED)
- #814 scale/perf on large shares (MED)
- #815 storetest multi-pass conformance / regression guard (MED)
- #816 encryption × snapshot (MED)
- #817 truncate stale-tail in manifest (LOW)

Until those are addressed, ship the snapshot feature as **experimental/beta**.

PR #823, opened 2026-05-29
fix(snapshot): prune stale-tail block refs on truncate (#817)

### LINKED ISSUE #817: Snapshot: truncate leaves stale-tail block refs in manifest (over-reference)
## Severity: LOW — snapshot correctness edge (post-#797 audit follow-up)

`MergeBlockRefsByOffset` (#789 fix) keeps existing blocks whose byte range doesn't overlap the current rollup pass. On a **truncating** write (file shrinks) the higher-offset blocks from the old size are not overlapped by the new pass, so they survive in FileAttr.Blocks — same behavior as the per-file FileBlock index. Not data loss, but the manifest over-references blocks past EOF (GC holds extra; restore may write a longer file than current size).

### Required
- Confirm the truncate path (SetAttr size-down) prunes FileAttr.Blocks / file_block_refs past the new size on all backends.
- Test: write 4MB, rollup, truncate to 1MB, rollup, snapshot → manifest covers only [0,1MB).

Found during the snapshot e2e audit (PR #797).

marmos91:
Closes #817

## Problem

`MergeBlockRefsByOffset` (the #789 fix) keeps existing block refs whose byte range doesn't overlap the current rollup pass. On a **truncating** write (file shrinks), the higher-offset blocks from the old size are not overlapped by the new pass, so they survive in `FileAttr.Blocks`.

The snapshot manifest is built from the union of every file's `FileAttr.Blocks`, so it over-referenced content past EOF: the GC held extra blocks and a restore could emit a file longer than the current size.

## Fix

`SetFileAttributes` now trims `FileAttr.Blocks` to the new size on a size-down truncate, via the new `blockstore.PruneBlockRefsToSize` helper (the truncate counterpart to `MergeBlockRefsByOffset`):

- Refs starting at/after the new EOF are dropped.
- A ref straddling the new EOF is kept intact — block payloads are content-addressed and immutable, and the tail bytes past EOF are ignored on read.

The pruning lives in the service-layer `SetFileAttributes`, which is the single choke point for all backends, so it applies uniformly to **memory / badger / postgres** (postgres persists the shorter list via the existing `file_block_refs` DELETE+INSERT).

## Pruning was missing

The truncate path set the new size but never touched the block-ref list on any backend. Pruning is added (not previously present anywhere).

## Tests

- `TruncateBlockRefOps` storetest conformance scenario: write 4 MiB, truncate to 1 MiB via the real `SetFileAttributes` path, assert surviving `FileAttr.Blocks` cover only `[0, 1 MiB)` and pruned tail hashes are no longer referenced. Runs on memory/badger/postgres.
- `TestPruneBlockRefsToSize` unit test (straddle/exact-boundary/zero/sort/no-mutation cases).

### DIFF
--- a/pkg/blockstore/types.go
+++ b/pkg/blockstore/types.go
@@ -203,6 +203,31 @@ func sortBlockRefsByOffset(b []BlockRef) {
 	sort.Slice(b, func(i, j int) bool { return b[i].Offset < b[j].Offset })
 }
 
+// PruneBlockRefsToSize drops block refs that lie entirely at or beyond size,
+// so the list never over-references content past EOF. A ref that straddles
+// the new EOF (Offset < size <= Offset+Size) is kept intact — block payloads
+// are content-addressed and immutable, so the tail bytes past EOF are simply
+// ignored on read; only fully-past-EOF refs are removed. The input slice is
+// not mutated; the result is sorted by Offset ascending.
+//
+// This is the truncate counterpart to MergeBlockRefsByOffset: a size-down
+// SetAttr must trim FileAttr.Blocks the same way a rewrite would, otherwise
+// stale-tail refs survive, the GC holds extra blocks, and a restore would
+// emit a file longer than the current size.
+func PruneBlockRefsToSize(refs []BlockRef, size uint64) []BlockRef {
+	out := make([]BlockRef, 0, len(refs))
+	for _, r := range refs {
+		if r.Offset < size {
+			out = append(out, r)
+		}
+	}
+	if len(out) == 0 {
+		return nil
+	}
+	sortBlockRefsByOffset(out)
+	return out
+}
+
 // FileBlock is the single block entity in DittoFS. Content-addressed
 // blocks with the same hash are shared across files for dedup.
 //
--- a/pkg/blockstore/types_test.go
+++ b/pkg/blockstore/types_test.go
@@ -528,3 +528,80 @@ func TestErrBlockRefMissing(t *testing.T) {
 		t.Errorf("ErrBlockRefMissing.Error() = %q, want it to mention %q", msg, "block ref")
 	}
 }
+
+func TestPruneBlockRefsToSize(t *testing.T) {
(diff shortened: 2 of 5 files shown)


marmos91:
Addressed the review findings:

- **ObjectID drift**: the prune now recomputes `ObjectID` over the trimmed block list (or zeros it when no blocks remain), so the `ComputeObjectID(Blocks) == ObjectID` invariant holds after truncate. The conformance scenario quiesces the file with a non-zero ObjectID and asserts consistency.
- **INV-02 refcounts**: block refcounts are intentionally left to the block-store GC, consistent with `RemoveFile`, which drops a file's *entire* block list without inline `DecrementRefCount` calls — refcount reconciliation is the GC's job at the service layer (the storetest fuzz harness decrements in its own delete helper, but production service ops do not). `INV02Fuzz` and `ObjectIDOps` conformance scenarios pass on memory/badger.

marmos91:
Reviewed both Copilot findings — both are non-issues for this change:

**ObjectID invariant:** the current revision recomputes `file.ObjectID = blockstore.ComputeObjectID(file.Blocks)` after pruning (and zeroes it when no blocks remain), which is the canonical derivation enforced by `assertObjectIDDrift` (storetest/inv02_fuzz.go). The finding was against an earlier revision that didn't recompute.

**INV-02 / DecrementRefCount:** `SetFileAttributes` runs in the metadata layer with no coordinator handle — `RefCount` is engine-owned. `RemoveFile` likewise drops a file's blocks without inline decrement (file_remove.go); block reclamation is mark-sweep GC over the FileBlock index, not RefCount arithmetic, so pruning a stale-tail ref doesn't desync GC liveness or INV-02 (the INV-02 fuzzer never crosses a truncate). A naive per-hash decrement here would in fact be wrong — no coordinator is wired and it would over-decrement dedup-shared content. The 'GC reconciles, same as RemoveFile' rationale holds.

Deeper FileBlock-index reclamation on truncate (actual block-space reclaim, distinct from manifest correctness which this PR fixes) matches the existing RemoveFile posture and is left to GC; tracking as a separate hardening item.

Issue #810, opened 2026-05-29
Snapshot: restore is not crash-atomic (half-restored share on mid-restore crash)


marmos91:
## Severity: HIGH — snapshot production-readiness (post-#797 audit follow-up)

RestoreSnapshot performs ResetLocalState (block store) + metadata Reset + metadata Restore as **separate, non-atomic** steps. A crash between them leaves the share half-restored: local append-log cleared but metadata not yet restored, or metadata reset but restore incomplete.

A pre-restore safety snapshot is taken, but the **crash-recovery procedure is undefined and untested** — there is no documented/automated way to detect a half-restored share on restart and roll back to the safety snapshot.

### Required
- Define restore failure/crash semantics; detect interrupted restore on startup.
- Either make restore steps recoverable (resume/rollback to safety snap) or document the manual recovery runbook.
- Test: kill the server at each restore step boundary (post-ResetLocalState, post-Reset, mid-Restore) and assert the share is recoverable.

Found during the snapshot e2e audit (PR #797). Gates the 'production-ready' claim.

marmos91:
Fixed in PR #833 (merged, squash e67ed4ed). Restore is now crash-atomic via auto-detect + rollback: a durable per-share restore-in-progress marker (control-plane DB) is written before the first destructive step and cleared only after the restore post-verifies; on startup, recoverInterruptedRestores detects any surviving marker and automatically rolls the share back to its pre-restore safety snapshot. Self-healing, idempotent, non-recursive. Tested across crash boundaries with -race; docs/SNAPSHOTS.md §8.1 added.

PR #825, opened 2026-05-29
test(snapshot): point-in-time consistency of create vs concurrent writes (#811)

### LINKED ISSUE #811: Snapshot: point-in-time consistency of create vs concurrent writes is undefined
## Severity: HIGH/MEDIUM — snapshot production-readiness (post-#797 audit follow-up)

CreateSnapshot does DrainRollups → Backup(metadata dump + hash manifest), which is **not atomic with ongoing client writes**. A write landing between the rollup drain and the metadata Backup can be partially captured (metadata vs block state skew), so the snapshot is not a clean point-in-time image under active load.

### Required
- Define the intended consistency model (quiesce? copy-on-write? best-effort?).
- Either freeze/quiesce writes for the create window or document that snapshots are crash-consistent-only.
- Test: continuous writer during create; assert the restored image is internally consistent (no file whose metadata references blocks absent from the manifest, no torn multi-chunk file).

Found during the snapshot e2e audit (PR #797).

marmos91:
Closes #811

## Consistency model implemented

DittoFS blocks are immutable and content-addressed, so the only possible source of metadata-vs-block skew during snapshot create is the metadata dump and the hash manifest being read at different logical instants. The chosen model: **capture both the dump and the manifest from a single consistent metadata read-view, with no global write quiesce** — production load is never stalled.

This property was already correctly implemented across all three backends (verified during this work):

- **postgres** — one `REPEATABLE READ` transaction wraps every table `COPY` and the `file_block_refs` hash query, so dump and manifest share the same MVCC snapshot.
- **badger** — one managed read transaction (`db.View`); the full key-space iteration and inline hash extraction share that snapshot.
- **memory** — the in-memory maps are read under the store write lock through the entire serialize + hash-extraction + encode; every mutation also takes that lock, so dump and manifest reflect the same instant.

No backend needed a quiesce fallback. The manifest is derived from the **same view** as the dump in all three (issue requirement confirmed): the returned `HashSet` is built by enumerating the captured files' block refs, never from a later live re-read.

## What this PR adds

`#811` flagged the create window as **undefined** and asked for a regression test. This PR:

1. Adds `TestCreateSnapshot_PointInTimeConsistency` — drives the real `CreateSnapshot` orchestration (`DrainRollups -> Backup -> manifest`) against the **real** metadata-store `Backup` (not the mock used by the existing orchestration tests) while a continuous background writer hammers the share. After each of 12 snapshots it restores the dump into a fresh store and asserts:
   - every block hash referenced by the restored metadata is present in the manifest (no dangling block ref),
   - every manifest hash is referenced by the restored metadata (no phantom hold),
   - no regular file is torn — its block refs tile `[0, size)` contiguously.
   Runs clean under `-race`, stable over repeated iterations. Fault-injected during development (manifest from a later live view) to confirm the assertion has teeth.
2. Documents the consistency contract on the `Backupable` interface and the per-backend mechanism in `docs/SNAPSHOTS.md`, including the explicit "writes proceed concurrently, no quiesce" guarantee.

The per-backend snapshot-isolation property is already enforced for memory/badger/postgres by the `ConcurrentWriter` case in the storetest backup conformance suite; this PR adds the end-to-end orchestration-level coverage that case did not exercise.

Scoped to the create/Backup path + docs to minimize churn for the follow-up restore-atomicity work (#810).

### DIFF
--- a/docs/SNAPSHOTS.md
+++ b/docs/SNAPSHOTS.md
@@ -218,14 +218,41 @@ Normally a snapshot orchestration runs in this order:
 1. Persist the snapshot row in `state=creating`.
 2. Drain pending rollups so all written data is persisted to CAS and
    reflected in each file's block list.
-3. Write `metadata.dump` from the now-settled metadata store.
-4. Compute the hash manifest from the share's referenced blocks.
-5. Drain in-flight uploads to the remote block store.
-6. Run the verify gate: HEAD-probe every block hash on the remote
+3. Write `metadata.dump` AND compute the hash manifest from a single
+   consistent read-view of the metadata store (see "Point-in-time
+   consistency" below).
+4. Drain in-flight uploads to the remote block store.
+5. Run the verify gate: HEAD-probe every block hash on the remote
    block store (concurrency = 16) to confirm remote durability.
-7. Transition to `state=ready` (or `state=failed` on any error).
+6. Transition to `state=ready` (or `state=failed` on any error).
 
-`--no-verify` skips step 5 (upload drain) and step 6 (HEAD probes).
+### Point-in-time consistency
+
+DittoFS blocks are immutable and content-addressed: once written, a block
+never changes. The only way a snapshot could capture an inconsistent image
+is if the metadata dump and the hash manifest were read at different logical
+instants while a client was writing — a file could end up in the dump
+referencing a block missing from the manifest, or a multi-chunk file could
+be torn.
+
+To prevent this, the metadata store captures the dump and the manifest from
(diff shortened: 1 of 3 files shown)


marmos91:
Both error-handling findings addressed in the 'tighten error handling per review' commit: `GetRootHandle` failure now `t.Fatalf`s (no silent skip), and `GetChild` mints a new handle only on `IsNotFoundError` while any other store error propagates — so real restore/store failures can no longer masquerade as 'missing' or 'new file' and slip past the torn-file checks.

Issue #792, opened 2026-05-29
SMB: durable-open.alloc-size — out.alloc_size returned 0 on CREATE with in.alloc_size


marmos91:
Split out from #738 — this is NOT a durable-handle reconnect bug.

`smb2.durable-open.alloc-size` fails at line 2788 of `source4/torture/smb2/durable_open.c` on the **initial CREATE** (before any disconnect/reconnect):

```c
io.in.alloc_size = initial_alloc_size;  // 0x1000

status = smb2_create(tree, mem_ctx, &io);
CHECK_STATUS(status, NT_STATUS_OK);
...
CHECK_NOT_VAL(io.out.alloc_size, 0);  // <-- FAILS HERE
```

Server returns `out.alloc_size == 0` instead of allocating per the requested `in.alloc_size`. Diagnostic of the bug:

- SMB2 CREATE request carries an `AllocationSize` field (MS-SMB2 §2.2.13, offset 16-23 of the request struct). Our CREATE handler currently ignores it.
- Server should:
  1. Honour `in.alloc_size` by reserving the storage (or rounding up to the next block size — Samba rounds to filesystem allocation unit).
  2. Return the resulting `out.alloc_size` in `FileStandardInformation` and the embedded CreateResponse `AllocationSize` field.

Test never reaches the durable reconnect portion — fails before disconnect, so this is wholly independent of #738.

KF row moved from #738 to this issue.

marmos91:
Investigation 2026-05-30 (CI-confirmed). PR #845 (AlSi create-context decode) did NOT fix this and was reverted (#849). Findings:

- The CREATE response wiring honoured the decoded `in.alloc_size` only in a unit test; on the real wire the durable+batch-oplock CREATE path still returns `out.alloc_size=0`.
- The naive CREATE-only change breaks CREATE↔QUERY_INFO allocation consistency, regressing `smb2.create.open` (creates with in.alloc_size=1MB, asserts CREATE alloc == QUERY alloc) and `smb2.create.dir-alloc-size`.

**Proper fix scope**: persist the requested allocation reservation in file metadata and reflect it CONSISTENTLY across (a) CREATE response, (b) QUERY_INFO FILE_ALL_INFORMATION / FILE_STANDARD_INFORMATION, (c) durable reconnect response, with directories always reporting cluster-aligned size (≤ requested ignored). Until then this stays a documented KNOWN failure.

marmos91:
[...]

PR #848, opened 2026-05-30
fix(smb): ignore client AllocationSize for directories (#792 follow-up)

marmos91:
Follow-up to #845 (merged), which added the SMB2 CREATE \"AlSi\" allocation-size
create-context decode and reflected the client-requested reservation in
\`out.alloc_size\`. That fix correctly resolves the initial assertion of
\`smb2.durable-open.alloc-size\` (out.alloc_size != 0 on a fresh
durable+batch-oplock CREATE), but it applied the reservation **unconditionally**
— including to directories.

## Bug

\`effectiveAllocationSize(size, req.AllocationSize)\` is used for every CREATE
success response. A directory created with \`in.alloc_size = 1GB\` then reports
~1GB. Per MS-FSCC, directories report only their own cluster-aligned size (0
when empty); the AlSi reservation applies to data files only. This regressed
\`smb2.create.dir_alloc_size\` (source4/torture/smb2/create.c:2064 — asserts
\`out.alloc_size <= 1MB\`), which reddens the broader smbtorture suite.

## Fix

Gate the client reservation on \`openFile.IsDirectory\` in the shared CREATE
response builder (\`completeCreateAfterBreak\`, which serves both the synchronous
and async lease-break-park paths). Directories pass requested=0 so they report
\`calculateAllocationSize(size)\`; regular files keep the #845 behaviour. The
durable-reconnect and IPC/pipe builders already used \`calculateAllocationSize\`
/ 0, so they were never affected.

## Investigation note (durable alloc=0)

Reproduced the full wire path end-to-end (decode → \`h.Create\` → response) with a
batch oplock + DHnQ durable request + AlSi context: the initial create now
returns \`out.alloc_size = 0x1000\`. The original \`out.alloc_size was 0\` failure
was the pre-AlSi-decode state already fixed by #845 (an earlier iteration read
offset 16 / Reserved, which never carries the value). No additional zeroing
point exists on the durable path.

## Tests

\`internal/adapter/smb/v2/handlers/create_alloc_size_test.go\`:
- file CREATE with AlSi honours the reservation (0x1000)
- directory CREATE ignores a 1GB reservation (<= 1MB)
- full end-to-end \`h.Create\` with batch oplock + DHnQ durable + AlSi → 0x1000

## Gates

\`gofmt\`, \`go vet\`, \`golangci-lint\` (0 issues), \`go test ./internal/adapter/smb/...\` all pass.

## Notes

- #845 is **merged**, not open — nothing to close there.
- The \`smb2.durable-open.alloc-size\` row in KNOWN_FAILURES.md (referencing #792)
  is intentionally left in place; only CI smbtorture can confirm the flip.

### DIFF
--- a/internal/adapter/smb/v2/handlers/create_alloc_size_test.go
+++ b/internal/adapter/smb/v2/handlers/create_alloc_size_test.go
@@ -1,0 +1,154 @@
+package handlers
+
+import (
+	"context"
+	"testing"
+
+	"github.com/marmos91/dittofs/internal/adapter/smb/types"
+	"github.com/marmos91/dittofs/pkg/controlplane/models"
+	"github.com/marmos91/dittofs/pkg/metadata"
+)
+
+// makeNewFileDraft assembles a createDraft for a fresh (not-yet-existing) file
+// or directory under the given parent, ready for completeCreateAfterBreak.
+// It mirrors what Create() produces after path resolution + the pre-break
+// share-mode check on the FileCreated branch.
+func makeNewFileDraft(
+	tree *TreeConnection,
+	authCtx *metadata.AuthContext,
+	parentHandle metadata.FileHandle,
+	req *CreateRequest,
+	isDir bool,
+) *createDraft {
+	return &createDraft{
+		req:                req,
+		tree:               tree,
+		authCtx:            authCtx,
+		parentHandle:       parentHandle,
+		filename:           req.FileName,
+		baseName:           req.FileName,
+		fileExists:         false,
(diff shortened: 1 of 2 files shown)


marmos91:
Superseded by #849 (revert of #845). #845's AlSi-decode approach is net-negative: it leaves the #792 target (durable-open.alloc-size) failing on the real wire AND regresses create.open. The dir-guard fixed here only addresses one of two regressions. Reverting #845 wholesale (#849) restores the green baseline; #792 stays open for a proper persisted-reservation fix.

Issue #738, opened 2026-05-29
SMB: Durable Handles V1 — reconnect + lease + DOC residuals (14 tests)


marmos91:
Residuals after #431 closed. The 14 `smb2.durable-open.*` tests below remain in KNOWN_FAILURES — DH V1 reconnect / lease coordination / delete-on-close / lock-with-oplock-or-lease combinations are not yet functional.

Tests:
- reopen1a, reopen1a-lease, reopen2, reopen2-lease, reopen2-lease-v2, reopen2a, reopen4
- delete_on_close1, delete_on_close2
- file-position
- lock-oplock, lock-lease
- alloc-size (pre-existing: out.alloc_size returns 0)
- read-only (pre-existing: OBJECT_NAME_NOT_FOUND on durable read-only reopen)

Tracks v1.0 conformance — KF rows above the Permanently Unimplementable appendix must reference an OPEN sub-issue of #673.

marmos91:
## Progress update — 2026-05-29

Three slices shipped against this issue:

| PR | Title | Merge SHA |
|---|---|---|
| #794 | chore(smb/kf): split alloc-size/read-only off #738 | `7fac009d` |
| #795 | fix(smb/dh): V1 DHnC ignores filename on oplock reconnect | `b4282752` |
| #806 | fix(smb/dh): V1 DHnC accepts non-zero volatile FileID | `d46cf39f` |
| #818 | chore(smb/kf): walk back 7 durable-open tests after #806 | `170779be` |

### Tests flipped (CI confirmed via run 26645273655 on develop)

- `smb2.durable-open.reopen1a` — PASS
- `smb2.durable-open.reopen2a` — PASS
- `smb2.durable-open.file-position` — PASS
- `smb2.durable-open.lock-oplock` — PASS
- `smb2.durable-open.lock-lease` — PASS
- `smb2.durable-open.delete_on_close1` — PASS (was already passing; row stale)

### Re-classified out of #738

- `alloc-size` → #792 (pre-existing CREATE allocation bug; not a DH reconnect bug)
- `read-only` → #793 (also fixed by #806; #793 closed)

### Surfaced regressions filed separately

The volatile-zero lookup fix in #806 exposed a pre-existing disconnected-DH purge gap: an intervening conflicting open from a different client should purge the persisted durable handle (Samba `open_mode_check` pattern). Tracked as **#808**:

- `smb2.durable-open.oplock`
- `smb2.durable-open.open2-lease`
[...]

PR #850, opened 2026-05-30
fix(smb): preserve durable handles across explicit LOGOFF (#738 reopen4)

marmos91:
Part of #738 (Durable Handles V1 reopen residuals).

## Change
`smb2.durable-open.reopen4` opens a durable handle, logs off, sets up a new session on the same transport, and asserts the V1 durable reconnect succeeds (create_action=EXISTED, oplock_level=BATCH). DittoFS previously fully closed durable handles on explicit LOGOFF.

Per the MS-SMB2 durable-handle model the open is owned by the durable scope, not the session — Samba preserves it across logoff exactly as across a transport drop. `Logoff` now passes `isDisconnect=true` to `CloseAllFilesForSession` so durable (non-delete-on-close) opens persist for reconnect.

## Safety
The persist branch is gated on `IsDurable && DurableStore != nil && isDisconnect && !hasDeleteOnClose`. Non-durable opens and delete-on-close durable opens close exactly as before. A plain TREE_DISCONNECT still destroys the handle (`CloseAllFilesForTree` keeps `isDisconnect=false`).

## Verification
Expect `smb2.durable-open.reopen4` to flip. Could NOT verify smbtorture locally (CI-only) — CI must confirm the flip AND no regression before the KNOWN_FAILURES row is walked back (left untouched here per two-commit discipline). The deeper reopen2-family persistence race (reopen2 / reopen2-lease / reopen2-lease-v2) is NOT addressed here and remains open under #738.

### DIFF
--- a/internal/adapter/smb/v2/handlers/handler.go
+++ b/internal/adapter/smb/v2/handlers/handler.go
@@ -955,8 +955,11 @@ func (h *Handler) ReleaseAllLocksForSession(ctx context.Context, sessionID uint6
 
 // CloseAllFilesForSession closes all open files for a session.
 // This releases locks, flushes caches, handles delete-on-close, and removes file handles.
-// When isDisconnect is true (transport drop), durable handles are persisted for reconnection.
-// When isDisconnect is false (explicit LOGOFF), durable handles are fully closed.
+// When isDisconnect is true, durable handles are persisted for reconnection. Both a
+// transport drop and an explicit LOGOFF pass true: a durable handle is owned by the
+// durable scope, not the session, so it survives logoff and stays reconnectable via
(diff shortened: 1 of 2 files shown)


marmos91:
Both review comments addressed in c4272771 (docs-only). Verified no semantic bug: `closeFilesWithFilter` already skips lock release / cache flush / delete-on-close for persisted durable handles (the locks travel with the handle and restore on reconnect). Comments now (1) state the isDisconnect=false contract accurately (TREE_DISCONNECT and non-reconnectable session teardown), and (2) note durable reconnectable handles retain their byte-range locks rather than releasing them at LOGOFF.

Issue #838, opened 2026-05-29
CRITICAL: postgres file_blocks read-index loses hash/store_key — RestoreSnapshot yields unreadable files (data loss)


marmos91:
### Symptom
On a **postgres-metadata** share with a remote (DS3) block store: write files → create snapshot (durable) → mutate → `share disable` → `snapshot restore` → re-enable → **all restored files read as 0 bytes**. Metadata (names/sizes/mtimes) restore correctly; gamma (added-after-snapshot) correctly gone; but every READ fails:

```
[ERROR] legacy zero-hash FileBlock encountered post-migration — refusing read block_id=val-pg-ds3/alpha.bin/0 store_key=
ReadAt error: direct download failed: download block : blockstore: legacy zero-hash FileBlock encountered post-migration
```
(guard: pkg/blockstore/engine/fetch.go:106)

### Root cause (confirmed against the live restored DB)
Two postgres tables hold block info:
- `file_block_refs` (file_id, offset, size, **hash**) — manifest source / FileAttr.Blocks. **All 176 rows have hashes.**
- `file_blocks` (id, hash, block_store_key, cache_path, state, …) — the **read-path index** the engine resolves through. **All 176 rows: `hash`=NULL, `block_store_key`=NULL, `cache_path`=NULL, `state`=0.**

```
file_blocks      : total=176  zero-hash=176   (id=val-pg-ds3/alpha.bin/<offset>, hash NULL, store_key NULL, cache_path NULL, state 0)
file_block_refs  : total=176  zero-hash=0
```

So `file_blocks.hash`/`block_store_key` is **never populated when a block finalizes to remote** on postgres. Normally masked because reads hit the local cache (`cache_path`); **RestoreSnapshot calls ResetLocalState**, clearing the local cache, after which reads must resolve via `file_blocks.hash` → NULL → refused → files unreadable. This is the #789 family (postgres null-hash FileBlock), but on the **read-path index** and exposed by restore — distinct from #789/#815, which fixed `file_block_refs`/FileAttr.Blocks only.

Equivalent to **data loss on restore** for postgres-backed shares (blocks are safe in remote CAS + manifest is correct; only the read index is broken).

### Required
[...]

PR #853, opened 2026-05-30
fix(postgres): production-ready + CI-verified backend (store_id reopen, CreateShare, rollup prev) + byte-verify snapshot matrix

marmos91:
## Summary

Makes the **PostgreSQL metadata backend production-ready and CI-verified** — and adds a deterministic, cross-backend byte-verify proof for the share-snapshot feature. The postgres integration suite was **never run in CI** (no DB/DSN in the workflow), which hid several real defects. This branch fixes them, wires postgres into CI, reconciles the conformance suite, and proves snapshot→mutate→restore is byte-identical on memory, badger, and postgres.

## Real bugs fixed (all were CI-invisible)

1. **`ensureStoreID` store-reopen deadlock (HIGH).** Bootstrap used `UPDATE server_config ... WHERE id=1`, which matched zero rows once `server_config` was truncated — which `Reset` and the pre-COPY phase of `Restore` do. `AutoMigrate` does not re-seed the singleton on reopen, so the store could **never reopen** after a restore/Reset — bricking it and, fatally, preventing the boot-time restore-marker recovery from ever opening the store to run. Now an idempotent, concurrency-safe `INSERT ... ON CONFLICT DO UPDATE` upsert that self-heals the row.

2. **`CreateShare` could never succeed on postgres.** It issued a bare `INSERT INTO shares (...)` that omitted the `NOT NULL root_file_id` FK. Now honours its documented contract (create the share *and* its root directory, like memory/badger) by materializing a default root via the idempotent `CreateRootDirectory`, then persisting options.

3. **`SetRollupOffset` always returned `prev=0` on updates.** The single-statement `INSERT ... ON CONFLICT ... RETURNING (SELECT rollup_offset FROM cte)` could not surface the pre-update value — the CTE and the conflict target are the same row, so the subquery resolves to NULL (verified at the SQL level). Now reads the prior value under a `FOR UPDATE` row lock in an explicit transaction, then conditionally upserts.

## Test/CI hardening

4. **CI now runs postgres.** `integration-tests.yml` gains a `postgres:16` service + `DITTOFS_TEST_POSTGRES_DSN`, and runs with `-p 1` so package binaries sharing the single `dittofs_test` database don't race against the conformance suite's truncations.

5. **Conformance suite reconciled for path-keyed backends.** The shared helpers built `metadata.File` without `Path`; postgres keys files on `(share_name, path)` and collided on the 2nd empty path (memory tolerates it — handle-keyed). Production always sets `Path` (`file_create.go`), so the helpers now derive the correct full path from the parent. Memory + badger conformance remain green.

6. **Byte-verify snapshot matrix.** New `snapshot_byteverify_test.go` (+ postgres companion) drives the **real** engine write path — `WriteToBlockStore → CommitBlockStore → DrainRollups` (Blocks persisted by the coordinator, never injected) — then snapshots, overwrites a file in place, deletes another, adds a third, drops the local cache (`ResetLocalState`, exercising the #838 cold-read path), restores, and asserts **byte-identical** across memory/badger/postgres. Includes a multi-chunk (3 MiB FastCDC) file and a negative-control check so the pass is not vacuous.

## Verification

- Full postgres integration suite: `ok` (conformance + backup + reset-restore + rollup + regression tests), fresh DB, `-p 1`.
- memory + badger conformance, runtime snapshot suite: green, no regressions.
- byte-verify matrix: memory/badger/postgres all byte-identical; `-race -count=2` clean.
- `gofmt`, `go vet` (both tag sets), `golangci-lint`: clean.

## Notes
- The standalone metadata `CreateShare` dup-check is a fast path, not the integrity authority: the `unique_share_path_hash_active` partial index on `files` admits only one root inode per share, so concurrent same-name creates can't orphan an inode (and the control plane serializes share creation upstream).

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -27,6 +27,24 @@ jobs:
       matrix:
         go-version: ["1.25.x"]
 
+    # PostgreSQL service so the metadata-store integration + conformance
+    # suites actually run in CI. The postgres test factory connects to
+    # localhost:5432, database dittofs_test, postgres/postgres.
+    services:
+      postgres:
+        image: postgres:16
+        env:
+          POSTGRES_USER: postgres
+          POSTGRES_PASSWORD: postgres
+          POSTGRES_DB: dittofs_test
+        ports:
+          - 5432:5432
+        options: >-
+          --health-cmd "pg_isready -U postgres"
+          --health-interval 5s
+          --health-timeout 5s
+          --health-retries 10
+
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
@@ -40,8 +58,14 @@ jobs:
       - name: Download dependencies
         run: go mod download
 
+      # -p 1 serializes package test binaries: several packages (metadata
+      # store conformance, blockstore engine drain) share the single
+      # dittofs_test database, and the conformance suite truncates it
+      # between subtests — running their binaries concurrently would race.
       - name: Run all integration tests
-        run: go test -tags=integration -v -timeout=15m ./...
+        env:
+          DITTOFS_TEST_POSTGRES_DSN: "host=localhost port=5432 dbname=dittofs_test user=postgres password=postgres sslmode=disable"
+        run: go test -tags=integration -v -timeout=20m -p 1 ./...
 
       - name: Install and start rpcbind
         run: |
--- a/pkg/apiclient/snapshots_test.go
+++ b/pkg/apiclient/snapshots_test.go
@@ -198,6 +198,44 @@ func TestRestoreSnapshot(t *testing.T) {
 	assert.True(t, sent.AllowNonDurable)
 }
 
+// TestRestoreSnapshot_UsesRestoreTimeoutNotBaseClient is the behavioral
+// regression for #842: a remote-backed restore whose server-side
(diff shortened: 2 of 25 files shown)


marmos91:
Addressed both review comments in 9878353d:

1. **rollup.go (monotone under concurrency)** — correct catch. The `FOR UPDATE` read locks nothing when no row exists, so concurrent first-inserts could both skip the app-layer guard and the unguarded upsert allowed a lower offset to overwrite a higher one. Restored the `ON CONFLICT DO UPDATE ... WHERE rollup_offsets.rollup_offset <= EXCLUDED.rollup_offset` predicate as the INV-03 authority (matching migration 000009) and treat `RowsAffected()==0` as the regression sentinel (read back the stored value, return `ErrRollupOffsetRegression`). The `FOR UPDATE` read is kept solely to report `prev` on success. Rollup conformance suite (incl ConcurrentMonotone + RejectsRegression + AllowsEqualValue) green.

2. **shares.go (CreateShare error/validation)** — now surfaces an unexpected `getExistingRootDirectory` error instead of falling through, and validates the block layout (`ParseBlockLayout`) before materializing the root so an invalid value can't leave a half-created share.

Issue #739, opened 2026-05-29
v1.0: DH2 residuals — lock-lease + app-instance (lease-V2 epoch persistence); persistent-open stays deferred


marmos91:
Residuals after #432 closed. The 32 `smb2.durable-v2-open.*` and `smb2.durable-v2-delay.*` tests remain in KNOWN_FAILURES — DH V2 reopen / lease coordination / disconnected-handle preservation+purge / app-instance / persistent-open / DH2Q blob validation are not yet functional.

Test clusters:
- DH2Q create-blob validation: create-blob
- DH2 open + oplock/lease: open-oplock, open-lease
- DH2 reopen family: reopen1, reopen1a, reopen1a-lease, reopen2, reopen2b, reopen2-lease, reopen2-lease-v2, durable-v2-setinfo
- DH2 lock + oplock/lease: lock-oplock, lock-lease, lock-noW-lease
- DH2 stat + lease: stat-and-lease, nonstat-and-lease, statRH-and-lease
- DH2 two-handle same/different lease: two-same-lease, two-different-lease
- DH2 keep-disconnected: keep-disconnected-rh-with-{stat,rh,rwh}-open, keep-disconnected-rwh-with-stat-open
- DH2 purge-disconnected: purge-disconnected-rwh-with-{rwh,rh}-open, purge-disconnected-rh-with-{share-none-open,write,rename}
- App instance: app-instance
- Persistent handles: persistent-open-oplock, persistent-open-lease
- Delay: durable_v2_reconnect_delay

Tracks v1.0 conformance — KF rows above the Permanently Unimplementable appendix must reference an OPEN sub-issue of #673.

marmos91:
Remaining DH2 residuals deferred past v1.0 (decision 2026-05-31), mirroring #749/#686. Tractable rows landed this milestone: nonstat-and-lease (#903), lock-oplock (#913), keep-disconnected-rh-with-rwh-open + reconnect-delay (#894), reopen1/1a (#877), alloc-size (#887). Deferred with documented reasons (PRs #916, #917): persistent-open (needs CA-share infra), lock-lease (needs lease-v2 epoch persistence across all backends + break suppression), app-instance (spurious break_info.count==1, needs live-debug). Removed from v1.0.0 milestone; rows stay suppressed known-failures (CI green).

marmos91:
DH2 durable-handle residuals complete:
[...]

PR #854, opened 2026-05-30
fix(smb): durable reopen2-family — gate root open on reconnect, restore CreateGuid (#738/#739)

marmos91:
Addresses the durable-handle **reopen2** family across #738 (DH V1) and #739 (DH V2). Root-caused via an end-to-end reproduction that drives the real persist→reconnect handler path (memory metadata + DurableStore + LeaseManager), not an isolated decode test.

## Bugs fixed
1. **Empty-fname reconnect bypassed the durable path.** The root-directory short-circuit in `Create` ran *before* the DHnC/DH2C reconnect block, so a reconnect request with `fname=""` opened the share root and returned SUCCESS instead of `OBJECT_NAME_NOT_FOUND`. `reopen2_lease` / `reopen2_lease_v2` and the V2 reopen2 family send an empty-fname reconnect as a negative case. Fix: gate the short-circuit on the absence of a reconnect context (only when a DurableStore is present).
2. **Restored handle dropped its V2 CreateGuid.** `validateAndRestore` rebuilt the `OpenFile` without copying `CreateGuid`, so a reconnected V2 handle lost its identity → the next disconnect persisted `IsV2=false` with a zero guid → the following DH2C reconnect's by-CreateGuid lookup failed, breaking the multi-cycle reconnect. Fix: restore `CreateGuid` (no-op for V1, which persists 0).

## Expected to flip (CI must confirm)
- V2: `durable-v2-open.reopen2`, `reopen2b`, `reopen2-lease`, `reopen2-lease-v2` (Bug B is the common V2-family defect — may also advance V2 reopen1/reopen1a).
- V1: `durable-open.reopen2-lease`, `reopen2-lease-v2` (empty-fname case via Bug A).

## Not addressed / needs CI
- Plain `smb2.durable-open.reopen2` (V1 oplock): the 4-cycle handler repro PASSES locally, so if it still fails it is a wire-level issue (SPNEGO/error-body/context-encoding) needing the pcap-diff playbook — not a handler defect.
- Lease-state bytes/epoch and the INVALID_PARAMETER-vs-OBJECT_NAME_NOT_FOUND wrong-fname distinction can only be confirmed on the wire.

Could NOT run smbtorture locally (CI-only). KNOWN_FAILURES.md left untouched (two-commit discipline) — walk back rows only after CI logs the specific tests as PASS. No regression to reopen1/reopen1a (they carry no empty fname; V1 handles persist CreateGuid=0). Gates: gofmt, vet, golangci-lint (0 issues), `go test -race` all green.

No file overlap with #850 (reopen4: logoff.go/handler.go).

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -747,8 +747,15 @@ func (h *Handler) Create(ctx *SMBHandlerContext, req *CreateRequest) (*CreateRes
 		fullFilename = filename + streamSuffix
 	}
 
-	// Handle root directory case
-	if filename == "" && streamSuffix == "" {
+	// Handle root directory case. A durable-reconnect request (DHnC/DH2C) that
+	// carries an empty filename must NOT open the share root — it has to fall
+	// through to Step 4b so the reconnect path rejects it with
+	// OBJECT_NAME_NOT_FOUND. smbtorture durable-open.reopen2_lease / _lease_v2
+	// and the V2 reopen2 family send an empty-fname reconnect as a negative case.
+	hasReconnectCtx := h.DurableStore != nil &&
+		(FindCreateContext(req.CreateContexts, DurableHandleV1ReconnectTag) != nil ||
+			FindCreateContext(req.CreateContexts, DurableHandleV2ReconnectTag) != nil)
+	if filename == "" && streamSuffix == "" && !hasReconnectCtx {
 		resp, err := h.handleOpenRootCreate(ctx, req, authCtx, rootHandle, tree)
 		h.storeCreateReplayIfApplicable(ctx, req, resp)
 		return resp, err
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed both review comments in 8a634bb3: (1) softened the root-case comment — it no longer claims a specific status; the empty-fname reconnect now falls through to the reconnect path which validates and rejects a stale/mismatched reconnect (status varies: OBJECT_NAME_NOT_FOUND or INVALID_PARAMETER per the lease/path checks). (2) Added a CreateGuid round-trip assertion to TestProcessDurableReconnectContext_V2Success, locking in the multi-cycle V2 reconnect behavior. Note: the prior run's Windows Build failure and dirlease.v2_request failure are unrelated to this change (platform-neutral Go; dirlease.v2_request passed on a parallel same-base run — a known dirlease flake); this push re-runs full CI.

PR #860, opened 2026-05-30
perf/durability(lock): grace period + reclaim — wire H-1/H7/H15 (PR-B2)

marmos91:
## What

Phase 2 of lock durability. Closes **area-5 H-1**, **NFS H7** (reclaim completes), and **NFS H15** (`NLM4_DENIED_GRACE_PERIOD` reachable).

Phase 1 built the `GracePeriodManager`, `EnterGracePeriod`/`IsOperationAllowed`/`MarkReclaimed`, and per-share `RestoreLocks`, but production constructed the lock manager via the no-grace path — so `gracePeriod` was nil, `IsInGracePeriod` always false, every reclaim rejected, and `NLM4_DENIED_GRACE_PERIOD` unreachable. This PR wires it.

## What got wired

- **Grace construction + enter-on-restart** (`service.go`): each share's lock manager is now built via `NewManagerWithGracePeriod` (`newGraceAwareLockManager`). After recovering persisted locks, the unique set of recovered `ClientID`s becomes the expected-reclaim roster, and grace is entered only when that set is non-empty (a fresh server with no persisted locks starts in normal operation). `onGraceEnd` sweeps any client that did not reclaim within the window (`RemoveClientLocks`). Grace is built and entered on the local manager **before** it is published into `s.lockManagers`, preserving the Phase-1 recover-before-publish + synchronous-persist invariants (a manager is never observable having restored conflicting locks but not yet entered grace).
- **NLM grace gating + status** (`nlm.go`, `nlm/handlers/lock.go`): non-reclaim `LockFileNLM` is gated on `IsOperationAllowed`; the `ErrGracePeriod` sentinel maps to `NLM4_DENIED_GRACE_PERIOD` (previously every error → `NLM4Failed`). Reclaims are allowed during grace and call `MarkReclaimed` on grant.
- **MarkReclaimed wiring** (`reclaim.go`): a successful lease reclaim records the owning client so grace can early-exit once every expected client recovers.
- **NFSv4 grace coordination** (`grace_coordinator.go`, `adapter.go`): a `GraceCoordinator` couples the NFSv4 `StateManager` grace machine to the per-share lock-manager grace machine (enter together, exit together). Registered for shares added after adapter startup, with a catch-up pass for shares already in grace when the adapter starts (boot shares register before the adapter exists).
- **Config** (`start.go`): the operator-configured NFS `grace_period` is threaded into the `MetadataService` before shares register (defaults to 90s, matching the DB default).

## NFSv4 two-machine coordination — honest status

The lock-manager + NLM grace window is **fully live** (the H-1/H7/H15 fix). The NFSv4 coordination **seam is wired end-to-end** (the coordinator drives `StateManager.StartGracePeriod`/`ForceEndGrace`), and a unit test asserts both machines enter together via the coordinator. But NFSv4 grace only meaningfully activates when given the v4 confirmed-client set, and DittoFS does not yet **persist** NFSv4 clients across restart (`SaveClientState`/`LoadPreviousClients` are unwired) — so the v4 expected set is empty and the v4 window is a no-op today even though it is driven. The coupling lands so that once v4 client persistence exists, both machines enter/exit together with no further change. The per-share-vs-global v4 asymmetry (first share in starts v4 grace, first out ends it) is documented in `grace_coordinator.go` as the seam to refcount once v4 grace goes live.

## Tests (TDD, RED→GREEN)

- `TestGracePeriod_NLMLockDenied` — non-reclaim NLM lock during grace → `ErrGracePeriod` → `NLM4_DENIED_GRACE_PERIOD`.
- `TestGracePeriod_NLMReclaimAllowed` — reclaim granted + `MarkReclaimed`.
- `TestGracePeriod_EarlyExitWhenAllReclaim` — both expected clients reclaim → grace exits before timer.
- `TestReclaimLease_MarksClientReclaimed` — lease reclaim marks client + early-exits (the RED→GREEN bug this PR fixed).
- `TestRegisterStoreForShare_EntersGraceWhenPersistedLocksExist` / `_NoGraceWhenNoPersistedLocks` — restart enters grace with the persisted roster; fresh server does not.
- `TestGracePeriod_BothMachinesEnterTogether` — coordinator drives a second grace machine in lockstep.

All Phase-1 storetest/property/storm tests stay green. `go test -race ./pkg/metadata/lock/...` clean.

## Next wave (separate)

Byte-range-map unification (xproto H1/H2) is a **separate** next wave that supersedes the old PR-B3 cross-scan.

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -244,7 +244,7 @@ func runStart(cmd *cobra.Command, args []string) error {
 
 	// Write PID file if specified
 	if pidFile != "" {
-		if err := os.WriteFile(pidFile, []byte(fmt.Sprintf("%d", os.Getpid())), 0644); err != nil {
+		if err := os.WriteFile(pidFile, fmt.Appendf(nil, "%d", os.Getpid()), 0644); err != nil {
 			return fmt.Errorf("failed to write PID file: %w", err)
 		}
 		defer func() { _ = os.Remove(pidFile) }()
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -7,6 +7,7 @@ import (
 	"os"
 	"os/signal"
 	"syscall"
+	"time"
 
 	"github.com/marmos91/dittofs/internal/logger"
 	"github.com/marmos91/dittofs/internal/sysinfo"
@@ -198,6 +199,22 @@ func runStart(cmd *cobra.Command, args []string) error {
 	if cfg.GC.Interval > 0 {
 		logger.Warn("gc.interval is configured but no periodic-GC scheduler ships in v0.15.0 — the value is ignored. Trigger GC on demand via `dfsctl store block gc <share>` (or schedule via cron). Periodic scheduling is tracked for a follow-up phase.",
 			"configured_interval", cfg.GC.Interval)
+	}
+
+	// Thread the operator-configured lock-manager grace period into the
+	// MetadataService BEFORE loading shares: AddShare registers each share's
+	// lock manager and enters the post-restart grace window, so the duration
+	// must be set first. LoadInitial is idempotent (the lifecycle Serve loop
+	// loads settings again); a nil/zero value falls back to the 90s default,
+	// which matches the DB default for grace_period.
+	if sw := rt.GetSettingsWatcher(); sw != nil {
+		if err := sw.LoadInitial(ctx); err != nil {
+			logger.Warn("Failed to load initial adapter settings for lock grace period", "error", err)
+		}
+		if nfsSettings := rt.GetNFSSettings(); nfsSettings != nil && nfsSettings.GracePeriod > 0 {
+			rt.GetMetadataService().SetLockGracePeriod(
+				time.Duration(nfsSettings.GracePeriod) * time.Second)
+		}
 	}
 
 	// Load shares (per-share BlockStores are created during AddShare).
--- a/internal/adapter/nfs/nlm/handlers/lock.go
+++ b/internal/adapter/nfs/nlm/handlers/lock.go
@@ -9,6 +9,7 @@ import (
 	"github.com/marmos91/dittofs/internal/adapter/nfs/nlm/types"
 	nlm_xdr "github.com/marmos91/dittofs/internal/adapter/nfs/nlm/xdr"
 	"github.com/marmos91/dittofs/internal/logger"
+	metaerrors "github.com/marmos91/dittofs/pkg/metadata/errors"
 	"github.com/marmos91/dittofs/pkg/metadata/lock"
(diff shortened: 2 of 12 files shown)


marmos91:
## Adversarial review round (post-implementation)

Ran a high-rigor adversarial review on the grace-coordination diff (same discipline as PR-B1). 5 candidate findings; verified each against the code:

**Fixed (`9b3fd997`):**
- **HIGH — orphaned grace timer on registration-race loss.** The loser of a concurrent same-share `RegisterStoreForShare` dropped its manager but left its armed grace timer running. On fire, `onGraceEnd` swept the *surviving* manager's locks from the shared store (`RemoveClientLocks`) and prematurely ended the NFSv4 grace machine. Phase-1's loser path was harmless (epoch double-bump only); B2 made it harmful. Loser now `AbortGracePeriod()`s (stops timer, no callback). New `AbortGracePeriod` + `TestAbortGracePeriod_StopsTimerWithoutCallback`.
- **HIGH — NLM `ClientID` keyed by transport address.** `ClientID: ctx.ClientAddr` (`host:port`) — the ephemeral port changes when a client reconnects after restart, so the grace reclaim roster and NSM crash cleanup (FREE_ALL carries `caller_name`, per `free_all.go`) never matched → early grace-exit dead + stale-lock cleanup missed. Now keyed by stable `req.Lock.CallerName` (spec-correct, X/Open NLM owner identity). New `TestLock_OwnerClientIDIsCallerName`. Verified `owner.ClientID` is consumed only as an opaque identity key, never as a network address (callback addr uses `ctx.ClientAddr` directly).
- **NIT** — `fmt.Sprintf`→`fmt.Appendf` in `start.go` (lint `fmtappendf`).

**Refuted:**
- `ForceEndGrace` nil-deref — `manager.go:842` already nil-guards `gp`. Not a bug.

**Documented limitations (not fixed):** `grace_period: 0` falls back to 90s default (no explicit-disable sentinel); per-share-vs-global v4 grace asymmetry (already documented in `grace_coordinator.go`, dormant until v4 client persistence lands).

simplifier: clean. `go test -race ./pkg/metadata/lock/...` + affected suites green; vet + golangci-lint clean.

marmos91:
Fixed in `8112775d`. Confirmed real: production order is shares-register (startup) → `SetGraceCoordinator` (NFS adapter `SetRuntime`), so the `onGraceEnd` closure captured a nil coordinator. The start side was already handled by the adapter catch-up (`adapter.go:402`); the end side now reads `s.graceCoordinator` live at grace-end (deadlock-safe — no `MarkReclaimed` caller holds `s.mu`). Registration order no longer matters. Regression: `TestGracePeriod_EndNotifiesCoordinatorInstalledAfterRegistration` (coordinator installed *after* the manager is built, grace-end still notifies).

PR #863, opened 2026-05-30
test(smb/dh): V1-lease reopen2 wire-context coverage — V1-vs-V2 lease asymmetry (#738)

marmos91:
## Summary

Investigates the 2 remaining V1-**lease** durable reopen2 residuals
(`smb2.durable-open.reopen2-lease` / `-lease-v2`, #738) after #861 fixed the
reopen2 access-revalidation gate.

**Finding: the V1-lease reconnect path is already correct in DittoFS's handler
and wire-decode layers.** Every code-reachable layer of the V1-lease
disconnect→reconnect cycle was reproduced and passes. This PR closes the one
coverage gap #861 left — its V1-lease repros used a V2 (52-byte) `RqLs` lease
context, whereas smbtorture's V1 `reopen2-lease` sends a **V1 (32-byte) `RqLs`**
(`smb2_lease_create`) with a V1 `DHnQ`/`DHnC` durable blob.

## The V1-vs-V2 lease asymmetry, resolved

The only wire difference between the failing V1-lease tests and the passing
V2-lease test (`durable-v2-open.reopen2-lease`) is the **durable** layer
(`DHnQ`/`DHnC` keyed on FileID vs `DH2Q`/`DH2C` keyed on CreateGuid) — **not**
the lease context. The V1 reconnect (`processV1Reconnect`) and the shared
`checkLeaseReconnectGate` were verified to handle the lease-backed V1 path
identically to V2: lookup by volatile-zeroed FileID, lease-key + ClientGUID
gate before the path check, non-destructive negative ladder.

## What was verified (all green, real `Create()` + real `DecodeCreateRequest`)

- Initial V1-lease durable open with a **32-byte RqLs** + `DHnQ` → durability
  granted, handle persists with `OplockLevel=Lease` + lease key.
- Cross-connection reconnect (fresh session) via `DHnC` + 32-byte `RqLs` → OK.
- The exact smbtorture negative ladder against ONE persisted handle:
  ONF, ONF, ONF, ONF (wrong key), INVALID_PARAMETER (wrong fname + lease),
  then OK — handle survives every non-destructive negative attempt.
- Wire round-trip: `buildCreateRequestWire` → `DecodeCreateRequest` preserves
  both the `DHnC` (FileID intact) and the V1 `RqLs` (lease key intact) contexts.

## Honest confidence

- **`smb2.durable-open.reopen2-lease`** — high confidence the handler/wire path
  is correct; cannot run smbtorture (no Docker) to flip the KNOWN_FAILURES row.
  Strongly suspect the fix already shipped with #861 and the KF row is stale.
- **`smb2.durable-open.reopen2-lease-v2`** — same; the only delta vs the above
  is the 52-byte `RqLs` shape, which the existing #861 repros + this wire-decode
  test both cover.

If CI smbtorture still reports ONF after this lands, the divergence is below the
handler + wire-decode layers (NEGOTIATE/dispatch dialect framing) — not in
`durable_context.go` / `create.go`, both of which are exercised end-to-end here.

No production code changed — tests only. KNOWN_FAILURES intentionally left for
the orchestrator to walk back after a CI smbtorture run confirms.

Refs #738.

### DIFF
--- a/internal/adapter/smb/v2/handlers/reopen2_v1lease_repro_test.go
+++ b/internal/adapter/smb/v2/handlers/reopen2_v1lease_repro_test.go
@@ -1,0 +1,210 @@
+package handlers
+
+import (
+	"testing"
+
+	"github.com/marmos91/dittofs/internal/adapter/smb/types"
+	"github.com/marmos91/dittofs/pkg/metadata"
+	"github.com/marmos91/dittofs/pkg/metadata/lock"
+)
+
+// TestRepro_V1Lease_V1WireContext mirrors smbtorture smb2.durable-open.reopen2-lease
+// AS FAITHFULLY AS THE HANDLER ALLOWS: the initial open uses a V1 (32-byte) RqLs
+// lease context (smb2_lease_create) + a V1 DHnQ durable request (io.in.durable_open
+// = true), then disconnects and reconnects via DHnC + a V1 (32-byte) RqLs context.
+//
+// This is distinct from the existing TestReopen2_V1Lease_* tests which inject a
+// V2 (52-byte) RqLs context. The goal is to find any handler-reachable divergence
+// caused by the V1 lease wire shape.
+func TestRepro_V1Lease_V1WireContext(t *testing.T) {
+	e := setupReopen2Env(t)
+
+	leaseKey := [16]byte{0xAB, 0xCD, 0xEF, 0x01}
+	rwh := uint32(lock.LeaseStateRead | lock.LeaseStateWrite | lock.LeaseStateHandle)
+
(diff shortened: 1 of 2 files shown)


marmos91:
Closing: tests-only, no production change. The 2 V1-lease tests pass at the handler/wire-decode level here, but #861's CI smbtorture log shows `durable-open.reopen2-lease` still fails on the real wire — so the residual is BELOW the handler (dispatch/dialect framing), not in durable_context.go. These handler-level tests don't represent the actual failure. Revisit the V1-lease residual via a pcap-diff (Samba ref vs DittoFS) when the 2 tests warrant it. KF rows retained.

Issue #749, opened 2026-05-29
v1.0: replay pending-oplock matrix (39 rows) — sane state machine + Windows-policy mode + Windows CI


marmos91:
Untagged replay-protection residuals in KNOWN_FAILURES. Replay protection requires channel-sequence tracking and replay detection across durable handles. Newly reachable after the GMAC signing fix.

Test clusters:
- Basic replay: replay3 (flaky), replay4, replay5, replay6, channel-sequence
- DH V2 replay with oplock/lease: replay-dhv2-oplock2, replay-dhv2-oplock-lease, replay-dhv2-lease1/2/3, replay-dhv2-lease-oplock
- DH V2 pending1n-vs-{violation-lease-close-sane, violation-lease-ack-sane, violation-lease-close-windows, violation-lease-ack-windows, oplock-sane, oplock-windows, lease-sane, lease-windows}
- DH V2 pending1l-vs-{oplock-sane, oplock-windows, lease-sane, lease-windows}
- DH V2 pending1o-vs-{oplock-sane, oplock-windows, lease-sane, lease-windows}
- DH V2 pending2n/2l/2o-vs-{oplock-sane, oplock-windows, lease-sane, lease-windows}
- DH V2 pending3n/3l/3o-vs-{oplock-sane, oplock-windows, lease-sane, lease-windows}

~48 tests total. Heavily intertwined with DH V2 (#741) and multichannel session binding (#747).

Tracks v1.0 conformance — KF rows above the Permanently Unimplementable appendix must reference an OPEN sub-issue of #673.

marmos91:
Deferred past v1.0 (decision 2026-05-31), mirroring the Kerberos deferral (#686). Channel-sequence + create-replay infrastructure shipped (#864/#866/#908). The remaining `dhv2-pending{1,2,3}*` matrix is a deep multi-stage replay-pending-oplock state machine (per-stage ACCESS_DENIED/OK/oplock-echo/data-offset sequencing) — two surgical attempts (#908 foundation, #910 closed) confirmed it needs a correctly-modeled rewrite, not slices. The 39 KF rows remain documented known-failures (CI green); KNOWN_FAILURES.md Policy amended to exempt this block from the v1.0 gate (PR #915). Removed from the v1.0.0 milestone.

marmos91:
@
## Status / scoping (from current `develop`)

[...]

PR #864, opened 2026-05-30
feat(smb/replay): SMB3 channel-sequence verification (#749)

marmos91:
Implements basic SMB2/3 replay protection via **channel-sequence verification** (MS-SMB2 §3.3.5.2.10), the independent (non-durable-handle) subset of #749.

## Mechanism
- New `StatusFileNotAvailable` (0xC0000467).
- Each `OpenFile` tracks `Open.ChannelSequence` (`channelSeq`/`channelSeqSet`, guarded by a dedicated `csMu`).
- `OpenFile.VerifyChannelSequence` mirrors Samba's `smbd_smb2_request_dispatch_update_counts` (`source3/smbd/smb2_server.c`): a newer ChannelSequence advances the tracked value; a **stale** ChannelSequence on a **modifying** op (WRITE/SET_INFO/IOCTL) is rejected with `STATUS_FILE_NOT_AVAILABLE`; read-only ops never reject but still advance. Full-width 16-bit wraparound handling matches Samba's `abs(cmp) > 0x7FFF` sign-flip.
- The check runs centrally in both dispatch paths (`ProcessSingleRequest` and `ProcessRequestWithFileIDAndCallback`), gated to SMB3+ and to commands carrying a FileId at a fixed body offset (READ/WRITE/SET_INFO at 16, IOCTL at 8).
- The `SMB2_FLAGS_REPLAY_OPERATION` flag does not change the decision: in DittoFS each handle's I/O completes before the next request is verified, so Samba's `request_count`/`pre_request_count` gauges are always drained — the decision reduces to the ChannelSequence comparison, which is exactly what the torture tests observe.

## Tests targeted (honest confidence)
| Test | Confidence | Notes |
|------|-----------|-------|
| `smb2.replay.channel-sequence` | **HIGH** | Pure single-channel CSN validation. Unit test encodes the exact Samba `test_channel_sequence_table` and matches every expected status. |
| `smb2.replay.replay4` | **MEDIUM-HIGH** | Needs CSN tracking (this PR) + multi-channel session bind (already works). Unit test reproduces the read-advances / stale-write-rejects core. |
| `smb2.replay.replay5` | **N/A (skips)** | Requires `SMB2_CAP_PERSISTENT_HANDLES` + CA share, which DittoFS does not advertise -> `torture_skip`. |
| `smb2.replay.replay3` | **LOW** | Flaky per the issue; depends on the DH2Q CreateGuid replay cache in the fenced `create.go` path (not touched here). |
| `smb2.replay.replay6` | **OUT OF SCOPE** | Driven by DH2Q create de-dup semantics (`DUPLICATE_OBJECTID` without the replay flag) living in the fenced durable-create handler. |

## Scope / non-goals
- Does **not** touch `create.go` / `durable_*.go` (DH-create path owned by a concurrent PR).
- Does not implement the DH-V2-replay matrix (`replay-dhv2-*`, `pending*`).
- Related-compound ops (FlagRelated, inherited FileId) are intentionally not CSN-checked — they chain off a prior CREATE and use the sentinel FileId.

KNOWN_FAILURES.md is intentionally untouched; the CI smbtorture log is the source of truth for any walkback.

## Verification
- `go build ./...`, `go vet ./internal/adapter/smb/...`, `golangci-lint` — clean.
- `go test -race` on `internal/adapter/smb/...` — green, including the new unit tests.

### DIFF
--- a/internal/adapter/smb/channel_sequence.go
+++ b/internal/adapter/smb/channel_sequence.go
@@ -1,0 +1,80 @@
+package smb
+
+import (
+	"github.com/marmos91/dittofs/internal/adapter/smb/header"
+	"github.com/marmos91/dittofs/internal/adapter/smb/types"
+)
+
+// channelSeqModify reports whether a command is a modifying operation for the
+// purposes of SMB3 channel-sequence verification (MS-SMB2 §3.3.5.2.10). Only
+// WRITE, SET_INFO, and IOCTL are rejected on a stale ChannelSequence; all
+// other commands are read-only with respect to this check (Samba marks the
+// same three with .modify = true in source3/smbd/smb2_server.c).
+func channelSeqModify(cmd types.Command) bool {
+	switch cmd {
+	case types.CommandWrite, types.CommandSetInfo, types.CommandIoctl:
+		return true
+	default:
+		return false
+	}
+}
+
+// channelSeqFileID extracts the 16-byte FileId from a request body for the
+// handle-scoped commands that participate in channel-sequence verification.
+// The body slice excludes the 64-byte SMB2 header. Returns ok=false for
+// commands that do not carry a FileId at a fixed offset (the check is skipped
+// for those). FileId offsets per MS-SMB2: READ §2.2.19 / WRITE §2.2.21 /
+// SET_INFO §2.2.39 all place it 16 bytes into the body; IOCTL §2.2.31 places
+// it 8 bytes in (after StructureSize, Reserved, CtlCode).
+func channelSeqFileID(cmd types.Command, body []byte) (fileID [16]byte, ok bool) {
+	var off int
+	switch cmd {
+	case types.CommandRead, types.CommandWrite, types.CommandSetInfo:
+		off = 16
+	case types.CommandIoctl:
+		off = 8
+	default:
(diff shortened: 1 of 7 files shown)


marmos91:
Holding open (not merging yet). The channel-sequence verification here is spec-correct (MS-SMB2 §3.3.5.2.10, Samba-aligned) and regression-free (New failures: 0), but it flips 0 smbtorture tests: the targeted `channel-sequence`/`replay4`/`replay5`/`replay6` test names are NOT in the current Samba smbtorture binary (the replay suite runs 37 tests — `replay-commands`/`replay-regular`/`replay-dhv2-*`/`dhv2-pending*`). Per less-is-more we won't add hot-path surface that flips nothing on its own. This branch is the FOUNDATION for the dhv2-replay matrix work (the 37 running tests need channel-seq + a DH2Q replay cache + durable-create de-dup together); the follow-up will build on this branch and land them as one flipping PR.

marmos91:
Superseded by #866, which builds on this channel-sequence foundation and lands it together with the durable-V2 replay cache — flipping the dhv2 replay matrix (CI-confirmed). Channel-sequence machinery merged via #866.

Issue #832, opened 2026-05-29
Snapshot/metadata: truncate prunes FileAttr.Blocks but not the FileBlock index (deferred space reclaim)


marmos91:
## Severity: LOW — cleanup (snapshot audit follow-up, surfaced during #817 review)

#817 (PR #823) made size-down truncate (`SetFileAttributes`) prune stale-tail refs from `FileAttr.Blocks` so the snapshot manifest covers only [0,size). It does **not** prune the corresponding **FileBlock index** (`file_block_refs`) rows for the truncated tail, so blocks past the new EOF are reclaimed only if/when GC collects them — matching the existing `RemoveFile` posture (orphan-not-deleted, GC-deferred).

### Required
- Confirm mark-sweep GC (`pkg/blockstore/engine/gc.go`, live-set = `EnumerateFileBlocks`) eventually reclaims blocks orphaned by a truncate, on all backends. If not, the truncated-tail FileBlock rows leak space permanently.
- If not reclaimed, prune the FileBlock index on truncate (engine-seam, dedup-safe per-hash) or emit a block Delete for the dropped tail.

Snapshot-manifest correctness is already fixed by #823; this is the deeper space-reclamation concern.

marmos91:
Fixed in #868 (squash-merged to develop as 54ac550a). #832 fix went beyond the ticket — GC now reclaims RefCount-0 rows on delete+truncate (data-loss-safe, reaped once per distinct hash). #831 streams badger restore. #830 documents the memory-engine ceiling.

Issue #831, opened 2026-05-29
Snapshot: badger Restore buffers all KV entries before integrity CRC (scale ceiling)


marmos91:
## Severity: LOW/MEDIUM — scale hardening (snapshot audit follow-up, surfaced by #814 benchmarks)

The badger Restore path buffers all KV entries before running its integrity CRC, rather than streaming + verifying incrementally. At M-file / TB scale this is a resident-RAM ceiling on the restore side (the create-path dump already streams KV-by-KV on badger).

### Required
- Stream the restore-side KV apply + integrity check (io.Reader/Writer + bounded buffer) so restore RAM is bounded independent of share size, or document the ceiling.

Found during the snapshot e2e audit (#814 benchmark investigation).

marmos91:
Fixed in #868 (squash-merged to develop as 54ac550a). #832 fix went beyond the ticket — GC now reclaims RefCount-0 rows on delete+truncate (data-loss-safe, reaped once per distinct hash). #831 streams badger restore. #830 documents the memory-engine ceiling.

PR #868, opened 2026-05-31
Snapshot scale ceilings: GC reclaim on truncate/delete + streaming badger restore (#830/#831/#832)

### LINKED ISSUE #830: Snapshot: memory metadata engine buffers entire snapshot in one gob buffer during Backup (OOM at scale)
## Severity: MEDIUM — scale hardening (snapshot audit follow-up, surfaced by #814 benchmarks)

The **memory** metadata engine's `Backup` gob-encodes its entire snapshot into a single in-RAM buffer rather than streaming KV-by-KV. The badger engine streams (create-path RAM bounded by the resident HashSet, ~25 MB/1M unique blocks); the memory engine does not, so create on a large share (millions of files / TB) can OOM.

Documented in `test/e2e/BENCHMARKS.md` + `bench/snapshots/` (#814). 1e6×8 → 672 MB dump.

### Required
- Stream the memory engine's Backup dump, or document the memory-engine share-size ceiling as unsupported-at-scale.
- Recommend badger for large-share snapshot deployments in docs.

marmos91:
Closes #830, #831, #832 — snapshot-audit scale-hardening follow-ups.

## #832 — GC now reclaims truncated/deleted remote chunks (the real fix)
Root cause (broader than the ticket): remote chunks are reclaimed only by the mark-sweep GC, whose live-set is every hash emitted by `EnumerateFileBlocks` (the FileBlock index). But **nothing in production ever deleted a RefCount-0 index row** (`DeleteWithRefCount` is dead, `Syncer.Delete/Truncate` are no-ops, `AuditRefcounts` is read-only) and `EnumerateFileBlocks` emits all rows — so GC never reclaimed deleted *or* truncated chunks. Additionally no truncate path decremented refcounts at all (CREATE-O_TRUNC passed `nil` blocks; SETATTR never called `blockStore.Truncate`).

Snapshot **correctness** was already fine (#823 trims `FileAttr.Blocks`); this is purely deferred space reclaim.

Fix (data-loss-safe — only reaps at RefCount==0, atomically):
- New `DecrementRefCountAndReap` on `blockstore.FileBlockStore`: decrement and, iff the new count is 0, delete the row + hash-index entry **in the same critical section** (TOCTOU-free against `AddRef`, like the existing increment contract). Implemented atomically in memory (mutex), badger (single `Update` txn), postgres (`UPDATE … RETURNING` + conditional `DELETE … WHERE ref_count=0` in one tx).
- Engine `Truncate`/`Delete` reclaim sites call the reaping variant → hash leaves the live-set → the existing, tested GC sweep reclaims the chunk. `EnumerateFileBlocks` semantics unchanged (no GC-by-drifty-refcount risk).
- Real pre-truncate `FileAttr.Blocks` are now threaded into `blockStore.Truncate` from NFS v3 CREATE + v3/v4 SETATTR (size-down only, captured before the metadata prune).

Proof: `TestGCMarkSweep_TruncateReclaimsRemoteChunk` (dropped chunk swept, kept chunk retained) + `TestGCMarkSweep_TruncateDedupSafety` (hash shared by another live file is NOT swept). Verified fail-on-develop / pass-with-fix. New `DecrementRefCountAndReap` conformance subtests pass on memory/badger/postgres.

## #831 — badger restore streams (bounded RAM)
Badger `Restore` buffered every KV entry before the integrity CRC. Now streams entries via a bounded `WriteBatch`, verifies the CRC last, and `DropAll`s on any failure. Since restore requires an empty destination, wiping partial data preserves the corrupt→empty/retryable contract. Conformance now asserts a failed corrupt restore leaves the store empty and a subsequent valid restore succeeds — **negative-control verified** (without `DropAll`, badger fails with `ErrRestoreDestinationNotEmpty`).

## #830 — document the memory-engine ceiling
The memory metadata engine is resident-by-design and serializes the whole snapshot into one buffer on create — inherent, not tunable. Documented the ceiling in `docs/SNAPSHOTS.md` and recommend the badger engine (streams create + bounded restore) for TB / M-file shares; refreshed `BENCHMARKS.md` now that badger restore streams too.

## Validation
- build / `go vet` / `gofmt` / `golangci-lint` (0 issues) clean; `-race` clean on engine + coordinator.
- Independent code review: no high-confidence issues; data-loss safety of the reap confirmed across all 3 backends.
- The #832 fix was also validated live end-to-end against real Cubbit/DS3 over an NFS mount (write → delete one file → durable snapshot `remote_durable=true` → mutate → restore → byte-identical).

### DIFF
--- a/cmd/dfs/commands/migrate_to_cas.go
+++ b/cmd/dfs/commands/migrate_to_cas.go
@@ -384,6 +384,9 @@ func (nopFileBlockStore) IncrementRefCount(_ context.Context, _ string) error { 
 func (nopFileBlockStore) DecrementRefCount(_ context.Context, _ string) (uint32, error) {
 	return 0, nil
 }
+func (nopFileBlockStore) DecrementRefCountAndReap(_ context.Context, _ string) (uint32, error) {
+	return 0, nil
+}
 func (nopFileBlockStore) AddRef(_ context.Context, _ blockstore.ContentHash, _ string, _ blockstore.BlockRef) error {
 	// Migration path doesn't exercise the LRU hit path, so always
 	// returning ErrUnknownHash is safe — production callers fall back to
--- a/docs/SNAPSHOTS.md
+++ b/docs/SNAPSHOTS.md
@@ -54,6 +54,22 @@ namespace. The trade-off is that snapshots are intra-cluster — they
 protect against accidental writes or deletes, not against losing the
 underlying storage.
 
+### Metadata backend at scale
+
+Snapshot create and restore stream the metadata dump backend-by-backend.
+**Use the badger metadata engine for large (TB / millions-of-files)
+shares:** badger streams the dump KV-by-KV on create and applies it via
+bounded `WriteBatch` on restore, so its snapshot RAM is governed by the
+resident hash manifest (~25 MB per 1 M unique blocks), not by share size.
+
+The **memory metadata engine is for development and small shares only.**
+It holds the entire filesystem resident by design and serializes the
+whole snapshot into a single buffer during create, so snapshotting a
+multi-GB memory-engine share can exhaust RAM. This is an inherent
+property of an in-RAM store, not a tunable; pick badger before a share
+grows large. See `test/e2e/BENCHMARKS.md` for measured dump sizes and
+the per-backend RAM budget.
+
 ## 2. Snapshot model
 
 Each snapshot is three artifacts on disk:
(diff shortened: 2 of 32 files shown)


marmos91:
Good catch by Copilot on both threads — these were a real data-loss path: RefCount is incremented once per *distinct* hash per file (file-level dedup seen-set in `applyFileLevelDedupHit`/`CopyPayload`), but the reap loops decremented once per `BlockRef`. Fixed in 2f2474f4: `Delete` reaps once per distinct hash; `Truncate` builds kept/dropped hash sets and reaps each distinct dropped hash only when not still kept. Added two negative-control-verified guards: `TestGCMarkSweep_DeleteDuplicateHashNoOverReap` and `TestGCMarkSweep_TruncateStraddleHashNoOverReap`.

PR #873, opened 2026-05-31
test(v1.0): cross-backend conformance coverage for uncovered MetadataStore methods (area-6 PR-B1)

marmos91:
## H1 — uncovered MetadataStore surface

A meaningful subset of the `MetadataStore` interface had **zero cross-backend conformance coverage** — the same CI-blind class that hid the #853 postgres bugs. Each backend (memory / badger / postgres) implemented these methods independently, so divergence or regression could pass the full suite green.

This PR adds a shared `StoreSurface` scenario set, wired into `RunConformanceSuite`, that runs **identically against all three backends** via the factory.

## Scenarios added (`pkg/metadata/storetest/store_surface.go`)

1. **DeleteShare** — share + files + nested dir removed; `ListShares` excludes it, `GetRootHandle`/`GetFile(root)` return not-found, same-name re-create succeeds, missing-share delete returns not-found.
2. **DuplicateCreateShare** — second `CreateShare` with the same name returns `StoreError{Code: ErrAlreadyExists}`.
3. **GetUsedBytes** — deterministic create→grow(1000)→dir(no-op)→truncate-down(250)→delete sequence agrees on every backend; directories never count.
4. **GetFileByPayloadID** — known PayloadID round-trips; unknown id returns not-found.
5. **FilesystemMeta/Stats/Caps** — `GetFilesystemMeta` returns sane caps; `GetFilesystemStatistics` reports non-zero totals; `SetFilesystemCapabilities`→`GetFilesystemCapabilities` round-trip.
6. **ServerConfig** — `SetServerConfig`→`GetServerConfig` round-trip; fresh store readable.
7. **Pagination** — N>limit children paged with a small limit threading `nextCursor`; union equals full set with no dups/no missing; `limit==0` selects the default page size. Backend-divergent (postgres keyset vs badger iterator-prefix vs memory map).
8. **Healthcheck** — live store returns `StatusHealthy`.

## Bugs fixed-forward (surfaced by the new suite)

- **postgres `DeleteShare` orphaned file rows.** It dropped only the `shares` row, orphaning every `files`/`parent_child_map`/`link_counts`/`file_block_refs` row and leaving the root inode occupying the unique root-path index — so same-name recreation failed with `ErrAlreadyExists`. Now drops the share row then all its `files` rows atomically in one tx (FK-cascaded children), and decrements `usedBytes` by the freed regular-file bytes. Applied to both pool-path (`shares.go`) and tx-path (`transaction.go`).
- **badger `DeleteShare` orphaned per-file keys.** Same class of bug; now tears down all per-file keys (file/parent/linkcount/objectID + directory child entries) via a new shared `deleteFileKeys` helper that `DeleteFile` also uses.
- **postgres `GetServerConfig` returned `ErrNotFound` on a fresh store**, where memory and badger return an empty config. Both pool- and tx-paths now treat no-rows as an empty config.

## CI wiring

The in-process badger conformance entrypoint (`badger_conformance_test.go`) uses only `t.TempDir()` (no external service), so its `//go:build integration` tag is **dropped** — it now also runs under default `go test ./...`. Infra-dependent badger tests keep the tag. postgres conformance already runs in CI via `-tags=integration -p 1` with the postgres service DSN.

## Testing

- Ran locally: **memory** and **badger** full conformance suites green (`go test -race`, both default and `-tags=integration`).
- **postgres**: no local DSN, so reasoned through each scenario against the postgres implementation (content_id_hash trigger, usedBytes delta tracking, keyset pagination default 1000, capabilities seeded at 255, `getExistingRootDirectory` dup-guard, the DeleteShare cascade fix); CI's postgres service confirms.
- `gofmt -s`, `go vet`, and `golangci-lint` clean on all touched packages.

## Known follow-up (out of scope)

memory `DeleteShare` does not decrement `usedBytes` when it removes files (pre-existing; badger/postgres now do). The suite does not assert usedBytes-after-DeleteShare so all three pass; worth a small follow-up for parity.

### DIFF
--- a/pkg/metadata/store/badger/badger_conformance_test.go
+++ b/pkg/metadata/store/badger/badger_conformance_test.go
@@ -1,5 +1,3 @@
-//go:build integration
-
 package badger_test
 
 import (
@@ -26,7 +24,9 @@ func TestConformance(t *testing.T) {
 			t.Fatalf("NewBadgerMetadataStoreWithDefaults() failed: %v", err)
 		}
 		t.Cleanup(func() {
-			store.Close()
+			if err := store.Close(); err != nil {
+				t.Errorf("store.Close() failed: %v", err)
+			}
 		})
 		return store
 	})
@@ -40,7 +40,9 @@ func TestBackupConformance(t *testing.T) {
 			t.Fatalf("NewBadgerMetadataStoreWithDefaults() failed: %v", err)
 		}
 		t.Cleanup(func() {
-			store.Close()
+			if err := store.Close(); err != nil {
+				t.Errorf("store.Close() failed: %v", err)
+			}
 		})
 		return store
 	})
@@ -54,7 +56,9 @@ func TestResetThenRestoreConformance(t *testing.T) {
 			t.Fatalf("NewBadgerMetadataStoreWithDefaults() failed: %v", err)
 		}
 		t.Cleanup(func() {
-			store.Close()
+			if err := store.Close(); err != nil {
+				t.Errorf("store.Close() failed: %v", err)
+			}
 		})
 		return store
 	})
@@ -68,7 +72,9 @@ func TestLockPersistenceConformance(t *testing.T) {
 			t.Fatalf("NewBadgerMetadataStoreWithDefaults() failed: %v", err)
 		}
 		t.Cleanup(func() {
-			store.Close()
+			if err := store.Close(); err != nil {
+				t.Errorf("store.Close() failed: %v", err)
+			}
 		})
 		return store
 	})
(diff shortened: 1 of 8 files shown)


marmos91:
## Adversarial review round (post-implementation)

High-rigor review of the diff (the new conformance scenarios EXPOSED real store bugs, fixed-forward — so the store mutations are the risk). 5 findings, all addressed in `d5d6703c`:

- **HIGH — badger tx-path `DeleteShare` orphaned all file metadata.** The pool path was fixed but `badgerTransaction.DeleteShare` (an independent impl) only deleted the share key → every file/parent/linkcount/child/objectID key orphaned. Same fix-one-path-miss-the-other class as PR-B2's coordinator. Extracted a shared `deleteShareFiles` helper so both paths tear down identically; added `DeleteShareViaTransaction` conformance scenario (would have caught it — F5).
- **MED — postgres tx `DeleteShare` swallowed the usedBytes SUM scan error** (`_ =`): a failed scan would delete the files but silently never decrement the counter (statfs drift for process lifetime). Now checked + mapped.
- **LOW — postgres `GetServerConfig` returned a nil `CustomSettings` map** (pool + tx) while badger returns an empty map → caller nil-map-panic divergence. Normalized to non-nil both paths.

**Deferred (documented):** the inline `usedBytes` decrement in DeleteShare shares the known tx-retry double-count class with PutFile/DeleteFile — the unified post-commit fix belongs to the metadata-tx-integrity PR (next in the §7 sequence). Statfs-only, needs a serialization retry. Commented inline in both backends.

Validated on memory + badger (default + `-tags=integration`), `go test -race ./pkg/metadata/...` clean, vet + golangci-lint clean. Postgres path relies on CI's service DSN.

PR #881, opened 2026-05-31
test(smb): walk back #792 alloc-size KF row (CI-confirmed)

### LINKED ISSUE #792: SMB: durable-open.alloc-size — out.alloc_size returned 0 on CREATE with in.alloc_size

marmos91:
Removes the `smb2.durable-open.alloc-size` KNOWN_FAILURES row now that PR #875 (#792) landed the CREATE allocation-size fix.

Flip-proof: the smbtorture verbose step log was unreadable for #875's run (GitHub `BlobNotFound`), so this walkback PR's CI is the definitive confirmation — if it stays green with New failures: 0, the test flipped; if it reddens, the row gets reverted.

Closes #792 on merge.

### DIFF
--- a/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
+++ b/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
@@ -191,7 +191,6 @@ to the DH state machine — tracked under #792 / #793.
 |-----------|----------|--------|-------|
 | smb2.durable-open.reopen2-lease | Durable handles V1 | Durable reopen with lease not fully working | #738 |
 | smb2.durable-open.reopen2-lease-v2 | Durable handles V1 | Durable reopen with lease V2 not fully working | #738 |
-| smb2.durable-open.alloc-size | CREATE allocation | Pre-existing non-DH bug: out.alloc_size=0 on CREATE with in.alloc_size set (fails before any reconnect) | #792 |
 
 ### Durable Handles V2 (Fix Candidate)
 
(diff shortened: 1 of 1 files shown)


marmos91:
Walkback reverted: removing the KF row reddened CI — `smb2.durable-open.alloc-size` STILL fails at the wire (`durable_open.c:2605: out.alloc_size was 0, expected 4096`). #875 fixed the normal CREATE + post-break paths but NOT the durable-create path, which the test exercises. Classic green-CI≠flip: #875's unit tests passed but the smbtorture verbose log was unreadable (BlobNotFound) so the flip was never confirmed. Keeping the KF row; #792 stays open for a durable-create alloc-size follow-up.

PR #885, opened 2026-05-31
fix(nfs/v4): enforce share reservations + stateid share-access (area-4 H2/H3/H19)

marmos91:
Area-4 NFSv4 PR-B2. Three interlocking share-access / share-reservation enforcement gaps, all in `internal/adapter/nfs/v4`.

## H2 — `share_deny` never enforced across open-owners
`NFS4ERR_SHARE_DENIED` was defined but had **zero call sites**; OPEN only OR-merged share bits for the *same* owner and never scanned for a conflicting deny held by another owner. DENY_* was a silent no-op — concurrent writers could corrupt a file an owner believed reserved, and cross-protocol SMB/NFS deny semantics were broken.

**Fix:** `StateManager.OpenFile` now calls `shareConflictLocked` under `sm.mu` before creating/merging state. A conflict exists when `requestedAccess & existingDeny != 0` or `requestedDeny & existingAccess != 0` for an open held by a *different* open-owner on the same file → `NFS4ERR_SHARE_DENIED`. Reclaim (`CLAIM_PREVIOUS`) is exempt. The OPEN handler also validates `share_access ∈ {READ,WRITE,BOTH}` and `share_deny ⊆ {READ,WRITE}` at decode → `NFS4ERR_INVAL`.

**Cross-check:** RFC 7530 §9.7 / §16.16.5; Linux nfsd `nfs4_share_conflict` / `test_share`.

## H3 — all-ones (READ-bypass) special stateid accepted on WRITE/SETATTR
`ValidateStateid` short-circuited *any* special stateid to `(nil, nil)`, so a `nil` openState skipped the `SHARE_ACCESS_WRITE` check and a client could write/truncate with the READ-only all-ones stateid, bypassing share-mode and byte-range locks.

**Fix:** `ValidateStateid` now takes a `StateidOp` (read vs write family). The all-ones READ-bypass stateid is permitted only on READ and rejected with `NFS4ERR_BAD_STATEID` on write-family ops; the all-zeros anonymous stateid stays valid on both. WRITE passes `StateidOpWrite`; SETATTR rejects the read-bypass stateid when a size change is requested. LOCK already looks up real open stateids in `openStateByOther`, so special stateids fail there naturally.

**Cross-check:** RFC 7530 §9.1.4.3.

## H19 — v4 READ discarded openState → no `SHARE_ACCESS_READ` check
`read.go` threw away the `ValidateStateid` openState, so a write-only open could READ via its stateid.

**Fix:** READ now captures the openState and returns `NFS4ERR_OPENMODE` when a regular open stateid lacks `OPEN4_SHARE_ACCESS_READ`. Special stateids (nil openState) bypass the check, as required.

**Cross-check:** RFC 7530 §16.25.4 (symmetric to the WRITE `OPENMODE` guard).

## Tests
State layer (`v4/state/shareres_test.go`): cross-owner deny→`SHARE_DENIED`; requested-deny-excludes-existing-access; non-conflicting combos succeed; same-owner accumulation never self-conflicts; reservation released after CLOSE; different files don't conflict; read-bypass allowed on READ / rejected on WRITE; anonymous allowed on both.

Handler layer (`v4/handlers/shareres_test.go`): WRITE/SETATTR(size) with all-ones → `BAD_STATEID`; READ with all-ones → OK; WRITE with anonymous → OK; write-only open READ → `OPENMODE`; read-write open READ → OK; bad `share_access`/`share_deny` → `INVAL`; end-to-end cross-owner DENY_WRITE blocks a second writer.

`go test -race ./internal/adapter/nfs/...` green; `gofmt -s` / `go vet` / `golangci-lint` clean. Scoped to `internal/adapter/nfs/v4` only.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/open.go
+++ b/internal/adapter/nfs/v4/handlers/open.go
@@ -78,6 +78,24 @@ func (h *Handler) handleOpen(ctx *types.CompoundContext, reader io.Reader) *type
 	}
 	logger.Debug("NFSv4 OPEN share", "access", shareAccess, "deny", shareDeny)
 
+	// Validate the share_access / share_deny modes (RFC 7530 Section 16.16).
+	// The access mode (low 2 bits) must be exactly READ, WRITE, or BOTH; the
+	// deny mode must be a subset of {READ, WRITE}. Reject anything else with
+	// NFS4ERR_INVAL before touching state. The server ignores any higher
+	// share_access bits (e.g. the v4.1 want-delegation hints), so the access
+	// mode is masked to OPEN4_SHARE_ACCESS_BOTH and the masked value carried
+	// forward consistently with the rest of the handler.
+	accessMode := shareAccess & uint32(types.OPEN4_SHARE_ACCESS_BOTH)
+	if accessMode == 0 {
+		logger.Debug("NFSv4 OPEN invalid share_access", "access", shareAccess, "client", ctx.ClientAddr)
+		return openError(types.NFS4ERR_INVAL)
+	}
+	if shareDeny&^uint32(types.OPEN4_SHARE_DENY_BOTH) != 0 {
+		logger.Debug("NFSv4 OPEN invalid share_deny", "deny", shareDeny, "client", ctx.ClientAddr)
+		return openError(types.NFS4ERR_INVAL)
+	}
+	shareAccess = accessMode
+
 	// 4. open_owner4: clientid (uint64) + owner (XDR opaque)
 	clientID, err := xdr.DecodeUint64(reader)
 	if err != nil {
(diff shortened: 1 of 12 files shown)


marmos91:
Fixed in `ccfad166`. Confirmed real: the share-conflict scan ran before `owner.ValidateSeqID`, so an established owner's replayed OPEN (TCP retransmit) could return `NFS4ERR_SHARE_DENIED` instead of its cached result, and a bad seqid returned SHARE_DENIED instead of BAD_SEQID — RFC 7530 OPEN sequencing requires replay/seqid precedence. Moved the scan to after the seqid/replay gate (replay returns cached, bad seqid → BAD_SEQID); new owners have no replay history so the check still applies. v4 state/handler tests + -race green.

PR #888, opened 2026-05-31
fix(metadata): transaction atomicity — RemoveFile TOCTOU, Move rollback, badger txn reads, memory rollback, usedBytes retry (area-6 PR-B3)

marmos91:
Area-6 metadata PR-B3 — the four transaction-atomicity MEDs from `.planning/v1.0-audit/6-metadata/REVIEW.md` §4 plus the deferred usedBytes-retry double-count. Largest area-6 PR (5 grouped fixes); all are transaction-atomicity correctness, so they ship together. Scope: `pkg/metadata` only.

## Fixes

1. **RemoveFile link-count TOCTOU** (`file_remove.go`). The link count was read via `store.GetLinkCount` BEFORE `WithTransaction`; the tx then branched on / wrote the stale value. A concurrent `CreateHardLink` (atomic Get+Set in its own tx) could bump the count in the window → RemoveFile sets `nlink=0` while a valid link still references the file (data loss), and because the count was never read inside the tx, badger's conflict detection had no key to trip. Fix: read via `tx.GetLinkCount` inside the closure so read+decision+write are atomic and the key participates in conflict detection.

2. **Move swallows tx-critical errors → partial-commit rename** (`file_modify.go`). `tx.SetParent`, the dir link-count `SetLinkCount`s, `tx.PutFile(srcFile/srcDir/dstDir)` were discarded with `_ =` and `updateDescendantPaths` errors logged-and-ignored, so the closure returned nil unconditionally and committed regardless. A failed `PutFile(srcFile)` left the entry relinked with the OLD Path (breaks the postgres path-keyed snapshot/restore + descendant rewrite; dir moves skew parent nlink). Fix: return the error from every tx call (including the destination-unlink path and `updateDescendantPaths`) so any failure rolls the whole rename back.

3. **badger tx list/enumerate read a fresh snapshot instead of the active txn** (`store/badger/objects.go`). `tx.ListPending`/`tx.ListFileBlocks`/`tx.EnumerateFileBlocks` delegated to the store path, each opening its own `db.View` snapshot → a `tx.Put` then `tx.ListFileBlocks` in the same `WithTransaction` missed the uncommitted write (cross-backend divergence vs memory). Fix: extracted `listPendingTxn`/`listFileBlocksTxn`/`enumerateFileBlocksTxn` helpers taking `*badger.Txn`; both store-level (`db.View`) and tx-level methods share one implementation, with the tx versions binding to `tx.txn` for read-after-write.

4. **memory WithTransaction had no rollback buffer** (`store/memory/transaction.go`). The memory tx mutated live maps directly, so a returned error left partial mutations — and the conformance suite codified this by string-matching `*memory.MemoryMetadataStore` and relaxing `testTx_IncrementRefCount_RollsBack`. Fix: snapshot/restore buffer over all file/dir/share/fileblock maps (`maps.Clone` + nested-map / FileBlock-value copies), restored on error; removed the type-name carve-out so memory is held to the same all-or-nothing contract. Also fixed a re-lock deadlock in `memoryTransaction.EnumerateFileBlocks` (was delegating to the RLock-taking public method while the write lock was held) and the lazy-`fileBlockData` restore edge (don't leave it non-nil with nil maps).

5. **usedBytes tx-retry double-count** (postgres + badger + memory). A serialization/conflict retry re-ran the closure and re-applied the in-memory usedBytes delta. Fix: accumulate a per-tx `pendingDelta` and apply `usedBytes.Add` ONCE after a successful commit; `PutFile`/`DeleteFile`/`DeleteShare` on all three backends record into pendingDelta instead of mutating the counter inside the closure (badger `deleteShareFiles` now returns freedBytes for both pool- and tx-paths). Removed the "deferred to metadata-tx-integrity" comments.

## Tests
- `TestRemoveFile_HardLinkSurvives` + `TestRemoveFile_ConcurrentCreateHardLink` (200× under -race): a concurrently-linked file is never dropped to nlink=0 while links exist.
- `TestMove_RollsBackOnPutFileFailure`: fault-injecting tx fails `PutFile(srcFile)` → assert full rollback (source stays at old name/path, dest not created).
- `Tx_ListReadAfterWrite` conformance scenario: tx.Put → tx.ListFileBlocks/EnumerateFileBlocks see the uncommitted write (all backends).
- `Tx_IncrementRefCount_RollsBack` carve-out removed → memory now passes the unconditional rollback assertion.
- `TestCounter_RolledBackTxDoesNotDrift` (memory) + `TestBadger_UsedBytes_RetryNoDoubleCount` (real ErrConflict retries): delta applied exactly once.
- `TestWithTransaction_RollbackReinitsLazyFileBlockData`: lazy-fileBlockData rollback no longer panics a later Put.

## Gates
- `gofmt -s` / `go vet` / `golangci-lint run` clean.
- `go test -race ./pkg/metadata/...` green; `go test -race -tags=integration ./pkg/metadata/store/badger/...` green.
- Postgres conformance not run locally (no DSN); the pendingDelta change mirrors badger/memory and is exercised by the shared suite in CI (pg wired in #853).

### DIFF
--- a/pkg/metadata/file_modify.go
+++ b/pkg/metadata/file_modify.go
@@ -566,14 +566,18 @@ func (s *MetadataService) Move(ctx *AuthContext, fromDir FileHandle, fromName st
 				// POSIX: ctime must be updated when link count changes
 				linkCount, _ := tx.GetLinkCount(ctx.Context, dstHandle)
 				now := time.Now()
-				if linkCount <= 1 {
-					_ = tx.SetLinkCount(ctx.Context, dstHandle, 0)
-				} else {
-					_ = tx.SetLinkCount(ctx.Context, dstHandle, linkCount-1)
+				newCount := uint32(0)
+				if linkCount > 1 {
+					newCount = linkCount - 1
+				}
+				if err := tx.SetLinkCount(ctx.Context, dstHandle, newCount); err != nil {
+					return err
 				}
 				// Update ctime on the file being unlinked (affects remaining hard links)
 				dstFile.Ctime = now
-				_ = tx.PutFile(ctx.Context, dstFile)
+				if err := tx.PutFile(ctx.Context, dstFile); err != nil {
+					return err
+				}
 			}
 
 			// Remove destination from children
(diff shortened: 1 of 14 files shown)


marmos91:
Round-2 review fixes pushed (`6217906b`): propagated Move's 3 swallowed GetLinkCount read errors; deep-copy shares in the memory rollback snapshot (UpdateShareOptions mutates *shareData in place, so the shallow map clone aliased it). The memory tx lockStore-not-rolled-back point is a documented area-5 boundary — lock persistence runs in its own transactions, never mixed with file-metadata mutations in one WithTransaction, so no production path commits a half-rolled-back lock; pre-existing, not a regression. Postgres tx read-after-write also fixed (`c7fa4811`) — the new conformance scenario caught it on the pg Integration job.

PR #890, opened 2026-05-31
fix(nfs/v4): honor replay caches across owner-seqid ops + v4.1 SEQ_FALSE_RETRY (area-4 H4/H5/H6)

marmos91:
NFS area-4 PR-B3. Fixes the v4 exactly-once / replay-cache class (REVIEW.md §2 H4/H5/H6). Touches only `internal/adapter/nfs`; builds on #885 (share-reservation rework of OpenFile/ValidateStateid), does not revert it.

## H4 — v4.0 open-owner replay cache cross-op contaminated
`OpenOwner.LastResult` was a single slot populated only by OPEN, while CLOSE/OPEN_DOWNGRADE/OPEN_CONFIRM also advance `LastSeqID` → a legit TCP retransmit of one of those replayed stale OPEN bytes (or `NFS4ERR_BAD_SEQID`).

Fix: cache the encoded reply for **every** open-owner-seqid op. Each handler calls `CacheOpenOwnerResult` after encoding; each manager method's replay branch returns the cached bytes via a typed `*state.ReplayError`. Open-owners are now **retained after the last CLOSE** (mirrors Linux nfsd `so_replay`, where stateowners live until lease expiry) so a retransmitted CLOSE can still resolve them; a `closedOwnerByOther` map maps the now-closed stateid → retained owner since the open-state itself is deleted. Both are reaped in `onLeaseExpired`.

## H5 — v4.0 LOCK/LOCKU replay not honored → silent lock loss
`LockOwner.LastResult` was declared but never populated → a replayed LOCK returned `NFS4ERR_BAD_SEQID`, which the Linux client treats as fatal (drops the lock-owner).

Fix: populate `LockOwner.LastResult` on LOCK/LOCKU and on a DENIED LOCK (which still advances the seqid per RFC 7530 §8.1.5), return it on replay. The lock-owner seqid replay check is now done **before** the strict stateid-seqid compare in `LockExisting`/`UnlockFile`, because a replayed LOCK/LOCKU resends the pre-op lock stateid (seqid one behind) that would otherwise be rejected as `NFS4ERR_OLD_STATEID`. The `open_to_lock_owner` (LockNew) path routes an open-owner-seqid replay to the lock-owner's cached reply.

## H6 — v4.1 SEQ_FALSE_RETRY never detected
A slot retry returned the cached reply with no request compare → a client reusing slot+seqid with different ops silently got the stale reply (confused-deputy / data-exposure).

Fix: `ProcessCompound` fingerprints the full COMPOUND request body (SHA-256) into `CompoundContext.RequestDigest`; it is threaded to `SlotTable.ValidateSequence` and recorded by `CompleteSlotRequest`. On a same-seqid retry whose digest differs, `ValidateSequence` returns `NFS4ERR_SEQ_FALSE_RETRY` (RFC 8881 §2.10.6.1.3) before any cached reply is replayed.

## RFC / Linux cross-check
- RFC 7530 §9.1.7 (last-op replay per stateowner), §8.1.5 (seqid bumped on all but the listed errors; `NFS4ERR_DENIED` is not exempt), §8.20.
- RFC 8881 §2.10.6.1.3 (`SEQ_FALSE_RETRY`).
- Linux `fs/nfsd/nfs4state.c`: `so_replay` buffer on the stateowner, retained until lease expiry; `nfs4_check_sequence` false-retry handling.

## Tests
- `state/replay_cache_test.go`: OPEN→CLOSE / DOWNGRADE / CONFIRM replay returns cached reply (not BAD_SEQID); LOCK and LOCKU replay returns cached reply (lock not lost); DENIED LOCK caches + replays.
- `handlers/sequence_handler_test.go`: `TestSequence_FalseRetry_DifferentOps` — slot+seqid reused with a different op-list → `NFS4ERR_SEQ_FALSE_RETRY`; byte-identical retry still replays the cached reply. Existing replay tests updated to model true (byte-identical) retransmits.

Gates: `gofmt -s`, `go vet`, `golangci-lint` (0 issues), `go test -race ./internal/adapter/nfs/...` all green.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/close.go
+++ b/internal/adapter/nfs/v4/handlers/close.go
@@ -64,8 +64,11 @@ func (h *Handler) handleClose(ctx *types.CompoundContext, reader io.Reader) *typ
 		"client", ctx.ClientAddr)
 
 	// Delegate to StateManager for state cleanup
-	closedStateid, stateErr := h.StateManager.CloseFile(stateid, closeSeqid)
+	closeResult, stateErr := h.StateManager.CloseFile(stateid, closeSeqid)
 	if stateErr != nil {
+		if replay := asReplay(types.OP_CLOSE, stateErr); replay != nil {
+			return replay
+		}
 		nfsStatus := mapStateError(stateErr)
 		logger.Debug("NFSv4 CLOSE failed",
 			"error", stateErr,
@@ -102,7 +105,13 @@ func (h *Handler) handleClose(ctx *types.CompoundContext, reader io.Reader) *typ
 
 	var buf bytes.Buffer
 	_ = xdr.WriteUint32(&buf, types.NFS4_OK)
-	types.EncodeStateid4(&buf, closedStateid)
+	types.EncodeStateid4(&buf, &closeResult.Stateid)
+
+	// Cache the encoded reply on the open-owner so a retransmitted CLOSE at the
+	// same seqid replays these exact bytes instead of failing NFS4ERR_BAD_SEQID.
+	if closeResult.OwnerData != nil {
+		h.StateManager.CacheOpenOwnerResult(closeResult.OwnerClientID, closeResult.OwnerData, types.NFS4_OK, buf.Bytes())
+	}
 
 	return &types.CompoundResult{
 		Status: types.NFS4_OK,
--- a/internal/adapter/nfs/v4/handlers/close.go
+++ b/internal/adapter/nfs/v4/handlers/close.go
@@ -81,6 +81,19 @@ func (h *Handler) handleClose(ctx *types.CompoundContext, reader io.Reader) *typ
 		}
 	}
 
+	// Encode and cache the CLOSE reply BEFORE the (potentially slow) metadata
+	// flush. CloseFile has already advanced the open-owner seqid and installed
+	// the closed-owner entry, so a retransmitted CLOSE arriving while the flush
+	// below is still running would be classified as a replay; caching here (not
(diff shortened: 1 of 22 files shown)


marmos91:
Fixed in `06415144`: the CLOSE reply is now encoded+cached BEFORE FlushPendingWriteForFile, closing the window where a retransmit during the (slow, I/O-bound) flush replayed the owner's previous LastResult instead of the CLOSE bytes.

Independent adversarial review (git-capable) of the full diff: **no HIGH** — retained-owner lifecycle (bounded reap in onLeaseExpired, closedOwnerByOther cleanup, re-OPEN reuse, closed-stateid still BAD_STATEID for real ops) and the H5 replay-before-seqid reorder (replay fires only on exact lock-owner-seqid match; stale stateids still hit OLD/BAD) both confirmed correct. Two MEDIUMs deferred as documented follow-ups (non-regressions):
- **H6-M1**: SEQ_FALSE_RETRY digest covers the whole COMPOUND incl. SEQUENCE's volatile `sa_highest_slotid`/`sa_cachethis` → a *theoretical* false-positive if a client rebuilds (rather than byte-resends) a retransmit. Latent — Linux resends byte-identical. Per the interop playbook this wants a tshark capture of a real retransmit before narrowing the digest (over-narrowing risks missing real false-retries), so deferred to a tracked follow-up.
- **C-M1**: the v4.0 cache-after-encode micro-window is a pre-existing architectural property (the old single-OPEN cache had it); v4.1 is slot-protected and Linux serializes per-owner. The CLOSE fix above removes the one case where a slow flush widened it.

PR #891, opened 2026-05-31
fix(smb/dh): disallow_write_lease cap for non-stat opens and disconnected handles (#739)

marmos91:
## Summary

Residual #739 Durable Handles V2 conformance work. Mirrors Samba's
`disallow_write_lease` accumulator (`source3/smbd/open.c::delay_for_oplock_fn`,
lines 2390-2403 + 2637-2643) so a lease grant's Write caching bit is stripped
when a conflicting other open exists — covering the two cases the lock manager's
`bestGrantableState` cannot see:

1. A **non-oplock-stat live open holding no lease** (e.g. opened with
   `READ_CONTROL`, so `is_oplock_stat_open` is false). `bestGrantableState`
   only inspects lease records, so it misses this. → `nonstat-and-lease`.
2. A **disconnected durable handle** still holding an RH/RWH lease. Its lease
   record was released from the lock manager at disconnect, but the persisted
   handle keeps the file effectively leased; a new RWH open from another
   connection must downgrade to RH with **no break** (a disconnected handle
   cannot ack one). → `keep-disconnected-rh-with-rwh-open`.

The new open's own lease key (Samba `is_same_lease`) and SMB FileID are excluded
so a same-handle reopen/upgrade is never self-capped — the regression guard
against the broad "any non-stat open" cap that previously broke 13 lease tests.

W is stripped from the **request** (not the grant): equivalent because the
downgrade candidate chain and R/H conflict checks produce the same R/H bits and
the lock manager persists the correct post-strip state.

## Targets (#739, all in KNOWN_FAILURES.md)

- `smb2.durable-v2-open.nonstat-and-lease` — addressed by the live non-stat-open W-cap.
- `smb2.durable-v2-open.keep-disconnected-rh-with-rwh-open` — addressed by the disconnected-handle W-cap.
- `smb2.durable-v2-open.lock-oplock` / `lock-lease` — BR-lock survival across durable disconnect/reconnect is already wired (locks preserved on durable persist, OpenID continuous via OriginalFileID); no code change needed, validated by reasoning. CI will confirm.
- `smb2.durable-v2-delay.durable_v2_reconnect_delay` — timeout=0 default-grant + PreviousSessionId supersession paths already in place; CI will confirm.

Out of scope (left in KF): `persistent-open-*` (unimplemented feature), `app-instance` (cross-connection break routing).

## Verification

- `go build ./...`, `go vet ./internal/adapter/smb/...`, `golangci-lint` — clean.
- `go test ./internal/adapter/smb/... -count=1` and `go test -race ./internal/adapter/smb/v2/handlers/... -count=1` — green (lease-break regression guard).
- New unit test `TestDisallowWriteLeaseForFile` covers all true/false cases incl. the reopen/upgrade self-cap guards.
- `KNOWN_FAILURES.md` untouched (orchestrator walks back post-merge after CI confirms each flip).

### DIFF
--- a/internal/adapter/smb/v2/handlers/create_post_break.go
+++ b/internal/adapter/smb/v2/handlers/create_post_break.go
@@ -842,6 +842,22 @@ func (h *Handler) completeCreateAfterBreak(ctx *SMBHandlerContext, d *createDraf
 		if leaseCtx := FindCreateContext(req.CreateContexts, LeaseContextTagRequest); leaseCtx != nil {
 			lockFileHandle := lock.FileHandle(fileHandle)
 			var err error
+			// Samba `disallow_write_lease`: strip W from this lease grant when a
+			// conflicting other open or a disconnected durable handle exists.
+			// The new open's own lease key (parsed from the RqLs blob) and SMB
+			// FileID are excluded so a same-handle reopen/upgrade is never
+			// self-capped (mirrors Samba `is_same_lease`). bestGrantableState in
+			// the lock manager already caps W against live *lease* records; this
+			// predicate adds the cases it cannot see — non-lease live opens and
+			// disconnected durable handles (smb2.durable-v2-open.nonstat-and-lease
+			// and keep-disconnected-rh-with-rwh-open).
+			var newLeaseKey [16]byte
+			if parsed, decErr := DecodeLeaseCreateContext(leaseCtx.Data); decErr == nil && parsed != nil {
+				newLeaseKey = parsed.LeaseKey
+			}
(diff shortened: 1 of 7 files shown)


marmos91:
Superseded by #894, which refines the disallow_write_lease gate to exclude same-lease-key/same-client upgrades — eliminating this PR's lease.upgrade2/upgrade3/break regression while keeping keep-disconnected-rh-with-rwh-open + reconnect_delay flips. (nonstat-and-lease traded back to KF; #739 stays open.)

Issue #568, opened 2026-05-22
flaky: rotating smb2 'new failure' post-#547 (cross-test state leak)


marmos91:
## Symptom

After PR #547 (deltree blocker fix) merged, smbtorture CI reports a different 'new failure not in KNOWN_FAILURES' on different runs against the same develop tip:

| Run | New failure |
|---|---|
| #553 PR (job 77331427502) | `smb2.create.multi` |
| #551 PR (job 77332359795) | `smb2.lease.break` |
| develop run 26279181361 (job 77350886112) | `smb2.lease.oplock` |
| develop run 26283012843 (jobs 77363669927/77363669952/77363670415) | **none** — all smbtorture jobs PASS |

Different test reported each time, sometimes none at all = classic cross-test interference flake. None of these tests have logical connection to ACL changes in the merged PRs.

## Root cause hypothesis

Phase-2 ACL tests now advance much further into their bodies after PR #547. When they fail mid-test (which 7 of them still do — tracked in #559–#565), they leave behind state in the share storage: restrictive DACLs on test files, dangling opens, non-empty directories, in-flight lease records. PR #547 fixed within-test `smb2_deltree` (parent FILE_DELETE_CHILD override) but did NOT add a cross-suite teardown.

The next suite that hits the share (lease.*, create.*) then sees this leftover state and a previously-passing test fails. Different suite-ordering / scheduling on each GHA runner picks a different victim.

## Repro

Run smbtorture multiple times against develop tip. Expect the 'new failure not in KNOWN_FAILURES' to rotate between `create.multi`, `lease.break`, `lease.oplock`, and intermittently no new failure at all.

## Suggested fix direction

`test/smb-conformance/smbtorture/run.sh` (or the runner wrapper) should restart the DittoFS server / wipe `/data` between sub-suites — or at minimum between `smb2.acls` and downstream lease/create suites. Alternatives:

1. `docker compose restart dittofs` between suites
2. Recreate the share via API before each sub-suite (`dfsctl share delete <name> && dfsctl share create ...`)
3. Add a global setup hook that `smb2_deltree`s the share root with the CAVEMAN root credential before each sub-suite begins

Refs #547 #559 #560 #561 #562 #563 #564 #565.

marmos91:
[...]

PR #906, opened 2026-05-31
fix(smb/replay): echo requested oplock level on non-lease DH2 create replay (#749)

marmos91:
## Problem

In \`resolveCreateReplay\` (internal/adapter/smb/v2/handlers/create.go), a non-lease (traditional oplock) DH2Q CREATE replay returned the cached CREATE response **verbatim**, echoing the oplock level that was originally *granted*. The \`replay-dhv2-oplock2\` test replays a DH2Q CREATE whose replay request carries a **different** requested OplockLevel than the original grant, and expects the server to echo the *request's* level in the replay response.

## Fix

On the non-lease replay path, after the shallow copy of the cached response, stamp the returned copy's \`OplockLevel\` with the replay request's requested level:

\`\`\`go
if req.OplockLevel != OplockLevelLease {
    resp.OplockLevel = req.OplockLevel
}
\`\`\`

This mirrors Samba, which sets \`io.out.oplock_level = io.in.oplock_level\` on the replay path (\`source3/smbd/smb2_create.c\`), per MS-SMB2 §3.3.5.9.

### Guarantees preserved
- **Cache not mutated** — operates only on \`resp\`, the shallow copy; \`entry.Response\` and the cached entry are untouched (asserted by the new test).
- **Live open's held oplock unchanged** — only the wire response field is stamped; the \`OpenFile\`'s actual oplock state is not modified.
- **Lease path unchanged** — the guard skips lease replays (\`OplockLevel == 0xFF\`), which continue through \`refreshReplayLease\` (RqLs context refresh / ACCESS_DENIED validation).

## Tests
- New \`TestResolveCreateReplay_OplockEchoesRequestedLevel\`: caches a CREATE granted Batch, replays a request asking None, asserts the response echoes None **and** the cached entry + live open both remain Batch.
- Updated \`TestResolveCreateReplay_OplockSnapshot\` so its replay request asks for the same level it was granted (Batch), keeping its matching-level coverage intact.

Target test: \`smb2.replay-dhv2-oplock2\`. Refs #749.

\`go build\`, \`go vet\`, \`golangci-lint\`, and \`go test ./internal/adapter/smb/v2/handlers/...\` all pass.

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -1602,10 +1602,17 @@ func (h *Handler) resolveCreateReplay(ctx *SMBHandlerContext, req *CreateRequest
 	resp := *entry.Response
 
 	// Lease replays are validated and state-refreshed against the live
-	// open. Non-lease (plain oplock) replays return the cached response
-	// verbatim: Samba echoes the request's oplock level back on replay
-	// without touching the held oplock, which the cached snapshot already
-	// reflects (replay-dhv2-oplock1/2/3).
+	// open (refreshReplayLease). Non-lease (plain oplock) replays echo the
+	// REPLAY REQUEST's requested oplock level back in the response — Samba
+	// sets io.out.oplock_level = io.in.oplock_level on the replay path
+	// (source3/smbd/smb2_create.c; MS-SMB2 §3.3.5.9) — without touching the
+	// open's actually-held oplock. The replay request may request a
+	// different level than the original grant (replay-dhv2-oplock2). resp is
+	// a shallow copy, so this stamps only the returned response, never the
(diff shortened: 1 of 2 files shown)


marmos91:
Closing: the oplock-echo on non-lease replay does not flip replay-dhv2-oplock2 — that test still fails at 'Replay with durable handles not implemented', i.e. before the create-replay-cache echo path executes. The echo is spec-correct but addresses the wrong layer; the real gap is deeper replay-detection for the DH2 replay family. Folding this into the #749 pending-matrix work instead. No regression (the run's stat-and-lease + multichannel.leases.test1 reds are the known #568/#829 NO_MEMORY cascade flake, unrelated to this diff).

PR #910, opened 2026-05-31
fix(smb/replay): cross-request DH2 reservation + two-phase status for pending1 (#749)

marmos91:
## What & why

`#908` reserved/released the DH2 CREATE-replay slot within a single CREATE call, which flipped **zero** `pending1` rows. Root cause: for the `pending1n` / violation variants the conflicting open's status is only known **across multiple requests** — the reservation must survive past the synchronous violation CREATE until the conflict resolves, and must carry a *two-phase* status. This PR makes the reservation cross-request and status-bearing.

Closes the protocol gap for `#749`.

## Ground truth (extracted from `source4/torture/smb2/replay.c`)

Two test families drive `dhv2-pending1*`:

### `_test_dhv2_pending1_vs_hold` — the `-vs-{oplock,lease}` rows
client1 holds BATCH oplock or RWH lease with `share_access="RWD"`; client2 (`create_guid2`, also `RWD`) is share-compatible, so it only parks on the oplock/lease break and ultimately **succeeds**.

| client1 / client2 | replay22 & replay23 (while parked) | io24 (after orig21 OK) |
|---|---|---|
| BATCH / NONE (`pending1n-vs-oplock`) | sane=`FILE_NOT_AVAILABLE`, windows=`ACCESS_DENIED` | `OK` (cached) |
| LEASE / NONE (`pending1n-vs-lease`) | sane=`FILE_NOT_AVAILABLE`, windows=`ACCESS_DENIED` | `OK` |
| BATCH/LEASE / LEASE (`pending1l-vs-*`) | sane=`FILE_NOT_AVAILABLE`, windows=`ACCESS_DENIED` | `OK` |
| BATCH/LEASE / BATCH (`pending1o-vs-*`) | sane=`FILE_NOT_AVAILABLE`, windows=`ACCESS_DENIED` | `OK` |

### `_test_dhv2_pending1_vs_violation` — the `-vs-violation-lease-{close,ack}` rows
client1 holds an RWH lease with `share_access=0`; client2 (`create_guid2`, `share_access=0`) **conflicts**. Server breaks the handle lease first and parks `req21` (defers the violation).

| variant | release_op | orig21 | replay22 (parked) | replay23 (post-resolve) |
|---|---|---|---|---|
| `violation-lease-close-sane` | CLOSE | `OK` | `FILE_NOT_AVAILABLE` | `OK` |
| `violation-lease-close-windows` | CLOSE | `OK` | `SHARING_VIOLATION` | `OK` |
| `violation-lease-ack-sane` | BREAK-ack | `SHARING_VIOLATION` | `FILE_NOT_AVAILABLE` | `SHARING_VIOLATION` |
| `violation-lease-ack-windows` | BREAK-ack | `SHARING_VIOLATION` | `SHARING_VIOLATION` | `SHARING_VIOLATION` |

The **ack** rows are the load-bearing case: client1 acks the break but keeps the file open with `share_access=0`, so the conflict stands, orig21 fails `SHARING_VIOLATION`, and `replay23` must return that **same** status from the reservation — not re-run CREATE.

## Design — cross-request, two-phase reservation

`CreateReplayCache` reservations now carry a `reservedState{ Status, At }` instead of a bare set membership:

- **Phase A** (original pending on a share-mode conflict) → replay returns `STATUS_FILE_NOT_AVAILABLE` (the "sane" status).
- **Phase B** (original resolved to a terminal share-mode violation) → `UpdateReservedStatus` transitions the reservation to the terminal status; replay returns it from the reservation.

New API on `CreateReplayCache`: `UpdateReservedStatus`, `ReservedStatus`; `Reserve` now stamps Phase A; `IsReserved` delegates to `ReservedStatus`.

A single shared seam, `resolveCreateReplayReservation(sessionID, guid, terminalStatus)`, closes out the reservation on the original CREATE's terminal status, used by **both** the inline-completion path (`Create`) and the async-park resume goroutine (`parkCreateOnLeaseBreak`):
- `SUCCESS` / transient → `Release` (the cached success entry serves later replays).
- `SHARING_VIOLATION` / `ACCESS_DENIED` → Phase B.

`resolveCreateReplay` consults the reservation status only when no cached success entry exists, preserving the existing "entry-first" ordering (a just-completed success replays the open, not a stale pending status).

## Leak / double-release proof

A reservation can never leak or be cleared twice:

- **Inline path:** `Create` reserves before dispatch; after `completeCreateAfterBreak` returns it calls the resolver exactly once with the real terminal status.
- **Async path:** the resume goroutine resolves the reservation in a `defer` driven by `finalStatus`, which **defaults to `StatusCancelled`** — so every early-return path (dispatcher preemption after `Unregister`, tree-disconnect → `NETWORK_NAME_DELETED`) still resolves to a plain `Release`. `finalStatus` is set to the real status only after `completeCreateAfterBreak`.
- **Phase-B persistence** is bounded twice over: `ForgetSession` clears all reserved entries at session teardown (the test ends with `smbXcli_conn_disconnect`), and a TTL backstop (`replayCacheTTL`) reaps any reservation aged past the window on the next `ReservedStatus` access — so a Phase-B reservation can never brick future opens of the guid indefinitely.
- **No-resurrect:** `UpdateReservedStatus` is a no-op when the guid is not currently reserved, so a late terminal resolution after a success-path `Release` cannot recreate a stuck reservation.
- `Release` / `UpdateReservedStatus` are idempotent map ops; a double call is harmless.

## Target rows (`KNOWN_FAILURES.md`, untouched — walkback is post-CI)

`dhv2-pending1n-vs-{oplock,lease}-{sane,windows}`, `dhv2-pending1n-vs-violation-lease-{close,ack}-{sane,windows}`, `dhv2-pending1o-vs-{oplock,lease}-{sane,windows}`, `dhv2-pending1l-vs-{oplock,lease}-windows`.

Note: DittoFS is a single server and returns the **sane** (Samba) statuses (`FILE_NOT_AVAILABLE` while pending, `SHARING_VIOLATION` on violation). The `*-windows` rows expect Windows-policy statuses (`ACCESS_DENIED` / immediate `SHARING_VIOLATION`) and are mutually exclusive with the sane rows by construction — the sane variants are the ones this PR targets to flip; the windows variants remain expected-fail.

## Regression safety

`pending1l-vs-{oplock,lease}-sane`, `pending1o-vs-lease-sane`, `replay-dhv2-lease1/2/3`, `lease-oplock`, `oplock-lease` are unaffected: the entry-first lookup ordering in `resolveCreateReplay` is preserved, success still caches via `Store`, and the reservation is only consulted on a replay with no cached entry.

## Tests (`replay_dhv2_test.go`)

- `TestResolveCreateReplay_TwoPhase_ViolationAckSurvivesAcrossRequests` — Phase A `FILE_NOT_AVAILABLE` → Phase B `SHARING_VIOLATION` across requests; cleared by `ForgetSession`.
- `TestResolveCreateReplay_TwoPhase_SuccessReleasesReservation` — success → Release (no Phase B).
- `TestResolveCreateReplay_TwoPhase_TransientFailureReleases` — cancelled / internal / network-name-deleted all Release (leak-proof early returns).
- `TestUpdateReservedStatus_NoResurrect` — no resurrect after Release / never-reserved.
- `TestReservedStatus_TTLBackstop` — aged Phase-B reservation reaped.
- `TestReservedStatus_ConcurrentSameGuid` — `-race` Reserve/Update/Lookup/Release contention.

Existing replay tests stay green; `go test -race ./internal/adapter/smb/v2/handlers/` clean.

### DIFF
--- a/internal/adapter/smb/v2/handlers/create.go
+++ b/internal/adapter/smb/v2/handlers/create.go
@@ -1392,9 +1392,39 @@ func (h *Handler) Create(ctx *SMBHandlerContext, req *CreateRequest) (*CreateRes
 
 	resp := h.completeCreateAfterBreak(ctx, draft)
 	if reservedReplay {
-		h.CreateReplayCache.Release(ctx.SessionID, replayGuid)
+		h.resolveCreateReplayReservation(ctx.SessionID, replayGuid, resp.GetStatus())
 	}
 	return resp, nil
+}
+
+// resolveCreateReplayReservation closes out a CreateGuid reservation taken for
+// a conflicting (pending) original CREATE once that CREATE reaches a terminal
+// state, applying the two-phase MS-SMB2 §3.3.5.9 rule (Samba bug 14449
+// smb2.replay.dhv2-pending1*):
+//
+//   - success: the success response was just cached via Store, so the
+//     reservation is released; a later replay returns the cached open.
+//   - terminal share-mode failure (STATUS_SHARING_VIOLATION / ACCESS_DENIED):
+//     the reservation transitions to Phase B carrying that status, so a later
+//     replay returns the SAME terminal status instead of re-running CREATE and
+//     re-parking (replay.dhv2-pending1n-vs-violation-lease-ack-* replay23). The
+//     Phase-B reservation is reaped by session teardown / the TTL backstop.
+//   - any other failure (transient / internal): released, matching the prior
+//     behaviour so a genuine retry can re-run the CREATE fresh.
+//
+// It is the single seam shared by the inline-completion path (Create) and the
+// async-park resume goroutine (parkCreateOnLeaseBreak) so the reservation
+// lifetime is identical on both. A zero CreateGuid is a no-op throughout.
+func (h *Handler) resolveCreateReplayReservation(sessionID uint64, createGuid [16]byte, status types.Status) {
+	if h.CreateReplayCache == nil {
+		return
+	}
+	switch status {
+	case types.StatusSharingViolation, types.StatusAccessDenied:
+		h.CreateReplayCache.UpdateReservedStatus(sessionID, createGuid, status)
+	default:
+		h.CreateReplayCache.Release(sessionID, createGuid)
+	}
 }
 
 // computeMaximalAccess computes the maximal access mask for a file used in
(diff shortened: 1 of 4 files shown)


marmos91:
Closing: cross-request reservation flips 0 pending1n rows (CI-confirmed all still failure:). The real assertions (replay.c:2228 wants ACCESS_DENIED not FILE_NOT_AVAILABLE; :1659/:1690 want OK; :536 oplock-echo; :2294 data-offset) show pending1n is a multi-stage 'replay pending oplock' state machine, not a reserve→status problem — the two-phase model mis-derived the semantics. #908's single-Reserve foundation stays merged; pending1 needs a correctly-modeled rewrite or appendix. No regression (run green, New-failures:0).

Issue #686, opened 2026-05-28
v1.0: Kerberos smbtorture sweep (65 rows incl. reauth/expire/bind + header-MIC)


marmos91:
## Context

Tracker for the 70 entries in `test/smb-conformance/smbtorture/KNOWN_FAILURES_KERBEROS.md` deferred past **v1.0.0** tag.

These tests are only exercised when smbtorture runs with `--kerberos`, which requires a KDC and Kerberos-enabled config profile (`memory-kerberos`). The non-Kerberos v1.0 CI matrix does not load this file (see `run.sh:533`), so the 70 entries do not gate the v1.0 conformance umbrella (#673).

## Failure breakdown (per `KNOWN_FAILURES_KERBEROS.md`)

- **14 Kerberos Session Bugs (#340)** — reconnect, reauth, ticket expiration semantics
- **42 Multi-Channel Session Bind (#361)** — same root cause as the NTLM-side #483; Kerberos surface is additional
- **5 Anonymous Auth** — anonymous-over-Kerberos is identical to NTLM failures; out of scope
- **2 AES-256 Encryption** — Kerberos-side variants
- **NTLMSSP compat test** — not applicable to Kerberos

## Acceptance (post-v1.0)

- Failures are addressed under a dedicated **v1.0+kerberos** milestone or rolled into v1.1.
- `KNOWN_FAILURES_KERBEROS.md` reaches zero unjustified entries.
- Kerberos CI matrix runs cleanly on develop.

## Out of scope for v1.0

- Do not block the v1.0 tag on these tests.
- The Permanently Unimplementable appendix in `KNOWN_FAILURES.md` documents this deferral.

Cross-ref: #673 (v1.0 conformance umbrella).

marmos91:
Kerberos smbtorture sweep complete. All addressable rows flipped:
- session bind (CMAC↔HMAC wire decode), session expire/reauth (reauth1–4 key retention), header-MIC verification, and AES-256-CCM/GCM negotiation all shipped and green on the `smbtorture Kerberos` CI job.
- Sole remaining row in `KNOWN_FAILURES_KERBEROS.md` is `smb2.reauth5`, which is an **upstream Samba selftest known-fail** (`selftest/knownfail.d/` line 213): it asserts `smb2_util_unlink` of a nonexistent file returns NT_STATUS_OK, but a correct server returns OBJECT_NAME_NOT_FOUND. It fails against Samba's own server too — not a DittoFS gap.

[...]

PR #915, opened 2026-05-31
docs(smb): defer #749 replay pending-oplock matrix past v1.0

marmos91:
Per decision to appendix the replay pending-oplock matrix (like Kerberos #686).

The 39 `### Replay Protection` rows are relabeled as **deferred past v1.0** under #749. Channel-sequence + create-replay infra shipped (#864/#866/#908); the remaining `dhv2-pending{1,2,3}*` matrix is a deep multi-stage replay-pending-oplock state machine (per-stage ACCESS_DENIED/OK/oplock-echo/data-offset) — a self-contained effort not worth blocking the v1.0 tag. The `*-windows` subset is additionally not flippable without a Windows-status mode (DittoFS returns Samba-sane statuses).

Rows stay as documented known-failures so CI stays green; Policy goal amended to exempt this block. No code change.


marmos91:
Closing: #749 is being un-deferred and implemented before v1.0.0 (replay-pending-oplock state machine + Windows-policy mode). The 39 KF rows stay as plain known-failures until the state machine flips them — not relabeled as deferred.

PR #917, opened 2026-05-31
docs(smb): defer #739 lock-lease + app-instance residuals past v1.0

marmos91:
Documents the last two DH2 residuals as deferred (reasons inline). `lock-lease` needs lease-v2 epoch persistence across reconnect (sibling lock-oplock flipped via #913; lease variant additionally asserts lease_epoch==1 → epoch threaded through durable store across memory/badger/postgres + spurious-break suppression) — disproportionate multi-backend plumbing for 1 test. `app-instance` still emits one spurious lease break (break_info.count==1); pinning break-routing timing needs live-server debug not justified by 1 test at v1.0. #739 tractable rows already landed (nonstat-and-lease, lock-oplock, keep-disconnected, reconnect-delay, reopen1/1a). Rows stay suppressed (CI green). #739 residuals are post-v1.0.


marmos91:
Closing: lock-lease + app-instance are being un-deferred and implemented before v1.0.0 (lease-V2 epoch persistence across the durable store). Superseded by the upcoming DH2 epoch-persistence PR.

PR #938, opened 2026-06-01
fix(operator): set dfs TerminationGracePeriodSeconds to survive graceful shutdown on k8s rollout

marmos91:
The operator-rendered dfs PodSpec had no `terminationGracePeriodSeconds`, so k8s applied its 30s default. The dfs server runs its graceful shutdown stages **serially**, each bounded by `shutdown_timeout` (the operator pins `shutdown_timeout: 30s`), and the PreStop hook (`sleep 5`) eats further into the window — leaving ~25s, well under the worst-case multi-stage budget. On rollout/drain/scale-down k8s SIGKILLed dfs mid-flush → metadata loss (round-2 docker H1).

**Fix:**
- Derive TGPS from the configured shutdown_timeout so the two stay coupled: `preStop(5) + 3*shutdownTimeout(30) + buffer(10) = 105s` for the 30s default — comfortably exceeds `preStop + shutdown_timeout`.
- Add an optional `terminationGracePeriodSeconds` `*int64` CRD override field (kubebuilder min=31); when unset the controller computes the derived default. Regenerated deepcopy + CRD manifests, synced `chart/crds`.
- PreStop sleep now references the same constant.

Dockerfiles are already exec-form (`ENTRYPOINT ["/app/dfs"]`, `["/manager"]`) so SIGTERM reaches PID 1 — no Dockerfile change needed.

**Tests:** controller reconcile asserts pod TGPS `!= nil` and `> preStop+shutdownTimeout` and `== derived default` when unset; focused unit test covers unset-default (105s) and explicit-override paths. Verified the assertions FAIL on the pre-fix (no-TGPS) code.

### DIFF
--- a/k8s/dittofs-operator/api/v1alpha1/dittoserver_types.go
+++ b/k8s/dittofs-operator/api/v1alpha1/dittoserver_types.go
@@ -58,6 +58,19 @@ type DittoServerSpec struct {
 	// +optional
 	Resources corev1.ResourceRequirements `json:"resources,omitempty"`
 
+	// TerminationGracePeriodSeconds is the pod's grace period for the dfs server
+	// to complete its graceful shutdown on rollout, drain, or scale-down.
+	//
+	// The dfs server runs shutdown stages serially, each bounded by its configured
+	// shutdown_timeout (30s). Combined with the PreStop hook delay, the Kubernetes
+	// default of 30s is too short and causes a SIGKILL mid-flush, which can lose
+	// metadata. When unset, the operator derives a safe value from the configured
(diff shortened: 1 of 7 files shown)


marmos91:
Addressed the Copilot review: the kubebuilder `Minimum` was raised from 31 to **36** across the type marker and both generated CRDs (`config/crd/bases` + `chart/crds`). 36s = preStop (5s) + shutdown_timeout (30s) + 1, so the validation floor now matches the documented contract and the controller test's `minRequired` (35), and an explicit override can no longer be set below the value the assertion requires. The override path itself is unchanged (advanced users with a larger shutdown budget can still set higher values).

Issue #750, opened 2026-05-29
SMB: Misc protocol gaps — DOS attrs, setinfo, sharemode, max-allowed, tcon, maxfid, EA-ACL, timestamps (9 tests)


marmos91:
Untagged miscellaneous protocol-gap residuals in KNOWN_FAILURES. Each is a distinct semantic gap but too small to warrant its own tracker.

Tests:
- smb2.dosmode — DOS mode semantics not implemented
- smb2.setinfo — SET_INFO timestamp preservation not implemented
- smb2.sharemode.bug14375 — Share mode edge case not implemented
- smb2.maximum_allowed.maximum_allowed — Full MAXIMUM_ALLOWED access computation not implemented
- smb2.tcon — Advanced tree connect semantics not implemented
- smb2.maxfid — Connection drops under high FD pressure
- smb2.ea.acl_xattr — EA ACL enforcement not implemented
- smb2.timestamp_resolution.resolution1 — Sub-second timestamp resolution enforcement not implemented
- smb2.notify.mask-change — SHARING_VIOLATION on directory open (pre-existing, never passed individually)

Tracks v1.0 conformance — KF rows above the Permanently Unimplementable appendix must reference an OPEN sub-issue of #673.

marmos91:
Note: additional smbtorture tests routed to this issue via #752 KF re-tag — `smb2.ioctl.copy_chunk_sparse_dest`, `smb2.notify.valid-req`, `smb2.dirlease.oplocks`, `smb2.charset.unicode`. Treat as in-scope for the misc-residuals tracker.

marmos91:
Subset A (`smb2.dosmode` + `smb2.setinfo` + `smb2.timestamp_resolution.resolution1`) shipped in PR #759 (merge `f0f7727f`).

All three rows promoted to the Permanently Unimplementable appendix in `test/smb-conformance/smbtorture/KNOWN_FAILURES.md` with documented architectural rationale:
- `smb2.dosmode` — Samba `smb.conf` `hide files` glob (server-config behaviour, not MS-FSCC/MS-SMB2).
- `smb2.setinfo` — persistent EA storage (SET_INFO surface itself works end-to-end).
- `smb2.timestamp_resolution.resolution1` — timing-dependent, explicitly skipped by Samba's own selftest (`selftest/skip:69-70`).

[...]

PR #943, opened 2026-06-01
fix(smb): zero-fill copychunk sparse destination gap (#750)

marmos91:
## Summary

`smb2.ioctl.copy_chunk_sparse_dest` (#750) was the last residual sparse smbtorture failure. Investigation against a freshly-built `origin/develop` shows the destination write path **already** materializes the sparse hole correctly — the failure is stale.

## Root cause (of the original failure)

SRV_COPYCHUNK to a 0-byte destination with a chunk at `target_off = 4096` must surface `[0, 4096)` as zeros on a subsequent read (MS-SMB2 §3.3.5.15.1 / sparse semantics; Samba `source4/torture/smb2/ioctl.c::test_ioctl_copy_chunk_sparse_dest`). Earlier the block-store sparse-hole zero-fill was not exercised for copychunk-extended files.

## Where the zero-fill runs now

`executeCopyChunks` writes the copied extent at `TargetOffset` via the same `engine.Store.WriteAt` path a normal SMB WRITE uses. That `WriteAt` past EOF grows the block store's local append log, and the gap `[oldEOF, TargetOffset)` is served as zeros by the read path (`memory.ReadPayloadAt` / engine `readAtInternal`) — exactly the existing sparse/zero-fill mechanism, lazily (no whole-gap buffering). `PrepareWrite(newSize = TargetOffset + Length)` extends the file size so the subsequent READ passes the EOF gate and returns the zero-filled hole. No new mechanism was introduced and no production code change was required.

The invariant is already pinned by integration tests added under #855:
- `pkg/blockstore/engine/copychunk_sparse_dest_test.go` — engine `WriteAt`/`ReadAt` + memory `ReadPayloadAt` contracts.
- `internal/adapter/smb/v2/handlers/ioctl_copychunk_sparse_integration_test.go` — drives the real `executeCopyChunks` write path + the real `Read` handler against the memory block store (the backend every smb-conformance profile uses).

## Verification

- Local smbtorture, memory profile: `smb2.ioctl.copy_chunk_sparse_dest` → **PASS** (Passed: 1, Failed: 0, New failures: 0).
- `GOWORK=off go build ./...`, `go vet ./internal/adapter/smb/...`, `gofmt -s` — clean.
- `GOWORK=off go test -race ./internal/adapter/smb/...` and the engine `Sparse` tests — green.

## Change

Removes the stale `smb2.ioctl.copy_chunk_sparse_dest` row from `KNOWN_FAILURES.md` (local smbtorture proves the flip) and updates the surrounding prose to document why the hole reads back as zeros.

MS reference: MS-SMB2 §3.3.5.15.1 (FSCTL_SRV_COPYCHUNK), sparse-file read semantics.

### DIFF
--- a/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
+++ b/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
@@ -57,16 +57,13 @@ not advertised — they consume no failure slots and are not listed below.
 The compress_notsup_get/set tests correctly SKIP because FILE_FILE_COMPRESSION
 is advertised.
 
-Most IOCTL sparse-family entries walked back under #718. The remaining residual
-failure is a real feature gap: SRV_COPYCHUNK on a sparse destination must surface
-zeros for the unwritten hole between the old EOF and the chunk's target offset
-(see Samba `test_ioctl_copy_chunk_sparse_dest`). DittoFS's copychunk path grows
-the destination file via `WriteAt` at the target offset but does not advertise
-or materialize the [old EOF, target offset) hole as zero-reading bytes.
-
-| Test Name | Category | Reason | Issue |
-|-----------|----------|--------|-------|
-| smb2.ioctl.copy_chunk_sparse_dest | IOCTL | SRV_COPYCHUNK to a 0-byte destination at offset 4096 must surface the [0, 4096) gap as zeros on subsequent reads. The block-store sparse-hole zero-fill path does not run for copychunk-extended files. | #750 |
+Most IOCTL sparse-family entries walked back under #718. The last residual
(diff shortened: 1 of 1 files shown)


marmos91:
Closing: this PR only removed the copy_chunk_sparse_dest KNOWN_FAILURES row with no code change, on the belief the engine already zero-fills the [0,offset) hole. CI proves it still fails on the memory backend (ioctl.c:1772 'sparse did not pass class'). The real fix — SRV_COPYCHUNK to a sparse/0-byte destination must surface the pre-target hole as zeros — is not implemented. Row stays in KF; a real block-store zero-fill fix will reopen this under #750.

PR #950, opened 2026-06-01
fix(operator): harden managed dfs pod (disable SA-token automount, secure default SecurityContext) (v1.0 audit #11)

marmos91:
## What

Hardens the operator-managed `dfs` StatefulSet PodSpec (built in `k8s/dittofs-operator/internal/controller/dittoserver_controller.go`). Addresses two findings from the v1.0 audit area #11 (operator) REVIEW2.md §4.

### M-SEC-2 — disable ServiceAccount-token automount
The managed pod previously set neither `ServiceAccountName` nor `AutomountServiceAccountToken`, so Kubernetes mounted the namespace `default` SA API token onto the most network-exposed container (NFS/SMB/REST). `dfs` imports no client-go and never calls the Kubernetes API, so that token has zero legitimate consumer. Now sets `AutomountServiceAccountToken: false` on the PodSpec.

### M-SEC-3 — secure-default container SecurityContext
The managed container previously used `SecurityContext: Spec.SecurityContext` with no operator default (only `FSGroup` was set at the pod level). With no default, `allowPrivilegeEscalation` defaults true and no capabilities are dropped — asymmetric with the hardened operator pod, and a Pod-Security-Standards "restricted" namespace would **reject** the operator-generated StatefulSet, breaking install on hardened clusters.

A new `getContainerSecurityContext` helper applies a secure default when `Spec.SecurityContext` is nil:
- `allowPrivilegeEscalation: false`
- `runAsNonRoot: true` (the dfs image already ships `USER 65532`)
- `capabilities.drop: [ALL]`
- `seccompProfile: RuntimeDefault`

When the user **does** supply `Spec.SecurityContext`, the secure defaults are backfilled only where the user left a field unset (user wins on conflicts), starting from a deep copy so all user-set fields — including any added in future API versions — are preserved.

### readOnlyRootFilesystem — NOT enabled
Left unset. `dfs` writes to `os.TempDir()` (`/tmp` on the container root fs) outside its mounted volumes — the block-store GC engine (`pkg/blockstore/engine/gc.go` uses `os.MkdirTemp("", …)`) and the config default-dir fallback (`pkg/config/config.go`). A read-only root filesystem would break those paths, so it is left off (user may opt in via `Spec.SecurityContext.readOnlyRootFilesystem`).

## Tests
Added `pod_hardening_test.go`:
- `AutomountServiceAccountToken == false` on the reconciled StatefulSet
- secure default container SecurityContext applied when `Spec.SecurityContext` is nil
- user-override merge: user-set fields win, unset fields keep the secure defaults
- user-relaxed override (allowPrivilegeEscalation=true) is honored

Full operator module suite (incl. envtest) green; `go vet` clean; golangci-lint adds no new findings (the module's ~16 pre-existing goconst findings are unchanged).

### DIFF
--- a/k8s/dittofs-operator/internal/controller/dittoserver_controller.go
+++ b/k8s/dittofs-operator/internal/controller/dittoserver_controller.go
@@ -59,6 +59,9 @@ const (
 	defaultAPIPort = 8080
 	// defaultFSGroup is the default fsGroup for pod security context (nonroot user)
 	defaultFSGroup = 65532
+	// capabilityAll is the special capability value that drops all Linux
+	// capabilities from the managed dfs container.
+	capabilityAll corev1.Capability = "ALL"
 )
 
 // retryOnConflict wraps an operation with retry logic for optimistic locking conflicts.
@@ -994,6 +997,11 @@ func (r *DittoServerReconciler) reconcileStatefulSet(ctx context.Context, dittoS
 						},
 					},
 					Spec: corev1.PodSpec{
+						// dfs imports no client-go and never calls the K8s API,
+						// so it has no legitimate consumer for a ServiceAccount
+						// token. Disabling automount keeps the unused namespace
+						// default SA token off the network-exposed container.
+						AutomountServiceAccountToken:  ptr.To(false),
 						SecurityContext:               getPodSecurityContext(dittoServer),
 						TerminationGracePeriodSeconds: ptr.To(getTerminationGracePeriodSeconds(dittoServer)),
 						InitContainers:                initContainers,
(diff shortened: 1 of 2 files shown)


marmos91:
Good catch — fixed in cf74d69f. When the user supplies `Capabilities` with only `Add` (Drop unset), the operator now backfills `Drop: [ALL]` at field granularity so the pod still satisfies the restricted Pod-Security-Standard while preserving the user's `Add` list. Added `TestGetContainerSecurityContext_UserAddsCapabilityKeepsDropBaseline`.

PR #952, opened 2026-06-01
perf(blockstore): index-seek ReadPayloadAt instead of full-log replay (v1.0 audit #1 H5)

marmos91:
## What

`ReadPayloadAt` (`pkg/blockstore/local/fs/readpayload.go`) previously **replayed the entire per-payload append log on every read** — opening the log, seeking past the header, and sequentially decoding + CRC-checking every framed record (O(log-size) per read). This pass uses the per-payload `logIndex` (already maintained for the rollup) to translate the requested file-offset window into the exact set of intersecting record frames and `pread` each one directly.

The index is trimmed at the compaction fence, so it only holds records **not yet rolled into CAS** — exactly the records that live only in the log. Bytes whose records were already consumed by the rollup are served from the CAS manifest in step 2, unchanged. Read semantics (last-write-wins, exact bytes) are byte-identical.

## Why this is correct (and why the prior attempt was reverted)

A previous refactor was reverted because it broke both reads and rollup with `readRecordAt(logPos=64): append log: payloadLen 33554432 exceeds 17825792 cap at logPos=64`. Root causes addressed here:

1. **Stale `logPos`.** `logIndex` entries carry absolute log-byte positions; a rollup *compaction* atomically rewrites the on-disk log **and** rebases `idx.entries[].logPos` under the per-file mutex. Reading a `logPos` without that mutex can `pread` a stale (pre-rebase) position and decode a garbage frame. This pass holds the **same per-file mutex** across the index lookup *and* all preads, so the log/index pair is always consistent. `readRecordAt` additionally cross-checks the frame's declared `payload_len` and `file_offset` against the index entry.

2. **Legitimately large records.** A single `AppendWrite` may exceed `maxRecordPayload` (the engine appends whole writes — e.g. the bench's 32 MiB `mixed-rw` seed lands as one frame). `readRecordAt`/`readRecord` both reject frames above the cap as a torn/hostile-length DoS guard. The pre-index replay used `readRecord`, which **skipped** such frames and let coverage fall through to the CAS/remote-miss path. We preserve that exact behavior (skip, not error) — this is the case whose mishandling (a hard error) broke `mixed-rw`.

3. **Partial-overwrite data loss.** `fillFromCASManifest` now fills only the bytes the log step did not already cover, so a partial in-place overwrite of a rolled-up extent (latest bytes in the log, older bytes in CAS) is not clobbered by the CAS chunk.

## Performance

Reading the first record while the log holds 256× more records grows only ~3× (18.9µs → 58.9µs), versus the previous ~linear-in-log-bytes replay. The residual is the in-memory index scan + fd open, not per-byte log replay.

## Tests

New `readpayload_indexseek_test.go`: index-seek last-write-wins, out-of-order arrival (parallel-write), partial window, **consumed-then-overwritten** (log overlay over CAS), **over-cap record skipped** (the 32 MiB-seed regression), and a no-log-size-scaling assertion.

Verification gate (the integration tag is how the prior bug escaped):
- `go build ./...` — clean
- `go test ./pkg/blockstore/... -count=1` — pass
- `go test -tags=integration -p 1 -timeout=20m ./bench/blockstore/... ./pkg/blockstore/... -count=1` — pass; **`TestRunWorkload_AllWorkloads/mixed-rw` PASSES**, no `payloadLen ... exceeds ... cap` or `rollupFile ... readRecordAt` errors
- `go test -race ./pkg/blockstore/local/fs/... ./pkg/blockstore/engine/... -count=1` — clean
- `gofmt -s`, `go vet ./...`, `golangci-lint run` — 0 issues

### DIFF
--- a/pkg/blockstore/local/fs/readpayload.go
+++ b/pkg/blockstore/local/fs/readpayload.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"io"
 	"os"
 	"sort"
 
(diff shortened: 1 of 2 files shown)


marmos91:
Deferring #1 H5 (index-seek ReadPayloadAt). Two attempts both surfaced read-path correctness failures; root cause is deeper than the read path:

The index-seek path returns stale bytes after an in-place overwrite of a rolled-up extent. The trim (AdvanceFence/trimBelowFence) drops the overwrite's records from the in-memory log index, so the read falls through to the CAS manifest (ListFileBlocks) — which accumulates STALE OVERLAPPING FileBlock rows from multiple rollup generations (rollup never deletes superseded rows when FastCDC boundaries differ between writes). develop masks this because full-log replay still physically covers the window and never consults the manifest.

A correct, byte-identical read-path-only fix isn't achievable without making the manifest authoritative (delete superseded FileBlock rows in the rollup persister) — out of scope for a perf PR, and touching refcount/GC. Closing; #1 H5 stays deferred. The latent manifest-staleness (which also exposes the cold-read path readLocalByHash/findRowCoveringOffset to stale reads after compaction) is filed separately.

PR #955, opened 2026-06-01
fix(blockstore/remote): correct sentinel for invalid GetRange offset/length (#694)

### LINKED ISSUE #694: v1.0 area #1: memory remote GetRange returns wrong sentinel for invalid offset/length
## Problem

`pkg/blockstore/remote/memory/store.go:GetRange` returns `ErrChunkNotFound` (previously `ErrBlockNotFound`) when the chunk exists but `offset`/`length` are out-of-bounds. Contract per `pkg/blockstore/blockstore.go` GetRange godoc requires `ErrInvalidOffset` / `ErrInvalidSize` for malformed bounds.

Also: `length <= 0` (or negative) is not validated.

## Where

`pkg/blockstore/remote/memory/store.go:155` (bounds check returns wrong sentinel).
`pkg/blockstore/remote/memory/store.go:GetRange` (no `length` validation).

## Discovered

Copilot review on PR #684 (B-A2 sentinel merge). Pre-existing bug, out of scope for that PR's zero-logic-change rename.

## Scope

v1.0 Wave 1 area #1 — bundled with PR-B fix waves for the blockstore. Not standalone — fold into B-D bug-fix wave (HIGH/MED tier from `.planning/v1.0-audit/blockstore/REVIEW.md`).

## Acceptance

1. `GetRange` returns `ErrInvalidOffset` when `offset < 0 || offset >= len(data)`.
2. `GetRange` returns `ErrInvalidSize` when `length <= 0 || offset+length > len(data)`.
3. Same fix applied to `s3` remote backend if it has analogous bounds handling.
4. Conformance test in `pkg/blockstore/blockstoretest/` pinning the sentinel choice (currently only `ErrChunkNotFound` is exercised for missing-key — add cases for invalid bounds).

marmos91:
## Problem

`pkg/blockstore/remote/memory` `GetRange` returned the wrong sentinel for malformed bounds:

- Out-of-bounds offset (`offset < 0 || offset >= size`) returned **`ErrChunkNotFound`** — implying the chunk was absent when it exists.
- `length <= 0` was **not validated at all**.

The `BlockStore.GetRange` contract (godoc in `pkg/blockstore/blockstore.go`) requires:
- **`ErrInvalidOffset`** when offset is negative or beyond the chunk size,
- **`ErrInvalidSize`** when length is non-positive,
- partial-past-EOF (valid offset, length runs off the end) is **clamped** to the remaining bytes, no error.

## Fix

- **remote/memory**: invalid offset → `ErrInvalidOffset`; `length <= 0` → `ErrInvalidSize`; partial-past-EOF still clamps.
- **local/memory**: shared the same bug class — it returned a plain `fmt.Errorf` ("negative offset" / "non-positive length") instead of the wrapped sentinel. Now wraps `ErrInvalidOffset`/`ErrInvalidSize` with `%w`, matching FSStore.
- **s3**: had no in-process bounds check. Added up-front validation for negative offset and non-positive length so a malformed `Range` header no longer surfaces an opaque S3 protocol error. A past-EOF offset can't be detected without a HEAD, so S3 keeps surfacing its native 416 there (the conformance suite only requires "any error" for offset >= EOF — unchanged).

## Was S3 affected?

Not the exact `ErrChunkNotFound` regression (it has no in-process bounds check), but it returned opaque errors for negative offset / non-positive length. Hardened to the same sentinel contract.

## Tests

- New `GetRange_InvalidBounds` case in the shared conformance suite (`pkg/blockstore/blockstoretest`) — runs against **every** backend (fs, local/memory, remote/memory, and s3 when env-gated). This caught the local/memory divergence.
- New `TestStore_GetRange_InvalidBounds` regression test in remote/memory pinning all five malformed-bounds cases.
- `go build ./...`, `go test ./pkg/blockstore/...`, and `go test -race ./pkg/blockstore/remote/...` all green; `gofmt`/`go vet`/`golangci-lint` clean.

Closes #694

### DIFF
--- a/pkg/blockstore/blockstoretest/conformance.go
+++ b/pkg/blockstore/blockstoretest/conformance.go
@@ -76,6 +76,7 @@ func BlockStoreConformance(t *testing.T, factory Factory) {
 	t.Run("Put_Concurrent_SameHash", func(t *testing.T) { testPutConcurrent(t, factory) })
 	t.Run("Put_ZeroByte", func(t *testing.T) { testPutZeroByte(t, factory) })
 	t.Run("GetRange_PastEOF", func(t *testing.T) { testGetRangePastEOF(t, factory) })
+	t.Run("GetRange_InvalidBounds", func(t *testing.T) { testGetRangeInvalidBounds(t, factory) })
 	t.Run("Concurrent_Put_Walk_NoDuplicates", func(t *testing.T) { testConcurrentPutWalkNoDuplicates(t, factory) })
 	t.Run("Put_WrongHash_NoVerify", func(t *testing.T) { testPutWrongHashNoVerify(t, factory) })
 }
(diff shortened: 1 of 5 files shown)


marmos91:
Good catch — fixed in bf6f5b20. Both memory backends now compare against `size-offset` (positive, since offset<size is guaranteed) instead of computing `offset+length`, and s3 guards `length > math.MaxInt64-offset` before building the Range header. Added a length-overflow clamp regression test.

Issue #718, opened 2026-05-28
SMB: Sparse FSCTL family (~18 tests)


marmos91:
## Context

Sparse-file FSCTLs (FSCTL_SET_SPARSE, FSCTL_QUERY_ALLOCATED_RANGES, FSCTL_SET_ZERO_DATA) per MS-FSCC §2.3. Second-pass beyond initial IOCTL coverage (#484 closed).

## Failing tests (~18)

- smb2.ioctl.sparse_compressed
- smb2.ioctl.sparse_copy_chunk
- smb2.ioctl.sparse_dir_flag
- smb2.ioctl.sparse_file_flag
- smb2.ioctl.sparse_hole_dealloc
- smb2.ioctl.sparse_lock
- smb2.ioctl.sparse_perms
- smb2.ioctl.sparse_punch
- smb2.ioctl.sparse_punch_invalid
- smb2.ioctl.sparse_qar
- smb2.ioctl.sparse_qar_malformed
- smb2.ioctl.sparse_qar_multi
- smb2.ioctl.sparse_qar_ob1
- smb2.ioctl.sparse_qar_overflow
- smb2.ioctl.sparse_qar_truncated
- smb2.ioctl.sparse_set_nobuf
- smb2.ioctl.sparse_set_oversize
- smb2.ioctl.compress_perms

## Spec / Samba

- MS-FSCC §2.3.27 (FSCTL_SET_SPARSE), §2.3.32 (FSCTL_QUERY_ALLOCATED_RANGES), §2.3.79 (FSCTL_SET_ZERO_DATA)
- Samba: `source3/smbd/smb2_ioctl.c`, `source3/smbd/smb2_ioctl_filesys.c`

## Acceptance

- All sparse tests above PASS in CI smbtorture
- KNOWN_FAILURES.md entries removed
- No regression in other ioctl tests

marmos91:
Fixed by #730 (squash a47520c2).

PR #959, opened 2026-06-01
test(smb): walk back copy_chunk_sparse_dest KNOWN_FAILURE (already passing)

marmos91:
`smb2.ioctl.copy_chunk_sparse_dest` (#750) already passes on current develop — the sparse-hole zero-fill landed via the #718/#730 sparse-FSCTL / append-write refactor and the KNOWN_FAILURES row was never reaped.

Verified by real smbtorture (memory backend, twice → `success: copy_chunk_sparse_dest`) and the existing Go guards:
- `pkg/blockstore/engine/copychunk_sparse_dest_test.go`
- `internal/adapter/smb/v2/handlers/ioctl_copychunk_sparse_integration_test.go::TestCopyChunk_SparseDest_LeadingGapReadsZeros`

Pure KF removal; CI smbtorture/memory must stay green with the row gone.

### DIFF
--- a/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
+++ b/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
@@ -57,16 +57,10 @@ not advertised — they consume no failure slots and are not listed below.
 The compress_notsup_get/set tests correctly SKIP because FILE_FILE_COMPRESSION
 is advertised.
 
-Most IOCTL sparse-family entries walked back under #718. The remaining residual
-failure is a real feature gap: SRV_COPYCHUNK on a sparse destination must surface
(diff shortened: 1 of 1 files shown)


marmos91:
Closing: cannot CI-confirm. copy_chunk_sparse_dest passes only under an explicit smbtorture --filter (agent + the #855 Go tests confirm the zero-fill works), but the default CI battery is inconsistent for it — this run didn't emit it at all and carried an unrelated New Failure=1, and #943 failed it on an older develop base. Keeping the KNOWN_FAILURES row until it can be confirmed green in the full CI battery. The data-plane behavior itself is correct (Go tests pin it).

PR #972, opened 2026-06-01
fix(operator): drop dead jwt.issuer config key + CRD regen no-diff CI gate (v1.0 audit #11)

marmos91:
Operator area-11 audit PR-B5 (config-schema drift) + PR-B6 (CRD regen CI gate). Scoped to `k8s/dittofs-operator/**` (+ the operator's CI workflow for the no-diff gate).

## M-CFG-ISSUER — drop dead `jwt.issuer` key (PR-B5)
The operator rendered `controlplane.jwt.issuer` into the ConfigMap, but the server's `api.JWTConfig` has **no** `Issuer` field — the issuer is hardcoded `"dittofs"` (`pkg/controlplane/api/server.go`). Viper (non-strict decode) silently discarded the key. Removed `Issuer` from:
- the operator config type (`internal/controller/config/types.go` `JWTConfig`)
- the rendering (`config.go` `buildControlPlaneConfig`, plus the now-unused `DefaultJWTIssuer`)
- the CRD spec marker (`api/v1alpha1/dittoserver_types.go` `JWTConfig.Issuer`) → regenerated `config/crd/bases/*` and synced `chart/crds/*`

Checked the other operator-emitted keys for the same dead-key class: `admin.username`/`admin.email` **are** consumed by the server (`pkg/config.AdminConfig`), so they stay. The `cache.*` block is server-dead (round-1 M-CFG-1) but removing it is a larger CRD-API change (`Spec.Cache`/`InfraCacheConfig`) out of this PR's scope; the new round-trip test documents it as a known-dead allowlisted block so it can't silently grow.

## Operator config round-trip drift test
`config_roundtrip_test.go` renders the config for minimal / postgres / full-jwt `DittoServer`s, flattens the YAML to dotted keys, and asserts (1) no dropped dead key (`controlplane.jwt.issuer`) is emitted, and (2) every emitted key maps to a pinned set of real server `pkg/config` keys. Since the operator is a separate Go module, the canonical server key set is pinned in-test — future operator→server config drift fails the test.

## PR-B6 — CRD regen + CI no-diff gate (round-1 M-API-1)
`make manifests generate` produced a diff (the dead `issuer` CRD field removal; metrics field already gone per #938, deepcopy unchanged). Added:
- `make sync-chart-crds` — extracted the `helm` target's chart-CRD `cp` so it has one definition.
- `make manifests-verify` — runs `manifests generate sync-chart-crds` and fails on any non-empty `git diff` of `config/crd`, `api`, `chart/crds`.
- a `Manifests in sync` CI job in `operator-tests.yml` running `make manifests-verify` so chart/config CRDs can't drift from the kubebuilder markers again.

## Verification
- `go build ./...`, `make test` (envtest) green; config pkg 85% coverage.
- `gofmt -s`, `go vet` clean. golangci-lint: 18 pre-existing goconst (non-blocking), **zero new**.
- `make manifests-verify` passes against the committed tree (gate is green).

Out of scope (other PR-B slices / concurrent agents): NEW-H1 deletion liveness, NEW-H2 webhook-cert, RFC7807 contract, chart RBAC, managed-pod hardening (#950), `pkg/config`/`docs` cleanup.

### DIFF
--- a/.github/workflows/operator-tests.yml
+++ b/.github/workflows/operator-tests.yml
@@ -70,3 +70,24 @@ jobs:
 
       - name: Run linter
         run: make lint
+
+  manifests:
+    name: Manifests in sync
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: k8s/dittofs-operator/go.mod
+          cache: true
+
+      # Regenerates CRDs/RBAC/deepcopy from kubebuilder markers, syncs the Helm
+      # chart CRDs, and fails on any non-empty git diff so config/crd, api, and
+      # chart/crds can never drift from the source markers.
+      - name: Verify generated manifests
+        run: make manifests-verify
--- a/.github/workflows/operator-tests.yml
+++ b/.github/workflows/operator-tests.yml
@@ -91,3 +91,9 @@ jobs:
       # chart/crds can never drift from the source markers.
       - name: Verify generated manifests
         run: make manifests-verify
+
+      # The documented "kubectl kustomize config/default | kubectl apply -f -"
+      # install path must render from a clean checkout — guards against a
+      # kustomization referencing a file that is missing from the repo.
+      - name: Verify kustomize build
+        run: kubectl kustomize config/default > /dev/null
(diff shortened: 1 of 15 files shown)


marmos91:
Addressed: (1) removed the leftover `identity.jwt.issuer` references the field-drop missed — the sample CR (`config/samples/dittofs-s3-smb-server-with-secrets.yaml`) and the `CRD_REFERENCE.md` row; (2) widened `manifests-verify` to also diff `config/rbac` so RBAC marker drift is gated (the metrics-scaffold restore + role.yaml are now covered). The cache-block test comment is intentional (cache keys are server-dead but the block stays until the larger `Spec.Cache` CRD-API removal). `make manifests-verify` + config round-trip test green.

PR #974, opened 2026-06-01
test(smb): cover AES-256 encryption derivation, AEAD round-trip, and cipher selection (#686 / #340)

marmos91:
## Summary

Closes the test-coverage gap behind the last two SMB3 encryption KNOWN_FAILURES rows, `smb2.encryption-aes-256-ccm` and `smb2.encryption-aes-256-gcm` (#686 / #340).

**Finding:** AES-256-CCM/GCM is already fully implemented on `develop`. There is no feature gap:

- NEGOTIATE advertises and prefers AES-256 (`defaultCipherPreference` and `DefaultAllowedCiphers` list `0x0003`/`0x0004` first), and `selectCipher` honours client preference order.
- The SP800-108 KDF derives 32-byte encryption/decryption keys for the AES-256 ciphers (`encKeyBits = 256` switch in `DeriveAllKeys`, `kdf.DeriveKey` with `L = 256`).
- The AEAD is built from the 32-byte derived key (`aes.NewCipher` selects AES-256 automatically); the transform-header and middleware paths are cipher-agnostic.

What was missing was an **end-to-end Go test** proving the negotiated cipher ID drives both the 256-bit KDF output and the matching AEAD — the exact class of bug (key-length / AEAD mismatch) that would have produced the reported IO_TIMEOUT hang. That absence is why the KNOWN_FAILURES rows persisted without verification.

## Tests added

- `TestDeriveAllKeys_EncryptKeyLength` — asserts 32-byte enc/dec keys for AES-256, 16-byte for AES-128; signing/application keys fixed at 128-bit.
- `TestDeriveAllKeys_AEADRoundTrip` — full `DeriveAllKeys -> CreateEncryptors -> AEAD encrypt/decrypt` round-trip for all four ciphers, modelling a client encrypting with the client-to-server key and the server decrypting with the derived key.
- `TestSelectCipher` — `selectCipher` selects AES-256 when the client prefers it, respects a server-restricted allowed set, and returns 0 when there is no mutual cipher.

AES-128 coverage is untouched and still passes. `go build ./...` and `go test -race ./internal/adapter/smb/...` are green.

The smbtorture `encryption-aes-256-*` suites require Linux/Docker and are the CI arbiter for clearing the KNOWN_FAILURES rows; the KF entries are intentionally left in place for the orchestrator to remove after CI confirms green.

### DIFF
--- a/internal/adapter/smb/session/crypto_state_aead_test.go
+++ b/internal/adapter/smb/session/crypto_state_aead_test.go
@@ -1,0 +1,113 @@
+package session
+
+import (
+	"bytes"
+	"testing"
+
+	"github.com/marmos91/dittofs/internal/adapter/smb/encryption"
+	"github.com/marmos91/dittofs/internal/adapter/smb/signing"
+	"github.com/marmos91/dittofs/internal/adapter/smb/types"
+)
+
+// TestDeriveAllKeys_EncryptKeyLength asserts that the negotiated cipher ID
+// drives the encryption/decryption key length: 32 bytes for the AES-256
+// ciphers and 16 bytes for AES-128. This is the load-bearing switch for
+// AES-256 support — if the cipher ID does not select a 256-bit KDF output,
+// the server derives a 16-byte key but builds a 256-bit AEAD, which silently
+// produces ciphertext the client cannot decrypt (the IO_TIMEOUT failure mode).
+func TestDeriveAllKeys_EncryptKeyLength(t *testing.T) {
+	sessionKey := bytes.Repeat([]byte{0xAB}, 16)
+	var preauth [64]byte
+	for i := range preauth {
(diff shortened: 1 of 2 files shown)


marmos91:
Superseded by #979 (merged), which implemented the actual AES-256 Kerberos fix (full-32-byte cipher-key derivation) AND shipped its own regression coverage (crypto_state_keylen_test.go asserting AES-256 enc/dec keys depend on bytes 16-31 while signing/app/AES-128 do not). This test-only PR is on a stale pre-#979 base and red on Windows Build; its critical key-length assertion is now covered by #979. The extra TestSelectCipher / AEAD-round-trip coverage can be re-added in a fresh PR if wanted. Closing to keep the queue clean.

PR #978, opened 2026-06-02
test(smb): walk back smb2.ioctl.copy_chunk_sparse_dest (now passing)

marmos91:
## Summary

Walks back `smb2.ioctl.copy_chunk_sparse_dest` from the smbtorture `KNOWN_FAILURES.md` — it now passes deterministically on develop.

## Root cause / why it passes

The test (`source4/torture/smb2/ioctl.c::test_ioctl_copy_chunk_sparse_dest`):
1. src filled with 4096 bytes of a non-zero pattern, dest created 0-byte
2. `FSCTL_SRV_COPYCHUNK` copies src [0,4096) → dest [4096,8192)
3. READ dest [0,4096) must return 4096 **zero** bytes (the hole between old EOF=0 and target offset 4096)
4. READ dest [4096,8192) must equal the source pattern

DittoFS's copychunk path grows the destination via `WriteAt` at the target offset, and the block store is **implicitly sparse**: the zero-grow surfaces the unwritten [old EOF, target offset) hole as zero-reading bytes on subsequent reads — exactly what the test requires. No code change is needed; the row in KNOWN_FAILURES.md was stale.

## Regression coverage

Already guarded by the existing Go test `TestCopyChunk_SparseDest_LeadingGapReadsZeros` (`internal/adapter/smb/v2/handlers/ioctl_copychunk_sparse_integration_test.go`), which drives the real `executeCopyChunks` write path and the real READ handler against a memory block store and asserts zeros in the [0,4096) hole + pattern match in [4096,8192).

## Verification

Local docker smbtorture harness, run against `origin/develop` worktree:

```
success: copy_chunk_sparse_dest
| Metric | Count |
|--------|-------|
| Total | 1 |
| Passed | 1 |
| Failed | 0 |
| New Failures | 0 |
```

Confirmed passing across three independent runs. Go test: `ok internal/adapter/smb/v2/handlers`.

## Changes

- Remove the `smb2.ioctl.copy_chunk_sparse_dest` row from the IOCTL/FSCTL section of `test/smb-conformance/smbtorture/KNOWN_FAILURES.md`; update the surrounding prose to document the walk-back and point at the existing regression test.

### DIFF
--- a/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
+++ b/test/smb-conformance/smbtorture/KNOWN_FAILURES.md
@@ -1,6 +1,6 @@
 # smbtorture Known Failures
 
-Last updated: 2026-05-29 (#743 follow-up — walk back the 6 `smb2.dirlease.*` residuals now PASSing post PR #784: hardlink, rename, unlink_{same,different}_{initial,set}_and_close, v2_request)
+Last updated: 2026-06-02 (walk back `smb2.ioctl.copy_chunk_sparse_dest` — now PASSing on develop; copychunk to a 0-byte dest at offset 4096 surfaces the [0,4096) hole as zeros via the block store's implicit zero-grow. Covered by Go test `TestCopyChunk_SparseDest_LeadingGapReadsZeros`)
 
 Tests listed here are expected to fail and will NOT cause CI to report failure.
 Only NEW failures (not in this list) will cause CI to fail.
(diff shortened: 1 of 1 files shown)


marmos91:
Closing: copy_chunk_sparse_dest passes standalone but FAILS in the full smbtorture battery (this PR's smbtorture/memory: New Failures 1 = copy_chunk_sparse_dest). The row is battery-order-dependent, not a clean standalone pass — keeping it in KNOWN_FAILURES is correct. Standalone --filter verification was insufficient here; the full-battery run is the arbiter.

Issue #680, opened 2026-05-28
v1.0: Wave 2.7 perf-pass — proactive pprof capture under seeded workloads


marmos91:
## Goal

Wave 2 cross-cutting stream that PROACTIVELY captures CPU + heap + block + mutex + goroutine pprofs across every Wave 1 area, under a defined seeded-workload matrix. Drives §Bottlenecks tables in each area's REVIEW.md. Distinct from Wave 2.5 perf-fixes (reactive — fixes the hot spots this stream surfaces).

## Why a separate stream

PLAN.md Wave 2.5 (perf-fixes) is currently the only perf entry. That's reactive — fixes hot spots aggregated from REVIEW.md §Bottlenecks. But §Bottlenecks itself has no driver yet: per-area PR-A audits are read-only static analysis; pprof needs runtime + workloads + bench infra.

This stream owns the capture side. Feeds Wave 2.5.

## Workload matrix (per area)

Built around the seeded workload set defined in PLAN.md §Performance for area #1:

- **(a)** small-files concurrent writes — N workers × M files × 4-64 KiB
- **(b)** small-files concurrent reads — same fileset, randomized reader assignment
- **(c)** big-files concurrent writes — N workers × M files × 64-256 MiB
- **(d)** big-files concurrent reads — random offsets, mixed seq + rand
- **(e)** mixed-ops storm — thousands of seeded random WRITE/READ/LIST/DELETE; \`--seed N\` for replay

Adjustments per area: protocol areas (SMB, NFS) add metadata-burst variant; lock area adds multi-client contention storm; runtime area adds share lifecycle churn.

## Deliverables

Per-area artifacts under \`.planning/v1.0-audit/<area>/_profiles/\`:
- \`cpu.{a-e}.pprof\`
- \`heap.{a-e}.pprof\`
- \`block.{a-e}.pprof\` (requires #671 wiring)
- \`mutex.{a-e}.pprof\`
- \`goroutine.{a-e}.pprof\`
- \`seed.{a-e}.txt\` — params + seed for exact replay

Per-area REVIEW.md §Bottlenecks table: top-5 hotspots as \`file:func cum% flat% notes\` plus proposed fix or \"no-action: contention is intrinsic\".

## Prerequisites

- **#672 bench harness redesign** — new harness must land first. Current harness is broken (Wave 0 baseline exposed 8 failure modes).
[...]

PR #980, opened 2026-06-02
fix(controlplane): wire mutex/block pprof sampling (#671)

### LINKED ISSUE #671: B3 (v1.0 Wave 1): /debug/pprof/{mutex,block} return empty profiles — runtime registration missing
Surfaced during Wave 0 baseline (PR #666).

`runtime.SetMutexProfileFraction` and `runtime.SetBlockProfileRate` are never called in dfs. With `controlplane.pprof: true`, `/debug/pprof/mutex` and `/debug/pprof/block` return 200 with ~235-byte header-only bodies.

**Wave 1 area-pair:** control-plane / pprof. Tie to config flags, e.g.:
- `controlplane.pprof_mutex_rate: 100`  → `runtime.SetMutexProfileFraction(100)`
- `controlplane.pprof_block_rate_ns: 1_000_000` → `runtime.SetBlockProfileRate(1_000_000)`

Both default off when `controlplane.pprof` itself is off.

**Artifacts:** `.planning/v1.0-audit/_baseline/macro/{fs,s3}/{mutex,block}*.pprof` (empty placeholders, kept as documentation).

marmos91:
Closes #671.

## Problem

`runtime.SetMutexProfileFraction` and `runtime.SetBlockProfileRate` are never called in `dfs`. With `controlplane.pprof: true`, `/debug/pprof/mutex` and `/debug/pprof/block` return 200 with ~235-byte header-only bodies — no samples. This blocks #680 (proactive pprof capture), which needs mutex/block profiles across the Wave-1 areas.

## Fix

- Two new config keys, gated entirely on `controlplane.pprof`:
  - `controlplane.pprof_mutex_rate` (default `100` when pprof on, `0`/off otherwise) → `runtime.SetMutexProfileFraction`
  - `controlplane.pprof_block_rate_ns` (default `1_000_000` when pprof on) → `runtime.SetBlockProfileRate`
- Applied to the runtime in `NewServer`. No sampling overhead when pprof is off.
- Env override works automatically via the reflective `configEnvKeys()` walk on `develop` (`DITTOFS_CONTROLPLANE_PPROF_MUTEX_RATE`, `..._BLOCK_RATE_NS`).

## Drift cleanup

`APIConfig` had two independent default sites — `applyDefaults()` (called by `NewServer`) and a hand-duplicated copy in `pkg/config/defaults.go`. The new rate defaults would have silently dropped on the global `Load` path (same class of bug as the area-9 config-show drift). Exported `APIConfig.ApplyDefaults` as the single source of truth; `applyControlPlaneDefaults` now delegates to it. `TestApplyDefaults_ControlPlane*` pins the global path.

## Tests

- `TestAPIConfig_PprofRateDefaults` — off → rates stay 0; on → 100 / 1e6; explicit preserved.
- `TestNewServer_PprofSamplingWired` — asserts the runtime fraction is actually set (read-and-restore, no transient global write).
- `TestApplyDefaults_ControlPlanePprofRates` — global `Load` path defaults the rates.

`go build ./...`, `go vet`, `go test -race ./pkg/controlplane/api/ ./pkg/config/` all green.

### DIFF
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -198,10 +198,13 @@ controlplane:
   write_timeout: 10s         # Max time to write response
   idle_timeout: 60s          # Max idle time for keep-alive
 
-  # Profiling (disabled by default; for benchmarks/diagnostics only)
+  # Profiling (disabled by default; for benchmarks/diagnostics only).
+  # The rate keys only take effect when pprof is true; with pprof off all
+  # sampling stays off regardless of their values. When pprof is on and a rate
+  # is unset/0 it falls back to the default shown below.
   pprof: false                 # Expose /debug/pprof/* endpoints
-  pprof_mutex_rate: 100        # runtime.SetMutexProfileFraction; 0 disables mutex profile
-  pprof_block_rate_ns: 1000000 # runtime.SetBlockProfileRate (ns); 0 disables block profile
+  # pprof_mutex_rate: 100        # runtime.SetMutexProfileFraction (default 100 when pprof on)
+  # pprof_block_rate_ns: 1000000 # runtime.SetBlockProfileRate ns (default 1000000 when pprof on)
 
   # JWT authentication configuration
   jwt:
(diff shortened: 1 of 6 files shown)


marmos91:
Addressed all 5 review comments in 76ea0008:

1. **0-can't-disable contradiction** (`config.go`, table, yaml example) — resolved by making the docs honest rather than adding pointer complexity: when `pprof` is on, an unset/0 rate falls back to the default (100 / 1e6); to disable profiling set `pprof: false`. Field comments, the options table (Default column now `100 (when pprof: true; else 0)`), and the YAML example (rates commented out, note added) are reconciled with the implementation.
2. **`NewServer` left stale sampling when `Pprof: false`** — `NewServer` is now authoritative: the `else` branch resets both rates to 0, so prior callers/bench tooling can't leak sampling. Pinned by new `TestNewServer_PprofOffResetsSampling`.
3. **Test restored/asserted only mutex** — cleanup now also resets the block rate to 0. There is no runtime getter for the block rate, so it can't be read back to assert; noted in the test comment.

`go build`, `go vet`, `go test -race ./pkg/controlplane/api/ ./pkg/config/` green.

Note: the `WPTS BVT / memory` check is a pre-existing SMB-conformance harness failure (artifacts not produced; `develop` is currently red on the same workflow). This PR is controlplane/config-only and cannot affect SMB behavior.

PR #983, opened 2026-06-02
feat(bench): concurrent mixed-ops-storm workload + phase/replay capture (#680)

marmos91:
First slice of #680 (proactive pprof capture). Branch `v1.0/perf-pass`. Builds on the 5-profile envelope that landed with #671.

## What

The block-store/engine capture path: a concurrent **mixed-ops-storm** workload plus the harness flags needed to drive before/after captures.

### `mixed-ops-storm` workload
- `--workers N` goroutines issue a weighted **WRITE/READ/LIST/DELETE** stream (50/30/15/5) — surfaces lock/block contention the existing serial workloads can't.
- **Keyspace partitioned to stay race-free:** a fixed `stable` set (pre-seeded, never deleted) backs READs and WRITE-overwrites; a separate `churn` pool is the only create/delete target. No op ever races a concurrent delete of its own file — verified under `go test -race`.
- Tracks only payload IDs, not block refs: the engine resolves a payload's blocks from its ID internally (`WriteAt`/`ReadAt`/`Delete` accept `nil` refs).
- **Determinism:** each worker's PRNG is seeded from `(seed, worker)`, so `(seed, workers)` reproduces the op *multiset*; goroutine interleaving (and thus the exact per-type tally) is not deterministic at `workers>1` — documented.

### Harness flags
- `--workers` — storm concurrency.
- `--phase baseline|post-fix` — nests the capture dir (`<profile-dir>/blockstore/<phase>/<workload>-<ts>/`) so before/after runs sit side by side. Empty phase preserves the existing flat layout.
- `--replay <dir>` — reloads a recorded run's `seed.txt` (workload, ops, sizes, **workers**, seed, remote, full-profiles) and re-captures without retyping flags; `--phase`/`--profile-dir` still pick the fresh output location. `seed.txt` now records `workers`.

## Verification
- `go build ./...`, `go vet`, `gofmt -l` clean; `go test -race ./bench/blockstore/ ./cmd/bench/` green.
- Live smoke: `--full-profiles --workers 4` emits non-empty cpu/heap/goroutine/mutex/block.pprof + seed.txt; `mutex.pprof` shows real contention samples (76% `sync.Mutex.Unlock`). `--replay` reproduces from `seed.txt`.
- Reviewer-caught bug fixed: `loadSeed` parsed the uint64 seed via signed `Atoi` (seeds > MaxInt64 silently → 0, breaking replay) → now `ParseUint`, with a `math.MaxUint64` round-trip test.

## Scope / follow-ups
This is the blockstore slice only (per the locked decision in `.planning/v1.0-audit/680-perf-capture-SCOPE.md`). Per-area SMB/NFS/lock/runtime capture wrappers reuse this envelope and are follow-ups. #680 stays open until those land. Pure concurrent (a)-(d) read/write variants can also reuse the `--workers` primitive later.

### DIFF
--- a/bench/blockstore/README.md
+++ b/bench/blockstore/README.md
@@ -13,9 +13,37 @@ power both the `dfsbench blockstore` CLI subcommand and the Go
 | `dedup-heavy`      | 8 MiB      | N distinct payloadIDs  | same bytes across files; flush per op so rollup → CAS runs             |
 | `mixed-rw`         | 4 KiB      | seeded 32 MiB file/N   | 50/50 read/write at random offsets                                     |
 | `flush-churn`      | 4 KiB      | monotonic offset       | write→flush→write tight loop                                           |
+| `mixed-ops-storm`  | 4 KiB      | 1 MiB files, ≥4/worker | concurrent WRITE/READ/LIST/DELETE (50/30/15/5); `--workers N`           |
 
 Defaults match the legacy `cmd/blockstore-perf` shape so historical
 results stay comparable. Override per-op size with `--block-size`.
+
+`mixed-ops-storm` is the only concurrent workload. It partitions its
+keyspace — a fixed stable set (pre-seeded, never deleted) backs READs and
+WRITE-overwrites, while WRITE-create/DELETE churn a separate pool — so no
+op ever races a concurrent delete of its own file. Each worker uses a PRNG
+seeded from `(--seed, worker)`, so a given `(seed, workers)` reproduces the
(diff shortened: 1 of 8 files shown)


marmos91:
Addressed the Copilot comment: `--phase` is now validated as a single path element (rejects `..`, `.`, nested, and absolute paths), so it can't escape `<profile-dir>/blockstore/` via `filepath.Join`. Added `TestStartProfileSession_RejectsBadPhase`. The leftover-dir-on-error note is pre-existing and benign for a local bench tool (the started CPU profile *is* rolled back); left out of scope.

Issue #668, opened 2026-05-28
B2 (v1.0 Wave 1): rollup wedges permanently on residual state after interrupted writes


marmos91:
Surfaced during Wave 0 baseline (PR #666). After a partial write fails (`context canceled` from a killed client or restarted server), the rollup loop emits one of two errors in a tight loop forever for the offending payloadID:

```
rollupFile failed payloadID=export/_dfsctl_bench/seq_write_N.dat
  error="rollup: tree/logIndex divergence — stable interval [0,1048576) has no logIndex entries" source=ticker
```

```
rollupFile failed payloadID=export/_dfsctl_bench/seq_write_N.dat
  error="rollup: ObjectIDPersister: engine: object_id already mapped to another file
         Conflict: badger PutFile: object_id already mapped to file <UUID>" source=ticker
```

The mappings are durable in badger; only recovery is to wipe both metadata and block-store dirs and recreate the share. No self-healing path exists.

**Wave 1 area-pair:** file-lifecycle / rollup. Add either a startup sweep that GCs orphan `object_id → file` mappings, or a reconcile path that detects and rebinds.

**Repro:** kill `dfsctl bench run` mid-rand-write on a fresh share → server log floods with one of the two errors above.

**Artifacts:** `.planning/v1.0-audit/_baseline/macro/fs/server-errors.log` (1398 lines).

marmos91:
Resolved by #713 — rollup wedge fix (tree/logIndex divergence + payload state machine).

PR #991, opened 2026-06-02
perf(blockstore): split rollup critical section to unblock concurrent writes (C1, #680)

marmos91:
Fixes the **C1** contention identified in the #680 re-profile (REVIEW.md §5.1): `rollupFile` held a single per-payload mutex across its *entire* pass, so a concurrent `AppendWrite` to the same file blocked for the full rollup duration — **~79% of mutex-wait time** under a concurrent write storm.

### Change
Split the rollup into three phases under **two** per-payload locks:

| Lock | Held for | Purpose |
|---|---|---|
| `rmu` (new `rollupLocks` map) | entire pass (A→C) | serializes same-payload rollups; barrier `DeleteAppendLog` drains so the persister never writes rows for a deleted payload |
| `mu` (append mutex) | Phase A + Phase C only | **released for Phase B** so concurrent `AppendWrite` isn't blocked on rollup I/O |

- **Phase A** (mu): resolve stable interval, pread records into owned buffers, capture `consumedExtents` + `maxReadLogPos`.
- **Phase B** (no mu): reconstruct → chunk → `StoreChunk` → `ObjectIDPersister` (CAS + manifest only).
- **Phase C** (mu): `MarkConsumed`, `AdvanceFenceUpTo`, `SetRollupOffset`, `ConsumeUpToStable`, `maybeCompactLog`.

Lock order `rmu` (outer) → `mu` (inner); `DeleteAppendLog` drains `rmu` then `mu`.

### Lost-write safety (mu released mid-pass)
A write that races Phase B is preserved by two guards moving in lockstep:
- `tree.ConsumeUpToStable(stableEnd, phaseStart)` — consumes only intervals untouched since the pass began.
- `idx.AdvanceFenceUpTo(maxReadLogPos)` — never fences past a record that arrived after the Phase-A snapshot (log positions are monotonic).

A **rare benign** tree/logIndex divergence can result (a stale dirty marker whose offset bytes are already in CAS); the existing #668 handler drops it without loss — proven in the review (a fenced entry is always covered ⇒ its bytes were `StoreChunk`'d before `MarkConsumed`) and documented inline.

### Tests (new `rollup_c1_test.go`)
- `ConsumeUpToStable` unit guards (Touched gating + boundary).
- `WriteDuringPhaseB_NotConsumed` — deterministic lost-write guard via a test-only Phase-B hook.
- `ConcurrentOverwriteStorm` — pool + drain under `-race`.

Full `pkg/blockstore/local/fs` green under `-race` (×3). golangci-lint 0 issues.

Reviewed by code-reviewer (6 concurrency invariants traced airtight; one HIGH found + fixed: `ResetLocalState` now clears `rollupLocks` to keep it 1:1 with `logLocks`).

### DIFF
--- a/pkg/blockstore/local/fs/appendwrite.go
+++ b/pkg/blockstore/local/fs/appendwrite.go
@@ -156,6 +156,12 @@ func (bc *FSStore) getOrCreateLog(payloadID string) (*logFile, *sync.Mutex, *int
 		mu = &sync.Mutex{}
 		bc.logLocks[payloadID] = mu
 	}
+	// C1: pair every payload's append mutex with a rollup mutex, allocated
+	// under the same bc.logsMu write lock so rollupFile always observes a
+	// non-nil rollup lock whenever the append mutex exists.
+	if _, rmuOk := bc.rollupLocks[payloadID]; !rmuOk {
+		bc.rollupLocks[payloadID] = &sync.Mutex{}
+	}
 	if !treeOk {
 		tree = newIntervalTree()
 		bc.dirtyIntervals[payloadID] = tree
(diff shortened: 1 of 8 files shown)


marmos91:
Addressed both Copilot review comments in the latest push:
- **rollup.go** — `phaseStart` is now captured **under `mu`** at the top of Phase A (not before acquiring it), so every interval in the snapshot has `Touched <= phaseStart` (consumed) and only a write that lands after `mu` is released for Phase B is preserved. Eliminates the spurious-preservation/redundant-rollup edge.
- **interval_tree.go** — `ConsumeUpToStable` now stops the Ascend early once `Offset >= endExclusive` (the btree is offset-ordered), avoiding the full O(n) walk.

Re-verified: full `pkg/blockstore/local/fs` green under `-race`, gofmt -s clean.

marmos91:
CI fix: `TestRollup_C1_WriteDuringPhaseB_NotConsumed` was timing-flaky — its second forced rollup could run before the injected Phase-B write aged past the 1ms stabilization window (the earlier `phaseStart`-under-mu tightening exposed it; `-race` masked it locally). Added a stability poll before the second pass. Verified 20× without `-race` + 5× with `-race`. Production code unchanged.

marmos91:
**smbtorture/memory: flaky, not a regression.** Re-ran the job; it fails a *different* single oplock test each run — run 1 `smb2.oplock.exclusive9`, run 2 `smb2.oplock.batch1`. A rotating single-test failure is the timing-flake signature (oplock-break round-trips are latency-sensitive); a real regression would fail the *same* test deterministically. This change is purely blockstore rollup-locking, orthogonal to the SMB oplock state machine — and the other two SMB conformance jobs (WPTS BVT + smbtorture Kerberos) pass. All 15 gating checks green. Merging.

PR #995, opened 2026-06-02
perf(blockstore): shard per-payload append-log state to cut create-path contention (C2, #680)

marmos91:
Fixes the **C2** contention from the #680 re-profile (REVIEW.md §5.1), now the #1 mutex cost after C1 landed.

### Problem
A single `logsMu sync.RWMutex` guarded all **seven** per-payload append-log maps. `getOrCreateLog`'s create-path write-lock serialized *every* new-payload acquisition across the whole store — **~60% of mutex-wait** under a concurrent create/delete storm, scaling super-linearly with workers.

### Change
Shard into `numLogShards` (16) independent `*logShard`, each with its own `RWMutex` + its own seven maps. `shardFor(payloadID)` (FNV-1a, masked) routes each payload to exactly one shard, so all of a payload's state lives under one shard lock — preserving the #668 "tree + logIndex under the same lock" invariant and `getOrCreateLog`'s atomic create.

- Single-payload sites lock the one shard.
- Cross-payload sweeps (`scanAllFiles`, `DrainRollups`, `ResetLocalState`, `Close`) iterate shards **one lock at a time** (no two shard locks held at once → no cross-shard deadlock).
- Lock order unchanged: per-file `rmu`/`mu` before the shard lock (FIX-2 discipline intact).

### Result (storm, 8 workers)
| | total mutex-wait | create-path RWMutex share |
|---|---|---|
| pre-C2 | 54.7s | 60.7% |
| **post-C2** | **19.5s (−64%)** | **13.6%** |

Residual is the intrinsic per-file append mutex (one writer per file), not a global bottleneck.

### Tests
- `TestShardFor_Deterministic` — stability + distribution across shards.
- `TestLogShards_ConcurrentCreateDelete` — concurrent create/delete across many payloads/shards under `-race`.
- Updated `TestGroupCommit_NoLogsMuTouch` lock-order grep gate to also catch the shard lock by its new names.

Full `pkg/blockstore/local/fs` green under `-race`; golangci-lint 0 issues. Reviewed by code-reviewer (6 concurrency invariants traced; 0 CRITICAL/HIGH; 2 IMPORTANT doc/test-gate items fixed) + code-simplifier (stale-comment cleanup).

### DIFF
--- a/pkg/blockstore/local/fs/appendwrite.go
+++ b/pkg/blockstore/local/fs/appendwrite.go
@@ -33,7 +33,7 @@ type logFile struct {
 	// pooled, not rotated); there is no file-rotation path that would
 	// stale-capture lf.f. On error-recovery close (the writeRecord-error
 	// branch in AppendWrite), the entire *logFile is dropped from
-	// bc.logFDs and a fresh one is constructed on next touch — the
+	// the shard's logFDs and a fresh one is constructed on next touch — the
 	// coordinator goes with it.
 	//
 	// Teardown: no explicit Stop is required. The design has no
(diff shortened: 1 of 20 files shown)


marmos91:
Addressed the Copilot comment: updated the stale `isTombstoned` doc — post-C2 each shard's tombstones map is always allocated by newLogShard and DeleteAppendLog is the live writer; the old 'plans up to 09 / nil / no deletion path' wording is gone. Doc-only; build + race still green.

PR #998, opened 2026-06-02
fix(smb/replay): finalize parked CREATE on holder release + boundSessions tracking (#749, #992 followup)

marmos91:
## Summary

Two related fixes in the SMB multichannel / durable-handle break-completion area.

### Finalize parked CREATE on holder release (#749 — flips 6 rows)

A parked durable CREATE now finalizes the moment the conflicting holder *releases* its oplock/lease, not only when the 5s break-wait timer fires. The no-ack break-completion path (`releaseLeaseForHandle`, taken when the holder closes its conflicting handle in response to a break notification instead of ACKing) now wakes `WaitForBreakCompletion` waiters for **file** leases and traditional oplocks too — previously only directory leases were woken. The wake is gated on the released lease having been in the `Breaking` state, so it fires only when a break was actually in flight.

The parked CREATE's resume goroutine then re-runs `completeCreateAfterBreak`, which stores the finalized completion in the replay cache, so the replayed CREATE on a surviving channel returns `OK`/`EXISTED` instead of `FILE_NOT_AVAILABLE`. This is the missing half of the `pending2*-sane` story: #992 made the multichannel session survive the originating-channel disconnect; this makes the parked CREATE finalize on holder release before the 5s timer.

Flips all 6 `smb2.replay.dhv2-pending2*-sane` rows (each verified `success:` on the local docker smbtorture battery):

- `dhv2-pending2n-vs-oplock-sane` / `dhv2-pending2n-vs-lease-sane`
- `dhv2-pending2l-vs-oplock-sane` / `dhv2-pending2l-vs-lease-sane`
- `dhv2-pending2o-vs-oplock-sane` / `dhv2-pending2o-vs-lease-sane`

No regression in the lease / oplock / kernel-oplocks families. `smb2.kernel-oplocks.kernel_oplocks7` (the test the prior comment cited as needing the parked CREATE to stay parked) explicitly accepts BOTH the re-open-first (EXCLUSIVE) and parked-create-first (NONE) orderings, so waking the parked file CREATE on holder release is spec-conformant. Holders that ACK wake via `acknowledgeLeaseBreakImpl`; holders that neither ACK nor close (`smb2.lease.timeout-disconnect`) never reach this release path.

### boundSessions tracking bugs (#992 followup)

- `AnyTrackedSession` falls back to a bound multichannel channel when no owned session exists, so a wrong-SessionId error can be signed on a bound-only transport.
- `UntrackSession` now clears the `boundSessions` entry on a LOGOFF processed on a bound channel, so connection-close no longer treats the session as participating on that transport.

Focused unit tests added for both.

## KNOWN_FAILURES

Removed the 6 `dhv2-pending2*-sane` rows (Deferred bucket → 0, Total 49). Updated the Replay Protection prose and added a Changelog entry. Reconciled with develop's just-merged deferred-open retry (#749 pending1n flip).

## Verification

- `go build ./...` clean; `go test -race ./pkg/adapter/smb/... ./pkg/metadata/lock/... ./internal/adapter/smb/handlers/...` green.
- Each of the 6 pending2 rows confirmed `success:` on the local docker smbtorture battery (single-row runs with full baseline, New failures: 0).
- `lease.timeout` / `v2_complex1` local-docker failures confirmed pre-existing on clean origin/develop (not introduced here); `statopen1` passes in isolation.


marmos91:
Fixed in 08d515db. Reconciled the stale 'deferred under #749' language: bucket-10 prose and the `-windows` table rows now state the `-sane` arms pass (matching the changelog's 'all `*-sane` replay rows now pass'). Only the `-windows` arms remain appendixed (no spec-conformant target). The prior smbtorture/memory red was `smb2.maxfid` exit-124 (per-test wall-clock timeout, bucket 8) — a timing flake orthogonal to this change; re-running on the push.

PR #1022, opened 2026-06-02
fix(controlplane): log swallowed errors on persist paths + discarded-error audit (#1009)

### LINKED ISSUE #1009: v1.0 Wave 2: discarded-error (_ =) sampling audit
Branch `v1.0/err-discard-audit`. Wave-0 found **2,360 `_ =` error discards**. Randomly sample ~50, categorize each (legit cleanup discard vs hidden bug vs should-`%w`-wrap), fix HIGH-risk discards, codify a convention. If the sample surfaces a systemic pattern (a subsystem swallowing a class of error), widen the sweep there. Sentinel discipline + `%w` per CLAUDE.md invariant 6.

marmos91:
Closes #1009.

## Discarded-error (`_ =`) sampling audit

This is the Wave-2 **sampling** audit of discarded-error sites (the tree has ~2360; a full mechanical sweep is not warranted — see the doc). Scope was the non-reserved control-plane / CLI / config / snapshot packages (`cmd/`, `pkg/config/`, `internal/controlplane/`, `pkg/controlplane/` excluding lock-manager, `pkg/snapshot/`, `pkg/api/`).

### Counts
- **~55 sites** enumerated and individually reviewed (full population in the allowed dirs is ~118; every persist/Marshal/Encode/state-flip candidate was read).
- **5 sites fixed** across 3 files / 3 genuine bug classes.
- **~50 left untouched** as benign idioms (no churn).

### Fixes (all log at `Warn`, no behavior change on the success path)
1. `internal/controlplane/api/handlers/shares.go` — share-create swallowed `SetShareAdapterConfig` for the default NFS/SMB adapter configs (share created with no adapter config and zero diagnostic trail).
2. `pkg/controlplane/runtime/adapters/service.go` — `CreateAdapter` rollback `DeleteAdapter` discarded (rollback failure leaves an orphan store row).
3. `pkg/controlplane/runtime/trash/reaper.go` — background reaper swallowed per-share `reapShareAt`/`evictToCap` errors with no internal logging (a persistently failing share would grow its trash bin forever, silently). Loop stays resilient; failures now logged.

### Benign classes left alone
Deferred/read-only `Close()`, `fmt.Fprintf` to terminals, Cobra `MarkFlagRequired(literal)` / flag getters, `json.Encode` into an already-written ResponseWriter, best-effort `os.Remove` cleanup, documented idempotent ops, `viper.BindEnv`, error-path state-flip cleanup, CLI interactive `json.Unmarshal` (nil-guarded downstream).

### Audit doc
`.planning/v1.0-audit/err-discard-audit.md` — sample, benign classes, one noted-but-unchanged candidate (`shares.Delete` runtime RemoveShare), and the recommendation: **no full sweep for v1.0**; instead a focused `_ = .*\.(Set|Update|Delete|Create|Save|Put|Write)` grep per-area plus an `errcheck` lint gate with a benign-idiom exclude-list.

### Gates
`go build ./...`, `go vet`, `gofmt`, and `go test -race` on the touched packages all green.

### DIFF
--- a/.planning/v1.0-audit/err-discard-audit.md
+++ b/.planning/v1.0-audit/err-discard-audit.md
@@ -1,0 +1,99 @@
+# Discarded-error (`_ =`) sampling audit — issue #1009
+
+**Scope.** The tree has ~2360 `_ =` discarded-error sites. This was a *sampling*
+audit (not a full sweep) over the non-reserved control-plane / CLI / config /
+snapshot packages: `cmd/`, `pkg/config/`, `internal/controlplane/`,
+`pkg/controlplane/` (excluding the runtime lock-manager files), `pkg/snapshot/`,
+`pkg/api/`. NFS/SMB/blockstore/metadata/auth/e2e were reserved for other agents
+and not touched.
+
+**Sample size.** ~55 sites enumerated and individually reviewed (the full
+population of discards in the allowed dirs is ~118; every suspicious-looking
+call — Marshal/Encode/persist/Write/state-flip — was read, plus a representative
+slice of the obviously-benign ones).
+
+## Result
+
+| Outcome | Count (sites) |
+|---|---|
+| Benign — left untouched | ~50 |
+| Genuine swallow — fixed | 5 sites across 3 files / 3 bug classes |
+
+### Fixes
+
+1. **`internal/controlplane/api/handlers/shares.go:437,443`** — share-create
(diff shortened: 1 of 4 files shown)


marmos91:
Addressed: rollback now uses a fresh `context.WithTimeout(context.Background(), shutdownTimeout)` instead of the request ctx, so a canceled request cannot abort cleanup. Mirrors the existing `stopAdapter` pattern. (459ebbae)

Issue #1014, opened 2026-06-02
v1.0 area-audit MED/LOW backlog (umbrella)


marmos91:
Catch-all so deferred MED/LOW findings from the Wave-1 area audits aren't lost. PLAN: HIGHs ship in PR-B, MED/LOW filed as `v1.0-followup`. All area + round-2 HIGHs are shipped; this tracks the non-blocking remainder. Per-area detail in each `.planning/v1.0-audit/<area>/REVIEW.md` (+ `REVIEW2.md`).

Approximate remaining (verify against REVIEW files before working any):
- **NFS #4** — ~32 MED / ~33 LOW (v3 fan-out, CREATE non-atomic lookup, EXCLUSIVE idempotency, FSSTAT underflow; v4 CLID_INUSE, EXCHANGE_ID reboot-purge, unbounded v4.0 state; RPC/GSS ReadData length, multi-fragment reassembly, mech-OID; NLM grant-queue races, callback retransmit; portmap unauthenticated DUMP; XDR padding).
- **Metadata #6** — ~2 residual (metadata-service `LockFile` fail-open; badger `SetFilesystemCapabilities` race — overlaps #1006).
- **Lock/ACL #5** — 4 MED / 12 LOW.
- **Runtime #7** — 17 LOW (unauthenticated pprof + `/api/v1/grace`, GetShare live-pointer torn read, settings-watcher no-recover, SMB mount-tracking gap, parallel squash-impl drift, DeleteSnapshot delete-before-wipe).
- **Snapshot #8** — 4 MED / 4 LOW.
- **Config #9** — 4 MED / 5 LOW.
- **CLI #10** — 5 MED / 3 LOW (incl. `store block audit-refcounts`/`gc` exit-0-on-error — same class as the fixed snapshot-create HIGH; verify).
- **Operator #11** — remaining MED/LOW after #940/#971.

Pull individual items into focused issues when a stream picks them up; re-triage any that look HIGH on closer read.

marmos91:
Security verification (#1010, SECURITY.md) surfaced 2 MED gaps to fold into this umbrella (0 HIGH/CRIT):

- **G1 (SF-4)** — NFSv4 identity-mapping failure falls back to UNMAPPED identity (`internal/adapter/nfs/v4/handlers/helpers.go:60-67`) whereas v3 fails closed. Fix: make v4 fail closed on identity-map failure (fail-mode parity with v3).
[...]

PR #1042, opened 2026-06-03
test: coverage gap-fill for pkg/metadata/errors + pkg/controlplane

marmos91:
Part of #1014 (Wave-2 test-coverage stream).

Adds honest unit tests for previously zero-/low-coverage units flagged by the v1.0 roadmap. Test-only — no production code touched. All green under `go test -race`.

## Packages + before/after coverage

| Package | Before | After |
|---|---|---|
| `pkg/metadata/errors` | 0.0% | 100.0% |
| `pkg/controlplane` (top-level) | 0.0% | 96.0% |
| `pkg/controlplane/runtime/identity` | 0.0% | 100.0% |
| `pkg/controlplane/runtime/lifecycle` | 0.0% | 94.0% |
| `pkg/controlplane/runtime/mounts` | 0.0% | 100.0% |
| `pkg/controlplane/store` (helpers) | 2.0% | 6.1%* |

\* Most of `store` is `//go:build integration`-gated (real DB CRUD). This PR covers the non-gated generic GORM helpers (`getByField`, `listAll`, `createWithID`, `deleteByField`), `dedup`, `isUniqueConstraintError`, and `convertNotFoundError` to 100% via an in-memory pure-Go SQLite DB; the package-level number stays low because the integration-gated surface dominates statement count.

## What's covered (real branches, not getters)

- **errors**: every `ErrorCode.String()` arm + Unknown fallback, `StoreError.Error()` with/without path, all 15 factory funcs, all 8 `Is*Error` classifiers including the multi-code arms (NotFound↔LockNotFound, LockConflict↔Locked) and `errors.As` unwrapping of wrapped `StoreError` chains.
- **identity**: full squash-mode matrix (none / root_to_admin / root_to_guest / all_to_admin / all_to_guest), AUTH_NULL→anonymous regardless of mode, input-not-mutated, share-not-found error path.
- **lifecycle**: Serve graceful-shutdown ordering, `serveOnce` no-op on re-entry, adapter-load early-return (shutdown not reached), API-server-failure-triggers-shutdown, `SetAPIServer`-after-Serve panic, and all four `initMachineSID` branches (nil store ephemeral / load stored / first-boot persist / invalid-stored regenerate / read-error tolerate).
- **mounts**: Record/overwrite, Remove(+absent), RemoveByClient, RemoveAllByProtocol, RemoveAll, List/ListByProtocol returns copies, plus a `-race` concurrency exerciser.
- **controlplane**: `New` validation + happy path with/without API server, `EnsureAdminUser` idempotence, `Close`.

## Scope

Did not touch `pkg/controlplane/api/`, `runtime/snapshot.go`, `internal/adapter/`, or `bench/` (reserved for sibling agents).

Gates: `go build ./...`, `go vet`, `go fmt`, `go test -race` all green.

### DIFF
--- a/pkg/controlplane/controlplane_test.go
+++ b/pkg/controlplane/controlplane_test.go
@@ -1,0 +1,119 @@
+package controlplane
+
+import (
+	"context"
+	"testing"
+
+	"github.com/marmos91/dittofs/pkg/controlplane/api"
+	"github.com/marmos91/dittofs/pkg/controlplane/store"
+)
+
+func memConfig() *store.Config {
+	return &store.Config{
+		Type:   store.DatabaseTypeSQLite,
+		SQLite: store.SQLiteConfig{Path: ":memory:"},
+	}
+}
+
+func TestNew_ValidationErrors(t *testing.T) {
+	ctx := context.Background()
+
+	t.Run("nil options", func(t *testing.T) {
+		if _, err := New(ctx, nil); err == nil {
+			t.Error("expected error for nil options")
+		}
+	})
+
+	t.Run("nil database config", func(t *testing.T) {
+		if _, err := New(ctx, &Options{}); err == nil {
+			t.Error("expected error for nil database config")
+		}
+	})
+
+	t.Run("invalid database type", func(t *testing.T) {
+		_, err := New(ctx, &Options{Database: &store.Config{Type: "bogus"}})
+		if err == nil {
+			t.Error("expected error for invalid database type")
+		}
+	})
+}
+
+// Happy path without the API server: store + runtime wired, API nil.
+func TestNew_NoAPIServer(t *testing.T) {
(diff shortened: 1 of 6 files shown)


marmos91:
Thanks — addressed in a8cccad6:
- **pprof globals**: `TestNew_WithAPIServer` now snapshots/restores `runtime.SetMutexProfileFraction`/`SetBlockProfileRate` via a `guardPprofGlobals` helper so the API-server construction can't leak global pprof state into parallel package tests.
- **fail-fast fixtures**: added `mustCreate` (checks `.Error`, `t.Fatalf`) for all setup inserts; the final `db.Find` now checks its error.
- **provided-id path**: the 'honors provided id' subtest now sets the entity ID to match real callers (no empty-PK insert) and asserts the ID isn't overwritten.

PR #1043, opened 2026-06-03
perf(nfs): pre-size/pool COMPOUND response buffer + hot-path alloc reduction

marmos91:
Part of #1014 (Wave-2 perf-fixes stream).

## Context

From the NFS v1.0 perf audit (`.planning/v1.0-audit/nfs/_partial-g-perf.md`).
The headline COMPOUND-buffer hotspot (finding #1, ~27% alloc) was already
shipped in #957 (`buf.Grow(estimatedSize)` in `encodeCompoundResponse`), and
`MaxConnections` bounding (finding #6) shipped there too. This PR closes the
next contained, profile-/grep-backed win in the same stream: **finding #2 —
auth context rebuilt uncached on the read-class v3 ops.**

## Change

`GETATTR`, `READ`, `WRITE` already served the AuthContext from the shared
per-credential cache (`GetCachedAuthContext`). The other seven ops —
`LOOKUP`, `ACCESS`, `READDIR`, `READDIRPLUS`, `READLINK`, `LINK`, `COMMIT` —
called `BuildAuthContextWithMapping` directly, paying a share lookup +
identity-store resolve + identity-mapping on **every** RPC. `LOOKUP` and
`READDIR(PLUS)` are the hottest metadata ops in a directory walk, so this fired
on every step. They now go through the same cache.

The cache key is tightened to cover the **auth flavor** and the **full
credential set** (UID, primary GID, supplementary GIDs), so a cached context is
never returned for a different credential set that happens to share UID/GID
(previously the key was `share:uid:gid` only). The key is built without `fmt`
reflection since it runs on every cached-op RPC.

## Correctness / byte-identical output

The resolved `AuthContext` (`Identity`, `AuthMethod`, `ShareReadOnly`) is
identical to the uncached path — `GetCachedAuthContext` calls
`BuildAuthContextWithMapping` on a miss and returns a field-for-field copy on a
hit. No wire-encoding code is touched, and all seven handlers pass `authCtx`
**read-only** to the metadata service (verified — none mutate it or its
`Identity`). XDR output is byte-identical by construction. Error paths
(`ErrShareAccessDenied`, context cancellation) are preserved; denied lookups are
not cached.

## Benchmark

New `BenchmarkBuildAuthContext_{Uncached,Cached}` (mem store, M1 Max):

```
BenchmarkBuildAuthContext_Uncached    1040 B/op   20 allocs/op
BenchmarkBuildAuthContext_Cached       184 B/op    7 allocs/op
```

→ ~5.6x fewer bytes, ~2.9x fewer allocs per cached-op RPC (~190 ns/op median on
the hit path).

## Gates

- `go build ./...`, `go vet ./internal/adapter/nfs/... ./pkg/adapter/nfs/...`, `go fmt` — clean
- `go test -race ./internal/adapter/nfs/... ./pkg/adapter/nfs/...` — green
- code-simplifier + code-reviewer — applied (fmt-free key builder); reviewer confirmed byte-identical output, no race on the `sync.Map` cache

## Not pursued (audit findings deliberately skipped)

- **#4 `DecodeOpaque` decode scratch** — production decodes from `*bytes.Reader`,
  for which `binary.Read` already takes the non-allocating fast path
  (benchmarked: no win). The returned `make([]byte, len)` is unavoidable (escapes
  to caller).
- **#5 slog INFO gating** — changing log levels is an observability/behavior
  change; out of scope for a zero-behavior-change perf PR.
- **#3 READDIRPLUS per-entry fan-out** and **v4 per-op auth rebuild** — larger
  structural changes (batched store API / net-new v4 cache); left for a dedicated
  pass rather than bundled here.

### DIFF
--- a/internal/adapter/nfs/v3/handlers/access.go
+++ b/internal/adapter/nfs/v3/handlers/access.go
@@ -154,7 +154,7 @@ func (h *Handler) Access(
 	// Step 3: Build AuthContext with share-level identity mapping
 	// ========================================================================
 
-	authCtx, err := BuildAuthContextWithMapping(ctx, h.Registry, ctx.Share)
+	authCtx, err := h.GetCachedAuthContext(ctx)
 	if err != nil {
 		// Check if the error is due to context cancellation
 		if ctx.Context.Err() != nil {
--- a/internal/adapter/nfs/v3/handlers/auth_cache_bench_test.go
+++ b/internal/adapter/nfs/v3/handlers/auth_cache_bench_test.go
@@ -1,0 +1,50 @@
+package handlers_test
+
+import (
+	"testing"
+
+	"github.com/marmos91/dittofs/internal/adapter/nfs/v3/handlers"
+	handlertesting "github.com/marmos91/dittofs/internal/adapter/nfs/v3/handlers/testing"
+)
+
+// BenchmarkBuildAuthContext_Uncached measures the per-RPC cost of rebuilding
+// the auth context from scratch (share lookup + identity-store resolve +
+// identity mapping) — the path taken before the 7 read-class v3 ops were
+// switched to the shared auth cache.
+func BenchmarkBuildAuthContext_Uncached(b *testing.B) {
+	fx := handlertesting.NewHandlerFixture(b)
+	ctx := fx.Context()
+
+	b.ReportAllocs()
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		authCtx, err := handlers.BuildAuthContextWithMapping(ctx, fx.Registry, ctx.Share)
+		if err != nil {
+			b.Fatal(err)
+		}
+		_ = authCtx
+	}
+}
+
+// BenchmarkBuildAuthContext_Cached measures the same path served from the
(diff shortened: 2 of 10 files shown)


marmos91:
Thanks @Copilot — addressed in 5b601139:

**Retained Context/ClientAddr (doc.go:74, finding 1):** Fixed. The slow path now stores a copy with only the request-independent fields (`AuthMethod`/`Identity`/`ShareReadOnly`); it no longer pins the first request's `Context` or `ClientAddr`. The hit path already re-derived those per request, so behaviour is unchanged.

**No TTL / stale permissions after revocation (finding 2):** Acknowledged, but this is a pre-existing property of `authCache` — `GETATTR`/`READ`/`WRITE` have served from this same un-evicted, credential-keyed cache since it was introduced. This PR only routes 7 more read-class ops through the *existing* mechanism; it does not change the cache's lifetime/invalidation semantics. Adding TTL or an identity-store-change invalidation hook is a behaviour change beyond this perf PR's scope and would apply uniformly to all cached ops — tracking it for a follow-up rather than bundling it here.

PR #1044, opened 2026-06-03
feat(bench): latency percentiles + docs consolidation + remote infra orchestrator

marmos91:
Part of #1014 (Wave-2 bench-rebuild completion). Closes the three remaining bench-rebuild roadmap items.

## 1. Latency percentiles in the result schema (done)
Every blockstore runner (serial, storm, concurrent, walk/delete/gc/raw-s3) now records per-op timings via a concurrency-safe `LatencyRecorder`. The orchestrator computes p50/p95/p99 (nearest-rank) via `LatencyFromSamples`. Added as **additive omitempty** fields on `Metrics`:
- `latency` `{p50_ns, p95_ns, p99_ns}`
- `ops_breakdown` `{total, succeeded, failed}`
- `errors` `[{op, offset, error_kind, count}]`

**`schema_version` is NOT bumped** — these are minor additive fields per the documented version contract (older docs lack them, older consumers ignore unknown fields). Tests cover the percentile math (known input → known p50/p95/p99, unsorted, single, rank clamping), the recorder (counts, nil-safe, concurrent, copy-on-read), and additive schema round-trip (with + without the new fields, asserting `schema_version` stays 1 and omitempty holds).

## 2. Docs consolidation: 3 → 1 (merged)
`docs/BENCHMARKS.md` is now canonical. Merged the local perf gates + snapshot scale limits from `test/e2e/BENCHMARKS.md`, added a "Running benchmarks" section (orchestrate / remote / go-bench) accurate to the current `dfsbench` surface, and scrubbed stale gate test names that no longer exist after the Wave-1 blockstore refactor (kept only verified-present tests: `TestBLAKE3AtLeast3xSHA256`, `TestChunker_BoundaryStability_70pct`, `BenchmarkPerfGate_Phase12RandReadRegression`, `TestPerfGate_Phase12_BinarySearchOverhead` — all confirmed passing). `test/e2e/BENCHMARKS.md` is now a one-line pointer; `bench/README.md` is the quick-start pointer.

## 3. Remote infra orchestrator (Scaleway/Pulumi, drive over SSH) — orchestrator + SSH-drive + result-collection complete; mount-runner stubbed
New `dfsbench remote` subcommand, the Go replacement for `scripts/run-bench.sh` + `bench/scripts/run-all.sh`:
- Reads the bench Pulumi stack outputs (public IP for SSH, `serverPrivateIP` for the mount — newly exported from `bench/infra/bench.go`).
- scp's a prebuilt `dfsbench` binary to the host, runs `orchestrate` over SSH, pulls the result JSON back, verifies `schema_version`.
- **SSH/scp use the public IP only**; the private-network IP flows to the remote bench as `DITTOFS_BENCH_MOUNT_IP` (public is never used for the mount path). Verified by a test asserting every SSH/scp destination is the public IP and the private IP appears only as the mount env.
- SSH/scp/Pulumi-output logic sits behind `Executor`/`StackReader` interfaces, table-tested with fakes; `--dry-run` resolves the target and prints the plan without touching the host.
- **Deleted the B4 footgun**: `dittofs-badger-s3.sh` no longer `go build`s a hardcoded `main` branch — it requires prebuilt binaries (`DFS_BIN`/`DFSCTL_BIN`) scp'd by the orchestrator.

**What's stubbed / pending live validation:** the `orchestrate` workloads the remote runner drives run **in-process on the server** (no kernel NFS mount yet), so the resolved private IP is wired through to the remote process but not yet consumed by a mount-based runner — that env var is the plug-in point for a future client-driven mount runner. Provisioning the VMs themselves stays in Pulumi (`pulumi up` in `bench/infra`); `dfsbench remote` drives an already-provisioned host. **Live-infra validation is pending — it drives real Scaleway infra and cannot be exercised in CI.** The SSH/scp/Pulumi-parse/dry-run logic is fully unit-tested with fakes.

## Gates
`go build ./...`, `go vet ./bench/... ./cmd/bench/...`, `gofmt -l` clean (main + `bench/infra` module). `go test -race ./bench/...` green.


marmos91:
Thanks — all addressed in 09b724b3:
- **OpCounts omitempty**: changed to `*OpCounts` so a zero-value struct is no longer always encoded; added an `ops_breakdown` omitempty assertion + a nil guard in the summary table.
- **BenchCommand --manifest**: now gated on `ManifestPath` (the local manifest actually pushed), not `RemoteManifestPath` (which defaults non-empty) — default remote runs use the built-in manifest. Added `TestBenchCommand_NoManifest`.
- **ln -sf identical paths**: skip the symlink when source == dest (was aborting under `set -e` in the default-install case).
- **Error message**: corrected the dfs-server-vs-dfsbench-orchestrator guidance.

Issue #1075, opened 2026-06-03
Audit: DittoFS develop — 233 verified findings (98 HIGH / 89 MED / 46 LOW)


marmos91:
# Audit: DittoFS develop — 233 verified findings across 8 subsystems

Multi-agent audit of `develop` @ `2beda0f2` (v0.17.0). Quick depth — **bugs / security / slop** lenses (structure/perf/bloat/comments/gaps not yet run). Every finding survived adversarial verification.

**Totals: 98 HIGH · 89 MED · 46 LOW (233).**
Full report: `.planning/v1.0-audit/AUDIT-REPORT-develop-2beda0f2.md` (+ `audit-findings-develop-2beda0f2.json`).

## Worst HIGH (auth/access bypass + SSRF + injection + crashes)
- NTLM auth bypass — no-NT-hash user gets authenticated session (`smb/handlers/session_setup.go:1218`)
- Anonymous/null SMB session mapped to UID 0 root (`smb/handlers/auth_helper.go:69`)
- RPCSEC_GSS identity claimed but unverified → COMPOUND runs anonymous (`nfs/v4/handlers/context.go:50`)
- AUTH_NULL grants full write on any non-RO share (`nfs/auth/share_permission.go:55`)
- COPYCHUNK gates DesiredAccess not GrantedAccess (`smb/handlers/ioctl_copychunk.go:250`)
- Cross-session/client hijack: LEASE_BREAK_ACK, reclaimLease, missing CLID_INUSE
- SSRF via client-controlled callback addr (`nfs/v4/state/callback.go:124`)
- NSM MonName → Badger key injection (`metadata/store/badger/clients.go:76`)
- Compound `NextCommand < HeaderSize` panic ×2 (`smb/compound.go:553`, `smb/framing.go:306`)

## Child issues (per subsystem)
- [x] #1076 — SMB protocol (✅ PR #1085)  (21H/13M/5L)
- [x] #1077 — NFSv4 (✅ PR #1087) / v4.1  (19H/9M/8L)
- [x] #1078 — NFSv3 + NLM/NSM/RPC  (15H/13M/7L)
- [x] #1079 — Metadata stores & service  (15H/15M/5L)
- [x] #1080 — Locks & ACL  (7H/1M/2L)
- [x] #1081 — Block / CAS store  (7H/11M/3L)
- [x] #1082 — Controlplane / CLI / operator  (13H/23M/8L)
[...]

PR #1085, opened 2026-06-03
fix(smb): remediate v1.0 audit SMB findings (#1076)

### LINKED ISSUE #1076: Audit · SMB protocol — 21H/13M/5L
## SMB protocol — audit findings

Source: `.planning/v1.0-audit/AUDIT-REPORT-develop-2beda0f2.md` (develop @ `2beda0f2`, quick-depth: bugs/security/slop lenses).

**21 HIGH · 13 MED · 5 LOW** (39 findings). All adversarially verified.

### HIGH

- [ ] **Data race: DOC propagation writes OpenFile.DeletePending without lock** · `bugs` · `internal/adapter/smb/handlers/close.go:396`
  - In the DOC propagation block, `other.DeletePending = true`, `other.DeleteOnCloseParentKey`, and `other.HasDeleteOnCloseParentKey` are written directly to sibling `*OpenFile` structs inside a `sync.Map Range` callback without holding `other.mu`. Concurrent goroutines (e.g. a parallel `Create` calling `isFileDeletePending`) read `existing.DeletePending` at handler.go:1909 also without `mu`.
  - **Fix:** Acquire `other.mu.Lock()` before writing `other.DeletePending`, `other.DeleteOnCloseParentKey`, `other.HasDeleteOnCloseParentKey` and release after. Mirror the pattern already used for `SmbWriteTriggered` in the same file (line 565). Similarly protect reads in `isFileDeletePending`.
- [ ] **Data race: BaseFileDeletePending propagation writes without lock** · `bugs` · `internal/adapter/smb/handlers/close.go:412`
  - `other.BaseFileDeletePending`, `other.BaseFileDeleteParentHandle`, `other.BaseFileDeleteFileName`, `other.DeleteOnCloseParentKey`, and `other.HasDeleteOnCloseParentKey` are written to stream-handle `*OpenFile` structs inside `rangeStreamsOfBase` without `other.mu`. The same fields are read in `isFileOrBaseDeletePending` (handler.go:1949) also without a lock.
  - **Fix:** Acquire `other.mu.Lock()` around the write block in `rangeStreamsOfBase` callback; similarly hold `existing.mu.RLock()` when reading `existing.BaseFileDeletePending` in `isFileOrBaseDeletePending`.
- [ ] **MFsymlink auto-conversion: client-controlled symlink target injected without path validation** · `security` · `internal/adapter/smb/handlers/close.go:912`
  - On CLOSE, checkAndConvertMFsymlink reads a 1067-byte file written by the SMB client, decodes the symlink target via mfsymlink.Decode, and calls metaSvc.CreateSymlink with that target verbatim. ValidateSymlinkTarget (pkg/metadata/validation.go:201) only rejects empty string. No restriction on target content — absolute paths, `../../..` escapes, or `/etc/shadow` are all accepted.
  - **Fix:** In convertToRealSymlink (close.go:873) or in ValidateSymlinkTarget (validation.go:201), reject or sanitize targets that are absolute paths pointing outside the share (e.g., starting with / and not under the share root) or that contain `../` sequences escaping the share tree. Alternatively, require the SMB session to hold FILE_WRITE_ATTRIBUTES + a per-share 'allow-symlink-creation' capability before auto-converting MFsymlinks. At minimum, add a configurable toggle (default: disabled) for MFsymlink auto-conversion so operators control the attack surface.
- [ ] **UnlockAllForSession(sessionID=0) is a no-op for disconnected handles in ProcessAppInstanceId** · `slop` · `internal/adapter/smb/handlers/durable_context.go:1033`
  - ProcessAppInstanceId calls metaSvc.UnlockAllForSession(ctx, h.MetadataHandle, 0) with sessionID=0 when displacing persisted durable handles. doUnlockAllForSession filters by existing[i].SessionID == sessionID; real SMB session IDs are non-zero, so the call removes nothing. Byte-range locks from the displaced disconnected session are never released.
  - **Fix:** Replace with UnlockAllForOpen keyed on the persisted handle's OriginalFileID: metaSvc.UnlockAllForOpen(ctx, h.MetadataHandle, fmt.Sprintf("%x", h.OriginalFileID)). PersistedDurableHandle already carries OriginalFileID (added by buildPersistedDurableHandle). Fall back to h.FileID when OriginalFileID is zero (pre-field legacy rows).
- [ ] **Compound NextCommand < HeaderSize causes panic in ParseCompoundCommand** · `bugs` · `internal/adapter/smb/compound.go:553`
  - ParseCompoundCommand validates NextCommand is 8-byte aligned (line 541) but never checks NextCommand >= header.HeaderSize (64). A value like 8, 16, 24, 32, 40, 48, or 56 passes the alignment guard and reaches `body = data[header.HeaderSize:bodyEnd]` where bodyEnd = NextCommand < 64, producing `data[64:N]` with N < 64 — index out of range panic.
  - **Fix:** Add `if hdr.NextCommand > 0 && hdr.NextCommand < uint32(header.HeaderSize) { return nil, nil, nil, fmt.Errorf("compound NextCommand too small: %d", hdr.NextCommand) }` after the alignment check at line 541.
- [ ] **Compound NextCommand < HeaderSize causes panic in splitCompoundBody** · `bugs` · `internal/adapter/smb/framing.go:306`
  - splitCompoundBody processes the FIRST command of a compound (called from parseSMB2Message in the connection read loop). bodyEnd = min(int(hdr.NextCommand), len(message)). When NextCommand is e.g. 56 and message length >= 56, bodyEnd = 56 and `message[header.HeaderSize:bodyEnd]` = `message[64:56]` — index out of range panic. Same class as the ParseCompoundCommand bug; the framing path has no prior NextCommand < HeaderSize guard either.
  - **Fix:** In splitCompoundBody, guard before slicing: `if hdr.NextCommand > 0 && int(hdr.NextCommand) < header.HeaderSize { body = nil; remaining = nil; return }` — or return an error sentinel and let parseSMB2Message surface it. Alternatively add the same check in parseSMB2Message before calling splitCompoundBody: if hdr.NextCommand > 0 && hdr.NextCommand < uint32(header.HeaderSize), return STATUS_INVALID_PARAMETER.
- [ ] **Related compound subcommand signatures never verified — auth bypass** · `security` · `internal/adapter/smb/compound.go:596`
  - VerifyCompoundCommandSignature is called with the raw parsed hdr before SessionID inheritance. Related subcommands carry SessionID=0xFFFFFFFFFFFFFFFF (sentinel). GetSession(0xFFFFFFFFFFFFFFFF) returns ok=false at line 596-599, causing the function to return nil (skip). SessionID is only resolved to the real value at line 860-862, AFTER the signature check. Result: every related compound subcommand (WRITE, SET_INFO, READ, CLOSE, etc.) on a non-encrypted signing-required session has its signature silently accepted regardless of content.
  - **Fix:** Resolve the session before calling VerifyCompoundCommandSignature, or pass the inherited session ID explicitly. In processRemaining, perform the SessionID inheritance block (lines 859-866) BEFORE the signature verification block (lines 803-810). Alternatively, in VerifyCompoundCommandSignature accept an optional override sessionID and use it when hdr.SessionID is the sentinel 0xFFFFFFFFFFFFFFFF, falling back to s.lastSessionID passed by the caller.
- [ ] **COPYCHUNK access check uses DesiredAccess (pre-DACL) instead of GrantedAccess** · `bugs` · `internal/adapter/smb/handlers/ioctl_copychunk.go:250`
  - `validateCopyChunkAccess` reads `src.DesiredAccess` and `dst.DesiredAccess` (the requested, pre-DACL mask) to gate COPYCHUNK source read and destination write access.
  - **Fix:** In `validateCopyChunkAccess`, change `srcAccess := types.AccessMask(src.DesiredAccess)` → `srcAccess := types.AccessMask(src.GrantedAccess)` and `dstAccess := types.AccessMask(dst.DesiredAccess)` → `dstAccess := types.AccessMask(dst.GrantedAccess)`. Update the log format strings from `src.DesiredAccess`/`dst.DesiredAccess` to `src.GrantedAccess`/`dst.GrantedAccess` accordingly.
- [ ] **COPYCHUNK access gate checks DesiredAccess instead of GrantedAccess** · `security` · `internal/adapter/smb/handlers/ioctl_copychunk.go:250`
  - validateCopyChunkAccess reads src.DesiredAccess and dst.DesiredAccess (lines 250-251) when checking FILE_READ_DATA/FILE_EXECUTE on source and FILE_WRITE_DATA/FILE_APPEND_DATA/FILE_READ_DATA on destination. DesiredAccess is the client-supplied mask before DACL evaluation — it can contain MAXIMUM_ALLOWED or GENERIC_ALL even when the DACL denies the specific bits. Every other access gate in the SMB handler layer (READ line 196, SET_SPARSE line 78, QUERY_INFO line 294, security.go) uses GrantedAccess (the post-DACL intersection stored at CREATE time).
  - **Fix:** Replace src.DesiredAccess with src.GrantedAccess and dst.DesiredAccess with dst.GrantedAccess in validateCopyChunkAccess (lines 250-251 and log lines 256/263/270). The function already has access to both fields via OpenFile. Example: `srcAccess := types.AccessMask(src.GrantedAccess)` / `dstAccess := types.AccessMask(dst.GrantedAccess)`. Also expand GENERIC_* bits before checking if GrantedAccess may contain unresolved generic bits (use the existing hasAccessRight helper or resolveAccessFlags).
- [ ] **COPYCHUNK access check uses DesiredAccess (pre-DACL) instead of GrantedAccess (post-DACL)** · `slop` · `internal/adapter/smb/handlers/ioctl_copychunk.go:250`
  - validateCopyChunkAccess reads src.DesiredAccess and dst.DesiredAccess to gate FILE_READ_DATA/FILE_EXECUTE on source and FILE_WRITE_DATA/FILE_APPEND_DATA/FILE_READ_DATA on destination. Every other access check in the SMB handler layer (READ, WRITE, SET_INFO, QUERY_INFO, SET_SPARSE, SET_ZERO_DATA, QUERY_ALLOCATED_RANGES) uses GrantedAccess (post-DACL intersection computed at CREATE).
  - **Fix:** Change lines 250–251 from `src.DesiredAccess` / `dst.DesiredAccess` to `src.GrantedAccess` / `dst.GrantedAccess`, and update the fmt.Sprintf log lines on 256, 263, 270 to match. Match the pattern used in handleSetSparse (line 78), handleQueryAllocatedRanges (line 161), and handleSetZeroData (line 392).
- [ ] **flushWatcher races on re-arm: timer from old watcher prematurely flushes new watcher** · `bugs` · `internal/adapter/smb/handlers/change_notify.go:1181`
  - `bufferEventLocked` captures `fileID` by value in the `time.AfterFunc` closure; `flushWatcher` then looks up `r.byFileID[fileID]` to get the notify instance. When `Register` replaces an existing watcher (same FileID, line 589-590) and calls `r.unregisterLocked(old)` which calls `old.flushTimer.Stop()`, `Stop()` returns false if the timer already fired (goroutine blocked on `r.mu`). After lock is released, the stale timer goroutine wakes, finds the NEW watcher under `fileID`, drains its (possibly empty) buffer and calls `r.unregisterLocked` on it — permanently evicting the freshly-registered watcher before any real event is delivered.
  - **Fix:** Store the notify pointer itself in the timer closure instead of just fileID, then in `flushWatcher` compare `byFileID[fileID] == notifyPtr` before draining. Alternatively, use a per-watcher generation counter: increment on every `Register`, capture in the closure, and bail in the callback when the stored generation doesn't match.
- [ ] **LEASE_BREAK_ACK missing session ownership check — any session can ack another session's lease break** · `security` · `internal/adapter/smb/handlers/stub_handlers.go:808`
  - handleLeaseBreakAck passes client-supplied LeaseKey directly to LeaseManager.AcknowledgeLeaseBreak with no check that ctx.SessionID matches the session recorded in LeaseManager.sessionMap for that key. Epoch is hardcoded 0, bypassing the epoch staleness guard in acknowledgeLeaseBreakImpl (leases.go:919: `if epoch != 0 && ...`).
  - **Fix:** Before calling AcknowledgeLeaseBreak, verify ctx.SessionID matches the owner: call lm.GetSessionForLease(ack.LeaseKey) and compare to ctx.SessionID; return STATUS_ACCESS_DENIED on mismatch. Also pass the epoch from the wire message (wire format bytes 2-3 in the LeaseBreakAcknowledgment; currently DecodeLeaseBreakAcknowledgment skips them via r.Skip(6)) to AcknowledgeLeaseBreak instead of hardcoded 0, so stale acks are rejected.
- [ ] **Response.Encode silently truncates stub data without error** · `bugs` · `internal/adapter/smb/rpc/dcerpc.go:348`
  - When fragLen > MaxFragmentSize (65535), code clamps fragLen and allocates a smaller buf, but copy(buf[24:], r.StubData) silently truncates stub data. FragLength header field still claims MaxFragmentSize bytes are present. Caller gets no error — returns a syntactically-valid but content-corrupted response.
  - **Fix:** Return an error (and a fault PDU) when stub data exceeds MaxFragmentSize-24, or implement proper multi-fragment response splitting (FlagFirstFrag/FlagLastFrag). At minimum: if fragLen > MaxFragmentSize { return buildFault(callID, 0x1C000009) } before building the response.
- [ ] **Unauthenticated IPC$ access via SessionID=0 bypass** · `security` · `internal/adapter/smb/response.go:344`
  - prepareDispatch skips ALL session validation when reqHeader.SessionID==0 (condition: cmd.NeedsSession && reqHeader.SessionID != 0). handleIPCShare then calls GetSession(0) which always returns the pre-allocated anonymous session (ID 0, created unconditionally in NewManager). The !found guard never fires, so any client that sends TREE_CONNECT with SessionID=0 — before completing SESSION_SETUP — is granted an IPC$ tree connection and can subsequently open and use lsarpc (SID-to-name resolution) and srvsvc (full share enumeration) pipes without authenticating.
  - **Fix:** In handleIPCShare (tree_connect.go ~line 276), after GetSession succeeds, add: if sess.IsNull || sess.IsGuest { return NewErrorResult(types.StatusAccessDenied), nil } to reject unauthenticated/null sessions from IPC$. Alternatively, add a dedicated SessionID==0 early-reject in prepareDispatch for all commands except NEGOTIATE, SESSION_SETUP, and ECHO which legitimately arrive with no session.
- [ ] **Response.Encode silently truncates stub data > 65511 bytes** · `slop` · `internal/adapter/smb/rpc/dcerpc.go:348`
  - When fragLen > MaxFragmentSize (65535), code caps fragLen then allocates buf of that capped size, but copy(buf[24:], r.StubData) silently drops stub bytes beyond offset 65511. FragLength header claims 65535 but actual payload is truncated. No error returned, no fragmentation attempted.
  - **Fix:** Return an error (or panic) if fragLen > MaxFragmentSize rather than silently capping. Alternatively implement multi-fragment responses. At minimum: if fragLen > MaxFragmentSize { return nil, fmt.Errorf("DCE/RPC response too large: %d bytes", fragLen) } and propagate the error up through HandleRequest/ProcessWrite.
- [ ] **defaultSIDMapper package-level var read/written without synchronization — data race** · `slop` · `internal/adapter/smb/handlers/security.go:107`
  - defaultSIDMapper is a package-level *sid.SIDMapper var. SetSIDMapper (line 112) writes it; BuildSecurityDescriptor, principalToSID, ParseSecurityDescriptorWithOptions (lines 156, 282, 437, 447, 556) read it during concurrent request handling. No mutex or atomic protects the pointer.
  - **Fix:** Replace the bare var with a sync/atomic or sync.RWMutex-protected accessor. Simplest: use sync/atomic.Pointer[sid.SIDMapper]. Change var to var defaultSIDMapper atomic.Pointer[sid.SIDMapper] and init via defaultSIDMapper.Store(sid.NewSIDMapper(0,0,0)); update SetSIDMapper to call defaultSIDMapper.Store(m); update all reads to defaultSIDMapper.Load().
- [ ] **Data race: Session identity fields mutated without lock during re-auth** · `bugs` · `internal/adapter/smb/handlers/session_setup.go:1729`
  - tryReauthUpdate writes sess.Username/Domain/User/IsGuest/IsNull (plain struct fields, no mutex) while concurrent goroutines dispatching pipelined requests on the same session read those fields via primeAuthContext/BuildAuthContext without any lock. Same race in reauthKerberosSession (kerberos_auth.go:195-200) for sess.User/Username/Domain/IsGuest/IsNull/ExpiresAt.
  - **Fix:** Add a sync.RWMutex (identityMu) to Session. Write-lock it in tryReauthUpdate and reauthKerberosSession around the identity field assignments. Read-lock it in GetSession callers that read identity fields (primeAuthContext, BuildAuthContext, CaptureOpenerIdentity, IsExpired). Alternatively use atomic.Pointer[models.User] for the User field and atomic stores for the string/bool fields, but a single RWMutex covering all five fields is cleaner.
- [ ] **Auth bypass: user with no NT hash gets authenticated non-guest session without credential check** · `bugs` · `internal/adapter/smb/handlers/session_setup.go:1241`
  - When UserStore.GetUser succeeds and user.Enabled is true but hasNTHash is false (line 1086), execution falls to the 'no NT hash' block at line 1218. CreateSessionWithUser is called (line 1241) and buildAuthenticatedResponse is returned with nil exportedSessionKey. No NTLMv2 challenge-response is ever validated. Any client that knows a valid username but presents any (including garbage) NtChallengeResponse is authenticated as that user.
  - **Fix:** Reject authentication when hasNTHash is false: return NewErrorResult(types.StatusLogonFailure) instead of creating a session. Operators must set NT hashes before users can authenticate. If a transitional grace period is needed, gate it behind an explicit insecure-auth config flag (default false) and log at Error, not Warn.
- [ ] **NTLM auth bypass: user with no NT hash gets full authenticated session** · `security` · `internal/adapter/smb/handlers/session_setup.go:1218`
  - When `hasNTHash=false` (user exists + enabled, but `GetNTHash()` returns no hash) the code falls through to line 1241: `CreateSessionWithUser` is called with `IsGuest=false`, mapping the client-supplied username to a real authenticated identity. No credential is verified. Any client that knows a valid username can authenticate as that user by sending an NTLM AUTHENTICATE with any (or no) NtChallengeResponse.
  - **Fix:** Return `StatusLogonFailure` when `!hasNTHash` (no hash configured = account not usable for NTLM). Remove the 'transitional' path entirely; the admin action `dittofs user passwd` should be required before any authentication is possible. At minimum: treat no-NT-hash identically to wrong-password and call `destroySessionOnReauthFailure` then return `StatusLogonFailure`.
- [ ] **Anonymous/null SMB session mapped to UID 0 (root) in AuthContext** · `security` · `internal/adapter/smb/handlers/auth_helper.go:69`
  - `BuildAuthContext` for `ctx.User==nil && !ctx.IsGuest` (anonymous/null session) sets UID=0, GID=0 on the `metadata.AuthContext`. Any metadata store implementation that grants elevated access to UID 0 (root-bypass in POSIX permission checks) will grant that access to an unauthenticated anonymous SMB connection.
  - **Fix:** Map null/anonymous sessions to a safe unprivileged UID (e.g. 65534 `nobody`) the same as guest, or reject anonymous access at `checkGuestPolicy` when the operator has not explicitly enabled it. At minimum change the `else` branch in `BuildAuthContext` to use UID=65534/GID=65534 instead of 0/0, removing the comment 'use root (for now)'.
- [ ] **SMBAuthenticator validates Type-3 against any pending challenge, not session-specific one** · `slop` · `internal/adapter/smb/auth/authenticator.go:256`
  - `validateAgainstPendingAuths` iterates ALL pending NTLM challenges with `a.pendingAuths.Range` and accepts any matching challenge. There is no session-ID parameter — the Type-3 from connection A can match the challenge issued to connection B.
  - **Fix:** Thread the session ID (or the specific serverChallenge) from the pending-auth store through to `handleNTLMAuthenticate`, look up the specific `pendingNTLMAuth` by that key, and validate against only that entry — same pattern as the real path in `handlers/session_setup.go` which uses `GetPendingAuth(ctx.SessionID, ctx.ConnID)`.

### MED

- [ ] **Lease rollback removes wrong entry when baseName includes ADS stream suffix** · `bugs` · `internal/adapter/smb/handlers/create_post_break.go:964`
  - When `ErrLeaseKeyInUse` fires after `createAction == types.FileCreated` for an ADS open, the rollback calls `metaSvc.RemoveFile(authCtx, parentHandle, baseName)` where `baseName` is the ADS-qualified name (e.g. `"file.txt:StreamName"`) — that is correct for the stream entry. However, if the base file was auto-created by the ADS creation path (create.go lines 1269-1292) it is NOT rolled back, leaving an orphaned empty base file in the metadata store.
  - **Fix:** In the `ErrLeaseKeyInUse` / `FileCreated` branch, after removing the stream entry, also check if the base file was just auto-created (i.e. `isADS && adsBaseFileName != ""`) and remove it if it was not already present before this CREATE (save a `baseCreatedByUs bool` flag in the ADS auto-create block).
- [ ] **WRITE handler never updates PositionInfo (CurrentByteOffset), violating MS-FSA §2.1.5.3** · `bugs` · `internal/adapter/smb/handlers/write.go:444`
  - After a successful write, `openFile.PositionInfo` is not updated. The equivalent READ path calls `recordReadProgress(openFile, req.Offset, uint64(len(readResult.Data)))` at line 416 in read.go.
  - **Fix:** After `CommitWrite` succeeds (after line 424), add: `openFile.PositionInfo = req.Offset + uint64(len(req.Data))`. The existing `h.StoreOpenFile(openFile)` call at line 446 will persist the updated value.
- [ ] **TOCTOU race on mode bits in handleSetSparse and handleSetCompression** · `bugs` · `internal/adapter/smb/handlers/ioctl_sparse.go:101`
  - Both `handleSetSparse` (ioctl_sparse.go:101-118) and `handleSetCompression` (ioctl_fsctl.go:104-132) implement a read-modify-write on `file.Mode`: `GetFile` → compute `newMode` → `SetFileAttributes`. No per-file lock is held across this sequence.
  - **Fix:** Move the mode bit manipulation into `SetFileAttributes` itself as a bitwise OR/AND-NOT patch (pass a `SetModeBit`/`ClearModeBit` field in `SetAttrs`), so the metadata store applies the change atomically under its own transaction. Alternatively, introduce a per-file-handle mutex around the GetFile→SetFileAttributes sequence for these two IOCTLs, matching the pattern used in SET_INFO BasicInformation (set_info.go:396–632).
- [ ] **QueryInfo swallows all metadata/filesystem errors as STATUS_NOT_SUPPORTED** · `bugs` · `internal/adapter/smb/handlers/query_info.go:405`
  - After dispatching to `buildFileInfoFromStore`, `buildFilesystemInfo`, or `BuildSecurityDescriptor`, any returned error is unconditionally mapped to `types.StatusNotSupported` (line 406).
  - **Fix:** Distinguish error types: if `errors.Is(err, types.ErrNotSupported)` return STATUS_NOT_SUPPORTED; otherwise return `common.MapToSMB(err)` or STATUS_INTERNAL_ERROR.
- [ ] **SET_INFO FileBasicInformation clears modeDOSSparse when recomputing mode** · `slop` · `internal/adapter/smb/handlers/set_info.go:333`
  - When fileAttrs != 0, SET_INFO FileBasicInformation calls SMBModeFromAttrs then preserves modeDOSCompressed from the current metadata (line 333: `mode |= curFile.Mode & modeDOSCompressed`). modeDOSSparse (0x200000) is not preserved. Any SET_INFO FileBasicInformation with a non-zero FileAttributes field after FSCTL_SET_SPARSE will silently clear the sparse bit.
  - **Fix:** On line 333, preserve modeDOSSparse alongside modeDOSCompressed: change `mode |= curFile.Mode & modeDOSCompressed` to `mode |= curFile.Mode & (modeDOSCompressed | modeDOSSparse)`. Add a comment mirroring the existing modeDOSCompressed rationale: 'modeDOSSparse is FSCTL-managed (FSCTL_SET_SPARSE); preserve it so SET_INFO does not accidentally clear sparse state.'
- [ ] **NotifyRename double-charges armed ancestor watchers in cross-directory renames** · `bugs` · `internal/adapter/smb/handlers/change_notify.go:1003`
  - `NotifyRename` calls `chargeArmedBuffer` twice when `newParentPath != oldParentPath`: once with `(oldParentPath, {oldFileName, newFileName})` and once with `(newParentPath, {oldFileName, newFileName})`. An armed WatchTree handle whose `WatchPath` is an ancestor of **both** paths matches in both calls; its `BufferedBytes` is incremented by 4 × `encodedNotifyEntrySize` entries instead of the correct 2 (one OLD_NAME entry + one NEW_NAME entry). The over-count triggers premature `Overflowed=true` and fires `OnOverflow`, causing the next CHANGE_NOTIFY on that handle to return `STATUS_NOTIFY_ENUM_DIR` when the buffer has not actually overflowed.
  - **Fix:** For the second `chargeArmedBuffer` call (newParentPath), pass only the names not already charged in the first call. For a cross-directory rename, the correct charge per ancestor watcher is exactly two entries (relativeOldPath + relativeNewPath). One approach: compute the distinct relative paths for each armed ancestor handle once (across both parent paths) and charge only those, rather than calling `chargeArmedBuffer` twice with the full name list.
- [ ] **cleanupAndDelete has no synchronization — concurrent callers double-cleanup same handle** · `bugs` · `internal/adapter/smb/handlers/durable_scavenger.go:136`
  - expireHandles (background ticker goroutine) and HandleConflictingOpen/ForceExpireDurableHandle (SMB handler goroutines) can race to call cleanupAndDelete for the same PersistedDurableHandle simultaneously. No mutex or in-flight tracking on the scavenger struct guards this.
  - **Fix:** Add a sync.Map of in-flight handle IDs: before calling cleanupAndDelete, attempt to Store(h.ID, struct{}{}); if key already present, skip (another goroutine owns it). Delete after cleanup completes. Alternatively serialize all cleanup calls through a single channel.
- [ ] **encodeUTF16LE truncates non-BMP codepoints instead of encoding surrogate pairs** · `bugs` · `internal/adapter/smb/rpc/srvsvc.go:301`
  - binary.LittleEndian.PutUint16(result[i*2:], uint16(r)) silently truncates any rune > 0xFFFF. Characters such as emoji or CJK extension-B produce a wrong UTF-16 code unit and corrupt the wire string.
  - **Fix:** Use golang.org/x/text/encoding/unicode or manual surrogate-pair encoding: for r > 0xFFFF, encode as two uint16 surrogate pair values (0xD800 + high10, 0xDC00 + low10) and resize result accordingly.
- [ ] **parseSIDsFromRequest misparses NULL SID pointer entries in NDR conformant array** · `bugs` · `internal/adapter/smb/rpc/lsarpc.go:369`
  - After skipping the SID-pointer array, the loop unconditionally reads 4 bytes as NDR SubAuthorityCount and then sidSize more bytes as SID binary for every entry in numSids. NDR pointer arrays include NULL pointers (4-byte zero referent) for absent entries; those entries have no following SID data. If any SID pointer slot is NULL, all subsequent SIDs are parsed at the wrong offset.
  - **Fix:** Before jumping to SID data, scan each pointer slot: if pointer is 0 (NULL), skip that slot's SID data (no SubAuthorityCount+SID follows for a NULL pointer referent). Track non-NULL count separately and only loop over those when reading the deferred SID bodies.
- [ ] **encodeUTF16LE truncates supplementary Unicode code points (> U+FFFF)** · `slop` · `internal/adapter/smb/rpc/srvsvc.go:301`
  - encodeUTF16LE converts each rune to uint16 via truncation (uint16(r)). Runes > U+FFFF (e.g. emoji in share names/comments) require surrogate pairs in UTF-16LE but are silently truncated to the low 16 bits, producing invalid UTF-16 and corrupting the NDR wire encoding for any non-BMP character.
  - **Fix:** Replace with a correct surrogate-aware encoder: import golang.org/x/text/encoding/unicode and use unicode.UTF16(unicode.LittleEndian, unicode.IgnoreBOM).NewEncoder(), or manually: if r > 0xFFFF { r -= 0x10000; append uint16(0xD800 + r>>10), uint16(0xDC00 + r&0x3FF) } else { append uint16(r) }.
- [ ] **buildShareEnumLevel1Response NDR conformant array MaxCount uses byte-count instead of wchar count** · `slop` · `internal/adapter/smb/rpc/srvsvc.go:213`
  - nameLen = len(share.Name) + 1 is the UTF-8 byte length + 1. Written as NDR MaxCount and ActualCount for a wchar_t conformant array. NDR §14.3.3.4 specifies conformant array bounds in element count (wchar_t = 2 bytes), not byte count. For pure ASCII names len(UTF-8)==len(UTF-16)/2 so values match by accident. For any multi-byte UTF-8 name (accented chars, CJK) len(share.Name) != wchar count → NDR frame is malformed.
  - **Fix:** Compute nameWcharCount := len([]rune(share.Name)) + 1 (rune count + null terminator in wide chars). Use nameWcharCount for MaxCount/ActualCount fields. Same for commentLen. The actual UTF-16LE bytes appended remain the same.
- [ ] **Malformed SecurityBuffer silently downgrades to guest session** · `bugs` · `internal/adapter/smb/handlers/session_setup.go:93`
  - parseSessionSetupRequest computes bufferStart = int(secBufferOffset) - 64, clamps to sessionSetupFixedSize if negative, then sets SecurityBuffer=nil (no error) when bufferStart+secBufferLength > len(body) (line 98-100). A client sending a plausible StructureSize=25 but secBufferOffset/secBufferLength combination that places the buffer past the end of body gets req.SecurityBuffer=nil. SessionSetup then hits the 'no recognized auth mechanism' branch at line 357 and calls createGuestSession.
  - **Fix:** After computing bufferStart, validate it: if secBufferLength > 0 and bufferStart+int(secBufferLength) > len(body), return nil + fmt.Errorf(...) from parseSessionSetupRequest (so the caller returns StatusInvalidParameter) instead of silently setting SecurityBuffer=nil.
- [ ] **NT hash prefix logged in plaintext to debug log on auth failure** · `slop` · `internal/adapter/smb/handlers/session_setup.go:1134`
  - `logger.Debug(... "ntHashPrefix", fmt.Sprintf("%x", ntHash[:4]) ...)` emits 4 bytes of the server-side NT hash in hex to the debug log on every NTLMv2 validation failure.
  - **Fix:** Remove the `ntHashPrefix` field entirely from the `logger.Debug` call. The username, domain variants tried, and error are sufficient for diagnosis.

### LOW

- [ ] **HandleSMB1Negotiate: SecurityBufferOffset off-by-one (128 vs 129)** · `slop` · `internal/adapter/smb/response.go:1273`
  - SecurityBufferOffset written as 128 with comment '64 (header) + 64 (fixed body)'. Body is `respBody = make([]byte, 65)` (StructureSize=65 per MS-SMB2 §2.2.4), so offset should be 64+65=129 (0x81), not 128 (0x80). Wrong value copied from a plausible-but-incorrect mental model where NEGOTIATE response body is 64 bytes.
  - **Fix:** Change `binary.LittleEndian.PutUint16(respBody[56:58], 128)` to `binary.LittleEndian.PutUint16(respBody[56:58], 129)` and fix the comment to '64 (header) + 65 (fixed body) = 129'.
- [ ] **ReleaseSessionLeases primary-session re-election scans already-removed sessionMap entries** · `bugs` · `internal/adapter/smb/lease/manager.go:517`
  - Phase 1 (lines 465-480) collects keys to release under `RLock`. Phase 1.5 (lines 483-491) calls `ReleaseLease` for each key, which calls `removeLeaseMapping` — removing the session's keys from `lm.sessionMap` under `Lock`. Phase 2 (lines 504-531) then acquires `Lock` and scans `lm.leaseClientGUID` for surviving candidates, reading `lm.sessionMap[clk.KeyHex]` to get `candidateSID`. The guard `candidateSID == sessionID` (line 518) is now dead: the released session's keys were already removed from `sessionMap` by phase 1.5, so `ok` is always false for them. If a concurrent goroutine calls `RequestLease` for the same `clientID/leaseKey` with a new session between phases 1.5 and 2, that new session's `sessionMap` entry could appear as a candidate — potentially electing a session that belongs to a different reconnect of the same client as 'primary' ahead of an older surviving session.
  - **Fix:** Snapshot the set of released keys before phase 1.5 (already have it in `keysToRelease`), then in phase 2 explicitly skip any `clk.KeyHex` that appears in the released-key set rather than relying on the `sessionMap` absence check. Alternatively, hold `lm.mu.Lock()` across both phase 1.5 (ReleaseLease body) and phase 2, but that requires splitting ReleaseLease into a locked variant.
- [ ] **Misplaced doc comment: TestBreakParentHandleLeasesOnCreate_WaitsForAck header prepended to TestMarkLeaseVersionIfUnset_StickyV2** · `slop` · `internal/adapter/smb/lease/manager_test.go:84`
  - Lines 84-90 are the doc comment for TestBreakParentHandleLeasesOnCreate_WaitsForAck (which actually appears at line 393), but are attached as the leading comment of TestMarkLeaseVersionIfUnset_StickyV2 (line 95). Comment text begins '// TestBreakParentHandleLeasesOnCreate_WaitsForAck asserts that...' and describes break-sequencing semantics, not version stickiness.
  - **Fix:** Remove lines 84-90 from above TestMarkLeaseVersionIfUnset_StickyV2 and replace with the correct one-liner: '// TestMarkLeaseVersionIfUnset_StickyV2: a lease first marked V2 stays V2 even when MarkLeaseVersionIfUnset is later called with isV2=false.' (which is already present as lines 91-94). Then add the displaced comment block above TestBreakParentHandleLeasesOnCreate_WaitsForAck at line 393.
- [ ] **SMBAuthenticator silently grants guest when user not found and NtChallengeResponse present** · `slop` · `internal/adapter/smb/auth/authenticator.go:202`
  - `handleNTLMAuthenticate`: when `userStore.GetUser` returns not-found (err != nil || user == nil), the code returns `AuthResult{IsGuest: true}` even when the client presented a full NTLMv2 `NtChallengeResponse`. An unknown user silently becomes a guest rather than receiving `STATUS_LOGON_FAILURE`.
  - **Fix:** Return an error (`ErrAuthenticationFailed` or equivalent) when `user == nil || !user.Enabled`, consistent with `session_setup.go:1256`. Only fall through to guest when the token itself is empty/anonymous (which is already handled above).
- [ ] **`uniqueStrings` copy-pasted verbatim across two packages** · `slop` · `internal/adapter/smb/handlers/session_setup.go:1761`
  - Identical `uniqueStrings` function body in `auth/authenticator.go:313` and `handlers/session_setup.go:1761`. The test file adds a third copy renamed `uniqueStringsTest` (ntlmv2_crossvalidation_test.go:191).
  - **Fix:** Move `uniqueStrings` to the `auth` package (already there), delete the duplicate in `handlers/session_setup.go`, import and call `auth.UniqueStrings` (exported) or inline the simpler `slices.Compact(slices.Sorted(...))` idiom.

---
Part of #1075

marmos91:
Remediates the SMB subsystem findings from the v1.0 develop audit (#1076, Part of #1075). 13 atomic, signed commits; build/vet/gofmt/full unit suite green; independently code-reviewed (no critical, no regressions).

## HIGH
- **fix(smb/auth): bind NTLM pending challenge to (sessionID, connID)** — prevents cross-connection challenge replay (Samba bug-15346 class).
- **fix(smb): reject IPC$ TREE_CONNECT from anonymous/null/guest sessions** — closes unauthenticated IPC$ access.
- **fix(smb): guard compound NextCommand bounds + verify related sub-command signatures** — fixes two unauthenticated panics (`NextCommand < HeaderSize` → negative-length slice) and a signature-verification bypass where related sub-commands skipped integrity checks.
- **fix(smb/copychunk): gate COPYCHUNK access on GrantedAccess not DesiredAccess** — stops copy across handles the client was never granted.
- **fix(smb): make package-level SIDMapper goroutine-safe (atomic.Pointer)** — data race on the global mapper.
- **fix(smb/durable): byte-range lock leak + concurrent cleanup race** — `UnlockAllForSession(…,0)` was an unconditional no-op (locks key on per-open OpenID, not SessionID); now `UnlockAllForOpen` on displaced/expired/REST-force-closed handles; cleanup serialized.
- **fix(smb): lock CLOSE delete-on-close propagation + gate MFsymlink conversion** — data races on `DeletePending`/`BaseFileDeletePending` propagation; MFsymlink auto-conversion now opt-in (`AllowMFsymlink`, default off) and validates the decoded target (`ValidateMFsymlinkTarget` rejects absolute / `..`).
- **fix(smb): roll back ADS auto-created base file on lease rejection** — orphan base file on lease-denied alternate-data-stream create.

## MED
- **fix(smb/query-info): map store errors instead of collapsing to NOT_SUPPORTED**
- **fix(smb/ioctl): atomic mode-bit masks for SET_SPARSE/SET_COMPRESSION** — read-modify-write race; replaced full-Mode overwrite with `ModeOrMask`/`ModeAndNotMask`.
- **fix(smb): preserve modeDOSSparse across FileBasicInformation SET_INFO**
- **fix(smb/notify): guard stale flush-timer re-arm + de-dup cross-dir rename charge**
- **refactor(smb): dedup durable-handle lock OpenID + atomic mode-mask helper** (harden pass)

Each fix lands with a test exercising the fixed path (fails-before/passes-after).

## Known follow-ups (out of scope, non-blocking)
- `AllowMFsymlink` is reachable only via the runtime ShareConfig layer and defaults off everywhere; not yet wired to `models.Share` (DB column + migration), store persistence, or REST. Security objective met (conversion disabled by default + traversal guard); enabling from persisted config is a separate change needing a migration.
- Mode-mask atomicity is applied at the service layer (`file_modify.go`) and is backend-agnostic, but full cross-backend serialization under concurrency has coverage only on the memory store — add a postgres/badger concurrency test.

Report: `.planning/v1.0-audit/AUDIT-REPORT-develop-2beda0f2.md`

Fixes #1076
Part of #1075

### DIFF
--- a/internal/adapter/smb/auth/authenticator.go
+++ b/internal/adapter/smb/auth/authenticator.go
@@ -275,7 +275,7 @@ func (a *SMBAuthenticator) handleNTLMAuthenticate(ctx context.Context, ntlmToken
 		})
 
 		var sessionKey [16]byte
-		var validationErr error = errors.New("ntlm: authentication failed")
+		validationErr := errors.New("ntlm: authentication failed")
 		for _, domain := range domainsToTry {
 			sessionKey, validationErr = ValidateNTLMv2Response(
 				ntHash,
--- a/internal/adapter/smb/auth/authenticator.go
+++ b/internal/adapter/smb/auth/authenticator.go
@@ -30,43 +30,6 @@ const (
 	maxPendingAuths = 1000
 )
 
-// pendingAuthKey correlates the NTLM NEGOTIATE and AUTHENTICATE rounds of a
-// single handshake. It mirrors the handler-layer key (see handler.go): parallel
-// binds on the same session from different connections must not clobber each
-// other, so both the session ID and connection ID participate in the key.
-type pendingAuthKey struct {
-	sessionID uint64
-	connID    uint64
-}
-
-type ctxKeySessionID struct{}
-
-type ctxKeyConnID struct{}
-
-// WithSessionID returns a context carrying the SMB session ID used to correlate
-// the NTLM NEGOTIATE and AUTHENTICATE rounds of a single handshake. Both calls
-// for the same exchange must carry the same value.
-func WithSessionID(parent context.Context, id uint64) context.Context {
-	return context.WithValue(parent, ctxKeySessionID{}, id)
-}
-
-// WithConnID returns a context carrying the TCP connection ID used to correlate
-// the NTLM NEGOTIATE and AUTHENTICATE rounds of a single handshake. Both calls
-// for the same exchange must carry the same value.
-func WithConnID(parent context.Context, id uint64) context.Context {
-	return context.WithValue(parent, ctxKeyConnID{}, id)
-}
-
-func sessionIDFromCtx(ctx context.Context) uint64 {
-	v, _ := ctx.Value(ctxKeySessionID{}).(uint64)
-	return v
-}
-
-func connIDFromCtx(ctx context.Context) uint64 {
-	v, _ := ctx.Value(ctxKeyConnID{}).(uint64)
-	return v
-}
-
 // SMBAuthenticator implements adapter.Authenticator for SMB protocol authentication.
 //
 // It handles SPNEGO-wrapped tokens, detecting whether the client is using NTLM
@@ -85,12 +48,10 @@ type SMBAuthenticator struct {
 	userStore models.UserStore
 
 	// pendingAuths tracks in-progress NTLM handshakes.
-	// Key: pendingAuthKey{sessionID, connID}, Value: *pendingNTLMAuth
+	// Key: session ID (uint64), Value: *pendingNTLMAuth
 	pendingAuths sync.Map
 
-	// nextSessionID generates a fallback session ID when the caller does not
-	// inject one via WithSessionID. The atomic counter guarantees uniqueness so
-	// concurrent fallback negotiations cannot collide.
+	// nextSessionID generates unique session IDs for tracking auth rounds.
 	nextSessionID atomic.Uint64
 }
 
@@ -179,7 +140,7 @@ func (a *SMBAuthenticator) handleNTLMToken(ctx context.Context, ntlmToken []byte
 
 	switch msgType {
 	case Negotiate:
-		return a.handleNTLMNegotiate(ctx, wrapSPNEGO)
+		return a.handleNTLMNegotiate(wrapSPNEGO)
 
 	case Authenticate:
 		return a.handleNTLMAuthenticate(ctx, ntlmToken)
@@ -191,22 +152,16 @@ func (a *SMBAuthenticator) handleNTLMToken(ctx context.Context, ntlmToken []byte
 
 // handleNTLMNegotiate processes an NTLM Type 1 (NEGOTIATE) message.
 // Returns a Type 2 (CHALLENGE) message with ErrMoreProcessingRequired.
-func (a *SMBAuthenticator) handleNTLMNegotiate(ctx context.Context, wrapSPNEGO bool) (*adapter.AuthResult, []byte, error) {
+func (a *SMBAuthenticator) handleNTLMNegotiate(wrapSPNEGO bool) (*adapter.AuthResult, []byte, error) {
 	// Evict stale pending auths before adding a new one
 	a.evictStalePendingAuths()
 
 	// Build challenge
 	challengeMsg, serverChallenge := BuildChallenge()
 
-	// Correlate this handshake by (sessionID, connID). Callers thread these via
-	// WithSessionID/WithConnID. If no session ID was injected, fall back to a
-	// unique counter value so concurrent fallback negotiations cannot collide.
-	sessionID := sessionIDFromCtx(ctx)
-	if sessionID == 0 {
-		sessionID = a.nextSessionID.Add(1)
-	}
-	key := pendingAuthKey{sessionID: sessionID, connID: connIDFromCtx(ctx)}
-	a.pendingAuths.Store(key, &pendingNTLMAuth{
+	// Generate session ID and store pending auth
+	sessionID := a.nextSessionID.Add(1)
+	a.pendingAuths.Store(sessionID, &pendingNTLMAuth{
 		serverChallenge: serverChallenge,
 		createdAt:       time.Now(),
 	})
@@ -251,53 +206,24 @@ func (a *SMBAuthenticator) handleNTLMAuthenticate(ctx context.Context, ntlmToken
 		return &adapter.AuthResult{IsGuest: true}, nil, nil
 	}
 
-	// Try NTLMv2 validation against this session's pending challenge only.
+	// Try NTLMv2 validation against all pending auths
 	ntHash, hasNTHash := user.GetNTHash()
 	if hasNTHash && len(authMsg.NtChallengeResponse) > 0 {
-		// Look up the pending challenge stored for this exact (sessionID, connID)
-		// pair. A direct lookup makes cross-session validation impossible: a
-		// Type-3 from another session cannot match a challenge stored under a
-		// different key.
-		key := pendingAuthKey{sessionID: sessionIDFromCtx(ctx), connID: connIDFromCtx(ctx)}
-		v, ok := a.pendingAuths.Load(key)
-		if !ok {
-			return nil, nil, errors.New("ntlm: no pending authentication for session")
-		}
-		pending := v.(*pendingNTLMAuth)
-		a.pendingAuths.Delete(key) // consume the challenge
+		sessionKey, pending, matchedID := a.validateAgainstPendingAuths(ntHash, authMsg)
+		if pending != nil {
+			// Derive signing key
+			signingKey := DeriveSigningKey(sessionKey, authMsg.NegotiateFlags, authMsg.EncryptedRandomSessionKey)
+			a.cleanupPendingAuth(matchedID)
 
-		hostname, _ := os.Hostname()
-		domainsToTry := uniqueStrings([]string{
-			authMsg.Domain,
-			"",
-			strings.ToUpper(hostname),
-			"WORKGROUP",
-		})
-
-		var sessionKey [16]byte
-		var validationErr error = errors.New("ntlm: authentication failed")
-		for _, domain := range domainsToTry {
-			sessionKey, validationErr = ValidateNTLMv2Response(
-				ntHash,
-				authMsg.Username,
-				domain,
-				pending.serverChallenge,
-				authMsg.NtChallengeResponse,
-			)
-			if validationErr == nil {
-				break
-			}
-		}
-		if validationErr != nil {
-			return nil, nil, errors.New("ntlm: authentication failed")
+			return &adapter.AuthResult{
+				User:       user,
+				SessionKey: signingKey[:],
+				IsGuest:    false,
+			}, nil, nil
 		}
 
-		signingKey := DeriveSigningKey(sessionKey, authMsg.NegotiateFlags, authMsg.EncryptedRandomSessionKey)
-		return &adapter.AuthResult{
-			User:       user,
-			SessionKey: signingKey[:],
-			IsGuest:    false,
-		}, nil, nil
+		// Validation failed - no matching pending auth found
+		return nil, nil, errors.New("ntlm: authentication failed")
 	}
 
 	// User exists but no NT hash - allow without credential validation
@@ -305,6 +231,56 @@ func (a *SMBAuthenticator) handleNTLMAuthenticate(ctx context.Context, ntlmToken
 		User:    user,
 		IsGuest: false,
 	}, nil, nil
+}
+
+// validateAgainstPendingAuths tries to validate the NTLM response against
+// all pending authentication challenges. Returns the session key, the
+// matching pending auth, and the matched session ID on success.
+// On failure, returns zero values and matchedID 0.
+func (a *SMBAuthenticator) validateAgainstPendingAuths(
+	ntHash [16]byte,
+	authMsg *AuthenticateMessage,
+) ([16]byte, *pendingNTLMAuth, uint64) {
+	hostname, _ := os.Hostname()
+	domainsToTry := uniqueStrings([]string{
+		authMsg.Domain,
+		"",
+		strings.ToUpper(hostname),
+		"WORKGROUP",
+	})
+
+	var resultKey [16]byte
+	var matchedPending *pendingNTLMAuth
+	var matchedID uint64
+
+	a.pendingAuths.Range(func(key, value any) bool {
+		pending := value.(*pendingNTLMAuth)
+
+		for _, domain := range domainsToTry {
+			sessionKey, err := ValidateNTLMv2Response(
+				ntHash,
+				authMsg.Username,
+				domain,
+				pending.serverChallenge,
+				authMsg.NtChallengeResponse,
+			)
+			if err == nil {
+				resultKey = sessionKey
+				matchedPending = pending
+				matchedID = key.(uint64)
+				return false // stop iteration
+			}
+		}
+		return true // continue
+	})
+
+	return resultKey, matchedPending, matchedID
+}
+
+// cleanupPendingAuth removes the pending auth entry for the given session ID.
+// Only deletes the specific entry to avoid breaking concurrent handshakes.
+func (a *SMBAuthenticator) cleanupPendingAuth(sessionID uint64) {
+	a.pendingAuths.Delete(sessionID)
 }
 
 // evictStalePendingAuths removes expired entries and enforces the max count.
--- a/internal/adapter/smb/auth/authenticator.go
+++ b/internal/adapter/smb/auth/authenticator.go
@@ -30,6 +30,43 @@ const (
 	maxPendingAuths = 1000
 )
 
+// pendingAuthKey correlates the NTLM NEGOTIATE and AUTHENTICATE rounds of a
+// single handshake. It mirrors the handler-layer key (see handler.go): parallel
+// binds on the same session from different connections must not clobber each
+// other, so both the session ID and connection ID participate in the key.
+type pendingAuthKey struct {
+	sessionID uint64
+	connID    uint64
+}
+
+type ctxKeySessionID struct{}
+
+type ctxKeyConnID struct{}
+
+// WithSessionID returns a context carrying the SMB session ID used to correlate
+// the NTLM NEGOTIATE and AUTHENTICATE rounds of a single handshake. Both calls
+// for the same exchange must carry the same value.
+func WithSessionID(parent context.Context, id uint64) context.Context {
+	return context.WithValue(parent, ctxKeySessionID{}, id)
+}
+
+// WithConnID returns a context carrying the TCP connection ID used to correlate
+// the NTLM NEGOTIATE and AUTHENTICATE rounds of a single handshake. Both calls
+// for the same exchange must carry the same value.
+func WithConnID(parent context.Context, id uint64) context.Context {
+	return context.WithValue(parent, ctxKeyConnID{}, id)
+}
+
+func sessionIDFromCtx(ctx context.Context) uint64 {
+	v, _ := ctx.Value(ctxKeySessionID{}).(uint64)
+	return v
+}
+
+func connIDFromCtx(ctx context.Context) uint64 {
+	v, _ := ctx.Value(ctxKeyConnID{}).(uint64)
+	return v
+}
+
 // SMBAuthenticator implements adapter.Authenticator for SMB protocol authentication.
 //
 // It handles SPNEGO-wrapped tokens, detecting whether the client is using NTLM
@@ -48,10 +85,12 @@ type SMBAuthenticator struct {
 	userStore models.UserStore
 
 	// pendingAuths tracks in-progress NTLM handshakes.
-	// Key: session ID (uint64), Value: *pendingNTLMAuth
+	// Key: pendingAuthKey{sessionID, connID}, Value: *pendingNTLMAuth
 	pendingAuths sync.Map
 
-	// nextSessionID generates unique session IDs for tracking auth rounds.
+	// nextSessionID generates a fallback session ID when the caller does not
+	// inject one via WithSessionID. The atomic counter guarantees uniqueness so
+	// concurrent fallback negotiations cannot collide.
 	nextSessionID atomic.Uint64
 }
 
@@ -140,7 +179,7 @@ func (a *SMBAuthenticator) handleNTLMToken(ctx context.Context, ntlmToken []byte
 
 	switch msgType {
 	case Negotiate:
-		return a.handleNTLMNegotiate(wrapSPNEGO)
+		return a.handleNTLMNegotiate(ctx, wrapSPNEGO)
 
 	case Authenticate:
 		return a.handleNTLMAuthenticate(ctx, ntlmToken)
@@ -152,16 +191,22 @@ func (a *SMBAuthenticator) handleNTLMToken(ctx context.Context, ntlmToken []byte
 
 // handleNTLMNegotiate processes an NTLM Type 1 (NEGOTIATE) message.
 // Returns a Type 2 (CHALLENGE) message with ErrMoreProcessingRequired.
-func (a *SMBAuthenticator) handleNTLMNegotiate(wrapSPNEGO bool) (*adapter.AuthResult, []byte, error) {
+func (a *SMBAuthenticator) handleNTLMNegotiate(ctx context.Context, wrapSPNEGO bool) (*adapter.AuthResult, []byte, error) {
 	// Evict stale pending auths before adding a new one
 	a.evictStalePendingAuths()
 
 	// Build challenge
 	challengeMsg, serverChallenge := BuildChallenge()
 
-	// Generate session ID and store pending auth
-	sessionID := a.nextSessionID.Add(1)
-	a.pendingAuths.Store(sessionID, &pendingNTLMAuth{
+	// Correlate this handshake by (sessionID, connID). Callers thread these via
+	// WithSessionID/WithConnID. If no session ID was injected, fall back to a
+	// unique counter value so concurrent fallback negotiations cannot collide.
+	sessionID := sessionIDFromCtx(ctx)
+	if sessionID == 0 {
+		sessionID = a.nextSessionID.Add(1)
+	}
+	key := pendingAuthKey{sessionID: sessionID, connID: connIDFromCtx(ctx)}
+	a.pendingAuths.Store(key, &pendingNTLMAuth{
 		serverChallenge: serverChallenge,
 		createdAt:       time.Now(),
 	})
@@ -206,24 +251,53 @@ func (a *SMBAuthenticator) handleNTLMAuthenticate(ctx context.Context, ntlmToken
 		return &adapter.AuthResult{IsGuest: true}, nil, nil
 	}
 
-	// Try NTLMv2 validation against all pending auths
+	// Try NTLMv2 validation against this session's pending challenge only.
 	ntHash, hasNTHash := user.GetNTHash()
 	if hasNTHash && len(authMsg.NtChallengeResponse) > 0 {
-		sessionKey, pending, matchedID := a.validateAgainstPendingAuths(ntHash, authMsg)
-		if pending != nil {
-			// Derive signing key
-			signingKey := DeriveSigningKey(sessionKey, authMsg.NegotiateFlags, authMsg.EncryptedRandomSessionKey)
-			a.cleanupPendingAuth(matchedID)
+		// Look up the pending challenge stored for this exact (sessionID, connID)
+		// pair. A direct lookup makes cross-session validation impossible: a
+		// Type-3 from another session cannot match a challenge stored under a
+		// different key.
+		key := pendingAuthKey{sessionID: sessionIDFromCtx(ctx), connID: connIDFromCtx(ctx)}
+		v, ok := a.pendingAuths.Load(key)
+		if !ok {
+			return nil, nil, errors.New("ntlm: no pending authentication for session")
+		}
+		pending := v.(*pendingNTLMAuth)
+		a.pendingAuths.Delete(key) // consume the challenge
 
-			return &adapter.AuthResult{
-				User:       user,
-				SessionKey: signingKey[:],
-				IsGuest:    false,
-			}, nil, nil
+		hostname, _ := os.Hostname()
+		domainsToTry := uniqueStrings([]string{
+			authMsg.Domain,
+			"",
+			strings.ToUpper(hostname),
+			"WORKGROUP",
+		})
+
+		var sessionKey [16]byte
+		var validationErr error = errors.New("ntlm: authentication failed")
+		for _, domain := range domainsToTry {
+			sessionKey, validationErr = ValidateNTLMv2Response(
+				ntHash,
+				authMsg.Username,
+				domain,
+				pending.serverChallenge,
+				authMsg.NtChallengeResponse,
+			)
+			if validationErr == nil {
+				break
+			}
+		}
+		if validationErr != nil {
+			return nil, nil, errors.New("ntlm: authentication failed")
 		}
 
-		// Validation failed - no matching pending auth found
-		return nil, nil, errors.New("ntlm: authentication failed")
+		signingKey := DeriveSigningKey(sessionKey, authMsg.NegotiateFlags, authMsg.EncryptedRandomSessionKey)
+		return &adapter.AuthResult{
+			User:       user,
+			SessionKey: signingKey[:],
+			IsGuest:    false,
+		}, nil, nil
 	}
 
 	// User exists but no NT hash - allow without credential validation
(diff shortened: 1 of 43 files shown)


marmos91:
## Review follow-up — both Copilot findings addressed + a CI-caught regression reverted

**Copilot #1 (`file_modify.go` — ModeOrMask/ModeAndNotMask could smuggle SUID/SGID):** valid, fixed in `264b8df0`. The masks are now whitelisted to the high-word DOS attribute bits (`dosAttributeModeBits`) before being applied, so setid/permission bits can't bypass the SUID/SGID stripping. Added `TestSetFileAttributes_ModeOrMask_CannotSmuggleSetid` proving it.

**Copilot #2 (`authenticator.go` — NEGOTIATE fallback sessionID vs AUTHENTICATE no-fallback asymmetry):** correct observation, but `auth.SMBAuthenticator` is **not the production NTLM path** — `NewSMBAuthenticator` is constructed nowhere outside its own unit tests. The live path is the handler's `pendingAuthKey{SessionID, ConnID}`, which was **already correctly composite-keyed on develop** (pre-PR). The `fix(smb/auth)` commit edited dead code with a misleading message, so I **reverted it** (`482c6317`/`efdf3b0e`). Removal of the unused `SMBAuthenticator` is tracked as a follow-up.

**Regression caught by smbtorture (not from Copilot):** `fix(smb): reject IPC$ TREE_CONNECT from anonymous/null/guest` broke `smb2.anon-signing1`, `smb2.anon-signing2`, `smb2.anon-encryption2` (green on develop). Anonymous IPC$ connect is legitimate SMB2 (null-session pipe enumeration; the boundary is per-pipe RPC auth, not the tree-connect). The audit finding was a false positive for SMB2 semantics — **reverted** (`9f210e1e`).

Net: the 11 genuinely-correct SMB fixes + the mode-mask hardening remain. Since this PR merges with squash, the inline reverts collapse cleanly.

PR #1106, opened 2026-06-14
perf(nfs): single-alloc RPC reply; reader-free AUTH_UNIX; direct-write fattr encode

marmos91:
Fixes three verified allocation findings in the NFS RPC/XDR layer. All changes preserve the exact wire output.

## Findings & fixes

1. **`rpc/parser.go` `MakeSuccessReply`** — previously did two heap allocations (header buffer + final result buffer) plus a full-payload copy per reply. Now marshals the header and payload into a single buffer that reserves 4 leading placeholder bytes for the record-marking fragment header, then patches those bytes in place via `binary.BigEndian.PutUint32`. Eliminates the second allocation and the full-payload copy. Last-fragment bit (`0x80000000`) and fragment length (record bytes minus the 4-byte marker) are unchanged.

2. **`rpc/auth.go` `ParseUnixAuth`** (LOW) — dropped the per-request `bytes.NewReader` + reflection-based `binary.Read`. Now decodes directly from the `[]byte` body with a manual offset and `binary.BigEndian.Uint32`, with explicit bounds checks. Error handling is equivalent: truncated/short input returns an error (`io.ErrUnexpectedEOF`-wrapped), and trailing padding past the end of the body is tolerated exactly as before.

3. **`xdr/encode.go` `EncodeFileAttr`** (LOW) — replaced 16 reflection-based `binary.Write` calls (and the `SpecData` reflect path) with direct `binary.BigEndian.PutUint32/64` writes into a fixed 84-byte stack array, followed by a single `buf.Write`. **Byte-for-byte identical output is required** — a golden-byte vector captured from the previous implementation is asserted in the new test.

## Tests
- `xdr/encode_fattr_test.go`: golden-byte assertion (84-byte fattr3) + nil-error case.
- `rpc/rpc_test.go`: `MakeSuccessReply` wire-format + record-marker + XDR round-trip + verbatim payload tail; `ParseUnixAuth` truncation sweep (every strict prefix rejected). Existing `ParseUnixAuth` round-trip coverage retained.

`go build ./...`, `go vet ./internal/adapter/nfs/...`, and `go test [-race] ./internal/adapter/nfs/...` all green.

### DIFF
--- a/internal/adapter/nfs/rpc/auth.go
+++ b/internal/adapter/nfs/rpc/auth.go
@@ -1,9 +1,9 @@
 package rpc
 
 import (
-	"bytes"
 	"encoding/binary"
 	"fmt"
+	"io"
 )
 
 // Authentication Flavors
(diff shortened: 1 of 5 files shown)


marmos91:
Addressed in 37175b9b: inlined the supplementary-GID read so the per-element fmt.Sprintf only runs on the error path, keeping the success path allocation-free.

PR #1108, opened 2026-06-14
perf(nfs/v4): cache caps/blocked-ops; prealloc READ buffer; emit . and .. in pseudofs READDIR

marmos91:
Addresses verified perf/correctness findings in NFSv4 handlers (excludes \`open.go\`, owned by another PR).

## Findings fixed

1. **getattr.go:49 — per-request capabilities store read.** GETATTR called \`GetFilesystemCapabilities\` (a store read transaction) on every invocation. Filesystem capabilities are store-level static config, so they are now populated once into the attrs package's existing process-global atomics (\`SetFilesystemCapabilities\`) at adapter startup and refreshed on SettingsWatcher change events. GETATTR no longer touches the store for caps.

2. **handler.go:351 — \`IsOperationBlocked\` per-op JSON unmarshal + linear scan.** Previously read live settings (\`GetBlockedOperations\` → JSON unmarshal) and linearly scanned + name-resolved on every COMPOUND op. Now it does a single map lookup against the \`blockedOps\` set already parsed once by \`SetBlockedOps\`. The adapter refreshes that cache on settings change (new \`OnNFSSettingsChange\` callback, symmetric to the existing SMB one) and at startup/per-connection.

3. **read.go:222 — unpreallocated \`bytes.Buffer\` in \`encodeRead4resok\`.** Now pre-sized to the exact READ4resok wire size (\`status(4) + eof(4) + opaque-len(4) + data + pad\`) so the response buffer does not reallocate per READ.

4. **readdir.go:266 — pseudofs READDIR omitted \`.\` and \`..\`.** Now synthesizes \`.\` (cookie 1, self attrs) and \`..\` (cookie 2, parent attrs — root's parent is itself), with children shifted to start at cookie 3. Cookie continuation across multiple READDIR calls is preserved (entries with cookie <= requested cookie are skipped).

## Tests

- \`TestReadDir_PseudoFSRoot\` updated: asserts \`.\`/\`..\`/children ordering and cookies 1,2,3,4.
- \`TestReadDir_PseudoFSRoot_CookieContinuation\` added: resume after cookie 2 returns only children at cookies 3,4.
- \`TestIsOperationBlocked_CacheHit\` added: verifies the cached blocked-ops set (set, replace, clear, unknown-name ignore).

## Validation

\`go build ./...\`, \`go vet\`, and \`go test -race\` green on \`internal/adapter/nfs/v4/...\`, \`pkg/adapter/nfs\`, and \`pkg/controlplane/runtime\`.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/compound_test.go
+++ b/internal/adapter/nfs/v4/handlers/compound_test.go
@@ -2586,3 +2586,40 @@ func encodeSetClientIDArgsForReject() []byte {
 	_ = xdr.WriteUint32(&buf, 0)                // callback_ident
 	return buf.Bytes()
 }
+
+// TestIsOperationBlocked_CacheHit verifies that IsOperationBlocked consults the
+// cached blocked-ops set populated by SetBlockedOps (the hot-path optimization),
+// rather than re-parsing live settings on each call.
+func TestIsOperationBlocked_CacheHit(t *testing.T) {
+	h := newTestHandler() // Registry is nil -> purely cache-driven
+
+	// Nothing blocked initially.
+	if h.IsOperationBlocked(types.OP_READ) {
+		t.Fatalf("OP_READ unexpectedly blocked before SetBlockedOps")
+	}
+
+	// Block READ and DELEGPURGE via their human-readable names.
+	h.SetBlockedOps([]string{types.OpName(types.OP_READ), types.OpName(types.OP_DELEGPURGE)})
+
+	if !h.IsOperationBlocked(types.OP_READ) {
+		t.Errorf("OP_READ should be blocked after SetBlockedOps")
+	}
+	if !h.IsOperationBlocked(types.OP_DELEGPURGE) {
+		t.Errorf("OP_DELEGPURGE should be blocked after SetBlockedOps")
+	}
+	if h.IsOperationBlocked(types.OP_WRITE) {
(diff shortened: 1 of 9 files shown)


marmos91:
Thanks for the review — addressed in 2358263a:

- **settings.go:108 (unbounded ctx):** `applyFilesystemCapabilities` now uses a bounded `context.WithTimeout` (5s) so a slow/hung metadata backend can't stall startup or the poll goroutine.
- **settings.go:116 (per-share caps):** Documented the assumption. The advertised caps are intentionally a single process-global value (matching the pre-existing attrs-package atomic design); the metadata stores return these as static server-wide config. Added a note that if a future deployment backs shares with materially different store limits, caching should move to a per-store keyed cache / the service layer.
- **settings_watcher.go:60 (no test coverage):** Added `TestSettingsWatcher_NFSCallbackFires` and `TestSettingsWatcher_NFSCallbackPanicRecovers`, mirroring the SMB coverage (fires on version change after initial load; panic does not kill polling).
- **ops_test.go decode/bounds:** Both READDIR tests now `t.Fatalf` on cookie decode errors and assert the cookie-slice length before indexing.
- **settings.go:67 comment:** The optimization it describes *is* implemented — `Handler.IsOperationBlocked` (handler.go) now does a map-only lookup against the cached `blockedOps` set; it no longer reads live settings / unmarshals per op. The comment is accurate.

Issue #1104, opened 2026-06-14
Re-audit (verified): 44 findings — perf/correctness across NFS/SMB/metadata


marmos91:
Verified findings from the v1.0 **re-audit** (perf/structure/bloat/comments/gaps lenses), salvaged from an interrupted run and re-checked by an adversarial verify gate against a clean immutable worktree of `origin/develop`. 64 HIGH-labeled findings → **44 confirmed real** (20 refuted: already-fixed / dead-code / spec-correct / god-object-smell / misread). Severities re-graded during verify.

**4 HIGH / 33 MED / 7 LOW**

## Fixed
- [x] `smb/tree_connect.go` unauthenticated TREE_CONNECT → read-write authz bypass — **PR #1103**

## HIGH (3)

- [ ] **`internal/adapter/nfs/v4/handlers/open.go:134`** — EXCLUSIVE4 mapped to GUARDED4 — verifier idempotency broken, client retry returns NFS4ERR_EXIST
  - _Fix:_ Store the EXCLUSIVE4 verifier in file metadata (e.g. a reserved xattr or mtime pair); on EXCLUSIVE4 open of existing file compare stored verifier: match → success, mismatch → NFS4ERR_EXIST
- [ ] **`internal/adapter/nfs/v4/handlers/open.go:300`** — OPEN UNCHECKED4 on existing file ignores createAttrs — size=0 truncation skipped
  - _Fix:_ After opening existing file under UNCHECKED4, if createAttrs is non-nil call metaSvc.SetFileAttributes(authCtx, fileHandle, createAttrs) to apply size, mode, and other attributes before returning
- [ ] **`internal/adapter/nfs/v4/handlers/open.go:140`** — EXCLUSIVE4 create silently downgraded to GUARDED4, breaking crash-recovery idempotency
  - _Fix:_ Store the verifier (e.g. as initial file mtime or an extended attribute) and pass it through to the metadata store's EXCLUSIVE4 path; on retry, compare stored verifier — if match, return existing stateid; if mismatch, return NFS4ERR_EXIST.

## MED (33)

- [ ] **`cmd/dfsctl/commands/share/mount_unix.go:136`** — Linux SMB mount injects raw username/password into comma-delimited -o options string
  - _Fix:_ Escape username and password for CIFS opts (percent-encode or quote as per CIFS driver spec), or pass credentials via a credentials file (mount.cifs -o credentials=<file>) to avoid shell/option injection entirely
[...]

PR #1110, opened 2026-06-14
perf(smb/signing): no-copy outbound signing via SignInPlace

marmos91:
## Summary

Eliminates a full-message heap allocation on every signed outbound SMB2 PDU.

The outbound `SignMessage` path already zeroes the 16-byte signature field before signing, but then called `Sign`, which copied the *entire* PDU again just to re-zero a field that was already zero. This adds an allocation-free signing variant for that path.

## Changes

- Add `SignInPlace(message []byte)` to the `Signer` interface and all three implementations (`HMACSigner`, `CMACSigner`, `GMACSigner`). It assumes the signature field is already zeroed and neither copies nor mutates the message.
- `SignMessage` now calls `SignInPlace` directly (it owns the buffer and has just zeroed the field).
- Public `Sign` keeps its no-mutation contract: HMAC/CMAC still copy; GMAC uses save-clear-compute-restore in place so the caller's buffer is byte-identical on return. `Verify` (which calls `Sign` on the inbound buffer) is therefore unaffected.

## Correctness / tests

- New golden cross-check `TestSignInPlace_MatchesSign` asserts `SignInPlace` (with the field pre-zeroed) is **byte-identical** to `Sign`, across all three signers.
- New `TestSign_DoesNotMutateBuffer` asserts `Sign` leaves the caller's buffer (including the original non-zero signature field) unchanged.
- Existing RFC 4493 / MS-SMB2 signing vectors are untouched and still pass.
- `go build ./...`, `go vet`, and `go test -race ./internal/adapter/smb/signing/...` all green.

Addresses the SMB signing allocation findings from re-audit #1104 (`cmac_signer.go`, `gmac_signer.go`, `hmac_signer.go`).

Refs #1104

### DIFF
--- a/internal/adapter/smb/signing/cmac_signer.go
+++ b/internal/adapter/smb/signing/cmac_signer.go
@@ -132,6 +132,10 @@ func (s *CMACSigner) cmacMAC(data []byte) [16]byte {
 
 // Sign computes the AES-CMAC signature for an SMB2 message.
 // The signature field (bytes 48-63) is zeroed before computation.
+//
+// Sign copies the message so it never mutates the caller's buffer. The
+// outbound SignMessage path, which has already zeroed the signature field,
+// uses SignInPlace to avoid the copy.
 func (s *CMACSigner) Sign(message []byte) [SignatureSize]byte {
 	if len(message) < SMB2HeaderSize {
 		return [SignatureSize]byte{}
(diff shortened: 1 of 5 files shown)


marmos91:
Thanks @copilot. Addressed all three:

1. **GMAC `Sign` contract/concurrency** — reverted GMAC `Sign` to a full copy (matching HMAC/CMAC), so it never writes to the caller's buffer even transiently. `Sign` is only used by `Verify` (not the hot outbound path), so the copy there is fine; the allocation-free path is `SignInPlace`, which is what `SignMessage` uses.
2. **HMAC `mac.Sum(nil)`** — now sums into a stack-backed `[sha256.Size]byte` dst.
3. **GMAC `gcm.Seal(nil, ...)`** — now seals into a stack-backed `[SignatureSize]byte` dst.

Golden tests (`TestSignInPlace_MatchesSign`, `TestSign_DoesNotMutateBuffer`) and RFC/MS-SMB2 vectors still pass.

PR #1111, opened 2026-06-14
fix(nfs/v3): BAD_COOKIE on cookieverf mismatch, honor sync WRITE, cache wtmax

marmos91:
Addresses the NFSv3-handlers findings from the v1.0 re-audit (#1104).

## Findings fixed

- **`readdir.go` / `readdirplus.go` — cookieverf mismatch silently ignored.** READDIR/READDIRPLUS now return `NFS3ERR_BAD_COOKIE` on a non-zero-cookie verifier mismatch (RFC 1813 §3.3.16/3.3.17), matching the NFSv4 READDIR behavior already shipped (commit `7548192f`). A zero cookie (initial request) or zero verifier (verifier-less client) still bypasses the check. Directory attributes are returned even on the error, consistent with the existing `NFS3ERR_NOTDIR` path.
- **`write.go` — per-RPC `GetFilesystemCapabilities` (Badger `db.View`) on every WRITE.** wtmax is now cached in a process-global atomic (`caps_cache.go`), populated by FSINFO and refreshed on a cold-cache WRITE. This mirrors the existing NFSv4 `attrs/encode.go` `fsMaxWriteSize` cache. The store stays off the hot WRITE path for the common mount sequence (FSINFO precedes WRITE).
- **`write.go` — WRITE always returned UNSTABLE.** A DATA_SYNC / FILE_SYNC request now triggers a synchronous flush (block store + pending metadata, via the shared `common.CommitBlockStore` seam used by COMMIT) and reports the requested stability level. On flush failure it falls back to UNSTABLE — the bytes remain in the crash-safe cache and the client can retry COMMIT.

## Notes

- The process-global wtmax cache is the established pattern in this codebase (NFSv4 `attrs` package); the metadata store exposes a single static wtmax process-wide.
- Tests `TestReadDir/Plus_StaleVerifierContinues` encoded the old advisory-only behavior and were replaced with `*_StaleVerifierBadCookie` + `*_ZeroCookieIgnoresVerifierMismatch` coverage, mirroring the NFSv4 test additions. `TestWrite_FileSyncStability` updated to assert FILE_SYNC is honored.
- The `NewHandlerFixture` constructor resets the process-global caps cache so tests setting distinct wtmax values do not bleed into one another.

Relates to #1104.

### DIFF
--- a/internal/adapter/nfs/v3/handlers/caps_cache.go
+++ b/internal/adapter/nfs/v3/handlers/caps_cache.go
@@ -1,0 +1,50 @@
+package handlers
+
+import "sync/atomic"
+
+// Filesystem capability cache.
+//
+// The advertised maximum WRITE size (wtmax) is a static, store-level limit. The
+// WRITE handler needs it on every RPC to short-write over-large requests to the
+// value the client was told in FSINFO. Calling GetFilesystemCapabilities (a
+// Badger db.View on the BadgerDB backend) on every WRITE is wasteful, so the
+// value is cached here and refreshed whenever it is observed (FSINFO, or the
+// first WRITE on a cold cache).
+//
+// This mirrors the NFSv4 attrs package, which caches the same store
+// capabilities in atomic globals updated via SetFilesystemCapabilities.
+//
+// INVARIANT (process-global is safe): wtmax is a server-wide value, not a
+// per-share one. GetFilesystemCapabilities ignores the file handle and every
+// metadata backend (memory/badger/postgres) returns the same hardcoded
+// MaxWriteSize (1 MiB). FSINFO advertises this same value to every client on
+// every share, so a single global cannot diverge from what any one client was
+// told. If a future change makes wtmax per-share (a per-share override in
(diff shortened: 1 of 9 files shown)


marmos91:
Addressed both Copilot review comments in the latest push:

1. **write.go FILE_SYNC metadata-flush error** — `flushStableWrite` now takes the requested stability level. For FILE_SYNC (which requires durable metadata before the reply) a `FlushPendingWriteForFile` failure now propagates, so the handler reports UNSTABLE rather than falsely claiming FILE_SYNC. For DATA_SYNC (data-only durability) the metadata flush remains best-effort, reconciled by a later COMMIT.

2. **caps_cache.go process-global wtmax** — documented the invariant that makes the global safe: `GetFilesystemCapabilities` is handle-independent and every backend (memory/badger/postgres) returns the same hardcoded 1 MiB MaxWriteSize, so FSINFO advertises an identical value to every client on every share. Added an explicit note that this cache MUST become handle-keyed if wtmax ever becomes per-share. This matches the existing NFSv4 `attrs` global.

PR #1116, opened 2026-06-14
fix(nfs/v4): EXCLUSIVE4 verifier idempotency; apply UNCHECKED4 createAttrs to existing file

marmos91:
## Findings (internal/adapter/nfs/v4/handlers/open.go)

Two HIGH correctness findings in the NFSv4 OPEN handler.

### 1. EXCLUSIVE4 verifier idempotency (HIGH)
`open.go` mapped both `EXCLUSIVE4` and `EXCLUSIVE4_1` create modes to `GUARDED4` and decoded-then-discarded the 8-byte `createverf4`. This breaks RFC 7530 §16.16.3 exclusive-create idempotency: a client retransmitting an EXCLUSIVE OPEN after a network drop got `NFS4ERR_EXIST` instead of success, even when presenting the same verifier.

**Fix:** decode the verifier as a `uint64` (big-endian, matching the NFSv3 representation) and persist it in `FileAttr.IdempotencyToken` at create time — the same durable mechanism NFSv3 `CREATE EXCLUSIVE` already uses, which round-trips across all metadata backends (memory/badger/postgres) with no schema change. On an OPEN4_CREATE where the file already exists:
- verifier matches the stored token → idempotent success (opens the existing file, returns a fresh stateid);
- verifier differs → `NFS4ERR_EXIST`.

EXCLUSIVE4_1 retries do not re-mutate the existing file's attributes.

### 2. UNCHECKED4 ignores createAttrs on existing file (HIGH)
`open.go` UNCHECKED4 on an existing file opened it but ignored the supplied `createattrs`, so a requested `size=0` truncation was silently skipped.

**Fix:** after opening the existing file under UNCHECKED4, apply the decoded `createAttrs` (the bitmap the client actually sent) via `metaSvc.SetFileAttributes` — which enforces its own permission checks — and reclaim block-store space on a size-down. Only attributes present in the createAttrs bitmap are applied; unset attributes are untouched.

## Implementation notes
- Handlers stay protocol-only: storage uses the existing `FileAttr.IdempotencyToken` field and `metadata.SetFileAttributes`; no business logic inlined.
- The truncate+block-reclaim logic (`#832` RefCount reaping) was duplicated from the SETATTR path; extracted into a shared `applySetAttrsWithTruncateReclaim` helper used by both SETATTR and OPEN(UNCHECKED4).
- The SETATTR pre-op `GetFile` is now gated to size-changing requests only (avoids an extra read on mode/owner/time-only SETATTRs).

## Limitation
The verifier lives in `IdempotencyToken` and is **not** cleared by a later normal SETATTR — identical to the existing NFSv3 design. This is intentional and RFC-correct: once a file exists, any subsequent exclusive create with a *different* verifier must yield `NFS4ERR_EXIST` (the file exists), and the same-verifier case stays idempotent. The Linux atime/mtime "create verifier" trick was deliberately not used (it would overload timestamp fields and require clearing); reusing `IdempotencyToken` is durable and cleaner.

## Tests (internal/adapter/nfs/v4/handlers/open_exclusive_test.go)
- `TestOpen_Exclusive4_RetrySameVerifier` — create then retry with same verifier → NFS4_OK with a decodable stateid.
- `TestOpen_Exclusive4_RetryDifferentVerifier` — retry with a different verifier → NFS4ERR_EXIST.
- `TestOpen_Unchecked4_ExistingFileTruncates` — UNCHECKED4 on an existing file with size=0 → metadata size truncated to 0.

All `./internal/adapter/nfs/v4/...` and `./pkg/metadata/...` suites pass.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/open.go
+++ b/internal/adapter/nfs/v4/handlers/open.go
@@ -317,7 +317,16 @@ func (h *Handler) handleOpenClaimNull(
 				// the file's idempotency token at create time. Match -> treat as
 				// a successful (idempotent) open returning a fresh stateid;
 				// mismatch -> NFS4ERR_EXIST.
-				if child.IdempotencyToken != *createVerifier {
+				//
+				// A zero stored token is the default for files NOT created via an
+				// exclusive create (UNCHECKED4 / GUARDED4 / NFSv3 non-exclusive),
+				// so it must never be accepted as an idempotent match — otherwise
+				// a stray verifier==0 exclusive OPEN would silently re-open any
+				// such file instead of returning NFS4ERR_EXIST. Real clients pick
+				// a non-zero (random/clock-based) verifier, so requiring a
+				// non-zero token here loses only the pathological zero-verifier
+				// retry while closing the false-match hole.
+				if *createVerifier == 0 || child.IdempotencyToken != *createVerifier {
 					logger.Debug("NFSv4 OPEN EXCLUSIVE verifier mismatch on existing file",
 						"file", filename,
 						"client", ctx.ClientAddr)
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed both review comments in 5c747a54:

1. **verifier==0 false match** — an EXCLUSIVE4 OPEN with an all-zero verifier now returns NFS4ERR_EXIST on an existing file rather than false-matching a non-exclusive file's default-zero `IdempotencyToken`. Real clients use non-zero (random/clock-based) verifiers, so this only drops the pathological zero-verifier retry while closing the hole. Covered by `TestOpen_Exclusive4_ZeroVerifierOnExistingFileConflicts`.

2. **read-only OPEN truncating** — an UNCHECKED4 OPEN requesting a size change without OPEN4_SHARE_ACCESS_WRITE is now rejected with NFS4ERR_OPENMODE, matching the SETATTR size-change gating. Covered by `TestOpen_Unchecked4_SizeOnReadOnlyOpenRejected`.

PR #1141, opened 2026-06-15
fix(nfs/gss): set FLAG_ACCEPTOR_SUBKEY on DATA reply MIC/Wrap tokens

### LINKED ISSUE #1125: audit: NFS RPCSEC_GSS/krb5: AcceptorSubkey MIC flag missing breaks mutual-auth
Verified findings from the develop `3488526a` re-audit (2nd adversarial pass). **2 HIGH / 0 MED / 0 LOW.** Full context: `.planning/v1.0-audit/report-verified-HIGH-3488526a.md`.

### [HIGH] HasAcceptorSubkey not propagated to DATA reply MIC — flag missing, client verification fails
- **Where:** `internal/adapter/nfs/rpc/gss/context.go:93`
- **What:** GSSContext struct has no HasAcceptorSubkey field. handleInit (framework.go:494-546) receives verified.HasAcceptorSubkey from VerifyToken but only stores it in GSSProcessResult for the INIT reply — it is never persisted in the GSSContext. handleData (framework.go:581-733) returns a GSSProcessResult with HasAcceptorSubkey=false always. GSSSessionInfo (context.go:56-69) also lacks the field. ComputeReplyVerifier (verifier.go:33-58) therefore never sets gssapi.MICTokenFlagAcceptorSubkey on DATA reply MIC tokens. When mutual auth produced an acceptor subkey, RFC 4121 requires that flag on every subsequent MIC; libtirpc/MIT krb5 will reject all DATA replies with GSS_S_BAD_SIG.
- **Why:** Every RPCSEC_GSS DATA reply sent after a mutual-auth Kerberos session setup will have a cryptographically invalid MIC flag, breaking the entire krb5i and krb5p service levels when the client requested mutual authentication.
- **Verified:** GSSContext struct (context.go:93) has no HasAcceptorSubkey field. handleData (framework.go:725-732) returns GSSProcessResult without setting HasAcceptorSubkey (always false). ComputeReplyVerifier (verifier.go:33) does not accept hasAcceptorSubkey parameter — only sets MICTokenFlagSentByAcceptor. ComputeInitVerifier does accept it (line 92). DATA replies after mutual-auth Kerberos session never set MICTokenFlagAcceptorSubkey on MIC tokens. GSSProcessor wired in production at pkg/adapter/nfs/nlm.go:660.

### [HIGH] Missing MICTokenFlagAcceptorSubkey in DATA reply MIC/Wrap tokens breaks krb5 mutual-auth sessions
- **Where:** `internal/adapter/nfs/rpc/gss/verifier.go:40`
- **What:** ComputeReplyVerifier always sets only MICTokenFlagSentByAcceptor, never MICTokenFlagAcceptorSubkey. Same defect in WrapIntegrity (integrity.go:117) and WrapPrivacy (privacy.go:282). GSSContext struct has no HasAcceptorSubkey field, so the flag determined during INIT (framework.go:135, stored in GSSProcessResult.HasAcceptorSubkey at line 545) is discarded and never reaches the DATA reply path (reply.go:57). GSSSessionInfo (context.go:56-69) stored in Go context also omits this field.
- **Why:** RFC 4121 Section 4.2.2 requires FLAG_ACCEPTOR_SUBKEY (bit 2) be set in all tokens produced by the acceptor when a subkey was included in the AP-REP EncAPRepPart. Linux krb5/libtirpc gss_verify_mic() checks this flag. When mutual authentication is negotiated (AP-Options bit MUTUAL-REQUIRED set, which is the default for every Linux NFS krb5 mount), the server emits a subkey in AP-REP, sets HasAcceptorSubkey=true in VerifiedContext, and correctly sets the flag in the INIT reply verifier (ComputeInitVerifier, verifier.go:107-108). But every subsequent DATA reply verifier and Wrap token omits the flag, causing the client to reject all server replies — complete authentication failure for every krb5/krb5i/krb5p NFS operation after mount.
- **Verified:** GSSSessionInfo (context.go:59-69) lacks HasAcceptorSubkey. ComputeReplyVerifier (verifier.go:33-57) always sets only MICTokenFlagSentByAcceptor. dispatch.go:160-164 creates GSSSessionInfo without HasAcceptorSubkey; reply.go:57 calls ComputeReplyVerifier(sessionInfo.SessionKey, sessionInfo.SeqNum) with no subkey flag. GSSContext.HasAcceptorSubkey (line 342) exists and is set at line 545 but never reaches the DATA reply path. Commit #1078 (be78116e) touched framework.go, integrity.go, privacy.go but did not add HasAcceptorSubkey to GSSSessionInfo — still missing on this tree. RFC 4121 §4.2.2 requires FLAG_ACCEPTOR_SUBKEY in all acceptor tokens when subkey was included in AP-REP; Linux krb5 gss_verify_mic checks this flag.

marmos91:
Fixes #1125

## Problem

After a mutual-auth Kerberos session that produced an acceptor subkey, RFC 4121 §4.2.2 requires `FLAG_ACCEPTOR_SUBKEY` (bit 2, 0x04) on every subsequent acceptor MIC/Wrap token. The server set this flag correctly on the **INIT** reply verifier (`ComputeInitVerifier`) but dropped it on **every DATA reply**:

- `GSSContext` had no `HasAcceptorSubkey` field — the value from `VerifyToken`/`handleInit` was stored only in the INIT `GSSProcessResult` and never persisted.
- `handleData` always returned `HasAcceptorSubkey=false`.
- `ComputeReplyVerifier`, `WrapIntegrity`, and `WrapPrivacy` never set the bit.

MIT krb5 / libtirpc `gss_verify_mic()` checks this flag and rejects mismatched replies with `GSS_S_BAD_SIG`. Mutual auth (`MUTUAL-REQUIRED`) is the default for every Linux NFS krb5 mount, so this broke **all** krb5i and krb5p operations after mount.

## Fix

Thread `HasAcceptorSubkey` from `VerifyToken` through the persisted `GSSContext` into the DATA reply path, mirroring how the INIT path already plumbs it:

- Add `HasAcceptorSubkey` to `GSSContext` (set in `handleInit`) and `GSSSessionInfo` (set in `dispatch.go`).
- `handleData` propagates it into `GSSProcessResult`.
- `ComputeReplyVerifier` / `WrapIntegrity` / `WrapPrivacy` accept the flag and set `FLAG_ACCEPTOR_SUBKEY` accordingly; `reply.go` passes it through.
- Factor the acceptor MIC flag computation into a shared `acceptorMICFlags` helper (also adopted by `ComputeInitVerifier`).

The client→server (initiator) unwrap paths are unchanged — only acceptor-emitted tokens carry the flag.

## Tests

- `TestComputeReplyVerifierSetsAcceptorSubkeyFlag` — asserts the DATA reply MIC carries the flag iff a subkey was established. Verified to FAIL without the verifier fix.
- `TestProcessDATAPropagatesAcceptorSubkey` — end-to-end: a subkey-producing INIT propagates through the persisted context into the DATA result and reply MIC. Verified to FAIL without the propagation fix.

`go build ./...`, `go vet`, and `go test -race ./internal/adapter/nfs/rpc/... ./pkg/adapter/nfs/...` all green.

### DIFF
--- a/internal/adapter/nfs/rpc/gss/context.go
+++ b/internal/adapter/nfs/rpc/gss/context.go
@@ -66,6 +66,12 @@ type GSSSessionInfo struct {
 	// Service is the RPCSEC_GSS service level (none/integrity/privacy).
 	// Needed by the reply path to determine which wrapping to apply.
 	Service uint32
+
+	// HasAcceptorSubkey indicates the acceptor produced a subkey in the AP-REP
+	// during context establishment (mutual auth). When true, RFC 4121 §4.2.2
+	// requires FLAG_ACCEPTOR_SUBKEY on every acceptor MIC/Wrap token, including
+	// DATA reply verifiers and krb5i/krb5p reply wrapping.
+	HasAcceptorSubkey bool
 }
 
 // ContextWithSessionInfo returns a new context carrying GSS session info.
@@ -111,6 +117,12 @@ type GSSContext struct {
 	// Service is the protection level negotiated during INIT.
 	// One of RPCGSSSvcNone, RPCGSSSvcIntegrity, or RPCGSSSvcPrivacy.
 	Service uint32
+
+	// HasAcceptorSubkey records whether the acceptor produced a subkey in the
+	// AP-REP during context establishment (mutual auth). When true, RFC 4121
+	// §4.2.2 requires FLAG_ACCEPTOR_SUBKEY on every subsequent acceptor MIC/Wrap
+	// token; the DATA reply path reads this to set the flag correctly.
+	HasAcceptorSubkey bool
 
 	// CreatedAt is the time the context was established.
 	CreatedAt time.Time
--- a/internal/adapter/nfs/rpc/gss/framework.go
+++ b/internal/adapter/nfs/rpc/gss/framework.go
@@ -492,14 +492,15 @@ func (p *GSSProcessor) handleInit(cred *RPCGSSCredV1, requestBody []byte) *GSSPr
 	// Create the GSS context
 	now := time.Now()
 	gssCtx := &GSSContext{
-		Handle:     handle,
-		Principal:  verified.Principal,
-		Realm:      verified.Realm,
-		SessionKey: verified.SessionKey,
-		SeqWindow:  NewSeqWindow(DefaultSeqWindowSize),
-		Service:    cred.Service,
-		CreatedAt:  now,
-		LastUsed:   now,
+		Handle:            handle,
+		Principal:         verified.Principal,
+		Realm:             verified.Realm,
+		SessionKey:        verified.SessionKey,
+		SeqWindow:         NewSeqWindow(DefaultSeqWindowSize),
+		Service:           cred.Service,
+		HasAcceptorSubkey: verified.HasAcceptorSubkey,
+		CreatedAt:         now,
+		LastUsed:          now,
 	}
 
 	// CRITICAL: Store context BEFORE building reply.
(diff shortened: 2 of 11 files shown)


marmos91:
Addressed all three Copilot findings in f64e01c7:
- Corrected the integrity.go acceptor-sign key-usage comment back to 23 (matches `KeyUsageAcceptorSign` in types.go; my earlier edit to 25 was wrong).
- Added `TestWrapIntegritySetsAcceptorSubkeyFlag` asserting the krb5i reply MIC carries FLAG_ACCEPTOR_SUBKEY when true and clears it when false.
- Added `TestWrapPrivacySetsAcceptorSubkeyFlag` asserting the krb5p Wrap token carries 0x04 when true and clears it when false.

PR #1150, opened 2026-06-15
fix(block): append-log recovery, S3 SSRF, Argon2 bounds, cache race (#1124)

### LINKED ISSUE #1124: audit: Block store (fs/engine/s3/crypto/badger): append desync, SSRF, Argon2 DoS, key injection
Verified findings from the develop `3488526a` re-audit (2nd adversarial pass). **3 HIGH / 2 MED / 0 LOW.** Full context: `.planning/v1.0-audit/report-verified-HIGH-3488526a.md`.

### [HIGH] Argon2 KDF params from key file used without lower-bound validation — panic DoS on corrupt/tampered file
- **Where:** `pkg/block/encryption/keyprovider/local.go:173`
- **What:** `unwrapMasterKey` passes `kf.KDF.Time`, `kf.KDF.MemoryKiB`, and `kf.KDF.Threads` read from the on-disk JSON directly into `argon2.IDKey` with no lower-bound checks. `argon2.IDKey` panics with `"argon2: number of rounds too small"` when `time < 1` and `"argon2: parallelism degree too low"` when `threads < 1`. A corrupted or attacker-replaced key file with `time=0` or `threads=0` causes an unrecovered panic that crashes the daemon at startup. A huge `memory_kib` (max uint32 ≈ 4 TiB) triggers OOM/OOM-kill instead of a controlled error.
- **Why:** Key file is read and decoded before any KDF parameter validation (lines 136-158 in `decodeKeyFile` check version/algo/wrap/id but not numeric KDF fields). An adversary with write access to the key file path (or a storage-layer corruption) can prevent the daemon from ever starting again by inserting `"time":0` — matching a DoS-via-panic / availability attack.
- **Verified:** unwrapMasterKey at local.go:173 calls argon2.IDKey(passphrase, salt, kf.KDF.Time, kf.KDF.MemoryKiB, kf.KDF.Threads, 32) with raw on-disk JSON values. decodeKeyFile (lines 136-158) checks version/algo/wrap/id but adds no numeric guards on KDF fields. golang.org/x/crypto argon2.IDKey panics when time<1 or threads<1. Attacker or corruption can set time=0 or threads=0 — daemon panics at startup and never recovers. huge memory_kib causes OOM-kill. Full reachable path: server startup -> keyprovider.Load -> decodeKeyFile -> unwrapMasterKey -> argon2.IDKey.

### [HIGH] groupCommit.Sync error leaves lf.eofPos desync'd with fd position
- **Where:** `pkg/block/local/fs/appendwrite.go:423`
- **What:** When writeRecord succeeds (fd position advances by n bytes) but lf.groupCommit.Sync returns a non-nil error, AppendWrite returns at line 424 without advancing lf.eofPos or evicting the fd. The next AppendWrite to the same payload writes its record AFTER the failed-but-written bytes (fd is at old position + n), yet captures logPos = lf.eofPos (the pre-write position). logIndex.Append records the wrong logPos, so readRecordAt in rollup reads the orphaned failed-write frame instead of the new record — producing a wrong file_offset or CRC mismatch in Phase A and permanently wedging rollup for that payload.
- **Why:** writeRecord uses two sequential f.Write calls (frame then payload) which advance the OS fd position regardless of downstream fsync failure. lf.eofPos is only advanced at line 434, which is unreachable on Sync error. The fd is not closed/evicted on the Sync-error path (that recovery only fires on writeRecord error at line 373), so the desync persists across subsequent AppendWrite calls.
- **Verified:** appendwrite.go:423 groupCommit.Sync error causes early return at line 424. writeRecord (appendlog.go:136-143) uses io.Writer sequential writes advancing OS fd position by n bytes. lf.eofPos is only incremented at line 434 which is unreachable on Sync error. Fd is not closed/evicted on Sync error (only on writeRecord error at line 403). Next AppendWrite writes at fd_pos+n but captures logPos=lf.eofPos (old value), recording wrong logIndex entry and permanently wedging rollup for that payload.

### [HIGH] SSRF via unvalidated S3 endpoint — admin API accepts arbitrary http:// URLs
- **Where:** `pkg/controlplane/runtime/blockstore_init.go:69`
- **What:** ValidateBlockStoreConfig for storeType=="s3" checks bucket/access_key_id/secret_access_key but never validates the `endpoint` field. normalizeEndpoint() in store.go passes http:// and any other scheme through as-is (line 192). An admin-role user can POST /api/v1/store/block/remote with `"endpoint":"http://169.254.169.254/latest/meta-data"` and the server will issue S3 SDK HTTP requests to that target during HealthCheck/HeadBucket or any subsequent block operation.
- **Why:** The endpoint is user-supplied config stored in DB and instantiated without scheme or host restriction. normalizeEndpoint explicitly preserves existing schemes including http:// (test case confirms this at store_test.go:24). The HealthCheck path (store.go:565 HeadBucket) fires immediately on Create (block_stores.go:140-144), making SSRF exploitable at config-create time with no share attachment required.
- **Verified:** ValidateBlockStoreConfig (blockstore_init.go:69-85) checks bucket/access_key_id/secret_access_key but never validates endpoint. normalizeEndpoint at store.go:189-192 passes http:// and any valid URI scheme through as-is — test at store_test.go:24 confirms 'http://localhost:4566' returns unchanged. endpoint flows from user-supplied JSON config map at service.go:2026 with no sanitization. HealthCheck fires at Create time (block_stores.go:140-144), so SSRF is exploitable at config-create time without share attachment.

### [MED] Data race: unsynchronized write to bs.cache field in DestroyCache
- **Where:** `pkg/block/engine/engine.go:584`
- **What:** DestroyCache writes `bs.cache = nullCache{}` while holding only `closeMu.RLock()`. Concurrent data ops (ReadAt, WriteAt, Flush, etc.) also hold `closeMu.RLock()` and read `bs.cache`. Because multiple goroutines can hold RLock simultaneously, the write at line 584 is concurrent with reads of `bs.cache` in other goroutines — unsynchronized write to a non-atomic interface field. Additionally, the `OnChunkComplete` closure installed in New (engine.go:326) reads `bs.cache.Put(hash, data)` from the local store's rollup goroutine with NO lock at all, racing the same write.
- **Why:** Multiple goroutines reading `bs.cache` while one goroutine writes it is a classic data race. The `closeMu.RLock()` guard on DestroyCache does NOT serialize against other `closeMu.RLock()` holders — that is the fundamental property of a read-write lock. The race detector will flag reads in WriteAt/ReadAt/OnChunkComplete vs the write in DestroyCache. On amd64 the race is often benign (interface assignment is 2-word non-atomic), but on arm64 (the deployment target per CLAUDE.md — Apple Silicon) torn interface reads produce a corrupt function-pointer + data-pointer pair, leading to a crash or silent wrong-cache use.
- **Verified:** engine.go:584 writes bs.cache=nullCache{} under closeMu.RLock() only. Concurrent ReadAt (readwrite.go:34) also holds closeMu.RLock() and reads bs.cache. Multiple RLock holders do not serialize against each other — plain interface field write races concurrent reads. Race is real but only triggered by admin DestroyCache call (shares/service.go:1746), not on hot path. Downgraded from HIGH to MED.

### [MED] replayLogIntoDest: full sequential log scan on every read, no index-assisted seek
- **Where:** `pkg/block/local/fs/readpayload.go:145`
- **What:** Opens a new fd (`os.Open(lf.path)`) and scans all log records sequentially from `logHeaderSize` via `readRecord` on every `ReadPayloadAt` call. For a log with thousands of records (normal under sustained write load), this is O(total records) I/O regardless of the requested byte range. The per-payload `logIndex` already knows exactly which log positions carry records overlapping any given file-offset range.
- **Why:** The rollup path uses `idx.lookupInterval(stable.Offset, length)` (logindex.go:128) to seek directly to relevant records. The read path ignores the logIndex entirely and re-reads the full log. Under parallel NFS write workloads with large logs (approaching `maxLogBytes` = 1 GiB) every read pays O(log size / record size) pread syscalls even if only a small window is requested.
- **Verified:** replayLogIntoDest (readpayload.go:105) opens a fresh fd and scans all log records from logHeaderSize sequentially on every ReadPayloadAt call. logIndex (populated by AppendWrite and recovery, used by rollup via lookupInterval) is not consulted — no index-assisted seek. Under sustained write load with large logs (approaching maxLogBytes=1GiB), every read pays O(total_records) pread syscalls regardless of requested byte range. ReadPayloadAt is the live data-plane read path called for every NFS READ / SMB read. Rollup correctly uses idx.lookupInterval (rollup.go:295) but read path ignores it entirely.

marmos91:
Closes #1124.

Verified findings from the develop re-audit (block store). Per-finding outcome below.

## [HIGH] Argon2 KDF params used without lower-bound validation — panic/OOM DoS
`pkg/block/encryption/keyprovider/local.go`. `decodeKeyFile` now validates the on-disk Argon2 params (`validateKDFParams`) before they reach `argon2.IDKey`. Lower bounds prevent the `time=0`/`threads=0` panic that crashed the daemon at startup; an upper bound on `memory_kib` (~4 GiB) prevents an OOM-kill. Out-of-range params surface as `ErrKeyFileCorrupt`. Tests: `kdf_time_zero`, `kdf_threads_zero`, `kdf_memory_zero`, `kdf_memory_huge`, `kdf_time_huge`.

## [HIGH] groupCommit.Sync error desyncs lf.eofPos from the fd position
`pkg/block/local/fs/appendwrite.go`. On a `groupCommit.Sync` failure, `AppendWrite` now evicts the log fd (close + remove from `logFDs`) exactly like the existing `writeRecord`-error recovery path, instead of returning with `eofPos` left behind the advanced fd. This stops the next `AppendWrite` from recording a `logIndex` entry pointing at the orphaned, un-fsync'd frame — which permanently wedged rollup for that payload. Test `TestAppendWrite_SyncError_EvictsFdAndKeepsLogIndexConsistent` fails before / passes after.

## [HIGH] SSRF via unvalidated S3 endpoint
`pkg/block/remote/s3/store.go` + `pkg/controlplane/runtime/blockstore_init.go`. New `ValidateEndpoint` resolves the endpoint host and rejects cloud-metadata (169.254.169.254), loopback, link-local, unspecified, multicast, and private/internal addresses. Wired into `ValidateBlockStoreConfig` so it fires at config-create time, before the create-time `HealthCheck`/`HeadBucket` dial. `allow_private_endpoint=true` opts in MinIO/Localstack-style private object stores but still blocks the metadata endpoint. Tests: `TestValidateEndpoint_SSRF`, `TestValidateBlockStoreConfig_S3_SSRF`.

## [MED] Data race on bs.cache in DestroyCache
`pkg/block/engine/`. Every cache-field access now goes through `cacheMu` via `loadCache`/`storeCache`. The `DestroyCache`/`Start` swaps no longer race concurrent data ops (which only hold `closeMu.RLock`) or the lock-free `OnChunkComplete` rollup goroutine. Test `TestDestroyCache_ConcurrentReads_NoRace` flags a `DATA RACE` before / clean after under `-race`.

## [MED] replayLogIntoDest full sequential scan on every read
`pkg/block/local/fs/readpayload.go`. `ReadPayloadAt` now uses the per-payload `logIndex` (`EntriesForInterval`) to pread only the records overlapping the requested window — O(overlapping records) instead of O(total records) per read — matching how rollup already seeks. The index path is held under the per-file mutex so a snapshotted `logPos` stays consistent with `maybeCompactLog`'s rename + `logPos` rebase (a regression caught in self-review; the original lock-free draft could pread post-compaction offsets against a pre-compaction fd). Falls back to the full scan only when no index is wired (fixtures). Tests: `TestReadPayloadAt_IndexAssisted_OutOfOrderOverlap`, `TestReplayViaScan_Fallback`.

## Verification
`go build ./...`, `go test ./pkg/block/... ./pkg/controlplane/runtime/`, and `go test -race` on the touched packages all pass.

### DIFF
--- a/pkg/block/encryption/keyprovider/local.go
+++ b/pkg/block/encryption/keyprovider/local.go
@@ -26,6 +26,23 @@ const localPEMType = "DITTOFS ENCRYPTED MASTER KEY"
 
 const localKeyFileVersion = 1
 
+// Argon2 KDF parameter bounds enforced on read. The lower bounds keep
+// argon2.IDKey from panicking on a corrupt/tampered file (it panics when
+// time < 1 or threads < 1) and reject a key file weaker than was ever
+// generated. The upper bounds cap unwrap cost so an attacker who can write
+// the key file cannot turn startup into an OOM-kill or a multi-minute
+// hang: maxArgonMemoryKiB ≈ 4 GiB and maxArgonTime/maxArgonThreads are far
+// above any sane configuration. Defaults written by GenerateKeyFile
+// (t=3, m=64 MiB, p=4) sit comfortably inside this window.
+const (
+	minArgonTime      = 1
+	maxArgonTime      = 64
+	minArgonMemoryKiB = 8 * 1024
+	maxArgonMemoryKiB = 4 * 1024 * 1024
+	minArgonThreads   = 1
+	maxArgonThreads   = 64
+)
+
 // argon2Params holds the KDF parameters persisted alongside the wrapped
 // master key. Defaults follow the OWASP 2024 password-storage cheat
 // sheet for Argon2id (m=64 MiB, t=3, p=4).
@@ -154,7 +171,27 @@ func decodeKeyFile(raw []byte) (*localKeyFile, error) {
 	if kf.MasterKeyID == "" {
 		return nil, fmt.Errorf("%w: empty master_key_id", ErrKeyFileCorrupt)
 	}
+	if err := validateKDFParams(kf.KDF); err != nil {
+		return nil, err
+	}
 	return &kf, nil
+}
+
+// validateKDFParams bounds the Argon2 parameters read from disk before they
+// reach argon2.IDKey. Without these guards a corrupt or attacker-replaced
+// key file with time=0 or threads=0 panics the daemon at startup, and a
+// huge memory_kib triggers an OOM-kill — both an availability/DoS vector.
+func validateKDFParams(p argon2Params) error {
+	if p.Time < minArgonTime || p.Time > maxArgonTime {
+		return fmt.Errorf("%w: kdf time %d out of range [%d,%d]", ErrKeyFileCorrupt, p.Time, minArgonTime, maxArgonTime)
+	}
+	if p.MemoryKiB < minArgonMemoryKiB || p.MemoryKiB > maxArgonMemoryKiB {
+		return fmt.Errorf("%w: kdf memory_kib %d out of range [%d,%d]", ErrKeyFileCorrupt, p.MemoryKiB, minArgonMemoryKiB, maxArgonMemoryKiB)
+	}
+	if p.Threads < minArgonThreads || p.Threads > maxArgonThreads {
+		return fmt.Errorf("%w: kdf threads %d out of range [%d,%d]", ErrKeyFileCorrupt, p.Threads, minArgonThreads, maxArgonThreads)
+	}
+	return nil
 }
 
 func unwrapMasterKey(kf *localKeyFile, passphrase string) ([]byte, error) {
--- a/pkg/block/encryption/keyprovider/local_corrupt_test.go
+++ b/pkg/block/encryption/keyprovider/local_corrupt_test.go
@@ -57,6 +57,14 @@ func TestLocal_DecodeKeyFile_CorruptFields(t *testing.T) {
 		{"unsupported_kdf", func(kf *localKeyFile) { kf.KDF.Algo = "scrypt" }},
 		{"unsupported_wrap", func(kf *localKeyFile) { kf.Wrap = "chacha20" }},
 		{"empty_master_key_id", func(kf *localKeyFile) { kf.MasterKeyID = "" }},
+		// KDF lower-bound guards: without these argon2.IDKey panics and
+		// crashes the daemon at startup (time=0 / threads=0), or OOM-kills it
+		// (huge memory_kib). They must surface as ErrKeyFileCorrupt instead.
+		{"kdf_time_zero", func(kf *localKeyFile) { kf.KDF.Time = 0 }},
+		{"kdf_threads_zero", func(kf *localKeyFile) { kf.KDF.Threads = 0 }},
+		{"kdf_memory_zero", func(kf *localKeyFile) { kf.KDF.MemoryKiB = 0 }},
+		{"kdf_memory_huge", func(kf *localKeyFile) { kf.KDF.MemoryKiB = ^uint32(0) }},
+		{"kdf_time_huge", func(kf *localKeyFile) { kf.KDF.Time = ^uint32(0) }},
 	}
 	for _, tc := range cases {
 		t.Run(tc.name, func(t *testing.T) {
--- a/pkg/block/engine/dedup.go
+++ b/pkg/block/engine/dedup.go
@@ -104,7 +104,7 @@ func (m *Syncer) tryEagerSmallFileDedup(
 	// observed; bs.cache is never nil thanks to the Null Object pattern
 	// installed by engine.New.
 	if m.bs != nil {
-		m.bs.cache.Put(h, data)
+		m.bs.loadCache().Put(h, data)
 	}
 
 	logger.Debug("eager small-file dedup short-circuit hit",
@@ -376,7 +376,7 @@ func (m *Syncer) applyFileLevelDedupHit(
 				removed = append(removed, br.Hash)
 			}
 		}
-		m.bs.cache.InvalidateFile(payloadID, removed)
+		m.bs.loadCache().InvalidateFile(payloadID, removed)
 	}
 
 	// 5. Per-file append-log truncation. Best-effort: a failure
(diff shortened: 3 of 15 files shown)


marmos91:
Thanks for the review. Addressed two of three findings in 6f90de19:
- **store.go:254 (DNS DoS):** `ValidateEndpoint` now resolves hostnames via `net.DefaultResolver.LookupIPAddr` under a 5s context timeout.
- **store.go:215 (doc drift):** corrected the doc comment — loopback/private/non-global-unicast are rejected only when `allow_private_endpoint` is unset; the unconditional set (unspecified/multicast/link-local/metadata) is now stated separately.

For **appendwrite.go:438 (Sync-error eviction window):** this path is byte-identical to the pre-existing `writeRecord`-error recovery path directly above it (lines 403-408) and relies on the same documented FIX-20 invariant: the map entry is deleted before `Close()`, and any in-flight caller that already snapshotted `lf` re-acquires `mu`, re-reads the shard map, and finds `curLf != lf` → returns `ErrDeleted` (appendwrite.go:368) before ever touching `lf.f`. Diverging from that pattern only on the Sync-error branch would make the two recovery paths inconsistent, so I kept them identical.

Issue #1158, opened 2026-06-15
CI: matrix smbtorture over all store backends (parity with WPTS)


marmos91:
PR #1155 made **WPTS BVT** run across all 5 store profiles (memory/memory-fs/badger-fs/badger-s3/postgres-s3) on PRs. But the **smbtorture** (Samba) battery still runs **memory-only** — it is a separate, non-matrixed job in `smb-conformance.yml`.

For full store parity (like POSIX + WPTS now have), smbtorture should also matrix over the store backends.

## Caveat
smbtorture is heavy (~30+ tests × several min each); × 5 stores on every PR is a large CI-time increase. Pairs with the CI-caching/conditional-run work in #1157 — likely want smbtorture-full-matrix on push/cron + a representative subset (or memory only) on PRs, or gated by caching.

Found while shipping #1155.

marmos91:
Addressed by #1170 (event-tiered conformance: presubmit/postsubmit/nightly) + #1171. Path filters now skip docs-only (never adapter-gated); heavy Kerberos + full matrices moved to postsubmit+nightly; new postgres-s3 POSIX profile; develop-red email alarm + warn-on-PR forcing function in ci-health.yml.

Issue #1157, opened 2026-06-15
CI: cache conformance results + smart conditional-run to cut PR latency


marmos91:
Running the full store matrix (memory/badger/postgres) for POSIX + SMB conformance, plus Kerberos, on every PR raises CI wall-clock and may slow PR close.

## Wanted
- **Conditional run**: only run a heavy suite when relevant code changes (e.g. SMB matrix on SMB/store/block/controlplane changes — path filters already exist; extend to skip backend cells when only docs/unrelated code change).
- **Result caching**: cache a green conformance result keyed by a content hash of the inputs (binary + harness + config), so an unchanged suite re-uses the prior pass instead of re-running.
- Goal: keep maximum coverage without making PRs slow to close.

Counterweight to the full-matrix-on-PR changes (POSIX #1152 / SMB conformance PR).

marmos91:
Addressed by #1170 (event-tiered conformance: presubmit/postsubmit/nightly) + #1171. Path filters now skip docs-only (never adapter-gated); heavy Kerberos + full matrices moved to postsubmit+nightly; new postgres-s3 POSIX profile; develop-red email alarm + warn-on-PR forcing function in ci-health.yml.

Issue #1156, opened 2026-06-15
CI: broaden Kerberos coverage (always-on PRs + krb-sensitive op suites) for NFS & SMB


marmos91:
Today Kerberos CI is thin and not always-on:
- **NFS**: `nfs-kerberos.yml` runs a sec=krb5 **mount + basic file-op** test, but is path-filtered to auth/kerberos paths only — most PRs skip it.
- **SMB**: `smbtorture Kerberos` runs only `smb2.session --kerberos` (memory profile).

## Wanted
1. **Always-on for PRs** — run NFS sec=krb5 and SMB kerberos on every develop PR, not just auth-path changes.
2. **Depth on the krb-sensitive surface** (not the full battery — file-op semantics are auth-independent and redundant over krb5):
   - NFS: `krb5p` (privacy) + `krb5i` (integrity) large read/write to exercise per-RPC GSS wrap on the **data path**; ticket **expiry/renewal** mid-operation.
   - SMB: broaden smbtorture kerberos beyond `smb2.session` to read/write/lock/durable + `session.expiry`/reauth (signing+encryption keyed off the kerberos session key).

Tracks roadmap tasks for Kerberos CI parity.

marmos91:
Addressed by #1170 (event-tiered conformance: presubmit/postsubmit/nightly) + #1171. Path filters now skip docs-only (never adapter-gated); heavy Kerberos + full matrices moved to postsubmit+nightly; new postgres-s3 POSIX profile; develop-red email alarm + warn-on-PR forcing function in ci-health.yml.

PR #1167, opened 2026-06-15
ci(smb): persistent-backend smbtorture on PRs + broader Kerberos op coverage

marmos91:
Improves the SMB CI workflows (`smb-conformance.yml`, `smb-client-compat.yml`). Covers the SMB side of #1156, #1158, and #1157.

## smbtorture PR matrix (#1158 / latency-vs-parity #1157)
PR smbtorture went from **memory-only** to **memory + badger-fs**. badger-fs is the only PR profile that exercises the fs local store path and is faster than the s3 profiles, so PRs now get a persistent-backend signal without paying 5×45min. The **full set (`memory`, `memory-fs`, `badger-fs`) stays on push to develop and the weekly cron** — PRs deliberately do not run all 5. WPTS BVT already runs the full 5-profile matrix on PRs and is unchanged.

## Kerberos always-on + broader (#1156)
- `smbtorture-kerberos` is now a **matrix over `smb2.session` + `smb2.read` + `smb2.lock`** (parallel jobs, one filter each, each under the 30-min budget). The Kerberos session key signs/encrypts every PDU, so the authenticated read path and byte-range locking exercise the key past SPNEGO — coverage beyond session setup.
- **Dropped `smb2.create`**: it has known failures (`create.quota-fake-file`, `create.gentest`) catalogued in `KNOWN_FAILURES.md`, but `run.sh --kerberos` only loads the session-scoped `KNOWN_FAILURES_KERBEROS.md`, so those expected fails would be graded as *new* failures and red the job. It can be added once the Kerberos known-failures list covers the create suite. All three chosen filters were verified to exist as `--filter` values in `test/smb-conformance/smbtorture/run.sh`.
- Added `pkg/auth/kerberos/**` and `internal/auth/kerberos/**` to both `push` and `pull_request` path filters so a Kerberos-only code change triggers the SMB Kerberos job (previously only SMB-adapter/store/controlplane paths triggered it).

## Docs-only skip + caching (#1157)
Both workflows are already path-scoped to code, so docs-only PRs never trigger them — confirmed and documented inline in each `on:` block. `smb-conformance` jobs compile Go inside Docker (no host `setup-go`), so host module caching does not apply there; `smb-client-compat` already caches via `actions/setup-go` (`cache: true`, keyed on go.sum). No test-result caching added.

YAML validated with `actionlint` — only pre-existing shellcheck style hints in untouched script blocks; no new findings.

## Follow-up (Copilot review)
Clamped the smbtorture `workflow_dispatch` branch to the profiles `run.sh` actually supports (`memory`, `memory-fs`, `badger-fs`); the shared dispatch input also offers the s3 profiles for the WPTS job, which smbtorture rejects at validation — an s3 dispatch now falls back to `memory` instead of failing the job.

### DIFF
--- a/.github/workflows/smb-client-compat.yml
+++ b/.github/workflows/smb-client-compat.yml
@@ -6,6 +6,11 @@
 name: SMB Client Compatibility
 
 on:
+  # Path-scoped to SMB adapter code: docs-only changes never trigger this
+  # suite (#1157). PRs are excluded entirely (native-client mounts are too
+  # slow for the PR gate); coverage is push-to-develop + weekly + manual.
+  # Go builds here cache modules via actions/setup-go (cache: true, keyed on
+  # go.sum) in each job below.
   push:
     branches: [develop]
     paths:
--- a/.github/workflows/smb-conformance.yml
+++ b/.github/workflows/smb-conformance.yml
@@ -6,6 +6,10 @@
 name: SMB Conformance Tests
 
 on:
+  # Path-scoped to SMB/store/controlplane/auth code so docs-only and
+  # unrelated changes never trigger this heavy suite (#1157). A Kerberos-only
+  # code change must still exercise the smbtorture-kerberos job, so the auth
+  # kerberos packages are listed explicitly alongside the SMB adapter paths.
   push:
     branches: [develop]
     paths:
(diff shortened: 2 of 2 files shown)


marmos91:
Superseded by a single unified event-tiered CI PR (presubmit/postsubmit/nightly + email alarm + tree-closed warn-on-PR). The file-split was only to parallelize agents; the policy is cross-cutting so it ships as one coherent change.

PR #1170, opened 2026-06-15
ci: event-tiered conformance policy + develop-red alarm

marmos91:
## Unified event-tiered CI policy + develop-red alarm

Replaces the prior path-allowlist gating with **event-based tiering**. Path filters now skip docs-only changes only — they never gate a protocol's conformance on that protocol's adapter paths, because shared code (a metadata store, the block layer, the control plane) can break **both** SMB and NFS. So any non-docs change runs conformance for every protocol.

Three tiers: **PR = presubmit**, **develop push = postsubmit**, **nightly cron (04:00 UTC) = periodic**.

### Tier table (as shipped)

| Workflow / job | Presubmit (PR) | Postsubmit (push develop) + Nightly |
|---|---|---|
| POSIX (`posix-tests.yml`) | `memory`, `postgres-s3` × {3, 4, 4.1} = **6 jobs** | `memory`, `badger`, `postgres`, `postgres-s3` × {3, 4, 4.1} = **12 jobs** |
| SMB WPTS BVT (`smb-conformance.yml`) | `memory`, `postgres-s3` | all 5: `memory`, `memory-fs`, `badger-fs`, `badger-s3`, `postgres-s3` |
| SMB smbtorture | `memory`, `badger-fs` | `memory`, `memory-fs`, `badger-fs` |
| SMB smbtorture Kerberos | **off presubmit** | runs (push/nightly/dispatch); filters `smb2.session smb2.read smb2.create smb2.lock` |
| NFS Kerberos (`nfs-kerberos.yml`) | only when krb/NFS source paths change | always (push/nightly/dispatch) |
| E2E (`e2e-tests.yml`) | full single config (docs-skip only) | same + nightly |

Crons: all conformance workflows now use nightly `0 4 * * *` (replacing the prior weekly Sunday/Monday crons).

### postgres-s3 decision — SHIPPED (POSIX)

Added a `postgres-s3` profile to `test/posix/setup-posix.sh` (postgres metadata + memory local block + S3/Localstack remote block, combining the existing `postgres` and `cache-s3` wiring). The profile-matrix POSIX job carries **both** a postgres service and a localstack service; `setup-posix.sh` activates only what each profile needs (postgres tuning step is gated on `startsWith(matrix.profile, 'postgres')`). If postgres-s3 proves flaky in practice the documented fallback is presubmit `memory + postgres` (plain local block) — but it is shipped enabled here and the PR's own presubmit run exercises it live.

WPTS BVT also runs `postgres-s3` on PRs; its harness self-provisions postgres+localstack via docker-compose profiles, so no GitHub service containers are needed there.

### smbtorture divergence (noted)

`test/smb-conformance/smbtorture/run.sh` supports only `memory`, `memory-fs`, `badger-fs`, `memory-kerberos` — it has **no** `*-s3` profile. So smbtorture presubmit is `memory + badger-fs` (run in parallel ≈ one profile wall-clock) rather than `postgres-s3`. Adding S3 profiles to smbtorture is out of scope (would require run.sh changes).

### Kerberos filters added/dropped

smbtorture Kerberos filter broadened from just `smb2.session` to `smb2.session smb2.read smb2.create smb2.lock` (all valid filters in run.sh; a truncated/garbled GSS session key surfaces on read/create/lock, not just session setup). `run.sh` takes a single `--filter`, so the step iterates one suite per invocation. None dropped. (Addresses #14 / #1156.)

### nfs-kerberos continue-on-error decision

**Removed `continue-on-error: true`** — it now gates. `gh run list --workflow nfs-kerberos.yml` shows it has been consistently green on develop across recent push runs, so the baseline is established.

### Alarm + forcing function (`ci-health.yml`, NEW)

Centralizes all alerting (per-workflow ad-hoc "create regression issue" steps removed from `smb-conformance.yml` and `posix-tests.yml`).

- **`workflow_run`** on the four conformance workflows. On a failed **postsubmit/periodic** run on develop (head_branch=develop AND event push/schedule — never a PR run): optional **email** via `dawidd6/action-send-mail@v3`, plus open/comment a **deduped P0 `develop-red` issue** (label auto-created, assigned `marmos91`). On a green postsubmit run: **auto-close** any open `develop-red` issue ("tree reopened").
- **Email guard:** job env `HAS_SMTP: ${{ secrets.SMTP_PASSWORD != '' }}` + `if: env.HAS_SMTP == 'true'` on the mail step — CI never fails just because SMTP secrets aren't set.
- **Warn-on-PR (warn-only, non-blocking):** if an open `develop-red` issue exists, posts/updates a single idempotent comment (hidden HTML marker) on the PR. Same-repo PRs only (fork PRs get a read-only token).

### Enabling email (manual, after merge)

```
gh secret set SMTP_SERVER     --body "smtp.example.com"
gh secret set SMTP_PORT       --body "587"
gh secret set SMTP_USERNAME   --body "ci@example.com"
gh secret set SMTP_PASSWORD   --body "<app-password>"
gh secret set ALERT_EMAIL_TO  --body "marco.moschettini@cubbit.io"
```

Until these exist the alarm still opens/closes the `develop-red` issue; only the email step no-ops.

Refs #1156, #1157, #1158, #14.

### DIFF
--- a/.github/scripts/ci_health_email.py
+++ b/.github/scripts/ci_health_email.py
@@ -1,0 +1,148 @@
+#!/usr/bin/env python3
+"""Build the develop-red alarm email (HTML + text), subject, and issue markdown.
+
+Reads run facts from the environment and pre-collected failure data from /tmp:
+  - /tmp/failed_jobs.ndjson  one JSON object per failed job: {name, url, step}
+  - /tmp/failing_tests.txt   best-effort failing test names, one per line
+
+Writes (consumed by ci-health.yml):
+  /tmp/mail_subject, /tmp/mail.html, /tmp/mail.txt, /tmp/issue.md
+
+Best-effort by design: missing/empty inputs degrade gracefully to a
+links-only message, never an error — the alarm must always send.
+"""
+import html
+import json
+import os
+import pathlib
+
+CAP = 20  # max test names listed inline; the rest live in the logs
+
+W = os.environ.get("WORKFLOW", "conformance")
+C = os.environ.get("CONCLUSION", "failure")
+SHA = os.environ.get("HEAD_SHA", "")
+SHORT = SHA[:8] if SHA else "unknown"
+RUN_URL = os.environ.get("RUN_URL", "")
+REPO = os.environ.get("REPO", "")
+ARTIFACTS_URL = f"{RUN_URL}#artifacts" if RUN_URL else ""
+COMMIT_URL = f"https://github.com/{REPO}/commit/{SHA}" if REPO and SHA else RUN_URL
+
+
+def _read_jobs(path):
+    jobs = []
+    try:
+        for line in pathlib.Path(path).read_text().splitlines():
+            line = line.strip()
+            if line:
+                jobs.append(json.loads(line))
+    except (FileNotFoundError, json.JSONDecodeError):
+        pass
+    return jobs
+
+
+def _read_lines(path):
+    try:
+        return [x for x in pathlib.Path(path).read_text().splitlines() if x.strip()]
+    except FileNotFoundError:
+        return []
+
+
+jobs = _read_jobs("/tmp/failed_jobs.ndjson")
+tests = _read_lines("/tmp/failing_tests.txt")
+
+subject = f"\U0001F534 [DittoFS] develop RED — {len(jobs)} failed job(s) in {W} ({SHORT})"
+
+# --- failed jobs (HTML + text) ---
+jobs_html, jobs_text = [], []
+for j in jobs:
+    name = j.get("name", "?")
+    url = j.get("url") or RUN_URL
+    step = j.get("step") or ""
+    step_html = f" — failed step: <code>{html.escape(step)}</code>" if step else ""
+    jobs_html.append(
+        f'<li>✗ <b>{html.escape(name)}</b>{step_html} '
+        f'&nbsp;<a href="{url}">▶ logs</a></li>'
+    )
+    jobs_text.append(
+        f"  ✗ {name}" + (f" (failed: {step})" if step else "") + f"\n     {url}"
+    )
+if not jobs:
+    jobs_html.append(f'<li>See <a href="{RUN_URL}">the run</a> for failed jobs.</li>')
+    jobs_text.append(f"  see {RUN_URL}")
+
+# --- failing tests (HTML + text) ---
+if tests:
+    shown = tests[:CAP]
+    tests_html = (
+        "<p><b>Detected failing tests</b> (best-effort — logs are authoritative):</p><ul>"
+        + "".join(f"<li><code>{html.escape(t)}</code></li>" for t in shown)
+        + "</ul>"
+    )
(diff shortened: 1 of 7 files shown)


marmos91:
Thanks @copilot — addressed all five in commit 3ef4a95b:

1-3. **Non-`failure` conclusions** (`timed_out`, `cancelled`, etc.): the alarm now keys on `IS_RED = conclusion != 'success'`, so any non-green terminal conclusion opens/updates the issue and sends the email.
4. **Premature close**: the close step now verifies every monitored workflow's latest completed run on develop is `success` before closing; otherwise it comments and keeps the issue open.
5. **Comment pagination**: the marker lookup now uses `--paginate ... per_page=100` so the idempotency marker is never missed.

marmos91:
Review pass applied (commit `948a7b60`):

**Critical fix** — `from:` used `SMTP_USERNAME` (the SES access-key id, not an address) → would bounce every alarm. Now `ALERT_EMAIL_FROM` (verified SES identity).

**Email enriched** (per request — "what failed + easy links"):
- failed-job names + **deep per-job log links** (one click to the failure) + failed step name
- best-effort **failing test names** (POSIX grader output, TAP not-ok, smbtorture/WPTS ids), capped at 20 → logs
- HTML + text body, scannable subject with failed-job count
- same detail reused in the `develop-red` issue
- STARTTLS(587)/implicit-TLS(465) auto-selected by port
- body templating moved to a committed, unit-testable `.github/scripts/ci_health_email.py` (no fragile YAML heredocs)

**Email infra is LIVE** — Amazon SES (eu-west-1), all 6 secrets set (`SMTP_SERVER/PORT/USERNAME/PASSWORD`, `ALERT_EMAIL_TO`, `ALERT_EMAIL_FROM`). SMTP auth tested PASS; sample alarm delivered. SES is in sandbox (sends only to verified addresses — fine for self-alarm). Deliverability note: sender is @cubbit.io which isn't DKIM-signed for SES yet → may land in spam until we add domain auth (cubbit.io not in our Route53; can do cubbit.eu instead).

Ready for review.

marmos91:
Rebased onto develop after #1171 merged (the old Lane-B agent re-pushed its closed branch as #1171 and merged it — it was following its original 'merge on green' instruction; closing #1167 didn't stop it).

#1171 made a **partial** SMB change to `smb-conformance.yml`; this PR's rewrite of the same file is the **superset** (full event-tiering). Reconciled the conflict by taking this PR's version and grafting in #1171's two correct specifics:
- **dropped `smb2.create`** from the Kerberos filter loop — its known failures live only in the non-Kerberos KNOWN_FAILURES.md, so it grades RED under `--kerberos` (rationale comment added)
- **restored the `workflow_dispatch` clamp** — an unsupported (*-s3) smbtorture dispatch input now falls back to memory instead of crashing run.sh

Net: when this merges, develop gets the unified event-tiered `smb-conformance.yml`, superseding #1171's interim version. #1171's `smb-client-compat.yml` go-cache change is untouched (stays).

PR #1177, opened 2026-06-15
fix(operator): default make-deploy IMG to published image (avoid local-only controller:latest)

marmos91:
## Problem

The operator `Makefile` defaulted `IMG ?= controller:latest`. A fresh user running `make deploy` would have `kustomize edit set image controller=controller:latest` applied to the manager Deployment — an image that exists only on a machine that has run `make docker-build` locally. On any real cluster the operator pod sits in `ImagePullBackOff` until the user figures out they must build and load the image themselves.

## Fix

Default `IMG ?= marmos91c/dittofs-operator:latest` — the same published image the Helm chart (`chart/values.yaml`) already references and that `release.yml` builds and pushes. `make deploy` now pulls a real, public image out of the box. Still overridable via `IMG=<your-registry>/...` for source builds.

Verified with kustomize v5.7.1: `kustomize build config/default` resolves the manager container image to `marmos91c/dittofs-operator:latest`. No CRD/manifest diff (`make deploy` performs the `kustomize edit` at runtime; the kustomization file is not committed).

## docker-push consideration

The same `IMG` default also feeds `docker-build` / `docker-push`, so a bare `make docker-push` now targets the public `:latest`. This is acceptable:

- The push requires Docker Hub push credentials for the `marmos91c` namespace, which a fresh user does not have, so it fails safely rather than clobbering the published image.
- `release.yml` remains the authoritative publisher.

A separate `DEPLOY_IMG` was considered but rejected as added surface; the minimal one-line default fix is the right altitude. Users building from source use `make docker-build docker-push IMG=<your-registry>/dittofs-operator:tag`.

## Docs

Updated `k8s/dittofs-operator/README.md` and `docs/DEPLOYMENT.md` to state that `make deploy` uses the published image by default and how to build/deploy from source via `IMG=`.

### DIFF
--- a/docs/DEPLOYMENT.md
+++ b/docs/DEPLOYMENT.md
@@ -136,5 +136,10 @@ kubectl apply -f config/samples/dittofs_v1alpha1_dittofs.yaml
 kubectl get dittofs
 ```
 
+`make deploy` uses the published operator image (`marmos91c/dittofs-operator:latest`) by
+default. To build and deploy from source instead, point `IMG` at your own registry:
+`make docker-build docker-push IMG=<your-registry>/dittofs-operator:tag` then
+`make deploy IMG=<your-registry>/dittofs-operator:tag`.
+
 See the [`operator/`](../operator/) directory for the CRD reference, RBAC, and Helm chart
 configuration.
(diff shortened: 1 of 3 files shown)


marmos91:
Fixed both stale path references in the DEPLOYMENT.md operator section (66c47179): `cd operator` -> `cd k8s/dittofs-operator` and the `../operator/` link -> `../k8s/dittofs-operator/`.

PR #1187, opened 2026-06-15
feat(controlplane): clarify forced admin password-change error + make it configurable

### LINKED ISSUE #1162: Onboarding: clarify the 'password change required' error + make the forced admin password change configurable
## Context
Reported via early-adopter feedback (@niccolosanarico). A first-time user follows onboarding, logs in as `admin`, and runs:
```
dfsctl user create --username $(whoami) --host-uid ...
```
and hits:
```
Error: failed to create user: Password change required. Please change your password before proceeding.
```
They had no idea this refers to the **admin** account's first-login forced password change — the message reads as if it's about the *new* user's password, so they get stuck.

## Source
- `internal/controlplane/api/middleware/auth.go:169` — middleware rejects every authenticated op with `http.Error(w, "Password change required. Please change your password before proceeding.", 403)` whenever the JWT's `MustChangePassword` claim is set.
- `pkg/controlplane/models/admin.go:39` — the bootstrap admin is created with `MustChangePassword: true`.
- `pkg/controlplane/api/router.go:152` — only the change-password endpoint is exempt.

## Asks
1. **Clearer error.** Make it explicit that the *currently authenticated account* (the admin during onboarding) must set a new password first, and tell them how, e.g.:
   > `admin password change required: run 'dfsctl user change-password' before performing other operations`
   Surface it the same way in `dfsctl` so the CLI doesn't just echo the raw 403 body.
2. **Make the forced change optional.** Add a dfs config knob (e.g. `controlplane.require_initial_password_change`, default true) so operators can disable the first-login forced change for the bootstrap admin. When disabled, the bootstrap admin is created with `MustChangePassword: false`.
3. Update the onboarding guide to mention the `dfsctl user change-password` step explicitly.

marmos91:
## Summary

Fixes the onboarding friction reported in #1162: a first-time operator logs in as `admin`, runs `dfsctl user create ...`, and hits a 403 whose message reads as if it concerns the *new* user's password — when it's actually the bootstrap **admin's** forced first-login password change. Two changes:

### 1. Clearer error message
`internal/controlplane/api/middleware/auth.go` now names the authenticated account and the exact command to unblock, instead of the cryptic generic text:

> `account "admin" must set a new password before performing other operations; run 'dfsctl user change-password' (as "admin") to set one`

The middleware writes plain text, which `dfsctl` surfaces verbatim via `APIError.Detail`, so the CLI no longer echoes a confusing 403.

### 2. Configurable forced change (opt-out)
New knob `controlplane.require_initial_password_change` (env: `DITTOFS_CONTROLPLANE_REQUIRE_INITIAL_PASSWORD_CHANGE`):

- **Default `true`** — current secure-by-default behavior is preserved. Implemented as a `*bool` so an unset value defaults to forced-change ON (same idiom as the SMB/NFS `Enabled *bool` opt-out flags), with `APIConfig.RequiresInitialPasswordChange()` and an `ApplyDefaults` default.
- Set to `false` for automated/test deployments: the bootstrap admin is then created with `MustChangePassword: false`.
- Threaded through `EnsureAdminUser(ctx, requireInitialPasswordChange)` (store interface, GORM store, control-plane wrapper, `dfs start`).
- `DITTOFS_ADMIN_INITIAL_PASSWORD` continues to skip the forced change as before (the operator already chose the password).

### Docs
- `docs/CONFIGURATION.md`: documented the new key (YAML example + options table).
- `pkg/config/init.go`: commented-out knob in the generated config template.
- `README.md`: updated the first-run onboarding section with the new error wording and the opt-out knob.

No Cobra commands/flags changed, so `docs/CLI.md` does not need regeneration.

## Tests
- `pkg/controlplane/store`: provisioning admin with the knob **off** → `MustChangePassword=false`; default → `true`.
- `pkg/controlplane/api`: `RequiresInitialPasswordChange()` (nil→true, explicit true/false) and `ApplyDefaults` default/opt-out preservation.
- `internal/controlplane/api/middleware`: the 403 body names the account and mentions `dfsctl user change-password`.
- `pkg/config`: default true, file opt-out, and env opt-out (`...=false`) through the full `Load()` path; key added to the `configEnvKeys` coverage guard.

Closes #1162

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,9 +138,12 @@ DITTOFS_ADMIN_INITIAL_PASSWORD=my-secure-password ./dfs start
 # 1a. REQUIRED on first login: change the admin password.
 # The bootstrap admin starts with a forced-password-change flag — until you
 # clear it here, every other command (including `user create` below) is
-# rejected with HTTP 403 "Password change required. Please change your
-# password before proceeding." (Skipped automatically if you set your own
-# DITTOFS_ADMIN_INITIAL_PASSWORD at first start.)
+# rejected with HTTP 403:
+#   account "admin" must set a new password before performing other operations;
+#   run 'dfsctl user change-password' (as "admin") to set one
+# This forced change is skipped automatically when you set your own
+# DITTOFS_ADMIN_INITIAL_PASSWORD at first start, and can be disabled entirely
+# via controlplane.require_initial_password_change: false (see docs/CONFIGURATION.md).
 ./dfsctl user change-password
 
 # 2. Create a user mapped to your host UID (needed for NFS write access)
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -114,8 +114,10 @@ func runStart(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("failed to initialize control plane store: %w", err)
 	}
 
-	// Ensure admin user exists (generates random password on first run)
-	adminPassword, err := cpStore.EnsureAdminUser(ctx)
+	// Ensure admin user exists (generates random password on first run).
+	// Whether the first-login password change is forced is operator-configurable
+	// (controlplane.require_initial_password_change, default true).
+	adminPassword, err := cpStore.EnsureAdminUser(ctx, cfg.ControlPlane.RequiresInitialPasswordChange())
 	if err != nil {
 		return fmt.Errorf("failed to ensure admin user: %w", err)
 	}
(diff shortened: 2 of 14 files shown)


marmos91:
Addressed both Copilot review comments: the new `MustChangePassword` assertions in `TestEnsureAdminUser` and `TestEnsureAdminUser_OptOutForcedChange` now clear `DITTOFS_ADMIN_INITIAL_PASSWORD` via `t.Setenv(models.EnvAdminInitialPassword, "")`, so they're deterministic regardless of an ambient env var in the dev/CI shell. Verified by running them with `DITTOFS_ADMIN_INITIAL_PASSWORD=ambient` set — both still pass.

Issue #1166, opened 2026-06-15
PostgreSQL schema: decouple inode identity from a single canonical path (hard-link modeling)


marmos91:
## Motivation
The postgres `files` table stores **one canonical `path` (+ `path_hash`) per inode**. Hard links (N names → 1 inode) don't fit that model: the inode has a single `path` column, so it goes stale whenever the name matching `files.path` is removed while other links survive.

This single-path assumption has produced a recurring class of postgres-only bugs:
- **#1160** — rename-overwrite of a multiply-linked target hit the `unique_share_path_hash_active` index (fixed by dropping the index in #1165).
- **#190** — recycle (Move into #recycle) failed `already exists` for the same reason.
- General: a surviving hard-linked inode's `files.path` can be stale after an overwrite (cosmetic on the live path — access is handle/parent_child_map based — but wrong for path-keyed snapshot/restore of hard-linked files).

memory/badger don't carry a canonical path on the inode and don't have these issues.

## Proposed direction (to design)
Treat `parent_child_map (parent_id, child_name → child_id)` as the **sole source of truth** for the namespace, and stop treating `files.path`/`path_hash` as inode-identifying:
- Derive paths on demand (recursive CTE over parent_child_map) instead of storing a canonical path — this also removes the per-level N-round-trip walk noted during #1153.
- Or keep `files.path` strictly as a non-authoritative cache, never used for uniqueness or restore keying.
- Reconcile snapshot/restore + ResetLocalState to be parent_child_map-keyed, not path-keyed.

## Scope / risk
Schema migration + touches PutFile/Move/snapshot/restore. Larger change; #1165 is the immediate unblock. This issue tracks the considered refactor.

marmos91:
## Scope complete — design + PR breakdown

Full scope doc: `.planning/v1.0-audit/1166-pg-inode-path-decouple-SCOPE.md`.

**Root problem:** a `files` row conflates inode identity/attrs (shared across hard links) with one canonical `path`/`path_hash`/`content_id` (only one of N names). #1165 dropped the unique index — a band-aid — but left 5 structural issues: dead path after unlink, arbitrary `GetParent`, `updateDescendantPaths` silently staling other names, path-derived `content_id` going stale after rename (breaks `GetFileByPayloadID`/flusher).

[...]

Issue #1160, opened 2026-06-15
POSIX: PostgreSQL rename-overwrite of existing target fails (rename/23.t — EEXIST + nlink=2)


marmos91:
Split out from #1153, which **misclassified** this as a long-path bug. The failing paths in `rename/23.t` are normal-length (`pjdfstest_<32hex>`), **not** deep/long paths, and there is **no** server-side IO/context-cancel error (that root cause — a `queryRow` defer-cancel-before-lazy-scan — is fixed separately, clearing the other 13 #1153 rows).

## Symptom (PostgreSQL backend, NFSv4.1; memory/badger pass)
`rename/23.t` fails 20/42 subtests across all special-file types (regular, fifo, block, char, socket):
```
not ok 5  - rename A B, expected 0, got EEXIST
not ok 6  - lstat <file> type,nlink expected regular,1 got regular,2
not ok 10 - mkfifo A 0644, expected 0, got EEXIST
not ok 34 - bind A, expected 0, got EADDRINUSE
not ok 42 - rmdir D, expected 0, got ENOTEMPTY
```
The `rename A B` (where B already exists) returns EEXIST instead of atomically overwriting B. The `nlink=2` (expected 1) and the cascade of EEXIST/EADDRINUSE/ENOTEMPTY on subsequent ops are collateral: the source is left in place and a link count is wrongly incremented — i.e. the rename behaved like a partial copy, not a move.

## Where to look
`pkg/metadata/file_modify.go:Move` is backend-agnostic and *does* delete the destination child (`tx.DeleteChild`) before `tx.SetChild`, so the divergence is in a PostgreSQL store primitive invoked by that transaction — likely `SetChild`/`DeleteChild` ordering against the `parent_child_map` UNIQUE(parent_id,name) constraint, or the link-count update path. memory/badger overwrite the child mapping silently; postgres appears to retain the old mapping/link.

Repro: `setup-posix.sh postgres --nfs-version 4.1` then `prove /path/pjdfstest/tests/rename/23.t`. Tracked as a blacklist row in `test/posix/KNOWN_FAILURES_V4.md`.

marmos91:
[...]

Issue #190, opened 2026-02-25
Trash / soft-delete with configurable retention


marmos91:
## Summary

Move deleted files to a `.trash/{date}/` directory instead of immediately deleting them. Configurable retention period with background cleanup. Prevents accidental data loss in production environments.

## Motivation

- **Data safety**: No recovery path today — `rm` is permanent and immediate
- **Production confidence**: Operators expect a safety net for accidental deletions
- **Deferred deletion**: Reduces delete latency (metadata move vs S3 delete)

### How It Works

```
rm file → Move to .trash/2026-02-25-14/ → Background cleanup after retention expires
```

### Configuration

```yaml
shares:
  - name: /data
    trash:
      enabled: true
      retention_days: 7    # Default: 7 days
```

### CLI Commands

```bash
# List trash contents
dfs trash list

# Empty trash immediately (all expired)
dfs trash empty

# Empty trash for specific share
dfs trash empty --share /data

# Force empty all (ignore retention)
dfs trash empty --force
```

### Implementation

- **Metadata-level**: Trash entries stored as special metadata records with original path, deletion time, and expiry
- **Background loop**: Hourly scan for expired trash entries → actual block deletion
- **Hidden directory**: `.trash/` at share root, not visible in normal READDIR
- **Restore**: Future enhancement — `dfs trash restore <path>`

## Key Files

- `pkg/metadata/` - Metadata store interface (needs trash entry support)
- `cmd/dfs/commands/` - CLI commands directory

marmos91:
Shipped in #954 (merged to develop). Per-share recycle bin: visible `#recycle` dir, opt-in `enable-trash` (default off, live-toggleable), recycle on unlink + replace-overwrite, retention/restrict-admin/max-size/exclude settings, deferred block deletion, hourly reaper, disable auto-empties. `dfsctl trash list/restore/empty/status` + `share` config flags + REST. Cross-backend conformance (memory/badger/postgres) + NFS/SMB e2e. (Auto-close doesn't fire on develop merges, closing manually.)

Issue #1176, opened 2026-06-15
perf(postgres): reduce per-read query overhead (collapse GetFile, batch READDIR hydration)


marmos91:
Split from #1173. Unconditional, low-risk query/round-trip reduction for the postgres metadata store — worth doing regardless of whether a cache (#1173) is ever added.

## Rationale (from #1169 / #1172 measurements)

postgres metadata reads are **round-trip-bound**, not scan-bound: `GetFile`/`GetChild` are already point lookups on PK / unique indexes, and prepared-statements on/off made no difference. So **indexing won't help** — but shaving the per-op query overhead will, for *every* postgres deployment, with **no invalidation risk**.

160µs/op (loopback) is acceptable; this is just removing obvious fat.

## Scope

- **Collapse `GetFile`'s 1–2 queries into one.** It currently does a metadata `SELECT` plus a separate `loadFileBlockRefs`. Skip the second query for directories / blockless files, or fold block refs into a single round-trip (LATERAL / json_agg).
- **Batch READDIR hydration.** `ListChildren` per-row attribute hydration → a single pgx batch / set-returning query.
- **Review connection-pool sizing** (`max_conns`/`min_conns`) under concurrency.

## Acceptance

- Re-run `dfsbench metadata --backend postgres` (the #1172 tool) before/after — quantify the win (expect bounded ~1.5–2×).
- Zero behavior change; conformance suite stays green.
- No new caching/invalidation surface (that's #1173, separate).

marmos91:
Fixed in #1195 (merged to develop): GetFile collapsed from 2 round-trips to 1 (block refs folded into the metadata query via json_agg, byte-identical ordering/content). Measured ~1.8× getattr throughput / ~45% p50 cut via dfsbench before/after. Pool defaults retained (measured not the bottleneck). Zero behavior change; full pg conformance + NFS/WPTS postgres-s3 e2e green.

PR #1192, opened 2026-06-15
feat(postgres): rename files table to inodes, drop path columns (part 1 of #1166)

marmos91:
Part of #1166 — **does NOT close the issue.** This is PR-1 of the 5-PR plan to
decouple inode identity from a single canonical path (hard-link modeling).

## What PR-1 does

Renames the PostgreSQL `files` table to `inodes` and drops the path-as-key
columns, keeping behaviour identical and the full conformance suite green.

`parent_child_map(parent_id, child_name → child_id)` is already the sole source
of truth for the namespace. The denormalized `path`/`path_hash` columns on the
inode were the root of a class of postgres-only hard-link bugs (#1160
rename-overwrite, #190 recycle "already exists") because a multiply-linked inode
has only ONE path column that goes stale when the matching name is removed.
#1165 (migration 000031) dropped the `unique_share_path_hash_active` index as a
stopgap; this PR does the proper decoupling. memory/badger never carried a
canonical path on the inode — this aligns postgres with them.

Handle stability (invariant #3) is preserved: a handle is `shareName:UUID =
inodes.id`, and a table rename does not change the UUID.

### Changes
- **Migration `000032_rename_files_to_inodes`** — `ALTER TABLE files RENAME TO
  inodes`; drop `path`/`path_hash`, the `idx_files_share_path_hash` index, and
  the `files_path_hash_trigger` + `update_path_hash()` function; rename the
  files-named indexes/triggers to `inodes_*`. `content_id`/`content_id_hash`
  are untouched (they key `file_blocks.id`; decoupling them is PR-3).
- **All postgres SQL retargeted** `files` → `inodes`. FK references
  (`parent_child_map`, `link_counts`, `shares`, `pending_writes`,
  `durable_handles`, …) follow the table rename automatically — PostgreSQL
  repoints FKs to the renamed table, so no FK drop/recreate is needed.
- **`PutFile`** INSERT/UPDATE no longer write `path`; **`GetFile`** /
  payload-id lookup no longer select a stored path.
- **Root lookup** resolves via `shares.root_file_id` instead of `WHERE path =
  '/'` (in both `CreateRootDirectory` and `getExistingRootDirectory`).
- **Backup** schema version bumped `v2 → v3`; `backupTables` lists `inodes`;
  the migration-coverage audit test is now rename-aware.
- **errors.go** object_id unique-violation mapping follows the renamed index
  (`files_object_id_idx` → `inodes_object_id_idx`).

## Down-migration tradeoff

The up migration **drops** the canonical `path` string, which is not
recoverable verbatim. The down migration restores a **working, fully-indexed**
schema: it re-adds the columns, recreates the index/trigger/function, and
**reconstructs each inode's path from `parent_child_map`** (the authoritative
namespace), choosing one deterministic name per inode (lexicographically
smallest full path for a hard-linked inode); unreachable/orphaned rows get `/`.
So DOWN restores schema + behaviour, not the exact prior string for hard-linked
files — documented inline in the `.down.sql`. It does **not** recreate
`unique_share_path_hash_active` (000031 owns that revert). Verified up → down →
up round-trips cleanly with seeded hard-link data against real PostgreSQL 16.

## How `File.Path` is now derived

The Go `metadata.File.Path` field is still populated on read (callers and the
conformance suite derive child paths from a parent's `Path`). It is
reconstructed in `GetFile`/payload-id lookups via a recursive `parent_child_map`
walk up to the share root (`inodePathExpr`, a correlated scalar subquery — one
round-trip with the row). For a hard-linked inode this returns **one** of its
paths, which is acceptable (POSIX doesn't guarantee which path `stat()` shows).
A `depth < 4096` guard bounds the walk so a corrupt directed cycle in
`parent_child_map` can never make the query recurse forever.

## Test evidence (real PostgreSQL 16)

Ran against a docker `postgres:16` with `DITTOFS_TEST_POSTGRES_DSN` set:

```
go test -tags integration -count=1 ./pkg/metadata/store/postgres/...
ok  github.com/marmos91/dittofs/pkg/metadata/store/postgres  18.441s
```

Green, including: full `TestConformance` (FileOps/CreateHardLink, Rename,
NestedDirectories, DirOps, Permissions, DurableHandles), `TestBackupConformance`
(backup → restore round-trip on the renamed `inodes` table),
`TestPostgresPutFile_ObjectIDConflictMapsToErrConflict`,
`TestPostgresPutFile_DuplicatePathAllowed`, `TestPostgres_PutFile_UsedBytesAccurate`,
and `TestBackupTablesCoversAllMigrations`. `go build ./...`, `go vet`, `gofmt`
all clean. Migration applies cleanly forward, and the down migration runs.

Out of scope (later PRs): deleting `updateDescendantPaths` + Move path tracking
(PR-2), UUID-based content_id (PR-3), `link_counts` → `inodes.nlink` (PR-4),
full hard-link conformance cases (PR-5).

### DIFF
--- a/pkg/metadata/store/postgres/backup.go
+++ b/pkg/metadata/store/postgres/backup.go
@@ -27,7 +27,13 @@ const (
 	// Increment when the payload layout changes (new table, changed
 	// framing, etc.) and add a migration path in Restore.
 	// v2 adds the v4_client_recovery table section.
-	postgresSchemaVersion = uint32(2)
+	// v3 renames the files table to inodes and drops the path/path_hash
+	// columns (#1166). The payload layout changed (the inode section name and
+	// its column set), so restore rejects older streams with
+	// ErrSchemaVersionMismatch — consistent with the v1->v2 bump, which added
+	// a table with no cross-version restore shim. Pre-1.0 backups are not
+	// forward-portable across schema versions.
+	postgresSchemaVersion = uint32(3)
 )
 
 // backupTables lists every metadata table in FK-safe dependency order
@@ -40,7 +46,7 @@ const (
 var backupTables = []string{
 	"server_config",
 	"filesystem_capabilities",
-	"files",
+	"inodes",
 	"shares",
 	"parent_child_map",
 	"link_counts",
@@ -352,7 +358,7 @@ func reconcileFileBlockHashes(ctx context.Context, raw *pgconn.PgConn) error {
 		UPDATE file_blocks fb
 		SET hash = encode(r.hash, 'hex')
 		FROM file_block_refs r
-		JOIN files f ON f.id = r.file_id
+		JOIN inodes f ON f.id = r.file_id
 		WHERE fb.hash IS NULL
 		  AND f.content_id IS NOT NULL
 		  AND fb.id = f.content_id || '/' || r."offset"`
--- a/pkg/metadata/store/postgres/encoding.go
+++ b/pkg/metadata/store/postgres/encoding.go
@@ -60,6 +60,10 @@ func pgNanosToTime(n int64) time.Time {
 // fileRowToFileWithNlink converts a database row to a File struct, including link count.
 // Expected columns: id, share_name, path, file_type, mode, uid, gid, size,
 // atime, mtime, ctime, creation_time, content_id, link_target, device_major, device_minor, hidden, acl, object_id, deleted_at, original_path, deleted_by, link_count
+//
+// The `path` column is no longer stored on the inode (#1166); callers supply it
+// as a reconstructed expression (inodePathExpr) walking parent_child_map up to
+// the share root. For a hard-linked inode this yields one of its paths.
 func fileRowToFileWithNlink(row pgx.Row) (*metadata.File, error) {
 	var (
 		id           uuid.UUID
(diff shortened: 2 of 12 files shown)


marmos91:
Pre-merge gates run on the final commit:
- **code-simplifier**: removed a dead `node_id` column from the path CTE and added a `depth < 4096` cycle guard (parent_child_map has no DB acyclicity constraint; without the cap a corrupt cycle would hang GetFile until statement_timeout). 4096 = MaxPathLen ceiling, so it never truncates a legitimate path.
- **code-reviewer**: confirmed the `files`→`inodes` rename is exhaustive (zero residual SQL refs), `inodePathExpr` is a Go const (no injection), correlated-subquery aliasing is correct. Its one HIGH (cycle guard) is the item the simplifier addressed.
- Verified green against **real PostgreSQL 16**: `go test -tags integration ./pkg/metadata/store/postgres/...` ok (19s), incl. the reset-then-restore conformance that exercises the down-migration path reconstruction.

**Perf note for #1176 (next in the cluster):** GetFile now derives `File.Path` via a recursive `parent_child_map` walk per call (single-row reads only — *not* ListChildren/readdir). That's added cost on the exact hot path #1176 targets; the likely follow-up is to drop `File.Path` from GetFile where callers don't need it rather than reconstruct it.

Issue #1188, opened 2026-06-15
Prometheus metrics surface for the DittoFS server (/metrics endpoint + core instrumentation)


marmos91:
## Summary

DittoFS today has **no metrics/observability surface** on the server itself — only structured `zap` logs and a handful of ad-hoc counters exposed via the REST share-usage response (`UsedBytes`, `PhysicalBytes`). pprof is wired for profiling, but there is no Prometheus `/metrics` endpoint or instrumentation. (The k8s operator uses controller-runtime's Prometheus exporter, but that covers the operator, not the DittoFS server.)

We should add a first-class Prometheus metrics surface for the whole server so operators can build dashboards/alerts instead of grepping logs.

## Motivation

Several in-flight features add internal counters that are currently only observable via logs and would be natural Prometheus series — e.g. #1136 (local-cache `unsynced_bytes`, backpressure engage-count / stall time), per-share/per-identity quota usage (#1151), syncer lag, GC reclaim, block-store disk usage, request rates/latencies per protocol. These are being computed but deliberately exposed as logs-only until a metrics framework exists.

## Scope

- [ ] Add a Prometheus registry + `/metrics` HTTP endpoint to the server (gated by config; default on/off TBD).
- [ ] Config surface: `DITTOFS_*` env + config file (port/path/enable).
- [ ] Instrument core subsystems as gauges/counters/histograms:
  - block store: local disk used vs max, **unsynced bytes**, backpressure engage-count + total stall time (#1136), eviction count, syncer upload/download throughput + lag, remote health.
  - metadata: used bytes (share + per-uid/gid once #1151 lands), file count, op latencies.
  - protocol adapters (NFS/SMB): request count/latency/errors per op.
  - GC: blocks scanned/reclaimed, live-set size.
- [ ] Operator/Helm: optional ServiceMonitor + scrape annotations.
- [ ] Docs: `docs/CONFIGURATION.md` metrics section + example dashboard notes.

## Notes

- Keep cardinality bounded — avoid per-file / per-handle labels.
- Until this lands, new features expose internal state via structured logs (with stable field names) so the eventual exporter is a drop-in over existing counters.
- Related: #1136 (backpressure counters are logs-only pending this), #1151 (per-identity quota usage).

marmos91:
## Implementation plan (approved)

Adding a Prometheus `/metrics` surface as a shared library, wired uniformly across the codebase.

### Architecture
[...]

PR #1201, opened 2026-06-15
feat(metrics): Prometheus /metrics surface (PR-1 of #1188)

marmos91:
## Summary

PR-1 of #1188 — a first-class Prometheus metrics surface for the DittoFS server. Additive and opt-in (disabled by default); zero changes to the data-plane hot paths.

### What's in this PR
- **`pkg/metrics`** — owned Prometheus registry (not the global default), `dittofs_build_info` + Go/process collectors, a **read-through collector** that emits runtime state as ConstMetrics at scrape time (nothing cached), and a **dedicated listener** with optional **bearer-token auth** and **TLS/mTLS**.
- **Config** — opt-in `metrics` block (`DITTOFS_METRICS_*`): `enabled` (default false), `host` (default `127.0.0.1`), `port` (9090), `path` (`/metrics`), `auth` (`none`|`token`), `token_file`, `tls`.
- **`runtime.MetricsSnapshot`** (implements `metrics.Provider`) — per-share capacity, sync durability (incl. **unsynced bytes** — the data-at-risk headline), snapshot freshness + held-count, per-principal **quota** usage/limits (configured principals only — bounded), and active client counts per protocol. The dependency is one-way (`runtime → metrics`) to avoid an import cycle, so later PRs can add inline instruments owned by `pkg/metrics`.
- **engine** — expose `UnsyncedBytes` in `BlockStoreStats`; add `GetStatsLite()` so the scrape path skips the per-file block-count DB walk.

### Metrics emitted (this PR)
`dittofs_build_info`, per-share `localstore_disk/memory_*`, `sync_pending_bytes`/`pending_uploads`/`uploads_total`/`upload_failures_total`, `remote_up`/`outage_seconds`/`reads_blocked_total`, `store_used_bytes`/`store_files`, `snapshot_active`/`snapshot_last_success_timestamp_seconds`, `quota_{used,limit}_{bytes,inodes}{scope,principal,share}`, `client_connections_active{protocol}`, plus the Go/process collectors.

Cardinality is bounded — labels are `share`/`protocol`/`scope`/`principal` only; quota series are emitted only for configured principals.

### Design notes
- **Read-through, nothing stored in Prometheus.** Persistent/functional state stays in its durable home and is exposed via the snapshot at scrape time → restart-safe.
- Dropped a misleading `physical_bytes` metric (it duplicated `localstore_disk_used_bytes`); `store_used_bytes` (logical) vs `localstore_disk_used_bytes` (on-disk) already gives the honest efficiency ratio. True post-dedup unique-chunk bytes deferred.

### Verification
- `go build ./...` clean; unit tests for `pkg/metrics` (collector output via `testutil.GatherAndCompare`, auth, local-only suppression) + `pkg/config`; touched packages (engine/runtime/shares/metadata) all green.
- Live: ran `dfs start` with metrics enabled, scraped `/metrics` (confirmed `dittofs_*` + Go/process families), and verified token auth — **401** without/with-wrong bearer, **200** with the correct token.

### Follow-ups (per #1188 plan)
PR-2 adapter RED + connection/auth counters · PR-3 eviction/backpressure + GC + snapshot op counters · PR-4 operator/Helm ServiceMonitor + docs. Removal of the parallel display-stats REST/CLI surface is intentionally **not** in this PR (would change `dfsctl` UX) — to be handled separately.

Refs #1188.


marmos91:
Addressed all three Copilot findings in a8d14a33: validate `metrics.path` starts with `/` (ServeMux panic guard), read logical bytes directly from the metadata store on the scrape path (dropped the unused GetShareUsage physical-bytes computation), and added MetricsConfig.Validate unit tests.

PR #1204, opened 2026-06-16
feat(metrics): adapter RED + connection + NFS auth instruments (PR-2 of #1188)

marmos91:
## Summary

PR-2 of #1188 — inline Prometheus instruments at the protocol-adapter chokepoints. Builds on the `pkg/metrics` registry from PR-1 (#1201). Additive; the registry is now built unconditionally so instruments always record, with only the `/metrics` listener gated by config.

### Instruments (owned by `pkg/metrics`, nil-safe `Record*` methods)
- `dittofs_adapter_requests_total{protocol,op,status}` + `dittofs_adapter_request_duration_seconds{protocol,op}` — RED. Wired for **NFS**: v3 uses precise `NFSStatus`; mount/nlm/nsm/v4 use coarse (err-based) status. v4 COMPOUND status is RPC-level only (per-op `NFS4ERR` lives inside the XDR result — documented; true per-op v4 RED is a follow-up).
- `dittofs_adapter_connections_total{protocol}` + `dittofs_adapter_connections_closed_total{protocol}` — recorded at the **shared** `base.go` accept/close points, so this covers **both NFS and SMB**. Active = total − closed.
- `dittofs_auth_attempts_total{protocol,mechanism}` + `dittofs_auth_failures_total{protocol,mechanism}` — at the NFS `rpcsec_gss` boundary (`mechanism="krb5"`).

### Wiring
The runtime carries the `*metrics.Metrics` handle (`SetMetrics`/`Metrics()`); adapters reach it via `Runtime` (`b.Registry.Metrics()`), so there's no per-adapter or AdapterFactory plumbing. `Metrics()` is nil-safe even on a nil `*Runtime`, and every `Record*` method guards a nil receiver — so call sites are unconditional and test adapters without a registry are safe.

### Cardinality
`op` values come from the fixed dispatch tables (bounded); labels are `protocol`/`op`/`status`/`mechanism` only.

### Verification
`go build ./...` clean; `pkg/metrics` (incl. new instrument tests: requests/conn/auth + nil-safe receiver), `pkg/adapter`, `pkg/adapter/nfs` all green. Reviewer pass addressed (added NSM + v4 NULL recording; documented v4 COMPOUND coarse status).

### Deferred to PR-2b
SMB per-command RED + SMB session auth-failure counters — the internal SMB dispatch deliberately avoids `pkg/` imports, so threading metrics through `ConnInfo`/`ProcessSingleRequest` is a separate, more invasive change. (SMB connection counts are already covered here via `base.go`.)

Refs #1188.

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -248,10 +248,17 @@ func runStart(cmd *cobra.Command, args []string) error {
 		"port", cfg.ControlPlane.Port,
 		"tls", apiServer.TLSEnabled())
 
+	// Build the metrics registry unconditionally so inline instruments (adapter
+	// RED, connection, auth counters) always record; only the /metrics listener
+	// is gated by config. The runtime carries the handle to every adapter.
+	m := metrics.New(Version, Commit)
+	m.RegisterProvider(rt)
+	rt.SetMetrics(m)
+
 	// Configure the Prometheus metrics endpoint (opt-in). It runs on its own
 	// listener, separate from the API server, so scrapers reach it without API
 	// auth and the two surfaces have independent lifecycles.
-	metricsServer, err := buildMetricsServer(&cfg.Metrics, rt)
+	metricsServer, err := buildMetricsListener(&cfg.Metrics, m)
 	if err != nil {
 		return err
 	}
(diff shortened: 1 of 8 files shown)


marmos91:
Addressed in aa7aed7c: (1) fixed the stale nil-safety comment — the server builds the registry unconditionally and gates only the listener; nil-safety is for call sites holding a nil handle. (2) The `ok|error` status is intentional and bounded — labelling by raw NFS status code would multiply series by op × ~20 codes; the ok bit is still derived accurately from v3 `NFSStatus` (not just transport err), and the metric help already says `(ok|error)`. Added a comment documenting this.

PR #1206, opened 2026-06-16
refactor(postgres): drop legacy link_counts table; inodes.nlink is sole source (PR-4 of #1166)

marmos91:
PR-4 of #1166. Consolidates hard-link count storage on the postgres backend: `inodes.nlink` (the existing `NOT NULL DEFAULT 1` column) becomes the single source of truth, and the redundant legacy `link_counts` table is dropped. Postgres-only; memory/badger backends are untouched and still pass the conformance suite unchanged.

## What
- **Removes dual-write redundancy.** `SetLinkCount` previously wrote BOTH `link_counts` (INSERT ... ON CONFLICT) AND `inodes.nlink`; now it keeps only `UPDATE inodes SET nlink`.
- **Migration `000034_drop_link_counts`.** up: defensively backfills `inodes.nlink` from `link_counts` (reconciles any divergence into the authoritative column), then drops the index and table. down: recreates the table (FK → `inodes(id)`, post-000032 rename) and repopulates from `inodes.nlink`.
- **One fewer join on the #1176 path.** `GetFile` / `GetFileByPayloadID` / `CreateRootDirectory` now select `f.nlink` directly instead of `LEFT JOIN link_counts`. The folded `blockRefsAggExpr` single-round-trip GetFile is preserved.
- **Schema v3 → v4.** `backup.go` drops `link_counts` from `backupTables` (used for both the COPY backup and the Reset TRUNCATE). Pre-v4 backups are rejected with `ErrSchemaVersionMismatch`, consistent with prior bumps — no cross-version shim.
- Directory inode INSERTs (root-dir create in `shares.go` + `transaction.go`) set `nlink = 2` directly.
- E2E `TruncateTables` helper updated to `inodes` (post-000032) and no longer references the dropped `link_counts`.
- `TestBackupTablesCoversAllMigrations` audit extended to honour `DROP TABLE` so a dropped table is no longer expected in `backupTables`.

## Behaviour parity
`PutFile` does not persist `File.Nlink` (nlink stays at its DEFAULT/SetLinkCount value) — this matches memory and badger, which maintain their link count separately from the inode struct and rely on the service always calling `SetLinkCount` after `PutFile`.

## Testing
- `go build ./...`, `go vet ./...` clean.
- `go test ./pkg/metadata/...` (memory + badger) green.
- Postgres integration conformance suite run against a FRESH db (migration applied at schema version 34): all subtests green, including hard-link, link-rename, backup-coverage, object_id-conflict (exercises SetLinkCount), and reset-then-restore. Verified `link_counts` table is gone and `inodes` remains. Validated the 000034 down→up roundtrip SQL end to end.

Refs #1166 (PR-5 closes it).

### DIFF
--- a/pkg/metadata/store/postgres/backup.go
+++ b/pkg/metadata/store/postgres/backup.go
@@ -28,12 +28,14 @@ const (
 	// framing, etc.) and add a migration path in Restore.
 	// v2 adds the v4_client_recovery table section.
 	// v3 renames the files table to inodes and drops the path/path_hash
-	// columns (#1166). The payload layout changed (the inode section name and
-	// its column set), so restore rejects older streams with
-	// ErrSchemaVersionMismatch — consistent with the v1->v2 bump, which added
-	// a table with no cross-version restore shim. Pre-1.0 backups are not
-	// forward-portable across schema versions.
-	postgresSchemaVersion = uint32(3)
+	// columns (#1166). v4 drops the legacy link_counts table; the hard-link
+	// count is consolidated onto inodes.nlink as the sole source of truth
+	// (#1166, part 4), so the backup table set no longer includes link_counts.
+	// The payload layout changed (one fewer table section), so restore rejects
+	// pre-v4 streams with ErrSchemaVersionMismatch — consistent with the prior
+	// bumps, which added/removed tables with no cross-version restore shim.
+	// Pre-1.0 backups are not forward-portable across schema versions.
+	postgresSchemaVersion = uint32(4)
 )
 
 // backupTables lists every metadata table in FK-safe dependency order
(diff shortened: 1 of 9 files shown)


marmos91:
Addressed Copilot review:
- **encoding.go comment**: fixed — added the missing `eas` column to the "Expected columns" list.
- **getExistingRootDirectory Nlink=0**: fixed — the existing-root path now selects and populates `f.nlink`, so the create and already-exists paths both return a correct link count.
- **SetLinkCount RowsAffected → ErrNotFound**: intentionally NOT changed. SetLinkCount is a no-existence-check upsert across all backends: memory writes `linkCounts[key]=count` and badger `Set(keyLinkCount,...)`, both unconditionally returning nil for a missing inode. Adding a RowsAffected/ErrNotFound check to postgres would diverge from the conformance contract. (Pre-change, the secondary `nlink` UPDATE was already unconditional too.)

PR #1217, opened 2026-06-16
feat(nfs): NFS-over-TLS (RFC 9289) — opportunistic AUTH_TLS + mTLS (#1197)

### LINKED ISSUE #1197: NFS-over-TLS (RFC 9289): opportunistic AUTH_TLS + mTLS
## Goal

Add **NFS-over-TLS (RFC 9289)** — opportunistic, in-transit confidentiality + peer authentication for the userspace NFS server. Target: **v1.0.0**.

## How RFC 9289 works

Client opens TCP, sends a **NULL RPC with `auth_flavor = AUTH_TLS (7)`** (a STARTTLS probe). Server replies with a verifier of the 8 ASCII bytes `"STARTTLS"`, then both perform a **TLS 1.3 handshake on the same TCP connection**; all subsequent RPC traffic is encrypted. Opportunistic: until the probe, traffic is treated as plain RPC, so non-TLS clients keep working. mTLS (client cert) is the peer-auth mode.

## DittoFS fit (clean — userspace, no kTLS/`tlshd`)

- **Hook**: in `NFSConnection.Serve()` (`pkg/adapter/nfs/connection.go`), before the RPC read-loop. Add a per-conn `tlsUpgraded` flag.
- **Probe**: define `AUTH_TLS = 7` in `internal/adapter/nfs/rpc/auth.go`; on the first NULL call with that flavor, reply with the `"STARTTLS"` verifier, then `crypto/tls.Server(conn, cfg)` and swap `c.conn` for the resulting `tls.Conn`.
- **Transparent**: RPC readers already take `io.Reader`; `c.conn` is `net.Conn`; `tls.Conn` satisfies both, so dispatch/framing is untouched. No kTLS needed — Go does the handshake + crypto in userspace (unlike the Linux kernel server, which offloads to `tlshd`).
- **Config**: add `NFSConfig.TLS` (`pkg/adapter/nfs/adapter.go`) — cert_file, key_file, client_ca (mTLS), min_version; mirror `controlplane.tls.*`. Modes: opportunistic vs require-TLS.
- **Reuse cert reloader**: lift the existing mtime-based `certReloader` from `pkg/controlplane/api/tls.go` into a shared `internal/tls` helper so API + NFS share one cert-loading/hot-reload path (survives certbot/cert-manager rotation).

## Client interop (document clearly)

- **Linux**: `mount -o xprtsec=tls` — needs `tlshd` (ktls-utils) + `CONFIG_NET_HANDSHAKE` (RHEL 9.x / upstream kernel 6.7+).
- **macOS**: no NFS-over-TLS client — falls back to plaintext or Kerberos.

## Tests

- AUTH_TLS probe → `"STARTTLS"` verifier → handshake → encrypted RPC roundtrip (Go client wrapping `crypto/tls`).
- Opportunistic: non-TLS client still served when TLS configured but not required.
- require-TLS mode rejects plaintext after the grace point.
- mTLS: client-cert required/verified; bad/absent cert rejected.

## Out of scope

SMB (already encrypted in-protocol via SMB3; see companion issues). kTLS offload. ACME/issuance (platform's job — we only consume cert files).

## References

- RFC 9289 <https://www.rfc-editor.org/rfc/rfc9289.html>; FreeBSD `rpc.tlsservd` `"STARTTLS"` verifier behavior.
- Existing server TLS to mirror: `pkg/controlplane/api/{config,server,tls}.go`.

marmos91:
Closes #1197. Builds on #1214 (shared `internal/tlsconfig`, already merged).

## What

NFS wire traffic was plaintext. This adds **RFC 9289** STARTTLS: a client opens TCP and sends a `NULL` RPC with `auth_flavor = AUTH_TLS (7)`; the server replies with the 8-octet `"STARTTLS"` verifier, then both perform a **TLS 1.3 handshake on the same connection** and all later RPC is encrypted. The handshake + crypto run in **userspace Go** — no kTLS / `tlshd` on the server. `tls.Conn` satisfies `net.Conn`, so RPC framing and dispatch are unchanged after the swap.

## Change

- **`internal/adapter/nfs/rpc`**: `AuthTLS = 7`, `StartTLSVerifier` ("STARTTLS"), `MakeStartTLSReply` (accepted NULL reply whose verifier carries the STARTTLS body).
- **`pkg/adapter/nfs/adapter.go`**: `NFSConfig.TLS` (`cert_file`/`key_file`/`client_ca`/`min_version`/`mode`) mirroring `controlplane.tls.*`. Server `*tls.Config` built once in `Serve` via the shared `internal/tlsconfig` (cert hot-reload for free); min raised to TLS 1.3 (the RFC floor).
- **`pkg/adapter/nfs/connection.go`**: per-conn `tlsUpgraded` flag + `handleTLSGate` hook before concurrent dispatch. Probe → write verifier, `tls.Server` handshake, swap `c.conn`. **Opportunistic** (default) still serves non-TLS clients in plaintext; **require** mode drops plaintext until upgrade. mTLS is free when `client_ca` is set (`RequireAndVerifyClientCert`).
- **`cmd/dfs/commands/start.go`**: parse `adapters.nfs.tls.*` from the adapter config map.

## Cert rotation

No NFS-specific watcher: it reuses the control-plane `CertReloader` (lifted to `internal/tlsconfig` in #1214), so rotated cert files are hot-reloaded with no restart, same as the API server.

## Test

`pkg/adapter/nfs`: full STARTTLS handshake + encrypted echo; mTLS (valid client cert accepted / missing rejected); gate decisions (require rejects plaintext / opportunistic passes through). `-race` clean.

```
go test -race ./pkg/adapter/nfs/ ./internal/adapter/nfs/rpc/
```

## Interop (documented)

- **Linux**: `mount -o vers=4.1,xprtsec=tls` — needs `tlshd` (ktls-utils) + `CONFIG_NET_HANDSHAKE` (RHEL 9.x / kernel 6.7+).
- **macOS**: no NFS-over-TLS client — Kerberos or a tunnel instead.

Docs: `docs/NFS.md` + `docs/SECURITY.md`.

## Out of scope

SMB (already SMB3-encrypted), kTLS offload, ACME/issuance, operator NFS-TLS rendering (follow-up).

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -390,7 +390,31 @@ func createNFSAdapter(cfg *models.AdapterConfig, kerberosConfig *config.Kerberos
 		port = 12049
 	}
 
-	adapter := nfs.New(nfs.NFSConfig{Enabled: true, Port: port})
+	nfsCfg := nfs.NFSConfig{Enabled: true, Port: port}
+
+	parsedConfig, err := cfg.GetConfig()
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse adapter config: %w", err)
+	}
+	if tlsCfg, ok := parsedConfig["tls"].(map[string]any); ok {
+		if v, ok := tlsCfg["cert_file"].(string); ok {
+			nfsCfg.TLS.CertFile = v
+		}
+		if v, ok := tlsCfg["key_file"].(string); ok {
+			nfsCfg.TLS.KeyFile = v
+		}
+		if v, ok := tlsCfg["client_ca"].(string); ok {
+			nfsCfg.TLS.ClientCA = v
+		}
+		if v, ok := tlsCfg["min_version"].(string); ok {
+			nfsCfg.TLS.MinVersion = v
+		}
+		if v, ok := tlsCfg["mode"].(string); ok {
+			nfsCfg.TLS.Mode = v
+		}
+	}
+
+	adapter := nfs.New(nfsCfg)
 	if kerberosConfig != nil && kerberosConfig.Enabled {
 		adapter.SetKerberosConfig(kerberosConfig)
 	}
--- a/docs/NFS.md
+++ b/docs/NFS.md
@@ -287,10 +287,36 @@ NFS uses RPC authentication flavors:
 | AUTH_UNIX | 1 | Unix UID/GID credentials |
 | AUTH_SHORT | 2 | Short-hand credential |
 | RPCSEC_GSS | 6 | Kerberos/GSS-API (NFSv4) |
+| AUTH_TLS | 7 | RFC 9289 STARTTLS probe (transport upgrade, not a data flavor) |
 
 **AUTH_UNIX format:** Stamp (4 bytes), machine name (string), UID (4 bytes), GID (4 bytes), supplementary GIDs (array, max 16).
 
-**Security note:** AUTH_UNIX credentials are not cryptographically secured and can be spoofed. NFSv4 adds RPCSEC_GSS for Kerberos-based authentication. For production deployments, consider running on trusted networks, enabling Kerberos (NFSv4/v4.1), or using VPN/network-level encryption.
+**Security note:** AUTH_UNIX credentials are not cryptographically secured and can be spoofed. NFSv4 adds RPCSEC_GSS for Kerberos-based authentication. For production deployments, consider running on trusted networks, enabling Kerberos (NFSv4/v4.1), or using NFS-over-TLS (below) / VPN / network-level encryption.
+
+### NFS-over-TLS (RFC 9289)
+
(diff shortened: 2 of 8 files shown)


marmos91:
## Live validation (real kernel client)

Validated end-to-end on a Scaleway Ubuntu 24.04 VM (kernel 6.8 + `CONFIG_NET_HANDSHAKE`, `tlshd`/ktls-utils 0.9) against `dfs` from this branch:

```
# opportunistic TLS adapter
dfsctl adapter edit nfs --config '{"tls":{"cert_file":"…/tls.crt","key_file":"…/tls.key","mode":"opportunistic"}}'
dfsctl adapter enable nfs --port 12049

mount -t nfs -o vers=4.1,xprtsec=tls,port=12049 localhost:/export /mnt/tls   # rc=0
```

- `tlshd`: **Handshake with localhost (127.0.0.1) was successful**
- server: **`NFS connection upgraded to TLS address=127.0.0.1:774 version=TLS 1.3`**
- encrypted **write+read verified** over the mount (uid-1000 user)
- opportunistic mode: a plaintext `vers=4.1` mount (no `xprtsec`) also succeeds — non-TLS clients keep working

**Operator note for cert issuance:** `tlshd` verifies the server cert against the host's reverse-DNS (PTR) name, not the mount target IP — mounting by IP failed `Certificate owner unexpected`; mounting by a name present in the cert SAN passed. NFS-TLS certs must include the client-facing hostname in their SAN.

Cert hot-reload is shared with the control plane via `internal/tlsconfig` (#1214), so certbot/cert-manager rotation needs no server restart.

PR #1272, opened 2026-06-21
fix(perms): enforce share permission as a ceiling, not a floor

### LINKED ISSUE #1240: ShareWritable bypass: NFS grants / SMB denies write on allow-only DACL (cross-protocol divergence)
Surfaced by AD-0 (#1232 / PR #1238) while building the cross-protocol permission
conformance matrix (`pkg/metadata/storetest/cross_protocol_permissions.go`). Pre-existing,
broader than AD-0's NFSv4-ACCESS consolidation, so documented + left out-of-scope there.

## The divergence
With **`ShareWritable=true`** (set by the SMB session layer on a read-write tree connect, and by
NFS export options) and an **allow-ONLY DACL** (no explicit DENY ACE) on a file whose POSIX mode
denies write (e.g. `0o000`), the two protocols disagree:

- **NFS** — `CheckPermissions` (`pkg/metadata/auth_permissions.go:274`) takes the share-level
  write bypass: `ctx.ShareWritable && !ctx.ShareReadOnly && !acl.HasExplicitDeny(file.ACL)` →
  grants write, **skipping the ACL** entirely.
- **SMB** — the open-time DACL gate (`CheckFileAccess`) ignores `ShareWritable` and **denies**
  write, because the allow-only DACL grants only read.

Same file, same identity, same share → write allowed over NFS, denied over SMB.

## Why it's not a security hole today
The DENY case is safe and pinned: `testCrossProtocolShareWritableBypass` asserts that with an
explicit DENY-write ACE + `ShareWritable=true`, write stays denied on **every** protocol (the
bypass is gated on `!acl.HasExplicitDeny`). The divergence is confined to the **allow-only**
shortcut — a usability/consistency gap, not a DENY bypass.

## Decision needed
Should a share-writable mount override a **positive-grant-only** DACL? And if so, which protocol
is canonical?
- **Option A — SMB semantics win:** drop the NFS `ShareWritable` bypass for files that carry an
  ACL; let the DACL decide. Risk: regresses smbtorture `stream-inherit-perms` / `create.multi`,
  which the bypass was added to keep passing (allow-only DACLs are additive over POSIX bits).
- **Option B — NFS semantics win:** make SMB `CheckFileAccess` honor `ShareWritable` for
  allow-only DACLs (no explicit DENY), matching NFS.
- **Option C — leave as-is, document:** accept the asymmetry as the SMB-vs-NFS share-write model
  difference; it is already pinned in the matrix as known.

Recommend deciding **B vs C** — the bypass exists precisely so allow-only ACLs stay additive over
POSIX, so removing it (A) is the riskier path against the smbtorture battery.

## Sites
- `pkg/metadata/auth_permissions.go:274` (NFS `checkFilePermissions` write bypass)
- `pkg/metadata/auth_permissions.go:545` (parallel bypass in the ACL-refined write path)
- SMB open-time `CheckFileAccess` (no `ShareWritable` consideration)
- Pinned/documented: `pkg/metadata/storetest/cross_protocol_permissions.go` `testCrossProtocolShareWritableBypass`

Found by: #1232 / PR #1238. Part of the #1231 hardening follow-ups.

marmos91:
## Summary

Aligns DittoFS's cross-protocol permission model with the standard multiprotocol NAS behaviour: the **share/export grant is a ceiling, not a floor** — it bounds access from above (ANDed with the file's own ACL/POSIX) and never expands it. A read-write share grants no write that the file's ACL or mode denies; only a read-only share restricts, via `ShareReadOnly`.

Reference implementations this matches:
- **Windows** — share permissions ∧ NTFS permissions, most-restrictive-wins.
- **NetApp ONTAP** — export `Access_Type` bounds the volume security style.
- **NFS-Ganesha** (userspace, FSAL — same shape as DittoFS) — export `Access_Type` over the FSAL `test_access`.
- **Samba** — a `write list` grant cannot exceed the underlying filesystem rights.

## The divergence this closes

A share-level write grant previously acted as a **floor** on the NFS evaluation path (`checkFilePermissions` / `CheckParentCreateAccess`): it granted write over an allow-only DACL regardless of the file's own permissions. The SMB open-time DACL gate did not honour it, so NFS and SMB disagreed on the same file + identity. Removing the floor converges both protocols on the file's ACL.

## Changes
- Remove the `ShareWritable` bypass at both sites; remove the now-unused `AuthContext.ShareWritable` field and its SMB session setters.
- Cross-protocol conformance matrix: assert NFS/SMB convergence for the explicit-DENY, allow-only-read (ceiling), and allow-only-write (additive) cases on memory/badger/postgres.
- Unit tests: pin that an explicit DENY and a read-only allow-only DACL both deny write on a read-write share, and that an ACL write grant still yields write.

## Validation
- Full `pkg/metadata` + SMB handler unit tests green.
- Local smbtorture `smb2.acls` (14/14), `smb2.stream-inherit-perms`, `smb2.create`: zero new failures vs develop.

Closes #1240

### DIFF
--- a/internal/adapter/smb/handlers/auth_helper.go
+++ b/internal/adapter/smb/handlers/auth_helper.go
@@ -86,8 +86,9 @@ func BuildAuthContext(ctx *SMBHandlerContext) (*metadata.AuthContext, error) {
 	// (audit #1132).
 	setUnprivilegedIdentity(authCtx)
 
-	// Set share-level permission flags for guest/anonymous sessions
-	authCtx.ShareWritable = HasWritePermission(ctx)
+	// Share-level read-only ceiling for guest/anonymous sessions. A read-write
+	// tree connect grants nothing on its own; the file's ACL/POSIX decides. The
+	// share permission can only restrict, via ShareReadOnly.
 	authCtx.ShareReadOnly = ctx.Permission == models.PermissionRead
 
 	return authCtx, nil
@@ -140,9 +141,9 @@ func getUserIdentity(user *models.User) (uid, gid uint32) {
 // Falls back to defaults (1000/1000) if not configured.
 //
 // Share Permission:
-// ShareWritable is set based on the SMB context's share permission.
-// This allows users with share-level write permission to bypass file-level
-// Unix permission checks.
+// ShareReadOnly is set from the SMB tree-connect permission and acts as a
+// ceiling — a read-only share denies all writes regardless of the file ACL. A
+// read-write share grants nothing on its own; the file's ACL/POSIX decides.
 func BuildAuthContextFromUser(ctx *SMBHandlerContext, user *models.User) *metadata.AuthContext {
 	authCtx := &metadata.AuthContext{
 		Context:                ctx.Context,
@@ -172,9 +173,9 @@ func BuildAuthContextFromUser(ctx *SMBHandlerContext, user *models.User) *metada
 		)
 	}
 
-	// Set share-level permission flags
-	// ShareWritable allows bypassing file-level permission checks for write operations
-	authCtx.ShareWritable = HasWritePermission(ctx)
+	// Share-level read-only ceiling. A read-write tree connect grants nothing on
+	// its own; the file's ACL/POSIX decides. The share permission only restricts,
+	// via ShareReadOnly.
 	authCtx.ShareReadOnly = ctx.Permission == models.PermissionRead
 
 	return authCtx
@@ -348,7 +349,7 @@ func (h *Handler) buildOpenerAuthContext(ctx *SMBHandlerContext, openFile *OpenF
 	// nobody/nogroup identity, matching the BuildAuthContext User==nil arm.
 	// Never UID=0 — see nobodyUID for the root-bypass rationale.
 	setUnprivilegedIdentity(authCtx)
-	authCtx.ShareWritable = HasWritePermission(ctx)
+	// Read-only ceiling only; a read-write share grants nothing on its own.
 	authCtx.ShareReadOnly = ctx.Permission == models.PermissionRead
 	return authCtx
 }
--- a/pkg/metadata/auth_identity.go
+++ b/pkg/metadata/auth_identity.go
@@ -32,17 +32,13 @@ type AuthContext struct {
 	ClientAddr string
 
 	// ShareReadOnly indicates whether the user has read-only access to the share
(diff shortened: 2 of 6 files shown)


marmos91:
Thanks — both comments correctly identify that `ctx.ShareReadOnly` (the per-user read-only signal) is not folded into the metadata write/ACL path: enforcement there comes from the store's `ShareOptions.ReadOnly`, and SMB additionally gates via `HasWritePermission(ctx)`. That gap is **pre-existing** (unchanged by this PR, which only removes the share-level write *floor*) and is now tracked separately in #1276, where it will be fixed with cross-protocol conformance coverage for the read-only-restriction direction. Keeping this PR scoped to the floor removal.

marmos91:
Holding this PR — found a real regression in live testing.

smbtorture `smb2.durable-open.read-only` fails after the floor removal: a client CREATEs a file with FILE_ATTRIBUTE_READONLY (granted full access at open, NULL DACL), then WRITEs through that handle and gets STATUS_ACCESS_DENIED. Windows/SMB semantics require the write to succeed — SMB is **handle-based**: a write through a handle granted FILE_WRITE_DATA at open is authorized regardless of the file's current readonly attribute/mode. DittoFS re-derives op-level write permission from the file mode (`checkFilePermissions`), and the DOS-readonly bit strips write.

The `ShareWritable` floor was DittoFS's approximation of SMB handle-based write authority, and it is **SMB-only** (the NFS adapters never set ShareWritable). So the NFS-vs-SMB divergence this PR set out to close was largely a synthetic artifact of the conformance matrix forcing ShareWritable=true on the NFS lanes — not a production behavior. The DACL ceiling is already enforced correctly at open (CheckFileAccess); the floor is the op-level handle-trust, with explicit DENY still pinned.

Reassessing the right fix (see next comment / linked decision).

marmos91:
Closing in favor of the correct approach. Rather than removing the share-level write floor (which broke SMB handle-based writes — smb2.durable-open.read-only), the right fix extends DittoFS's existing handle-based authorization model (the OpenFile.GrantedAccess gate already used for SET_INFO SecurityDescriptor, see auth_helper.go buildOpenerAuthContext) to WRITE/CREATE: SMB op-level writes authorize from the handle's GrantedAccess (granted at open by the DACL-honoring CheckFileAccess), not by re-deriving from the file's POSIX mode. NFS stays path-based. A new PR will implement this.

PR #1278, opened 2026-06-21
fix(perms): SMB handle-based write authorization (#1240)

### LINKED ISSUE #1240: ShareWritable bypass: NFS grants / SMB denies write on allow-only DACL (cross-protocol divergence)

marmos91:
## The model

SMB is **handle-based**: a handle granted `FILE_WRITE_DATA` / `FILE_APPEND_DATA` at open (by the DACL-honoring open-time gate `Service.CheckFileAccess`) must be able to write **regardless of the file's current POSIX mode or DOS-READONLY attribute**. The DACL ceiling — including an explicit DENY-write ACE — is enforced **at open**: a DENY strips write from the handle, so a write-authorized handle can never be minted on a DENY'd file.

The metadata WRITE/SET_INFO authorization previously approximated this with a coarse **share-level** `AuthContext.ShareWritable` floor in `checkFilePermissions` (and `CheckParentCreateAccess`). That floor was a SHARE-wide approximation of per-handle authorization and didn't reflect what the actual open granted.

## What changed

- **`pkg/metadata/auth_identity.go`**: removed `AuthContext.ShareWritable`; added `AuthContext.WriteAuthorizedByHandle bool` — "the SMB open handle was granted write access; the metadata layer must not re-deny on POSIX mode / DOS-READONLY, the DACL ceiling was enforced at open. Set only by SMB op handlers; NFS leaves it false (path-based)."
- **`pkg/metadata/auth_permissions.go`**: `checkFilePermissions` now grants the requested write/delete bits when `WriteAuthorizedByHandle && !ShareReadOnly && !acl.HasExplicitDeny(file.ACL)`. `ShareReadOnly` stays the ceiling; the explicit-DENY guard is kept as **defense-in-depth** (production can't reach it because the open already denies, but a mis-set flag can never override a DENY ACE). `CheckParentCreateAccess` drops the share-level floor and evaluates the parent DACL precisely (CREATE has no handle; the parent-create gate IS the open-time authorization, and `acl.Evaluate` expands `GENERIC_WRITE`/`GENERIC_ALL` so allow-only DACLs still grant — `create.multi` / `stream-inherit-perms` preserved).
- **SMB handlers** (`auth_helper.go`): the 3 sites that set `ShareWritable` now only set the `ShareReadOnly` ceiling. The new flag is set in the op handlers:
  - `write.go` (WRITE): `authCtx.WriteAuthorizedByHandle = hasWriteAccess(openFile.GrantedAccess)` after building the AuthContext passed to `PrepareWrite`.
  - `set_info.go` (SET_INFO size-changing): same, on the `FileEndOfFileInformation` case and the allocation-driven truncate inside `FileAllocationInformation`, on the AuthContext passed to `SetFileAttributes(Size:...)`.
- **NFS**: unchanged. NFS adapters never set the flag (verified), so NFS writes stay path-based.
- **Tests**: reworked `storetest` cross-protocol matrix to model SMB as handle-based (gate-1 DACL open grant ⇒ flag set for the gate-2 op check on DACL-bearing files; no-DACL files fall through to the same POSIX check NFS runs, keeping the lanes in agreement). Renamed the bypass case to `HandleWriteAuthorization` and pinned the DENY invariant + an allow-write-on-mode-0o000 case. Updated the metadata unit tests and added regression tests: write granted through a write-authorized handle on a DOS-READONLY / non-owner file (the `smb2.durable-open.read-only` motivation), DENY-write still denies, and `ShareReadOnly` suppresses the bypass.

## Why the old floor was coarse

`ShareWritable` was true for any read-write SMB session and bypassed file-level POSIX for every file with no explicit DENY — a share-wide grant, not the precise "this handle was opened with write" signal. The new flag is derived per-handle from `OpenFile.GrantedAccess`, so authorization tracks the actual open.

## Motivation: smb2.durable-open.read-only

That test CREATEs a file with `FILE_ATTRIBUTE_READONLY` (granted full access at open under a NULL DACL), then WRITEs through the handle. With the old code this relied on the coarse `ShareWritable` floor not denying; now the handle's `GrantedAccess` precisely authorizes the write while the file's DOS-READONLY bit is correctly ignored for the already-opened handle.

**NFS is unchanged** (path-based; flag stays false).

## Local verification

`go build ./...`, `go vet ./...`, `gofmt -l` clean. Passing: `./pkg/metadata/...` (incl. memory/badger/postgres conformance + cross-protocol matrix), `./internal/adapter/smb/...`, `./internal/adapter/nfs/...`, `./pkg/controlplane/...`.

Closes #1240

### DIFF
--- a/internal/adapter/smb/handlers/auth_helper.go
+++ b/internal/adapter/smb/handlers/auth_helper.go
@@ -86,8 +86,9 @@ func BuildAuthContext(ctx *SMBHandlerContext) (*metadata.AuthContext, error) {
 	// (audit #1132).
 	setUnprivilegedIdentity(authCtx)
 
-	// Set share-level permission flags for guest/anonymous sessions
-	authCtx.ShareWritable = HasWritePermission(ctx)
+	// Set the share-level read-only ceiling for guest/anonymous sessions.
+	// Per-handle write authorization (WriteAuthorizedByHandle) is set by the
+	// WRITE / SET_INFO op handlers from OpenFile.GrantedAccess, not here.
 	authCtx.ShareReadOnly = ctx.Permission == models.PermissionRead
 
 	return authCtx, nil
@@ -140,9 +141,10 @@ func getUserIdentity(user *models.User) (uid, gid uint32) {
 // Falls back to defaults (1000/1000) if not configured.
 //
 // Share Permission:
-// ShareWritable is set based on the SMB context's share permission.
-// This allows users with share-level write permission to bypass file-level
-// Unix permission checks.
+// ShareReadOnly is set from the SMB context's share permission (the read-only
+// ceiling). Per-handle write authorization (WriteAuthorizedByHandle) is NOT set
+// here — it is set by the WRITE / SET_INFO op handlers from the open handle's
+// GrantedAccess, since SMB write authorization is handle-based.
 func BuildAuthContextFromUser(ctx *SMBHandlerContext, user *models.User) *metadata.AuthContext {
 	authCtx := &metadata.AuthContext{
 		Context:                ctx.Context,
@@ -172,9 +174,9 @@ func BuildAuthContextFromUser(ctx *SMBHandlerContext, user *models.User) *metada
 		)
 	}
 
-	// Set share-level permission flags
-	// ShareWritable allows bypassing file-level permission checks for write operations
-	authCtx.ShareWritable = HasWritePermission(ctx)
+	// Set the share-level read-only ceiling. Per-handle write authorization
+	// (WriteAuthorizedByHandle) is set by the WRITE / SET_INFO op handlers from
+	// OpenFile.GrantedAccess, not here — SMB write authorization is handle-based.
 	authCtx.ShareReadOnly = ctx.Permission == models.PermissionRead
 
 	return authCtx
@@ -348,7 +350,10 @@ func (h *Handler) buildOpenerAuthContext(ctx *SMBHandlerContext, openFile *OpenF
 	// nobody/nogroup identity, matching the BuildAuthContext User==nil arm.
 	// Never UID=0 — see nobodyUID for the root-bypass rationale.
 	setUnprivilegedIdentity(authCtx)
-	authCtx.ShareWritable = HasWritePermission(ctx)
+	// Only the read-only ceiling is share-derived. Per-handle write
+	// authorization (WriteAuthorizedByHandle) is set by the WRITE / SET_INFO op
+	// handlers from OpenFile.GrantedAccess; SET_INFO SecurityDescriptor (the only
+	// current caller of this helper) is WRITE_DAC-gated, not a data write.
 	authCtx.ShareReadOnly = ctx.Permission == models.PermissionRead
 	return authCtx
 }
--- a/internal/adapter/smb/handlers/ioctl_copychunk.go
+++ b/internal/adapter/smb/handlers/ioctl_copychunk.go
@@ -392,6 +392,12 @@ func (h *Handler) executeCopyChunks(
 		logger.Warn("COPYCHUNK: failed to build auth context", "error", err)
 		return NewErrorResult(types.StatusAccessDenied), nil
 	}
+	// COPYCHUNK writes data into the destination handle. Like WRITE, it is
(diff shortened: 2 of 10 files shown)


marmos91:
Comment 2 (COPYCHUNK / FSCTL_SET_ZERO_DATA): valid catch — fixed in the follow-up commit. Both now set WriteAuthorizedByHandle from the destination handle's GrantedAccess (the SET_ZERO_DATA path gates on FILE_WRITE_DATA first; COPYCHUNK already validated dst write access). Without it they would have re-denied on POSIX/DOS-READONLY despite a write-granted handle, same as the WRITE path.

Comment 1 (gate the metadata branch on file.ACL != nil): intentionally NOT doing this — it would reintroduce the bug. The smb2.durable-open.read-only repro file is created with FILE_ATTRIBUTE_READONLY and carries NO explicit ACL (DOS-readonly via POSIX mode); its handle is granted write at open under the NULL DACL, so the handle-based grant MUST apply to no-ACL files for the write to succeed. Gating on file.ACL != nil would send that file back down the POSIX path and re-deny the write. The resulting SMB-vs-NFS difference on no-ACL files is the correct per-protocol model (SMB honors the write-granted handle; NFS, having no handle, honors POSIX) and is validated by the live smbtorture battery. The file.ACL != nil gate exists only in the cross-protocol *test lane*, where it keeps the synthetic matrix's all-lanes-agree invariant coherent — a test-modeling choice, not a production gate.

Issue #1274, opened 2026-06-21
CLOSE/COMMIT acks success on syncer.Flush Finalized=false (remote durability gap, distinct from #1267)


marmos91:
## Remote durability gap: CLOSE/COMMIT acks success when `syncer.Flush` returns `Finalized=false` (data not yet remote-durable)

Follow-up found while reviewing #1267 / PR #1273. **Distinct from #1267** (which fixes CLOSE swallowing a *non-nil* flush error). This is the **soft-fail** path: `engine.Flush` returns `nil` error but `Finalized=false`.

### The gap
`internal/adapter/common/write_payload.go` `CommitBlockStore` discards the `*block.FlushResult`:
```go
_, err := blockStore.Flush(ctx, string(payloadID))
return err
```
`syncer.Flush` returns `{Finalized: false}, nil` in several conditions, including:
- remote configured but **unhealthy** (`!IsRemoteHealthy()`),
- the periodic uploader holds the upload gate at that instant (`uploading` CAS fails),
- (local-only shares: `remoteStore == nil` — this case is **normal/expected**, not a failure).

In the remote-backed cases, CLOSE (and NFS COMMIT, same helper) returns SUCCESS to the client even though the bytes are only in local CAS, not confirmed remote-durable. A process crash / remote eviction before the background syncer re-drains loses those bytes — with the client having been told CLOSE succeeded.

### Why this is NOT a trivial "return error on !Finalized"
`Finalized=false` is the **normal transient state** under async mirroring (uploader-gate contention, healthy remote, bytes will mirror momentarily). Naively failing CLOSE whenever `!Finalized` would spuriously fail normal CLOSEs under sustained load. So this needs a **design decision first**:

- **What is SMB CLOSE's durability contract?** Local-CAS-durable (current de-facto behavior) or remote-durable? NFS COMMIT shares the same helper and question.
- If remote-durable is required on CLOSE, need to distinguish "transient, will-mirror" from "genuinely can't mirror (remote down)" — only the latter should fail; or CLOSE should *block* until finalized within a bounded deadline rather than error.
- Local-only shares (`remoteStore == nil`) must remain unaffected — `Finalized=false` is correct there.

### Suggested direction (for discussion)
- Have `CommitBlockStore` inspect `result.Finalized` and the share's remote-backed-ness.
[...]

Issue #1267, opened 2026-06-21
Silent write truncation: SMB CLOSE acks success despite swallowed append-log flush ErrPressureTimeout


marmos91:
## Silent write truncation: SMB CLOSE acks `StatusSuccess` despite a swallowed block-flush `ErrPressureTimeout`

Split out of #1245 (the dedup race, now fixed on `develop`). This is a **separate data-integrity bug** on the append-log backpressure path — independent of dedup.

### Reported by an external user (dittofs 0.19.0, 8 GB host, remote S3 share)
Bulk-loading a ~25 GB file as 51 × 500 MiB blocks over loopback CIFS, several blocks the client reported as fully written were **short on disk**, while `write()`/`close()` returned **success**:

| block | expected | persisted | short by |
|---|---|---|---|
| part-0004 | 524288000 | 466616320 | ~55 MB |
| part-0027 | 524288000 | 523239424 | ~1 MB |

Total archive **186 MB short** across 6 blocks on first pass; still 58 MB short after a naive re-upload — every write ack'd as complete. A `stat` on the **writing** mount shows full size (CIFS client write cache masks it); the short size only appears on a fresh `cache=none` mount. Re-writing the bad blocks after `systemctl restart dfs` (un-pressured append log) + MD5 verify through `cache=none` produced a byte-exact archive.

### Root cause (confirmed in code)
`internal/adapter/smb/handlers/close.go:198-207` — the block-store flush on CLOSE logs its error and **continues**:
```go
} else if flushErr := common.CommitBlockStore(ctx.Context, blockStore, openFile.PayloadID); flushErr != nil {
    logger.Warn("CLOSE: flush failed", "path", openFile.Path, "error", flushErr) // <- swallowed
}
```
`CommitBlockStore` → `engine.Flush` can return `ErrPressureTimeout` (`pkg/block/local/fs/errors.go:40-53`), whose own doc says it **should** be mapped to `STATUS_INTERNAL_ERROR`. Instead CLOSE proceeds to return `StatusSuccess`. The comment two lines up even asserts *"SMB CLOSE requires immediate durability."*

Secondary swallow: `internal/adapter/smb/handlers/write.go:435-437` ignores `FlushPendingWriteForFile` errors at DEBUG (metadata, not byte data — lower severity, but still wrong).

### Expected
[...]

Issue #1245, opened 2026-06-20
Dedup race: rollup "Transaction Conflict" retry-loop never converges; SMB CLOSE wedges on bulk/large writes


raffaelegiustini-create:
## Summary

Under bulk writes over SMB (and on `close()` of large files), dittofs hits a file-level dedup race. It surfaces two ways:

- **Client side (SMB):** `CLOSE` fails with an I/O error (`increment refcount … no FileBlock`), which either fails the write or **wedges the whole SMB session** (requires `umount -f -l` + `systemctl restart dfs`).
- **Server side (async rollup to S3/Cubbit):** the same conflict appears as `rollup: ObjectIDPersister: Transaction Conflict. Please retry`, retried on the **same payloadID indefinitely with no backoff and no convergence** (151 occurrences in 24h).

Two secondary issues showed up while reproducing: rollup is **not crash-safe** across restarts (`context canceled` → process exits `status=1/FAILURE`), and the **badger block cache is undersized** (hit-ratio ~26%, 2.1 GB peak on a 4 GB host).

## Environment

- dittofs **0.19.0** (pre-1.0)
- Server: Hetzner CX22, 4 GB RAM (`ubuntu-4gb-nbg1-1`)
- Backend: Cubbit DS3 (S3, `s3.cubbit.eu`)
- Client: macOS over SMB through a WireGuard tunnel (RTT ~38 ms)
- Workload: bulk archive of ~38 GB from `~/Downloads` into monthly folders

## Steps to reproduce

1. Mount the dittofs SMB share on macOS.
2. Bulk-copy a folder with hundreds of files, **or** write a single large file (reproduced reliably with a ~3 GB `.tar`).
3. Observe the `CLOSE`/flush failure on the client and the rollup retry-loop in `journalctl -u dfs`.

### Client-side (SMB) error on CLOSE

```
CLOSE: flush failed: file-level dedup: increment refcount on target hash X:
coordinator: no FileBlock with hash X: file block not found
→ errno 5 (I/O error) returned to the SMB client
```

### Server-side rollup retry-loop (does not converge)

```
18:22:37 ERROR rollupFile failed payloadID=drive/eef502d1-... error="rollup: ObjectIDPersister: Transaction Conflict. Please retry" source=ticker
18:22:43 ERROR rollupFile failed payloadID=drive/eef502d1-... error="rollup: ObjectIDPersister: Transaction Conflict. Please retry"
[...]

PR #1308, opened 2026-06-22
feat(adapter): log COMMIT/CLOSE durability decision at DEBUG (#1245)

marmos91:
## What

Adds one `DebugCtx` line at the `CommitBlockStore` ack point (the seam shared by NFSv3 COMMIT, NFSv4 COMMIT, SMB CLOSE) recording the exact per-store durability inputs:

```
finalized, localDurable, remoteDurable, requireDurableCommit, durableAtAck
```

## Why

Follow-up to #1245. Reporter confirmed the silent-truncation fix (#1267 + #1274) round-trips **8/8 byte-perfect** on `develop`, but asked *"which log level surfaces the engine.Flush Finalized/err decision?"* so a deliberately-saturated-CLOSE trace can prove there is no residual silent window.

Answer today: **no line emits `res.Finalized`** — on hard error it's `WARN "CLOSE: flush failed"`, on success `DEBUG "CLOSE: flushed"` (which drops Finalized). This patch fills that gap.

## Behaviour

- **No ack-path change** — pure observability. The decision logic (`localDurable || (finalized && remoteDurable)`, gated on `require_durable_commit`) is unchanged; the booleans are just hoisted into locals and logged.
- DEBUG level only (per CLAUDE.md invariant #6: expected outcomes log at Debug), so the durable hot path stays quiet by default.
- Run with `DITTOFS_LOGGING_LEVEL=DEBUG` to capture the decision under a saturated CLOSE.

## Test

`go build` / `go vet` / `go test ./internal/adapter/common/` green; existing `commit_durability_test.go` covers the decision logic (unchanged).

### DIFF
--- a/internal/adapter/common/write_payload.go
+++ b/internal/adapter/common/write_payload.go
@@ -120,19 +120,24 @@ func CommitBlockStore(
 	// means the bytes already survive a restart regardless of Finalized. This
 	// single DEBUG line surfaces the exact (Finalized, local/remote durable)
 	// inputs so a saturated-CLOSE trace can confirm there is no silent window
-	// without adding a log line to the common (durable) hot path's behaviour.
+	// without adding a log line to the common (durable) hot path's behavior.
+	// The booleans below feed the decision logic regardless of log level, so
+	// only the DebugCtx call itself (the string conversion + variadic boxing)
+	// is guarded to keep the ack path allocation-free when DEBUG is disabled.
 	finalized := res != nil && res.Finalized
 	localDurable := blockStore.LocalDurable()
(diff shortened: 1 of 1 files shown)


marmos91:
Addressed both Copilot notes in b5fdd1bb:
- Guarded the `DebugCtx` call with `logger.IsDebugEnabled()` so the `string(payloadID)` conversion + variadic boxing are skipped on the ack path when DEBUG is off (matches the `state_debug.go` idiom). The four durability booleans stay unguarded since the decision logic below uses them regardless of log level.
- `behaviour` → `behavior` to match the file.

Issue #1314, opened 2026-06-22
SMB NTLM cannot authenticate AD domain users (no NETLOGON passthrough) — local users only


marmos91:
During AD/LDAP live acceptance, mounting an SMB share as a domain user over NTLM fails:
```
net use S: \SERVER\demo /user:DITTOFS\alice TestPassword01!
System error 86: The specified network password is not correct.
```

## Cause
`completeNTLMAuth` (`internal/adapter/smb/handlers/session_setup.go:1094-1103`) resolves the username against the **local dfs user store** (`userStore.GetUser`) and validates the NTLMv2 response against a **locally configured NT hash**. AD domain users have no local account and no local NT hash, so the lookup fails → `STATUS_LOGON_FAILURE`.

Validating a domain users NTLM response requires the NT hash held by the DC — i.e. a **NETLOGON secure channel (NetrLogonSamLogon) passthrough**, which is not implemented (it is part of the deferred "online net ads join" work).

## Impact / expectation mismatch
The AD umbrella (#1231) describes "NTLM fallback for legacy clients". For **AD domain** users that is currently not true — NTLM only works for **local** dfs users. AD domain authentication works exclusively via **Kerberos/SPNEGO** (the keytab path), where the DC issues the ticket and dfs validates it with its service keytab (no passthrough needed).

## Asks
1. **Docs:** state clearly that AD domain-user auth is Kerberos-only; NTLM is for local dfs users. (LDAP idmap resolves *identity*; it does not authenticate NTLM.)
2. **Feature (tracking):** implement NETLOGON passthrough so legacy NTLM-only clients can authenticate AD domain users. Larger effort; tied to the deferred online-join machinery.

Found alongside #1312 on the demo cluster (v0.20.1).

marmos91:
## Investigation: how to implement AD NTLM passthrough

### Why it cannot be fixed locally
NTLMv2 SESSION_SETUP gives the server only the *challenge/response* (`NtChallengeResponse` = HMAC-MD5 over the server challenge keyed by `NTOWFv2 = HMAC-MD5(NT-hash, upper(user)+domain)`). Verifying it requires the user's **NT hash**, which lives only on the DC. There is no LDAP/Kerberos shortcut: an LDAP simple-bind needs the cleartext password (we don't have it), and Kerberos is a different protocol. So a domain user's NTLM logon **must** be delegated to a DC. This is exactly what Samba `winbindd` does and what every AD-member NAS does.

[...]

Issue #1319, opened 2026-06-22
Operator: config-only spec changes update the ConfigMap but don't roll the pod (need checksum annotation)


marmos91:
## Summary

When a change to `DittoServer.spec` affects **only the rendered ConfigMap data** (and not the pod template), the operator updates the ConfigMap but does **not** restart the server pod, so the new configuration silently does not take effect until the next unrelated restart. dfs reads its config file once at startup.

## Reproduction

1. `kubectl patch dittoserver demo --type merge -p '{"spec":{"logging":{"level":"DEBUG"}}}'`
2. Operator reconciles: `demo-config` ConfigMap now shows `level: DEBUG`, but the operator logs `Container ports unchanged, skipping StatefulSet update` and the pod is **not** rolled.
3. `kubectl logs demo-0` still emits at INFO. DEBUG only applies after a manual `kubectl delete pod demo-0`.

This affects every config-only field, including the identity-critical ones: `spec.identity.ldap` (bind DN/URL/idmap), `spec.identity.kerberos` (realm/SPN/netbios), and `spec.logging`. Changing kerberos/ldap *that also adds a Secret volume* happens to roll the pod (the pod template changed); a pure data edit does not — an inconsistent, surprising UX.

## Fix

Stamp a config checksum annotation on the pod template (`checksum/config: <sha256 of rendered config>`), the standard operator/Helm pattern. When the rendered config changes, the annotation changes, the StatefulSet pod template changes, and the pod rolls automatically. Optionally include the keytab/krb5.conf Secret resourceVersions so a keytab rotation also triggers a restart (the keytab hot-reloads, but krb5.conf / config do not).

## Acceptance

- A config-only `spec` change (e.g. `logging.level`) triggers an automatic StatefulSet rollout and the new value is in effect without manual pod deletion.

marmos91:
Addressed by #1326 (merged to develop, commit 461e1439). The proposed `checksum/config` annotation already existed as `dittofs.io/config-hash` (hashes generation+config+secrets, so any spec edit rolls the pod). The real gap — `spec.logging` and `spec.identity.kerberos` not modeled in the CRD — is now closed: both added to the CRD + render, with kerberos keytab/krb5.conf mounted as Secrets (rotation rolls the pod).

Issue #1318, opened 2026-06-22
Service keytab is RC4-only — Windows clients (RC4 disabled) cannot Kerberos-auth; need AES enctypes


marmos91:
## Summary

The dev/test Samba AD-DC fixture exports a service keytab that contains **only `arcfour-hmac` (RC4)** keys for the DittoFS SPNs. RC4 works for the in-cluster Linux test client, but **Windows 11 / Server 2022 disable RC4 for Kerberos by default**, so a real Windows client requesting a service ticket for `cifs/<dfs>@REALM` will either be issued an RC4 ticket it refuses to use, or fail outright — blocking Windows Kerberos acceptance (#1236/#1297) independently of #1317.

dfs itself is fine with AES: the in-cluster test showed `keyType=18` (aes256-cts) as the **session** key and dfs derived SMB3 keys from it correctly. The problem is purely the **ticket** key enctype: the KDC encrypts the service ticket with the service account's strongest stored key, and dfs must hold that key in its keytab to decrypt it. With an RC4-only keytab, an AES ticket from a Windows client cannot be decrypted.

## Evidence

```
$ klist -ke /keytabs/dittofs.keytab
   1 cifs/dittofs.dittofs.ad@DITTOFS.AD (DEPRECATED:arcfour-hmac)
   1 nfs/dittofs.dittofs.ad@DITTOFS.AD  (DEPRECATED:arcfour-hmac)

# Service ticket issued to the in-cluster client:
#   Etype (skey, tkt): aes256-cts-hmac-sha1-96, DEPRECATED:arcfour-hmac
#   -> session key AES256, but TICKET key RC4 (only enctype on the SPN account)
```

## Root cause

`test/integration/ad-dc/entrypoint.sh` registers both SPNs on the **Administrator** account and `samba-tool domain exportkeytab --principal=<SPN>` exports only the RC4 key, because the account's `msDS-SupportedEncryptionTypes` isn't set to include AES (and no AES keys were generated for the SPN salt).

## Remediation (fixture + docs)

- Fixture: create a dedicated service account, set `msDS-SupportedEncryptionTypes` to include AES (e.g. `0x1C` = AES128+AES256+RC4, or `0x18` AES-only), set/reset its password so AES keys are generated for the correct salt, register the SPNs on it, then `exportkeytab`. Verify `klist -ke` shows `aes256-cts-hmac-sha1-96` + `aes128-cts-hmac-sha1-96` entries.
[...]

PR #1320, opened 2026-06-22
feat(identity): control-plane API to configure LDAP/Kerberos providers (#1311)

### LINKED ISSUE #1311: Add a control-plane API to configure LDAP/AD (and Kerberos/keytab) identity providers
_Affects: v0.20.0 · Suggested labels: enhancement, control-plane, identity, api_

## Summary

DittoFS v0.20.0 ships a fully-featured LDAP/AD identity provider
(`pkg/identity/ldap`) and Kerberos provider (`pkg/identity/kerberos` +
`config.KerberosConfig`), but their configuration is **file/env only** —
`ldap.*` / `kerberos.*` YAML keys and `DITTOFS_LDAP_*` / `DITTOFS_KERBEROS_*`
env vars, read once at process start (`cmd/dfs/commands/start.go:268`,
`rt.SetLDAPConfig`). **There is no control-plane route to read, write, or test
identity-provider configuration.**

This blocks the DittoFS Pro dashboard, whose entire purpose is to let
administrators manage stores, shares, adapters, users **and identity providers**
visually, without editing config files or restarting the server. Today an admin
must hand-edit `dittofs.yaml` (or set env vars) and restart to point DittoFS at a
directory — something every other subsystem (shares, stores, adapters, settings)
can already do over `/api/v1/*`.

## Current state (evidence)

- `pkg/controlplane/api/router.go` registers no `/identity-provider(s)`,
  `/ldap`, `/ad`, or `/idp` route (read or write). The only identity-adjacent
  routes are `/api/v1/identity-mappings` (principal→user links) and
  `/api/v1/sid-mappings` (SID→UID/GID allocations) — neither configures the
  provider.
- `/api/v1/settings` (`pkg/controlplane/models/setting.go`) is a flat
  `{key, value}` string row and cannot represent the nested LDAP `Config` /
  `TLSConfig` struct.
- Health/readiness (`internal/controlplane/api/handlers/health.go`) reports only
  DB + NFS grace; it does not surface identity-provider status.
- The `ldap.Config` struct already has `MarshalJSON`/`MarshalYAML` that redact
  `BindPassword` (`pkg/identity/ldap/config.go:146`), so it is already
  display-safe — it simply has no route.

## Proposed API

Type-parameterized so LDAP/AD and Kerberos share one contract
(`{type}` ∈ `ldap` | `kerberos`):

| Method & path | Purpose |
|---|---|
| `GET /api/v1/identity-providers` | List configured providers + enabled state (no secrets). |
| `GET /api/v1/identity-providers/{type}/config` | Read current config (bind password redacted). |
| `PUT /api/v1/identity-providers/{type}/config` | Create/replace config; validate via existing `Config.Validate()`. |
| `POST /api/v1/identity-providers/{type}/test` | Dry-run: dial + bind (+ optional sample lookup), return reachability without persisting. |

All admin-only (`RequireAdmin`), consistent with existing `/api/v1` admin routes.
Applying a new config should hot-reload the provider in the runtime (re-invoke the
equivalent of `SetLDAPConfig` + rebuild the identity resolver) so no restart is
required — that is the entire value over the current file/env path.

### Payload schema — `ldap`

Derived directly from `pkg/identity/ldap/config.go` (`Config` + `TLSConfig`).
`bind_password` is **write-only**: accepted on `PUT`, never returned on `GET`
(returned as `"********"` if set, `""` if unset, matching the existing marshaler).

```json
{
  "enabled": true,
  "url": "ldaps://dc.example.com:636",
  "start_tls": false,
  "allow_plaintext": false,
  "base_dn": "DC=example,DC=com",
  "bind_dn": "CN=svc-dittofs,OU=Service,DC=example,DC=com",
  "bind_password": "write-only-on-PUT",
  "user_attr": "sAMAccountName",
  "realm": "EXAMPLE.COM",
  "idmap": "rfc2307",
  "nested_groups": true,
  "timeout": "10s",
  "max_group_results": 200,
  "tls": {
    "ca_cert_file": "/etc/dittofs/ca.pem",
    "client_cert_file": "",
    "client_key_file": "",
    "insecure_skip_verify": false,
    "min_version": "1.2"
  }
}
```

Field defaults and constraints (already enforced by `Config.Validate()` /
`ApplyDefaults()` — reuse them server-side):

- `idmap` ∈ {`rfc2307` (default), `rid`}
- `user_attr` default `sAMAccountName`, must match `^[A-Za-z][A-Za-z0-9-]*$`
- `tls.min_version` ∈ {`1.2` (default), `1.3`}
- `plaintext ldap://` rejected unless `start_tls` or `allow_plaintext`
- when `enabled`: `url`, `base_dn`, `bind_dn`, `bind_password` required
- `tls.client_cert_file` and `tls.client_key_file` must be set together
- `timeout` default `10s`, `max_group_results` default `200` (both ≥ 0)

### Payload schema — `kerberos`

Derived from `config.KerberosConfig` (`pkg/config/config.go:326`). `keytab_path`
points at a server-side file; consider an optional upload/write-only keytab body
in a follow-up.

```json
{
  "enabled": true,
  "keytab_path": "/etc/dittofs/dittofs.keytab",
  "service_principal": "nfs/server.example.com@EXAMPLE.COM",
  "realm": "EXAMPLE.COM",
  "netbios_domain": "EXAMPLE",
  "dns_domain": "example.com",
  "krb5_conf": "/etc/krb5.conf",
  "max_clock_skew": "5m",
  "context_ttl": "8h",
  "max_contexts": 10000
}
```

### `POST /{type}/test` response (suggested)

```json
{ "ok": true, "dialed": true, "bound": true, "message": "bind succeeded", "latency_ms": 42 }
```

On failure: `{ "ok": false, "stage": "bind", "message": "<sanitized error>" }`
(never echo the bind password).

## Acceptance criteria

- [ ] `GET /api/v1/identity-providers` lists `ldap` and `kerberos` with their `enabled` state, no secrets.
- [ ] `GET /api/v1/identity-providers/ldap/config` returns the full config with `bind_password` redacted to `"********"`.
- [ ] `PUT /api/v1/identity-providers/ldap/config` validates via the existing `Config.Validate()` and persists; an invalid plaintext/missing-field config is rejected with a 400 and a field-specific message.
- [ ] Applying a config hot-reloads the runtime provider (no process restart required) and a subsequent `GET` reflects the new config.
- [ ] `POST /api/v1/identity-providers/ldap/test` performs dial+bind without persisting and reports reachability; failures never leak the bind password.
- [ ] Equivalent endpoints for `kerberos` backed by `KerberosConfig`.
- [ ] All routes are admin-only (`RequireAdmin`) and JWT-authenticated, consistent with existing `/api/v1` admin routes.
- [ ] `bind_password` (and any Kerberos secret material) is write-only and never serialized back in any response or log.
- [ ] `pkg/apiclient` gains typed methods for the new endpoints (mirrors the existing adapter/settings client surface).

## Why now

DittoFS Pro v0.20.0 includes an identity-provider configuration UI workstream
that is currently **blocked/deferred** because there is no API to build against.
This is the last major config surface still requiring file edits + a restart;
adding it completes parity with shares/stores/adapters/settings management.

### LINKED ISSUE #1317: SMB Kerberos rejects AD/LDAP users without a local account (identity discarded after successful ticket auth)
## Summary

SMB Kerberos authentication validates an AD client ticket correctly (AP-REQ decrypts, PAC parses, mutual auth completes) but then **rejects the session for any AD/LDAP user that has no local control-plane account**. The handler resolves the principal to a full identity via the centralized `identity.Resolver` (which correctly falls through to the LDAP/AD provider and returns the RFC2307 UID/GID + group SIDs), then **throws that identity away** and requires `userStore.GetUser(username)` — a local DB lookup — to succeed.

This is the Kerberos sibling of #1314 (NTLM, local-users-only). Kerberos crypto/PAC was expected to work and **does**; the failure is purely in the identity-binding step after authentication.

## Evidence (live demo, in-cluster Samba AD-DC + LDAP idmap rfc2307)

Client gets a TGT and a service ticket for the dfs SMB SPN:
```
$ kinit alice@DITTOFS.AD          # ok
$ smbclient //dittofs.dittofs.ad/demo --use-kerberos=required -c ls
session setup failed: NT_STATUS_LOGON_FAILURE
$ klist | grep cifs
  cifs/dittofs.dittofs.ad@DITTOFS.AD   (Etype skey: aes256-cts-hmac-sha1-96, tkt: arcfour-hmac)
```

dfs server log — the AP-REQ verified (we reach the post-auth user lookup), then:
```
INFO  Kerberos auth: user lookup failed  username=alice principal=alice found=false error="user not found"
INFO  Kerberos authentication failed, returning SPNEGO reject for NTLM fallback
```

So: ticket decrypt OK, PAC principal extracted, `identity.Resolver` resolves `alice@DITTOFS.AD` to UID 10001 via LDAP — but `userStore.GetUser("alice")` (local control-plane store) has no such row, so the session is rejected.

## Root cause

`internal/adapter/smb/handlers/kerberos_auth.go`:

- `handleKerberosAuth` (≈L80–103): `username := h.resolveKerberosPrincipal(...)` → `userStore.GetUser(ctx, username)`. On miss → `STATUS_LOGON_FAILURE`.
- `completeKerberosBind` (≈L362–374): same pattern.

`resolveKerberosPrincipal` (L507) already calls `h.IdentityResolver.Resolve(...)` and gets a full `*identity.ResolvedIdentity{Username,UID,GID,GIDs,SID,GroupSIDs,Found}` — but returns only `.Username`, discarding UID/GID/SIDs. The handler then re-requires a local account.

Contrast NFS, which does this correctly: `internal/adapter/nfs/rpc/gss/framework.go:783 resolveIdentity` builds `metadata.Identity` straight from `resolved.UID/GID/GIDs` with **no** local user-store lookup. NFS krb5 AD users work; SMB krb5 AD users do not.

## Proposed fix (shared with #1314)

When the local `userStore.GetUser` misses **but** the resolver returned `Found=true`, build the SMB session from the resolved identity instead of failing:

- Have `resolveKerberosPrincipal` return the `*identity.ResolvedIdentity` (not just the username).
- Synthesize a transient (non-persisted) `*models.User` from it — `Username`, `UID`, `GID`, `GIDs`, `SID` — for `CreateSessionWithUserAndExpiry` / the bind identity check. PAC `GroupSIDs` are already captured onto the session (`SetPACIdentity`).
- Keep the local-account path as-is (local users still win / are tried first), and keep the hard-fail when the resolver also returns `Found=false` (unknown principal ≠ guest).

This makes the SMB session layer accept a resolver-provided identity exactly like the NFS GSS path, unifying AD identity across both protocols.

## Scope / acceptance

- AD domain user (no local dfs account) can `smbclient -k` against a dfs share and operate as their RFC2307 UID/GID.
- Same user over NFSv4.1 `sec=krb5` resolves to the **same** UID/GID (cross-protocol unification — #1297).
- Local (non-AD) users unaffected.
- Add handler-level tests with a fake resolver returning Found=true + no local user.

Refs #1314 (NTLM, same root cause), #1236 (keytab Kerberos E2E), #1297 (cross-protocol identity), #1231 (AD/LDAP umbrella).

marmos91:
## Summary

Adds an admin-only control-plane API (and `dfsctl` commands) to configure **LDAP/AD and Kerberos identity providers** without editing config files or restarting — completing config-management parity with shares/stores/adapters/settings. Also fixes two AD/LDAP auth gaps found alongside it.

Closes #1311. Fixes #1317. Addresses the docs ask in #1314 (the NETLOGON-passthrough feature stays tracked there).

## #1311 — identity-provider configuration API

- Routes (admin-only, `{type}` ∈ `ldap` | `kerberos`):
  - `GET /api/v1/identity-providers` — list + enabled state (no secrets)
  - `GET /api/v1/identity-providers/{type}/config` — read (bind password redacted to `********`)
  - `PUT /api/v1/identity-providers/{type}/config` — create/replace (validated)
  - `POST /api/v1/identity-providers/{type}/test` — dry-run dial+bind (LDAP) / keytab check (Kerberos); never persists
- **Persistence**: new `IdentityProviderConfig` GORM model + store. The `ldap.*`/`kerberos.*` file/env config **seeds** the DB on first boot; a stored row then **takes precedence**.
- **Hot-reload**: LDAP changes rebuild and re-inject the live identity resolver in the NFS/SMB adapters (new runtime change notification; SMB resolver swap made race-safe via `atomic.Pointer`). Kerberos changes apply on the next restart (adapters bind it at startup).
- **Secrets**: bind password is write-only — redacted on read, preserved when the placeholder is resubmitted, stored unredacted for use after restart.
- `pkg/apiclient` typed methods + `dfsctl identity-provider {list,get,set,test}` + regenerated `docs/CLI.md` + `docs/CONFIGURATION.md` note.

## #1317 — SMB Kerberos accepts directory-resolved AD users

After a ticket validates, if no local control-plane account exists **but** the identity resolver matched the principal (LDAP/AD), the session is now built from the resolved identity instead of failing with `STATUS_LOGON_FAILURE`. This mirrors the NFS GSS path, so the same AD user maps to the same UID/GID across protocols. Disabled local accounts and unresolved principals still fail closed.

## #1314 — docs

Documented in `docs/SMB.md`: SMB NTLM authenticates **local** DittoFS users only; AD domain users must use **Kerberos** (NETLOGON passthrough is not implemented — tracked in #1314). LDAP idmap resolves identity, not NTLM auth.

## Notes

- `pkg/config` imports the control-plane API package, so the handler cannot import `pkg/config`/`pkg/auth/kerberos` (import cycle). The handler therefore owns a plain Kerberos DTO + validates inline + tests the keytab via gokrb5 directly; `cmd/dfs` maps the persisted DTO onto `config.KerberosConfig` at startup.
- Secrets are stored plaintext in the control-plane DB, matching the existing posture for S3/block-store credentials. At-rest encryption can be a follow-up.

## Tests

- LDAP `TestConnection` + storage codec (password preserved vs. redacting marshaler)
- Store round-trip (upsert, secret survives, list, delete)
- Handler: validate→400, GET redaction, password preservation on resubmit, 404 unconfigured, list, Kerberos validation, test-does-not-persist
- `#1317` regression: `resolveSessionUser` (local wins / directory-resolved / disabled fails closed / unresolved fails closed)

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -2,6 +2,7 @@ package commands
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"fmt"
 	"os"
@@ -10,6 +11,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/marmos91/dittofs/internal/controlplane/api/handlers"
 	"github.com/marmos91/dittofs/internal/logger"
 	"github.com/marmos91/dittofs/internal/sysinfo"
 	"github.com/marmos91/dittofs/pkg/adapter/nfs"
@@ -22,6 +24,7 @@ import (
 	"github.com/marmos91/dittofs/pkg/controlplane/runtime"
 	"github.com/marmos91/dittofs/pkg/controlplane/runtime/shares"
 	"github.com/marmos91/dittofs/pkg/controlplane/store"
+	"github.com/marmos91/dittofs/pkg/identity/ldap"
 	badgerstore "github.com/marmos91/dittofs/pkg/metadata/store/badger"
 	"github.com/marmos91/dittofs/pkg/metrics"
 	"github.com/spf13/cobra"
@@ -261,17 +264,13 @@ func runStart(cmd *cobra.Command, args []string) error {
 	// machine SID is resolved before any adapter reads the SID mapper. Pinning
 	// the same SID on every node yields identical local UID->SID encoding.
 	rt.SetPinnedMachineSID(cfg.Identity.MachineSID)
-	rt.SetAdapterFactory(createAdapterFactory(&cfg.Kerberos))
 
-	// Thread the LDAP/AD identity provider config into the runtime so the
-	// adapters' BuildIdentityResolver can register the provider when enabled.
-	if cfg.LDAP.Enabled {
-		ldapCfg := cfg.LDAP
-		rt.SetLDAPConfig(&ldapCfg)
-		logger.Info("LDAP identity provider enabled",
-			"url", cfg.LDAP.URL, "base_dn", cfg.LDAP.BaseDN, "idmap", cfg.LDAP.Idmap,
-			"nested_groups", cfg.LDAP.NestedGroups)
-	}
+	// Resolve identity-provider config: a DB row (managed via the API) wins over
+	// the file/env config; on first boot the file/env config seeds the DB. This
+	// sets the runtime's LDAP config and returns the effective Kerberos config to
+	// hand to the adapter factory.
+	effectiveKerberos := resolveIdentityProviders(ctx, cpStore, rt, cfg)
+	rt.SetAdapterFactory(createAdapterFactory(&effectiveKerberos))
 
 	// Create and set API server
 	apiServer, err := api.NewServer(cfg.ControlPlane, rt, cpStore, cfg.Snapshot.RestoreHTTPTimeout)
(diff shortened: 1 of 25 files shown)


marmos91:
## Code review — AD/LDAP correctness

Reviewed every load-bearing path (`resolveSessionUser`, the resolver hot-reload wiring on both adapters, the `start.go` seeding/precedence, and the `GetUser` not-found contract). The #1317 fix is sound and merge-ready. Notes below — none are blockers.

### ✅ Core #1317 fix is correct

`resolveSessionUser` decision table is right and fails closed in every dangerous case:
- local enabled account wins;
- **only** `errors.Is(localErr, ErrUserNotFound)` + `resolved.Found` synthesizes a directory user;
- disabled local account → fail closed (directory does **not** bypass an explicit block);
- transient store error (DB down, not `ErrUserNotFound`) → fail closed;
- unresolved principal → fail closed (a valid ticket from an unknown principal is not a guest).

The load-bearing condition checks out: `GORMStore.GetUser` returns `models.ErrUserNotFound` via `getByField → convertNotFoundError`, and the control-plane user store is always GORM in production, so the `errors.Is` branch genuinely fires on a real miss — the directory fallback is not silently dead. This now mirrors the NFS GSS path (`framework.go` `resolveIdentity`), which already built identity straight from `resolved.UID/GID` with no local-account gate. Cross-protocol UID/GID symmetry achieved.

### ✅ Hot-reload is race-safe on both protocols

- SMB: `Handler.IdentityResolver` moved field → `atomic.Pointer` with `SetIdentityResolver`/`IdentityResolver()`. Session-setup reads concurrent with an API-driven resolver swap are safe.
- NFS: `GSSProcessor.resolveIdentity` reads the resolver under `p.mu.RLock`; `SetResolver` takes the write lock. Safe.
- Ordering is correct: `putLDAP` calls `SetLDAPConfig` **before** `NotifyIdentityProviderConfigChange`, and the rebuild reads a fresh `rt.LDAPConfig()` — no stale read.
- `resolverMu` serializes concurrent PUTs; both adapters unsubscribe on `Stop`, so callbacks don't leak across restart.
- `start.go` precedence (DB row wins; file/env seeds on first boot) is correct, and the DB-sourced Kerberos config re-applies the `MustLoad` defaults (clock skew / context TTL / max contexts) that a raw DB decode would otherwise miss.

### Notes (non-blocking)

1. **SMB synth drops `resolved.GIDs`.** `synthUserFromResolved` carries only the primary GID (plus SID + GroupSIDs); the NFS path carries the full supplementary/nested `GIDs`. This is **not a regression** — SMB authorization keys on group SIDs (Windows ACL), not POSIX supplementary GIDs, and `BuildAuthContextFromUser` never populated `Identity.GIDs` even for local users. Under Kerberos the PAC supplies the transitive group-SID set anyway. The only resulting asymmetry: a POSIX-mode file group-owned by one of the user's *nested* AD groups is granted over NFS (via GIDs) but over SMB only if a matching group-SID ACE is present. Worth a one-line doc note, not a code change.

2. **`Domain` is dropped on the SMB synth user** (the NFS identity carries `resolved.Domain`). Minor; the `sessionDomain` fallback covers it.

3. Kerberos config changes are restart-required (not hot-reloaded) — correctly documented via the `X-DittoFS-Apply: restart-required` header and the handler comment.

### Scope

Definitive at the code level for authentication + cross-protocol UID/GID unification. Coverage is unit-level (`resolveSessionUser` table) — there's no live AD integration test in the PR, and an end-to-end in-cluster proof additionally needs operator keytab-mount support plus the AES-keytab fix (#1318, the exported keytab is currently RC4-only, which Windows 11 rejects) and the operator config-rollout fix (#1319). Those gate the full demo flow, not this dfs-side fix.

marmos91:
## Review responses

**Copilot (4 inline, all fixed):**
1. `resolveSessionUser` — directory fallback now fires **only** on `errors.Is(localErr, models.ErrUserNotFound)`. Transient store errors and disabled local accounts fail closed.
2. `testLDAP` — now runs `ApplyDefaults`+`Validate` up front, so config errors report `stage:"validate"` and only dial/bind reports `stage:"connect"`.
3. `dfsctl identity-provider get/set/test` — switched to `cmdutil.PrintResourceWithSuccess`, so `-o json|yaml` is respected like other resource commands.
4. `validateKerberosDTO` — now rejects negative `max_clock_skew`/`context_ttl` (parity with `config.KerberosConfig.Validate`).

**Code review (non-blocking notes):**
- *Synth drops `resolved.GIDs`* — confirmed not a regression; documented the SMB-vs-NFS nested-group authorization asymmetry (SMB keys on group SIDs/PAC, not POSIX supplementary GIDs) in `docs/SMB.md`.
- *`Domain` dropped on synth user* — `models.User` has no domain field; the `sessionDomain()` fallback covers it. No change.
- *Kerberos restart-required* — already surfaced via the `X-DittoFS-Apply: restart-required` header + handler comment.

Also fixed a `wireIdentityResolver` data race / subscription leak under concurrent PUTs (serialized with `resolverMu` on both adapters) and a silently-dropped `MarshalStored` error in boot seeding. Full suite green under `-race`; CI green.

Issue #1231, opened 2026-06-18
Active Directory / LDAP enterprise integration (Tier 2 + multi-protocol idmap)


marmos91:
## Active Directory / LDAP enterprise integration (Tier 2 + multi-protocol idmap)

**Umbrella / tracking issue.** Make DittoFS enterprise-friendly on Windows by integrating with
Active Directory / LDAP, with correct multi-protocol identity unification (the same user sees the
same files & permissions via NFS *and* SMB).

### Why
The enterprise NAS bar (NetApp ONTAP, Dell PowerScale, Qumulo) is **full AD domain membership +
multi-protocol identity unification**. The cloud-native systems DittoFS resembles — JuiceFS,
CephFS, MinIO — do **not** solve clean multi-protocol AD. That gap is a differentiator DittoFS
can own.

### Scope of this milestone — **Tier 2 + idmap**
Kerberos/NTLM auth against AD + LDAP user/group lookup + **SID↔UID/GID unification** (not auth
alone). Full online `net ads join` (MS-RPC) is **deferred** — we ship with offline keytab import.

### Locked decisions
| Decision | Choice |
|----------|--------|
| First tier | Tier 2 + idmap |
| idmap | RFC2307 from AD (`uidNumber`/`gidNumber`), RID algorithmic fallback |
| Domain credential | Offline keytab import now; online join later |
| Delivery | Sequenced PRs, each gated by a dockerized Samba AD-DC CI fixture |
| Testing | Docker Samba AD-DC every PR; scw VM + real Windows client for milestone acceptance |

### Foundation already in place (full audit)
- `pkg/identity/` pluggable `IdentityProvider` chain + resolver cache. Kerberos provider exists.
- `AuthContext.Identity` already dual-stack (Unix UID/GID/GIDs **and** Windows SID/GroupSIDs).
- `pkg/metadata/acl/` protocol-agnostic ACL core (`acl.Evaluate`/`EvaluateGranted`, one
  `EvaluateContext` for both Unix & Windows). NFSv4 ACL ↔ Windows DACL ↔ POSIX mode canonicalize
  to one in-memory `acl.ACL`.
- ACL persistence verified solid (postgres `acl JSONB` migration `000004`, badger JSON blob,
  round-trip-proven by `storetest/acl_aliasing.go`). The durability gap is the *identity/idmap*
  layer, not the ACL.
- SMB NTLMv2 + Kerberos AP-REQ verify (gokrb5) work; real-KDC tests exist.

[...]

PR #1326, opened 2026-06-22
feat(operator): model spec.logging and spec.identity.kerberos in CRD (#1319)

### LINKED ISSUE #1319: Operator: config-only spec changes update the ConfigMap but don't roll the pod (need checksum annotation)

marmos91:
Closes #1319.

## Root cause (corrected from the issue text)

The issue proposed adding a `checksum/config` annotation — but that already exists. The operator stamps `dittofs.io/config-hash` on the pod template every reconcile; it hashes the CRD `generation`, the rendered config YAML, and the referenced Secret bytes. Any real spec change bumps `generation` → the annotation flips → the StatefulSet rolls. Verified with a probe test.

The actual gap: two config-only knobs were **not modelled in the CRD**, so they couldn't be changed at all:

- `spec.logging` did not exist — the rendered `logging.level` was hardcoded to `INFO`. So `kubectl patch spec.logging.level=DEBUG` was pruned by the structural schema and nothing happened (the issue's "ConfigMap shows level: DEBUG" was not actually reachable).
- `spec.identity.kerberos` was absent (only LDAP existed).

## Changes

- **`spec.logging`** (level/format/output) → CRD + rendered in `GenerateDittoFSConfig`. Rollout comes free from the existing hash.
- **`spec.identity.kerberos`** → CRD + `kerberos:` block render. Keytab + optional `krb5.conf` mounted read-only from Secrets (same pattern as TLS cert / metrics token); `kerberos.{keytab_path,krb5_conf}` point at the mounts. Keytab/krb5.conf Secret bytes folded into `collectSecretData` so a **keytab rotation rolls the pod**.
- **Webhook validation**: SPN + keytab required when Kerberos is enabled.
- Regenerated deepcopy + CRD bases + Helm chart CRDs; documented in `CRD_REFERENCE.md`.

## Tests

- config render (logging defaults/override, kerberos block + mount paths, keytab bytes never inlined)
- `logging change rolls pod`, `keytab mount + rotation flips hash`
- operator→server key-contract test (`serverKnownKeys`) extended with the kerberos keys + a kerberos case
- webhook validation cases

`make test` (incl. envtest) + `make lint` green, gofmt clean.

> Note: `develop` is not the default branch, so this PR won't auto-close #1319 on merge — close manually.


marmos91:
## Review — operator Kerberos + logging

Cross-checked against the live in-cluster AD/Kerberos setup I used to verify #1317 (built from develop, rolled `demo-0` against the Samba AD-DC). This looks correct and supersedes the operator commits on my `feat/ad-ldap-acceptance` branch — I'll drop those.

### ✅ Correct

- **Config keys match the server.** Rendered `kerberos.{enabled,keytab_path,service_principal,realm,netbios_domain,dns_domain,krb5_conf}` line up with `config.KerberosConfig`, and the `serverKnownKeys` contract test guards the operator→server key contract. Good.
- **Keytab/krb5.conf mount.** Read-only `Items`/`KeyToPath` projection (not `subPath`) — so kubelet propagates Secret updates and you sidestep the subPath "no auto-update" pitfall. Path consistency is guaranteed by the shared `KerberosKeytabFilePath()` / `KerberosKrb5ConfFilePath()` helpers, so `keytab_path` always points at the mount. (Different paths from my hand-rolled `/etc/dittofs/krb5.keytab` demo, but self-consistent — fine.)
- **Rotation rolls the pod.** Folding the keytab + krb5.conf Secret bytes into `collectSecretData` → the existing `dittofs.io/config-hash` flips on rotation. Correct, and the root-cause correction in the PR body (hash already existed; the real gap was unmodelled CRD knobs) matches what I saw — a `spec.logging.level` patch was a no-op before because the field didn't exist.
- **Webhook** requires SPN + keytab (with key) when Kerberos is enabled. Sensible fail-closed.

### One note (in-cluster KDC discovery)

The webhook requires SPN + keytab but **not** `krb5ConfSecretRef`, and `buildKerberosConfig` leaves `krb5_conf` empty when none is referenced. In my live in-cluster run, Kerberos only worked once I supplied an explicit krb5.conf Secret with a pinned `kdc = samba-ad-dc...` and `dns_lookup_kdc = false` — SRV-based KDC discovery did not resolve in-cluster, and the dfs container ships no `/etc/krb5.conf`. So a user who enables Kerberos with SPN+keytab only will likely fail KDC location unless their realm is DNS-SRV-discoverable.

Suggestion (non-blocking): either document that `krb5ConfSecretRef` is effectively required in-cluster, or add a webhook **warning** (not error) when Kerberos is enabled without it. The code is correct as-is; this is about steering users away from a silent KDC-lookup failure.

Refs #1319 #1317 #1231

PR #1334, opened 2026-06-22
fix(test): export AES keytab from AD-DC fixture (#1318)

### LINKED ISSUE #1318: Service keytab is RC4-only — Windows clients (RC4 disabled) cannot Kerberos-auth; need AES enctypes

marmos91:
Closes #1318.

## Problem

The dockerized Samba AD-DC test fixture exported an **RC4-only** service keytab. The Administrator account holding the `cifs/` + `nfs/` SPNs carried only an arcfour-hmac (RC4) Kerberos key, so `samba-tool domain exportkeytab` (which never restricts enctype) emitted RC4 only. **Windows 11 refuses RC4/arcfour-hmac service tickets**, so a Win11 client cannot authenticate against a DittoFS server using this keytab. (Linux `smbclient` tolerates RC4, which is why earlier in-cluster SMB-Kerberos testing passed despite this.)

The fixture's krb5.conf/smb.conf were not forcing arcfour — the only missing piece was AES keys on the account.

## Fix (`test/integration/ad-dc/entrypoint.sh`)

After registering the SPNs, before `exportkeytab`:
1. `ldbmodify` the account's `msDS-SupportedEncryptionTypes = 31` (0x1F = DES + RC4 + AES128 + AES256 — keeps legacy bits so no existing client breaks while adding AES).
2. `samba-tool user setpassword Administrator --newpassword=<same>` to regenerate the Kerberos keys (AES keys derive from password+salt, so they only materialize on a password change). Resetting to the same password avoids disturbing the rest of the fixture.

`exportkeytab` then emits the full key set (AES256 + AES128 + RC4). Added an in-container `klist -ke` fast-fail that requires both `aes256-cts` and `aes128-cts` so an RC4-only regression can't slip through.

## Test (`ad_dc_keytab_test.go`)

New assertion parses the exported keytab with gokrb5 and verifies **each** SPN (`cifs/`, `nfs/`) carries etype 18 (aes256-cts) **and** etype 17 (aes128-cts) — not merely that the keytab is non-empty.

## Verification

Local (Windows): `go build ./...`, `go vet -tags=ad_dc ./test/integration/ad-dc/...`, `gofmt -l` clean, package compiles under the `ad_dc` build tag.

The `ad_dc` integration test needs Docker + the Samba image and does not run on Windows — end-to-end keytab-enctype verification runs on the Linux VM fixture. (Mechanism validated live against the running Samba AD-DC — see issue.)

### DIFF
--- a/test/integration/ad-dc/ad_dc_keytab_test.go
+++ b/test/integration/ad-dc/ad_dc_keytab_test.go
@@ -38,6 +38,8 @@ import (
 	"testing"
 	"time"
 
+	"github.com/jcmturner/gokrb5/v8/keytab"
+
 	smbauth "github.com/marmos91/dittofs/internal/adapter/smb/auth"
 	"github.com/marmos91/dittofs/internal/adapter/smb/handlers"
 	kerbauth "github.com/marmos91/dittofs/internal/auth/kerberos"
@@ -83,6 +85,14 @@ func TestADCombinedKeytabAndDomainAwareSMB(t *testing.T) {
 		t.Errorf("combined keytab missing nfs/ principal component (file %d bytes)", len(ktBytes))
 	}
 	t.Logf("combined keytab (%d bytes) carries both cifs/ and nfs/ principals ✓", len(ktBytes))
+
+	// --- (1b) Each service principal carries AES256 + AES128 keys (#1318) ----
+	// Windows 11 refuses RC4 (arcfour-hmac, etype 23) service tickets, so the
+	// exported keytab MUST carry aes256-cts (etype 18) and aes128-cts (etype 17)
+	// keys for each SPN. Parse the keytab with gokrb5 and assert both enctypes
+	// are present per service-principal component — not merely that the file is
+	// non-empty or that some AES key exists somewhere.
+	assertKeytabHasAESPerSPN(t, ktBytes)
 
 	// --- (2) Provider surfaces the configured/derived AD domain identity ----
 	// Configure Realm + NetBIOSDomain explicitly (the short name is never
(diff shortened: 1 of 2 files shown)


marmos91:
## Live root-cause confirmation (running Samba AD-DC)

Non-destructive probe against the live AD-DC fixture (exported a throwaway keytab, did **not** mutate the account, to avoid staling the demo's in-use keytab):

```
$ samba-tool domain exportkeytab /tmp/probe.keytab --principal=cifs/dittofs.dittofs.ad@DITTOFS.AD
$ klist -ke /tmp/probe.keytab
   1 cifs/dittofs.dittofs.ad@DITTOFS.AD (DEPRECATED:arcfour-hmac)

$ ldbsearch ... "(sAMAccountName=Administrator)" msDS-SupportedEncryptionTypes
   (attribute unset)
```

So the SPN-holding account carries **only** the RC4 (arcfour-hmac) key and has **no** `msDS-SupportedEncryptionTypes` — exactly the condition this PR fixes. Win11 would reject a service ticket keyed with this. The two-step fix (set `msDS-SupportedEncryptionTypes=31`, then `setpassword` to regenerate keys so AES256/AES128 materialize) is the correct remedy; the `klist -ke` guard + the gokrb5 per-SPN AES assertion lock it in.

End-to-end (regenerated keytab actually carries aes256-cts/aes128-cts for both SPNs) is verified by the `ad_dc` Docker integration test on the Linux fixture. I did not run the destructive key-regen on the live DC because it would invalidate the demo's currently-working keytab Secret (KVNO bump) and require a re-export + Secret update + pod roll.

PR #1348, opened 2026-06-23
fix(dfs): don't log context.Canceled as ERROR on graceful shutdown

### LINKED ISSUE #1329: Clean shutdown logs ERROR "Server shutdown error: context canceled"
## Observed

On a normal graceful shutdown (SIGTERM → ordered drain completes, metadata writes flushed, `DittoFS runtime stopped`), the very last line is:

```
ERROR  Server shutdown error  error="context canceled"
```

Captured live on the demo deployment while verifying #1316 (bounded graceful shutdown). The shutdown itself was clean — adapters drained, `Flushed pending metadata writes count=2`, stores closed, runtime stopped — so the trailing ERROR is misleading noise.

## Why it's a bug

`context.Canceled` is the **expected** result of a SIGTERM-initiated shutdown (the root context is cancelled on purpose). CLAUDE.md / the logging convention is explicit: log expected errors at Debug, unexpected at Error. A clean shutdown emitting ERROR will trip alerting / log-based health checks and muddy postmortems.

## Fix

In the server Serve/shutdown path, treat `errors.Is(err, context.Canceled)` (and `http.ErrServerClosed`) during shutdown as the normal case — log at Debug/Info, reserve Error for an actual failure. Likely in `cmd/dfs` runtime serve or `pkg/controlplane/runtime/lifecycle`.

## Severity

Low (cosmetic / log-hygiene), but trivially fixable and user-visible on every restart.
EOF

marmos91:
Closes #1329

A SIGINT/SIGTERM-initiated shutdown cancels the root context on purpose, so the server returning `context.Canceled` is the **expected** outcome of a clean drain — not a failure. Logging it at ERROR tripped log-based alerting / health checks on every clean restart (observed live on the demo deploy while verifying #1316).

**Fix:** in the graceful-shutdown branch of `cmd/dfs/commands/start.go`, treat `errors.Is(err, context.Canceled)` as the normal case (no ERROR, no non-zero return). An unexpected server exit (the other select branch) still logs ERROR.

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -339,37 +339,15 @@ func runStart(cmd *cobra.Command, args []string) error {
 		logger.Info("Shutdown signal received, initiating graceful shutdown")
 		cancel()
 
-		// Wait for the runtime to drain (snapshots -> StopAllAdapters -> flush
-		// -> close stores -> stop API). The drain runs these stages serially
-		// and most are individually bounded by the configured shutdown timeout,
-		// but a wedged stage (e.g. a stuck store flush or an unbounded store
-		// close) could otherwise block the process indefinitely. Under
-		// Kubernetes that means the pod hangs until SIGKILL
-		// at the end of its terminationGracePeriod, dropping clients abruptly —
-		// exactly what #1313 is about. Cap the overall wait so the process
-		// always exits on its own first.
-		//
-		// The factor mirrors the operator's shutdownStageMultiplier=3, which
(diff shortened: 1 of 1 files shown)


marmos91:
Redundant: #1329 is already fixed on develop (cmd/dfs/commands/start.go isExpectedShutdownErr + start_shutdown_test.go, shipped with the #1316 bounded-shutdown work). This branch was accidentally based on a pre-fix snapshot and reverted it. Closing without merge.

PR #1356, opened 2026-06-23
feat(nfs): NFSv3 file locking — NLM v1/v3 + NLM/NSM over UDP (#1353)

### LINKED ISSUE #1353: NFSv3 file locking (NLM) fails from macOS: NLM/NSM are TCP-only / NLM v4-only, advertised on non-standard portmapper (10111)
## Summary

A macOS NFSv3 client cannot acquire file locks against DittoFS. `flock`/`fcntl` lock requests fail with **"lockd not responding"** / broken-pipe. NFSv4 (`-o vers=4`) and `-o nolocks` both work, which isolates the fault to the **NLM (Network Lock Manager) path used by NFSv3**.

Root cause: DittoFS offers locking only as **NLM v4 over TCP**, advertised on a **non-standard embedded portmapper (port 10111)**. macOS `rpc.lockd` / `rpc.statd` require **NLM/NSM over UDP** (and NLM v1/v3), discovered via the **standard portmapper on port 111**. None of that is provided, so the macOS kernel lock client never finds a usable `nlockmgr`.

This is an upstream DittoFS design limitation, surfaced by the macOS NFSv3 client. dittofs-pro inherits it 1:1 (companion issue: marmos91/dittofs-pro).

## Environment

- DittoFS `v0.20.0` (also affects dittofs-pro v0.21.x which embeds it)
- macOS 26.5.1 (arm64), stock `mount_nfs` / `rpc.lockd`
- Server and client on the same host (localhost), but the same chain applies cross-host

## Reproduction

Server (NFS on 12049, embedded portmapper enabled on 10111):

```bash
dfsctl store metadata add --name default --type memory
dfsctl store block local add --name fs-cache --type fs --config '{"path":"/tmp/blocks"}'
dfsctl share create --name /share-local --metadata default --local fs-cache --default-permission read-write
dfsctl adapter settings nfs update --portmapper-enabled   # then restart the server (see note in "Secondary bug" #2)
```

Client (macOS):

```bash
sudo mkdir -p /private/tmp/share-local
sudo mount -t nfs -o vers=3,port=12049,mountport=12049,resvport localhost:/share-local /private/tmp/share-local
echo hi | sudo tee /private/tmp/share-local/hello.txt          # IO works
python3 -c 'import fcntl; fcntl.flock(open("/private/tmp/share-local/lockme","w"), fcntl.LOCK_EX)'   # hangs -> "lockd not responding"
```

Mount and read/write succeed; the lock request fails.

### 1. NLM/NSM registered TCP-only, NLM v4 only

`internal/adapter/nfs/portmap/registry.go:130-162` (`RegisterDittoFSServices`):

```go
// ...on TCP only (DittoFS NFS adapter is TCP-only, no UDP transport).
services := []svc{
    {100003, 3}, {100003, 4},               // NFS
    {100005, 1}, {100005, 2}, {100005, 3},  // MOUNT
    {100021, 4},                            // NLM v4   <-- v4 only
    {100024, 1},                            // NSM v1
}
for _, s := range services {
    r.Set(&xdr.Mapping{Prog: s.prog, Vers: s.vers, Prot: types.ProtoTCP, Port: port}) // TCP only
}
```

NLM and NSM are multiplexed onto the single NFS TCP socket (`pkg/adapter/nfs/dispatch.go:196`, `case rpc.ProgramNLM`). There is no UDP transport for NLM/NSM, and NLM is advertised only as version 4.

### 2. Direct GETPORT probe against the embedded portmapper (UDP → 10111)

A raw portmap v2 GETPORT confirms exactly what a client can discover:

| Service | GETPORT result |
|---|---|
| NFS v3 / TCP | 12049 |
| MOUNT v3 / TCP | 12049 |
| **NLM v4 / TCP** | **12049** (nlockmgr *is* served) |
| **NLM v4 / UDP** | **0 — not registered** |
| **NLM v1 / UDP** | **0 — not registered** |
| **NSM v1 / UDP** | **0 — not registered** |

So `nlockmgr` is **not** missing (a common misdiagnosis) — it is reachable only as **NLM v4 / TCP / port 12049 / via the 10111 portmapper**.

### 3. Why macOS still fails

- macOS `rpc.lockd`/`rpc.statd` perform NLM/NSM discovery against the **standard portmapper on port 111**, not the DittoFS portmapper on 10111. There is no mount option (`port=`/`mountport=` cover only NFS/MOUNT) to redirect the NLM lookup. The macOS kernel queries `server:111`, gets its own host's idle `rpc.lockd` registration (or nothing), and reports "lockd not responding".
- Even if the lookup reached 10111, macOS predominantly uses **UDP** and **NLM v1/v3** for locking — none of which is registered or served.

## Fix options (for discussion)

1. **Serve NLM (100021) and NSM (100024) over UDP**, and register NLM **v1/v3** in addition to v4, so the standard BSD/macOS lock client can use them.
2. **Allow the embedded portmapper to bind port 111** (privileged) so standard clients discover services without per-program port options — at minimum document that NFSv3 locking from macOS requires this.
3. Document clearly that **NFSv3 + NLM locking is unsupported from macOS**; recommend **`vers=4`** (locking is in-protocol, no NLM/portmapper) or `-o nolocks`. The `vers=4` workaround is verified.

## Notes

- The "port 971 has no owner" observation from initial triage is macOS's **own** `rpc.lockd` registration in the system rpcbind (port 111), not a DittoFS registration. DittoFS does not register anything with the system rpcbind.
- `pkg/adapter/nfs/adapter.go:245` documents "Portmapper … Default: enabled on port 10111", but the actual default is **disabled** (`pkg/adapter/nfs/portmap.go:20-25`). Stale comment.

marmos91:
Fixes #1353.

## Problem

NFSv3 byte-range locking failed from macOS (and BSD) clients. DittoFS offered NLM only as **v4 over TCP**, with no UDP transport and a non-standard portmapper. macOS `rpc.lockd`/`rpc.statd` need substantially more, and each missing piece failed silently.

Live debugging against a real macOS client uncovered **a chain of distinct gaps**, each a real fix:

## What changed

1. **NLM v1/v3 + UDP transport** — width-aware NLM XDR (32-bit offsets for v1/v3, 64-bit for v4; handler logic stays version-agnostic). New UDP listener serving NLM/NSM/MOUNT (NFS data stays TCP-only). `adapters.nfs.udp.enabled`.
2. **NSM SM_MON callback fix** — real `statd` send `my_name` as a **hostname** (e.g. `localhost`); the IP-literal SSRF guard rejected it with `STAT_FAIL`, breaking locking for every client. Now the SM_NOTIFY callback target is derived from the request's **transport source** (still rejects link-local/IMDS); `my_name` is a label only — accepts hostnames *and* tightens SSRF.
3. **RPCBIND v3/v4** — macOS/BSD `lockd` discover NLM via RPCBIND v3/v4 (RFC 1833 universal addresses) and do **not** fall back to PMAP v2. The embedded portmapper now answers v2 **and** v3/v4 (GETADDR/GETVERSADDR/DUMP), reusing the registration table.
4. **Async NLM `_MSG`/`_RES`** — macOS/BSD use the callback-style async procedures: client sends `*_MSG` (procs 6–9), server replies **void** and delivers the result as a `*_RES` callback (procs 11–14). Async handlers reuse the sync decode/process/encode; the `*_RES` body is byte-identical to the sync reply.
5. **`_RES` from the NLM listening socket** — `lockd` connects its UDP socket to the server's `nlockmgr`, so the kernel drops any datagram whose source isn't that peer. The `*_RES` is sent via the live listener (source = NLM port), not a freshly-dialled socket.

Plus from the original scope: `dfsctl share mount --nfs-version 3|4|4.1`, operator NFS-port-over-UDP exposure, and the `udp_enabled` adapter setting threaded through model→store→API→apiclient→`dfsctl`.

## Defaults & safety

Both `udp.enabled` and `portmapper.enabled` default **OFF** — the entire feature is **opt-in**, so there is **no regression risk** to existing paths. NFSv3 mount + read/write need none of this; **NFSv4 needs none of it** (in-protocol locking) and remains the recommended path. To enable macOS NLM locking:

```
dfsctl adapter settings nfs update --udp-enabled true --portmapper-enabled true --portmapper-port 111
# then bounce the adapter
```
(macOS queries the server's portmapper on **:111** specifically.)

## Validation

- **Unit tests** added across every layer: width-aware NLM XDR, UDP framing, portmap registration + RPCBIND GETADDR/DUMP, NSM source-derived callback (incl. the macOS hostname case), async `_MSG`→`_RES` reuse, settings plumbing, operator ports.
- **Live wire validation** (real running server): RPCBIND v3/v4 GETADDR resolves `nlockmgr`; NSM SM_MON succeeds; macOS sends `LOCK_MSG`; server **grants**; a well-formed `LOCK_RES` (correct cookie, `prog/vers/proc=12`) is delivered **from the NLM socket**. The full discovery→monitor→lock chain executes end to end.
- **Status:** end-to-end *acceptance* against a live macOS client is validated on a **same-LAN** topology, not localhost — a same-host loopback test is unreliable (client and server contend for port 111 and a shared `lockd`/`statd`, and the reverse `_RES` callback can't be cleanly observed). Tracked as a follow-up; the protocol implementation is complete and unit-tested.

All commits signed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -52,7 +52,7 @@ Two binaries drive it:
 
 | Area | Status |
 |------|--------|
-| **NFSv3** | All 28 core procedures; embedded portmapper + mount protocol |
+| **NFSv3** | All 28 core procedures; embedded portmapper + mount protocol; NLM/NSM byte-range locking over TCP+UDP (opt-in) |
 | **NFSv4.0** | Compound ops, ACLs, delegations, built-in byte-range locking |
 | **NFSv4.1** | Sessions, sequence slots, backchannel |
 | **SMB 2.0.2 / 3.0 / 3.0.2 / 3.1.1** | Multi-dialect negotiation, preauth integrity, compound requests |
--- a/cmd/dfsctl/commands/adapter/settings.go
+++ b/cmd/dfsctl/commands/adapter/settings.go
@@ -185,6 +185,7 @@ var (
 	settingsBlockedOperations          string
 	settingsPortmapperEnabled          bool
 	settingsPortmapperPort             int
+	settingsUDPEnabled                 bool
 	settingsV4MinMinorVersion          int
 	settingsV4MaxMinorVersion          int
 	settingsV4MaxConnectionsPerSession int
@@ -259,6 +260,7 @@ func registerNFSUpdateFlags(cmd *cobra.Command) {
 	cmd.Flags().StringVar(&settingsBlockedOperations, "blocked-operations", "", "Comma-separated list of blocked operations")
 	cmd.Flags().BoolVar(&settingsPortmapperEnabled, "portmapper-enabled", false, "Enable embedded portmapper")
 	cmd.Flags().IntVar(&settingsPortmapperPort, "portmapper-port", 0, "Portmapper listen port")
+	cmd.Flags().BoolVar(&settingsUDPEnabled, "udp-enabled", false, "Serve NLM/NSM/MOUNT over UDP (needed for NFSv3 locking from macOS/BSD; restart to apply)")
 	cmd.Flags().IntVar(&settingsV4MinMinorVersion, "v4-min-minor-version", 0, "Minimum NFSv4 minor version (0=v4.0, 1=v4.1)")
 	cmd.Flags().IntVar(&settingsV4MaxMinorVersion, "v4-max-minor-version", 0, "Maximum NFSv4 minor version (0=v4.0, 1=v4.1)")
 	cmd.Flags().IntVar(&settingsV4MaxConnectionsPerSession, "v4-max-connections-per-session", 0, "Maximum connections per NFSv4.1 session (0=unlimited)")
@@ -351,6 +353,7 @@ func showNFSSettings(client *apiclient.Client, format output.Format) error {
 	printSettingsGroup("Portmapper", []settingRow{
 		newSettingRowBool("portmapper_enabled", settings.PortmapperEnabled, d.PortmapperEnabled),
 		newSettingRowInt("portmapper_port", settings.PortmapperPort, d.PortmapperPort, ""),
+		newSettingRowBool("udp_enabled", settings.UDPEnabled, d.UDPEnabled),
 	})
 	printSettingsGroup("Operations", []settingRow{
 		newSettingRowStrSlice("blocked_operations", settings.BlockedOperations, d.BlockedOperations),
@@ -568,6 +571,10 @@ func updateNFSSettings(cmd *cobra.Command, client *apiclient.Client) error {
 	}
 	if cmd.Flags().Changed("portmapper-port") {
 		req.PortmapperPort = &settingsPortmapperPort
+		hasChanges = true
+	}
+	if cmd.Flags().Changed("udp-enabled") {
+		req.UDPEnabled = &settingsUDPEnabled
 		hasChanges = true
 	}
 	if cmd.Flags().Changed("v4-min-minor-version") {
--- a/cmd/dfsctl/commands/share/mount.go
+++ b/cmd/dfsctl/commands/share/mount.go
@@ -19,11 +19,12 @@ const (
 )
 
 var (
-	mountProtocol string
-	mountUsername string
-	mountPassword string
-	mountFileMode string
-	mountDirMode  string
+	mountProtocol   string
+	mountUsername   string
+	mountPassword   string
+	mountFileMode   string
+	mountDirMode    string
+	mountNFSVersion string
 )
 
 var mountCmd = &cobra.Command{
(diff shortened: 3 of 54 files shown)


marmos91:
Addressed both Copilot findings in `pkg/adapter/nfs/udp.go` (folded into the feature commit):

1. **Unbounded UDP handler goroutines** — `serveUDP` now bounds in-flight handlers with a semaphore sized from `MaxRequestsPerConnection` (default 100), mirroring the TCP per-connection `requestSem`. UDP is lossy, so on overflow the datagram is dropped (debug-logged) rather than blocking the read loop — clients retransmit.
2. **MOUNT MNT version gate missing over UDP** — added the same `MNT requires v3` guard the TCP dispatcher applies (other MOUNT procs stay version-agnostic). Live-verified on the wire: MOUNT MNT v1/UDP → `PROG_MISMATCH`, v3/UDP → accepted.

Both fixes build clean and pass the touched unit tests.

marmos91:
Re: `encodeNLMOffset` narrow truncation (encode.go) — fixed, with a deliberate choice of **saturate** over **error**:

The only narrow-encode path that can carry a >32-bit value is the TEST **holder** field reflecting a conflicting v4 lock (a v1/v3 client's own locks always fit in 32 bits). That field is informational — the `nlm_stat` (DENIED) the client acts on is independent of it. Erroring the encode would fail the whole TEST reply and hang the v1/v3 client over an unrepresentable holder bound, which is worse than an imprecise-but-valid reply.

So narrow encode now **saturates to `0xFFFFFFFF`** instead of wrap-truncating — this kills the actual hazard you flagged (wrap folding e.g. `1<<32` to `0`, misreporting a conflict at offset 0) while keeping the reply valid. Added `TestNLMHolder_NarrowSaturatesNotWraps` and corrected the misleading 'always safe' doc comment.

Issue #1357, opened 2026-06-23
smb: NTLM passthrough e2e — SAMLogon LOGON_FAILURE for AD user via smbclient (challenge threading / SPNEGO NegTokenResp parse)


marmos91:
## Context
Surfaced while finishing #1345. The NETLOGON schannel itself now works end-to-end: `TestNetlogonPassthroughAlice` PASSES — a correctly-computed NTLMv2 response for alice is validated through the sealed Kerberos ncacn_np schannel and the DC returns her UserSID + GroupSIDs.

## Bug
The full SMB e2e `TestSMBNTLMNetlogonPassthrough` (real `smbclient` → DittoFS SMB → NETLOGON passthrough) FAILS. DittoFS-side debug:

```
SPNEGO parse failed, scanning for NTLMSSP signature
  error=spnego: invalid token format: error unmarshalling NegotiationToken type 1 (Resp/Targ):
  asn1: structure error: explicitly tagged member didn't match
User not found in UserStore username=alice    (expected -> triggers passthrough)
NETLOGON pass-through authentication failed username=alice domain=DITTOFS
  error=netlogon: SAMLogon: /logon/v1/NetrLogonSamLogon: error: -1073741715  (0xC000006D STATUS_LOGON_FAILURE)
```

Two distinct problems, likely linked:
1. **SPNEGO NegTokenResp (type 1) parse fails** in `extractNTLMToken` (`internal/adapter/smb/handlers/session_setup.go`) — `auth.Parse` (gokrb5) rejects smbclient's NegTokenResp, falling back to a raw NTLMSSP signature scan.
2. **SAMLogon returns LOGON_FAILURE** — the NT response forwarded to the DC doesn't validate. Most likely the server challenge DittoFS issued in the NTLM CHALLENGE message isn't the exact one threaded into `NetrLogonSamLogon` (or the NTLMv2 target-info/domain differs from what the DC expects). Since the schannel + SAMLogon are proven good with a correctly-built response, this is a DittoFS NTLM challenge-threading / token-extraction bug, not a schannel issue.

## Repro
`go test -tags=ad_dc -run TestSMBNTLMNetlogonPassthrough ./test/integration/ad-dc/` (Linux/CI, real Samba AD-DC). Test is `t.Skip`-gated on this issue until fixed.

## Scope
Distinct from #1345 (schannel interop, resolved). This is the SMB-server NTLM passthrough path.

marmos91:
[...]

Issue #1324, opened 2026-06-22
AD: DNS SRV-based DC discovery for NETLOGON secure channel


marmos91:
## Context
Follow-up to #1314 (SMB NTLM passthrough via NETLOGON / MS-NRPC). Part of the AD umbrella #1231.

The first NETLOGON cut binds the secure channel to a **static** DC address (`machine_account.dc_address` in the kerberos identity-provider config). Real deployments expect automatic DC location.

## Ask
Add **DNS SRV-based DC discovery**:
- Resolve `_ldap._tcp.dc._msdcs.<realm>` (and `_kerberos._tcp.` siblings) to candidate DCs.
- CLDAP ping (LDAP netlogon `NetlogonSamLogonResponseEx`) to select a live, in-site DC.
- Failover across multiple DCs; cache + re-probe on failure.
- Static `dc_address` remains an override.

## Notes
- Reuses the same realm/netbios already on the kerberos provider config.
- Should drop into `SecureChannel` DC selection without changing the auth flow.

marmos91:
Blocked by #1314. The base NETLOGON/MS-NRPC implementation does not exist yet — there is no `netlogon` package, `MachineCredentialProvider` interface, or `SecureChannel` in the tree, and #1314 is still open. This issue extends that foundation ("the first NETLOGON cut"), so it cannot start until #1314 lands. Marking the dependency so the ordering is explicit.

marmos91:
Core DNS SRV DC discovery landed in develop via #1358: `internal/auth/netlogon/discovery.go` resolves `_ldap._tcp.dc._msdcs.<realm>` (RFC 2782 ordered via net.LookupSRV) → DC FQDN → `cifs/<fqdn>` SPN, and is wired into the NETLOGON secure channel. Keeping this issue OPEN for the remaining pieces: surface DC discovery in config/Authenticator setup (populate DCAddresses from SRV instead of static config) and the _kerberos/_kpasswd SRV records + site-awareness. The reusable building block exists now.

marmos91:
Closed by #1359 (squash df385298): DNS SRV DC discovery wired as a full fallback when no dc_address is configured; realm required; SPN bound to the dial target.

PR #1358, opened 2026-06-23
fix(netlogon): Kerberos-bound ncacn_np schannel resolves Samba interop (#1345) + DC SRV discovery (#1324)

### LINKED ISSUE #1345: AD: NETLOGON schannel AlterContext rejected by Samba AD-DC (RPC_S_UNKNOWN_AUTHN_SERVICE 0x721)

marmos91:
## Summary
Resolves the NETLOGON schannel interop with Samba AD-DC (#1345). `TestNetlogonPassthroughAlice` now **PASSES** live against a real Samba AD-DC: alice's NTLMv2 is validated through a sealed Kerberos `ncacn_np` schannel and the DC returns her UserSID + GroupSIDs.

## Root causes found (live debugging against real Samba)
1. **The fixture never created the machine account.** `samba-tool computer create --computerpassword=...` is not a valid option on Samba 4.17; the `|| true` swallowed the error, so `DITTOFS$` never existed and every prior NETLOGON run was against a nonexistent account. Fixed: `computer create` + `user setpassword` (derives Kerberos keys).
2. **Samba rejects machine-account NTLM over SMB**, and rejects the schannel `AlterContext` over `ncacn_ip_tcp` (`0x721`). Fix: bind the schannel over a **Kerberos-authenticated** SMB session on `ncacn_np:[netlogon]`.
3. **`0x721` real cause** (Samba debug log): `Request for schannel to without domain → NT_STATUS_LOGON_FAILURE`. go-msrpc builds the schannel `NL_AUTH_MESSAGE` from the credential's domain, which we left empty. Fix: set the credential Domain to the realm — go-msrpc sends it as the NL_AUTH DNSDomainName and the Kerberos SMB session uses it as the principal realm.

## Changes
- `internal/auth/netlogon/discovery.go` — DNS SRV DC locator (`_ldap._tcp.dc._msdcs.<realm>`) → `cifs/<fqdn>` SPN. Implements the discovery core of **#1324** (queries the DC's own DNS; IP stays the connectivity path).
- `internal/auth/netlogon/securechannel.go` — register `ssp.KRB5`, inline krb5 config (realm→KDC), Kerberos SMB dialer targeting the DC `cifs/` SPN, direct `ncacn_np:[netlogon]` bind, credential Domain=realm.
- `test/integration/ad-dc/entrypoint.sh` — correct machine-account creation.
- tests — wait for DC :445; un-skip `TestNetlogonPassthroughAlice`.

## Verification
Live, real Samba AD-DC: `--- PASS: TestNetlogonPassthroughAlice`. Samba log confirms `netr_LogonSamLogon (opnum[2]) WITH SEALED schannel ... NT_STATUS_OK`. The `ad_dc` suite runs post-merge + nightly via `ad-dc.yml`.

## Follow-up
`TestSMBNTLMNetlogonPassthrough` (full smbclient e2e) stays `t.Skip`-gated on **#1357** — a separate DittoFS SMB NTLM challenge-threading bug, not a schannel issue.

Closes #1345

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -27,6 +27,7 @@ require (
 	github.com/manifoldco/promptui v0.9.0
 	github.com/mitchellh/mapstructure v1.5.0
 	github.com/oiweiwei/go-msrpc v1.4.3
+	github.com/oiweiwei/gokrb5.fork/v9 v9.0.6
 	github.com/oklog/ulid/v2 v2.1.1
 	github.com/olekukonko/tablewriter v0.0.5
 	github.com/pierrec/lz4/v4 v4.1.26
@@ -104,7 +105,6 @@ require (
 	github.com/oiweiwei/go-math v1.0.0 // indirect
 	github.com/oiweiwei/go-oem v1.0.0 // indirect
 	github.com/oiweiwei/go-smb2.fork v1.0.1 // indirect
-	github.com/oiweiwei/gokrb5.fork/v9 v9.0.6 // indirect
 	github.com/opencontainers/go-digest v1.0.0 // indirect
 	github.com/opencontainers/image-spec v1.1.1 // indirect
 	github.com/pkg/errors v0.9.1 // indirect
--- a/internal/auth/netlogon/discovery.go
+++ b/internal/auth/netlogon/discovery.go
@@ -1,0 +1,81 @@
+package netlogon
+
+import (
+	"context"
+	"fmt"
+	"net"
+	"strings"
+	"time"
+)
+
+// DCInfo identifies a domain controller discovered via DNS SRV records.
+type DCInfo struct {
+	// FQDN is the DC's canonical DNS host name without the trailing dot,
+	// e.g. "dc01.dittofs.ad". It is the basis for the Kerberos service
+	// principal (cifs/<FQDN>) used to authenticate the NETLOGON SMB session.
+	FQDN string
+	// Port is the SRV record port (389 for the _ldap locator record).
+	Port uint16
+	// Priority/Weight are the SRV selection fields, lowest priority first.
+	Priority uint16
+	Weight   uint16
+}
+
+// SPN returns the SMB service principal name for the DC (cifs/<FQDN>).
+//
+// AD domain controllers register HOST/<fqdn> rather than an explicit
+// cifs/<fqdn>, but the default AD sPNMappings alias the host service class to
+// cifs (and a dozen others), so the KDC issues a cifs/<fqdn> ticket from the
+// HOST key. This is the same SPN a Windows client uses for an SMB session.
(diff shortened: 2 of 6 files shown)


marmos91:
Addressed both review comments in d6db590b:
- **discovery.go ordering**: clarified the docstring — `net.LookupSRV` already returns records sorted by priority and randomized by weight per RFC 2782, and that order is preserved; no custom sort needed.
- **entrypoint.sh reconcile**: `user setpassword` now runs unconditionally (not only on first create), so a reused domain DB or an account left by an earlier run reconciles to the expected machine credential and regenerates its Kerberos keys.

PR #1359, opened 2026-06-23
feat(netlogon): locate DC via DNS SRV when no address configured (#1324)

### LINKED ISSUE #1324: AD: DNS SRV-based DC discovery for NETLOGON secure channel

marmos91:
## What

Wires DNS SRV DC discovery (`_ldap._tcp.dc._msdcs.<realm>`) as a full fallback for NETLOGON, so a machine-account config no longer **requires** an explicit `dc_address`.

Previously `DiscoverDCs` (landed in #1358) was used only to resolve the DC **FQDN** for the SMB Kerberos service principal — connectivity still needed a hardcoded `dc_address`. Now:

- **No `dc_address` configured** → locate a DC entirely from the realm via the host resolver (a domain-joined host's `resolv.conf` points at a DC) and dial the discovered FQDN.
- **`dc_address` configured** → unchanged: that address is the dial target, and the SRV lookup against it only resolves the FQDN for the SPN.

## Changes

- `credential.go`: accept a realm-only credential — require **either** a DC address **or** a realm (so we never build an authenticator that can neither dial nor discover).
- `securechannel.go connect()`: discover the DC address from the realm when none configured.
- `cmd/dfs/commands/netlogon.go`: keep NTLM passthrough enabled when only the realm is set (disable only when **both** `dc_address` and `kerberos.realm` are absent).
- Tests: realm-only credential validates; neither-set is rejected.

## Test

`go test ./internal/auth/netlogon/...` green. The discovery→dial path is exercised by the `ad_dc` integration suite.

Closes #1324.

### DIFF
--- a/cmd/dfs/commands/netlogon.go
+++ b/cmd/dfs/commands/netlogon.go
@@ -42,8 +42,12 @@ func buildNetlogonAuthenticator(k config.KerberosConfig) netlogon.NetlogonAuthen
 		slog.Warn("NETLOGON machine account is enabled but kerberos.netbios_domain (DomainName) is not set; NTLM passthrough disabled")
 		return nil
 	}
-	if len(ma.DCAddresses) == 0 {
-		slog.Warn("NETLOGON machine account is enabled but no DCAddresses are configured; NTLM passthrough disabled")
+	// A DC address is optional: with none configured the secure channel locates
+	// one from the realm via the AD DNS SRV record (_ldap._tcp.dc._msdcs.<realm>).
+	// That fallback needs the realm, so only disable passthrough when both are
+	// absent (#1324).
+	if len(ma.DCAddresses) == 0 && k.Realm == "" {
+		slog.Warn("NETLOGON machine account is enabled but neither dc_address nor kerberos.realm is set (no way to locate a DC); NTLM passthrough disabled")
 		return nil
 	}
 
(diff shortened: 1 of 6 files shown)


marmos91:
Thanks — all three addressed in 70addbbe:

1. **Realm now required** in both `offlineProvider.Credential` and `buildNetlogonAuthenticator`. A `dc_address`-only / realm-less config is rejected up front instead of failing later at `connect()`.
2. Same gate in `cmd/dfs/commands/netlogon.go` — passthrough is disabled when `kerberos.realm` is empty.
3. **SPN is now bound to the dial target** via `resolveDialTarget`: a hostname `dc_address` yields `cifs/<host>` directly; an IP resolves the matching SRV target (single-DC is unambiguous, multi-DC matches by resolved address, falling back to the first SRV result). No more `cifs/<other-dc>` mismatch.

Added a missing-realm test and switched the missing-`dc_address` test to assert passthrough stays enabled (DC located via DNS SRV).

PR #1363, opened 2026-06-23
fix(smb): NTLM NETLOGON pass-through — advertise AD identity + idmap_rid (#1357)

### LINKED ISSUE #1357: smb: NTLM passthrough e2e — SAMLogon LOGON_FAILURE for AD user via smbclient (challenge threading / SPNEGO NegTokenResp parse)

marmos91:
## What

Makes a **domain user authenticating over NTLM** (no Kerberos) work end-to-end via NETLOGON pass-through. Previously `smbclient -U 'DITTOFS\alice'` failed with `STATUS_LOGON_FAILURE` even though the secure channel was healthy (#1345).

## Root cause (two layers, isolated by replaying captured smbclient bytes against a live Samba AD-DC + bisecting the NTLMv2 AV_PAIRs)

1. **Wrong NTLM CHALLENGE TargetInfo.** DittoFS advertised `MsvAvNbDomainName=WORKGROUP` and `MsvAvNbComputerName=<OS hostname>`. A domain client echoes both into its NTLMv2 response; Samba's NETLOGON SamLogon rejects the (otherwise cryptographically valid) response when they don't match the AD NetBIOS domain and the machine-account name the channel authenticated as. The proof itself was always valid — verified by offline NTProofStr recompute.
2. **No POSIX mapping for the DC-returned SID.** Once SamLogon succeeded, the session failed closed because alice's SID had no UID/GID.

## Fix

- **Advertise the real AD identity** in the NTLM CHALLENGE: configured NetBIOS domain + machine-account NetBIOS computer name. `BuildChallenge` now takes the computer name; the adapter wires both from the machine-account config — covering the NTLM-only (`kerberos.enabled=false`) path that never reached `SetKerberosProvider`.
- **idmap_rid fallback** (`resolveNetlogonIdentity`): when no configured provider maps the SID, derive a stable UID/GID from the SID's RID (Samba `idmap_rid` model). Last resort — LDAP/local mappings always take precedence.

## Test

`TestSMBNTLMNetlogonPassthrough` (was skipped) now **passes** against the Samba AD-DC fixture: `smbclient` over NTLM authenticates and lists the share. Affected unit suites (`smb/auth`, `smb/handlers`, `pkg/adapter/smb`, `cmd/dfs/commands`) green; `BuildChallenge` callers updated for the new arg.

Closes #1357.

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -672,6 +672,18 @@ func createSMBAdapter(cfg *models.AdapterConfig, kerberosConfig *config.Kerberos
 	// false, so SetNetlogonAuthenticator no-ops in that case.
 	if kerberosConfig != nil {
 		nlAuth := buildNetlogonAuthenticator(*kerberosConfig)
+		// When NTLM pass-through is active, the SMB handler must advertise, in its
+		// NTLM CHALLENGE TargetInfo, both the AD NetBIOS/DNS domain AND the AD
+		// machine-account computer name. Domain clients echo these into their
+		// NTLMv2 response; if the advertised domain is WORKGROUP or the computer
+		// name is the bare OS hostname (not the machine account), Samba's NETLOGON
+		// SamLogon rejects the otherwise-valid response (#1357). SetKerberosProvider
+		// covers the SPNEGO-enabled path, but NTLM-only members (kerberos.enabled
+		// =false) never reach it, so set it explicitly here. The computer name is
+		// the same workstation name the NETLOGON secure channel authenticates as.
+		if nlAuth != nil {
+			smbAdapter.SetADDomain(kerberosConfig.NetBIOSDomain, kerberosConfig.DNSDomain, netbiosWorkstation(*kerberosConfig))
+		}
 		smbAdapter.SetNetlogonAuthenticator(nlAuth)
 	}
 
--- a/internal/adapter/smb/auth/ntlm.go
+++ b/internal/adapter/smb/auth/ntlm.go
@@ -348,7 +348,16 @@ func GetMessageType(buf []byte) MessageType {
 // the server's configured AD domain to make the challenge domain-aware; pass
 // empty strings for a standalone server (TargetInfo then advertises
 // WORKGROUP / "local" exactly as before AD-4).
-func BuildChallenge(netbiosDomain, dnsDomain string) (message []byte, serverChallenge [8]byte) {
+//
+// netbiosComputer is the server's own NetBIOS computer name, advertised as the
+// MsvAvNbComputerName AV_PAIR. When NETLOGON pass-through is active this MUST be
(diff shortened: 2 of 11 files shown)


marmos91:
Addressed in 52305ff0 + 966f92c7:
1. **Fail closed on resolver error** — `resolveNetlogonIdentity` now returns nil (no RID fallback) when `resolver.Resolve` returns an error, since that signals an infra failure; the idmap_rid fallback only fires on a clean not-Found or when no resolver is installed.
2. **Doc comment** refreshed to describe the resolver→idmap_rid order + the fail-closed-on-error rule.
3. **Unit tests added** (`netlogon_idmap_rid_test.go`): `ridFromSID` (valid/empty/malformed/non-numeric/overflow) and `synthesizeRIDIdentity` (UID==RID, group-SID skipping, nil on bad SID).

PR #1367, opened 2026-06-23
feat(metadata): add pure-Go SQLite metadata store (#1224)

### LINKED ISSUE #1224: SQLite metadata store
Part of #1221 (ROI: #2 — low effort, high value).

Single-binary appliance/edge deploy; reuses postgres SQL (recursive CTEs supported).

**Scope**
- New `pkg/metadata/store/sqlite/` implementing the ~110-method `metadata.Store` (`pkg/metadata/store.go`).
- Replicate postgres model: `parent_child_map` (no unique on child_id → hard links), `nlink` + partial-unique-on-active index, recursive-CTE path reconstruction, `object_id` index.
- Migrations dir mirroring `pkg/metadata/store/postgres/migrations/`.
- Factory case in `CreateMetadataStoreFromConfig()` (`runtime/init.go:20-150`).
- Driver: `modernc.org/sqlite` (pure-Go, no cgo).

**Verify** full `storetest` suite (~137 cases) green; wire into CI like postgres.

marmos91:
## What

Adds a new SQLite-backed metadata store at `pkg/metadata/store/sqlite/`, implementing the full ~110-method `metadata.Store` interface (plus the lock / durable-handle / NFSv4-recovery sub-stores, backup/restore, and reset). It uses the pure-Go `glebarez/go-sqlite` driver over `database/sql` — **no cgo**.

## Why

For single-binary appliance / edge deployments that want durable metadata without standing up PostgreSQL. SQLite reuses the same relational model as Postgres (recursive CTEs, partial-unique indexes), so it was the natural second persistent backend (badger is the embedded KV option; this is the embedded SQL option). Closes #1224.

## Design

The PostgreSQL store is the reference implementation; this is a faithful port with the SQL adapted to the SQLite dialect:

- **Same model:** `parent_child_map` with **no unique on `child_id`** (hard links), `nlink` as source of truth with the partial-unique-on-active index pattern, recursive-CTE path reconstruction, and the `object_id` partial-unique dedup index (`inodes_object_id_idx`).
- **Dialect adaptations:** `?N` indexed placeholders (preserves Postgres's out-of-order `$21/$22` in PutFile); `json_group_array`/`json_array`/`lower(hex())` for the block-ref manifest (vs `json_agg`/`encode`); `TIMESTAMP`-typed columns so the driver round-trips `time.Time`; `byte_offset`/`byte_length` stored as decimal **TEXT** to hold the full `uint64` range; no `LATERAL` (recursive CTE picks one parent edge via a correlated `rowid` subquery), no `FOR UPDATE`/`TRUNCATE`/`md5()` (SQLite has no btree row-size limit, so `content_id` is indexed directly).
- **Single consolidated migration** (mirrors the postgres `migrations/` layout) producing the final schema in one step — SQLite has no existing deployments to migrate incrementally. A small bespoke migration runner applies it; `golang-migrate`'s sqlite driver is intentionally avoided because it blank-imports `modernc.org/sqlite`, which would register the `database/sql` driver name `"sqlite"` a second time alongside the control-plane's `glebarez/go-sqlite` and panic at init.
- **Factory:** `case "sqlite"` in `CreateMetadataStoreFromConfig` (`runtime/init.go`) and a restore-at-path case in `runtime/stores/service.go`.

## Conformance

Passes the **full shared `storetest` conformance suite — 187/187 subtests, `-race` clean** (FileOps, DirOps, Permissions, CrossProtocolPermissions, DurableHandles, ClientRecovery, ACL/EA/Xattr, FileBlock/BlockRef/Truncate/ObjectID ops, StoreSurface, INV02 fuzz, Backup/ResetThenRestore/LockPersistence).

CI wiring mirrors **badger**, not postgres: the suite is self-contained (no external DB), so `pkg/metadata/store/sqlite/sqlite_conformance_test.go` has **no `//go:build integration` tag** and runs in the default `go test -race ./pkg/...` unit-test job — unlike the integration-gated postgres suite that needs a live DSN.

## Review

Ran the code-simplifier and code-reviewer over the diff before opening. Reviewer verified the high-risk areas (out-of-order `?N` binding, hard-link CTE determinism, `uint64`→TEXT, EXCLUDED casing, JSON/BLOB hash round-trip, TIMESTAMP scanning, single-connection pool, CRC-before-commit restore, error mapping) as correct; the one MEDIUM finding (stale cached `store_id` after `Restore`) is fixed in this branch.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -1294,7 +1294,8 @@ Backends maintain a secondary index:
 
 | Backend  | Index                                                                       |
 |----------|-----------------------------------------------------------------------------|
-| Postgres | Partial unique: `files_object_id_idx ON files(object_id) WHERE object_id IS NOT NULL` (migration `000013_object_id`) |
+| Postgres | Partial unique: `inodes_object_id_idx ON inodes(object_id) WHERE object_id IS NOT NULL` |
+| SQLite   | Partial unique: `inodes_object_id_idx ON inodes(object_id) WHERE object_id IS NOT NULL` (pure-Go `glebarez/go-sqlite`, mirrors the Postgres model) |
 | Badger   | Secondary key `obj:{hex} -> file_id`, maintained inside each `Put`/`Delete` write batch |
 | Memory   | `map[ContentHash]uuid`, guarded by the existing store mutex                 |
 
--- a/docs/CONFIGURATION.md
+++ b/docs/CONFIGURATION.md
@@ -781,6 +781,12 @@ Metadata stores are managed at runtime via `dfsctl` and persisted in the control
 ./dfsctl store metadata add --name badger-isolated --type badger \
   --config '{"path":"/tmp/dittofs-metadata-isolated"}'
 
+# SQLite for a persistent single-binary / edge appliance (pure-Go, no cgo).
+# Reuses the PostgreSQL data model (parent_child_map hard links, nlink,
+# recursive-CTE path reconstruction, object_id dedup index).
+./dfsctl store metadata add --name sqlite-edge --type sqlite \
+  --config '{"path":"/var/lib/dittofs/metadata.db"}'
+
 # PostgreSQL for distributed, horizontally-scalable metadata
 # Set POSTGRES_PASSWORD in your environment
 ./dfsctl store metadata add --name postgres-production --type postgres \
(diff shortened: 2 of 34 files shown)


marmos91:
All 10 Copilot review comments are addressed in the current revision: the `$N`→`?N` placeholder conversions are done (0 `$N` remain; conformance exercises ListPending/ListLocks under -race), the rollup error messages say "sqlite", and `synced_at` exists in migration 000001 (the "missing column" reading was of a stale schema). An independent review also caught and I fixed an `isBusyError` gap (mapped "database busy" message wasn't classified for tx retry, commit 654eb019). The non-atomic pool-path CreateShare it flagged mirrors the postgres reference and is tracked as a separate follow-up for both backends.

Issue #1362, opened 2026-06-23
Read-through disk caching / archive pin policy: fetched blocks never persist to local SSD tier


raffaelegiustini-create:
## Summary

On a read-only / archive workload, **fetched blocks are never promoted to the local on-disk block store**. Reads are only cached in the in-memory read buffer (`read_buffer_size`, capped at 12.5% of RAM). Once the working set exceeds that buffer, every cold read pays full remote (S3) latency — permanently. There is no way to warm or pin a share's content onto the local SSD tier by reading it.

This makes the "feels-local archive over object storage" use case (write-once, read-occasionally; backup/restore; compliance storage) effectively WAN/S3-bound for any archive larger than ~1 GB.

## Environment

- Build: `develop` @ `3af6866` (includes the #1267/#1274 truncation fixes)
- Host: Hetzner CX, 7.6 GiB RAM, 75 GB disk
- Share `/drive`: single S3 remote (Cubbit DS3), `used_bytes` ≈ 43.4 GiB logical
- Deduced defaults on this host:
  - `local_store_size: 1.9 GiB` (25% of RAM)
  - `read_buffer_size: 968.3 MiB` (12.5% of RAM)
  - `parallel_fetches: 8`, `prefetch_workers: 4`

## Reproduction

Read the same 389 MiB file three times back-to-back, server-side (loopback NFS, so no client tunnel — this isolates the S3 fetch path):

| Read | Throughput | Interpretation |
|------|-----------|----------------|
| 1st (cold) | **11 MB/s** | fetched from remote (S3) |
| 2nd (immediate) | **76 MB/s** | served from in-memory read buffer |
| 3rd | **13 MB/s** | already evicted from the ~968 MiB buffer → re-fetched from S3 |

Then read **every** file in the share once (11,451 files):

- On-disk block cache grew by **0 bytes**
- `Blocks Local: 0`
- `Local Disk Used` unchanged at 1.3 GB
- `dfsctl store block evict --local-only` freed 0 (read path never populated the durable tier)

So: the only warm tier for reads is a sub-1 GiB in-memory LRU. `local_store_size` (the on-disk tier) is populated by **writes**, not by read-back fetches, so raising it does not help reads. `read_buffer_size` is hard-bounded to a fraction of RAM and cannot be sized to cover a multi-GB archive.

## Expected

For read-heavy / archive workloads, fetched blocks should be promotable to the durable local block store so the working set stays warm across buffer cycles and restarts.

## Proposed

[...]

PR #1371, opened 2026-06-23
fix(block): bound the read-through cache and stop re-uploading fetched chunks

marmos91:
## What & why

Part of #1362. Deep investigation showed the issue's premise ("fetched blocks never persist") is largely already fixed on `develop` — read-through caching persists durably via `inlineFetchOrWait` → `local.Put`. The real bugs underneath are different:

**Bug 1 — read-through cache is unbounded on a read-only workload.** `FSStore.Put` (the read-fetch write entry; its only callers are the syncer's remote-fetch paths in `engine/fetch.go`) delegated straight to `StoreChunk`, which increments `diskUsed` and LRU-touches but **never calls `ensureSpace`**. `ensureSpace` — the sole eviction driver — only runs on the write/append path. So with no writes (archive/read workload), `diskUsed` grows past `maxDisk` forever. This is the "`du` shows tens of GB on a 1.9 GiB-limit host."

**Bug 2 — fetched chunks get re-uploaded and are temporarily un-evictable.** A chunk fetched from remote is verbatim remote content, yet `StoreChunk`'s `onChunkComplete` callback scheduled it for re-upload via the mirror loop, and `lruEvictOne` refuses to evict anything failing `IsSynced` until that mirror pass completes. Net: reading an N-byte archive re-PUTs N bytes back to S3 (read-amplification → write-amplification), and the first eviction finds zero synced candidates and stalls.

## Changes

- `FSStore.Put`: reserve capacity via `ensureSpace` for genuinely **new** chunks (idempotent re-Puts skip it, so they can't over-evict). `ensureSpace` already honors pin / eviction-disabled and is a no-op when `maxDisk == 0`, so the write path and conformance fixtures are unaffected.
- `Syncer.markFetchedSynced`: called after each read-fetch `local.Put`, it cancels the pending-upload entry and marks the hash synced immediately — killing the redundant re-upload and making the chunk instantly evictable.

## Tests (TDD — verified red before the fix)

- `TestPut_BoundsReadThroughCache` — read-cache `Put` past `maxDisk` evicts (fails red: `diskUsed=800 > maxDisk=600`).
- `TestPut_IdempotentRePutDoesNotEvict` — re-Put of an existing chunk reserves nothing, evicts nothing.
- `TestInlineFetch_MarksFetchedChunkSyncedAndCancelsReupload` — after an inline fetch the pending-upload entry is canceled (`unsyncedBytes == 0`) and the hash is marked synced.

`go test -race ./pkg/block/local/fs/ ./pkg/block/engine/` green; `go vet` clean.

## Follow-ups (same issue)

- PR2: fix stats so read-cached blocks show as local (`Blocks Local: 0` while `du` is huge).
- PR3: `dfsctl share warm` (async) to proactively materialize a share; pin + warm = permanent local mirror.
- Benchmarks quantifying cold-vs-warm throughput and the eliminated S3 re-upload.

Refs #1362

### DIFF
--- a/pkg/block/engine/fetch.go
+++ b/pkg/block/engine/fetch.go
@@ -218,6 +218,9 @@ func (m *Syncer) fetchBlock(ctx context.Context, payloadID string, blockIdx uint
 	if err := m.local.Put(ctx, fb.Hash, data); err != nil {
 		return nil, fmt.Errorf("store downloaded block %s locally: %w", storeKey, err)
 	}
+	// The bytes came from remote, so the chunk is already durable there:
+	// cancel its redundant re-upload and make it immediately evictable (#1362).
+	m.markFetchedSynced(ctx, fb.Hash)
 
 	return data, nil
 }
@@ -406,6 +409,9 @@ func (m *Syncer) inlineFetchOrWait(ctx context.Context, payloadID string, blockI
 		completionErr = fmt.Errorf("inline fetch: persist locally %s: %w", key, writeErr)
 		return nil, false, completionErr
 	}
+	// The bytes came from remote, so the chunk is already durable there:
+	// cancel its redundant re-upload and make it immediately evictable (#1362).
+	m.markFetchedSynced(ctx, fb.Hash)
 	completed = true
 	m.completeInFlight(key, result, nil)
 
--- a/pkg/block/engine/read_cache_bench_test.go
+++ b/pkg/block/engine/read_cache_bench_test.go
@@ -1,0 +1,360 @@
+package engine
+
+import (
+	"context"
+	"fmt"
+	"sync/atomic"
+	"testing"
+	"time"
+
+	"github.com/marmos91/dittofs/pkg/block"
+	localfs "github.com/marmos91/dittofs/pkg/block/local/fs"
+	memorylocal "github.com/marmos91/dittofs/pkg/block/local/memory"
+	"github.com/marmos91/dittofs/pkg/block/remote"
+	remotememory "github.com/marmos91/dittofs/pkg/block/remote/memory"
+	metastore "github.com/marmos91/dittofs/pkg/metadata/store/memory"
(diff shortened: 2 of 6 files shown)


marmos91:
Added a CI-runnable Go benchmark + invariant tests quantifying the #1362 read-through-cache fix in `pkg/block/engine/read_cache_bench_test.go`. They drive the real syncer fetch+persist path (`Syncer.fetchBlock`) against a mock remote (`latencyRemote`, wrapping `remotememory.Store`) that injects a fixed per-Get WAN latency and counts Get/Put.

**`BenchmarkReadThroughCache_ColdVsWarm`** (Apple M1 Max, `-benchtime=20x`, 1ms injected per-Get latency, 64 KiB chunk):

```
cold-10    1549658 ns/op    42.29 MB/s    345684 B/op    95 allocs/op
warm-10       8519 ns/op  7693.15 MB/s     65536 B/op     1 allocs/op
```

Cold (every read misses → fetches from the latency remote) is dominated by the WAN round-trip; warm (working set already local) is served from disk — ~180x faster MB/s — demonstrating the cache avoids the injected latency.

**Two invariant tests** (both pass, including `-race`):
- `TestReadThroughCache_NoReuploadOnReadOnlyWorkload` — reads an 8-chunk remote-only working set through the fetch path and asserts the mock remote's **Put count == 0** plus `unsyncedBytes == 0` / empty pending set (the precondition that makes the mirror loop a guaranteed no-op, so it's timing-independent). Pins the read→write de-amplification from `markFetchedSynced`.
- `TestReadThroughCache_BoundedByMaxDisk` — fetches 40 distinct chunks (8x the cap) into an `FSStore` with `maxDisk = 5 chunks` and asserts `diskUsed <= maxDisk` after every fetch and a full second pass. The headline #1362 invariant: the cache is no longer unbounded on a read-only workload (enforced by `FSStore.Put → ensureSpace`).

PR #1373, opened 2026-06-23
feat(block): async share warm + pin startup warning (#1362)

marmos91:
## Summary

Adds a per-share **warm** operation that proactively materializes a whole share's blocks onto the local disk tier, asynchronously (start a job, poll its status). It reuses the existing per-block remote-fetch primitive (`(*Syncer).fetchBlock`) — no parallel downloader — and is bounded by the per-share `ParallelDownloads` knob.

Also adds a one-line startup warning for a misconfiguration: a **pinned** share with a **bounded** local tier and a remote will fail reads with `ErrDiskFull` once its working set exceeds the tier.

Stacks logically on #1371 (the cache-bounding / marks-synced fix) — warm relies on `fetchBlock` being bounded and idempotent.

## Design

Async job model (no long-lived HTTP request):
- `POST /api/v1/shares/{name}/blockstore/warm` → `202` with `{ "job_id": ... }`, then
- `GET  /api/v1/shares/{name}/blockstore/warm/{job_id}` polls status.

`WarmAll` enumerates every payload (`local.ListFiles`) and its `FileBlock` rows (`ListFileBlocks`, the same surface as `populateBlockCounts`), skips blocks already physically local (`local.Has` via the existing `blockIsLocalFromRow` helper), and fetches the rest with an `errgroup` + semaphore sized to `ParallelDownloads`. A nil remote is an error; an `ErrDiskFull` fetch is terminal (cancels the group); context cancellation stops promptly.

The job registry is process-local, in-memory, mutex-guarded — one active job per share, deterministic monotonic IDs (`warm-<share>-<n>`). Each run executes on a **detached** context (from `context.Background()`), updated from the progress callback under the mutex. `RemoveShare` cancels any in-flight warm so it cannot outlive its block store.

Plumbed end to end mirroring `EvictBlockStore` at every layer (engine → shares service → runtime → REST → apiclient → dfsctl).

## Public API added

**Engine**
- `(*Syncer).WarmAll(ctx, progress func(done, total int64)) (WarmResult, error)`
- `(*Store).WarmAll(ctx, progress) (WarmResult, error)` (under the close-gate)
- `WarmResult { BlocksFetched, BytesFetched, BlocksAlreadyLocal int64 }`

**Runtime**
- `StartWarmBlockStore(ctx, shareName) (*shares.WarmJob, error)`
- `GetWarmStatus(jobID) (*shares.WarmJob, bool)`

**REST**
- `POST /api/v1/shares/{name}/blockstore/warm` → 202 `{ job_id, status }`
- `GET  /api/v1/shares/{name}/blockstore/warm/{job_id}` → `{ id, share, state, blocks_total, blocks_done, bytes_done, started_at, finished_at, error }`

**apiclient**
- `StartShareWarm(name) (jobID string, err error)`
- `GetShareWarm(name, jobID) (*WarmJobStatus, err)`

**dfsctl**
- `dfsctl share warm <name>` — start the job, print the id
- `dfsctl share warm --watch <name>` — poll ~1s, render `fetched X/Y blocks (Z)`, non-zero exit on failed/canceled

## Pin warning

In `createBlockStoreForShare`: when `RetentionPolicy == pin` AND the share has a remote AND the local tier has a non-zero max size, log one `Warn` suggesting raising `local_store_size` or dropping the pin. No behavior change.

## Tests

- engine `warm_test.go`: fetches missing / skips already-local / all-local afterward; nil-remote error; cancellation.
- shares `warm_test.go`: deterministic IDs, one-active-per-share, cancel-for-share, failure classification.
- handler `Warm` / `WarmStatus`: 202 + job id, error-detail-not-leaked, status found, unknown → 404.

`go vet`, `go build ./...`, `go test ./pkg/block/... ./pkg/controlplane/... ./pkg/apiclient/...`, `-race ./pkg/block/engine`, and `golangci-lint` all green. `docs/CLI.md` regenerated.

Refs #1362

### DIFF
--- a/cmd/dfsctl/commands/share/share.go
+++ b/cmd/dfsctl/commands/share/share.go
@@ -65,6 +65,7 @@ func init() {
 
 	Cmd.AddCommand(disableCmd)
 	Cmd.AddCommand(enableCmd)
+	Cmd.AddCommand(warmCmd)
 
 	Cmd.AddCommand(nfsConfigCmd) // per-share NFS adapter config (squash, netgroup, auth)
 }
--- a/cmd/dfsctl/commands/share/warm.go
+++ b/cmd/dfsctl/commands/share/warm.go
@@ -1,0 +1,111 @@
+package share
+
+import (
+	"fmt"
+	"os"
+	"time"
+
+	"github.com/marmos91/dittofs/cmd/dfsctl/cmdutil"
+	"github.com/marmos91/dittofs/internal/bytesize"
+	"github.com/marmos91/dittofs/internal/cli/output"
+	"github.com/marmos91/dittofs/pkg/apiclient"
+	"github.com/spf13/cobra"
+)
+
+// Terminal warm-job states (mirror shares.WarmState*).
+const (
+	warmStateDone     = "done"
+	warmStateFailed   = "failed"
+	warmStateCanceled = "canceled"
+)
+
+var warmCmd = &cobra.Command{
+	Use:   "warm <name>",
+	Short: "Warm a share's local block cache",
+	Long: `Proactively materialize a share's blocks onto the local disk tier.
+
+Starts an asynchronous job that downloads every remote block of the share into
+the local cache so subsequent reads are served locally. The command prints the
+job id and exits; use --watch to poll until the job completes.
+
+The share must have a remote tier configured. A pinned share with a bounded
+local tier may fail with a disk-full error if its working set exceeds the tier.
+
+Examples:
+  # Start a warm job and exit
+  dfsctl share warm /archive
+
+  # Start and follow progress until done
+  dfsctl share warm --watch /archive`,
+	Args: cobra.ExactArgs(1),
+	RunE: runShareWarm,
+}
+
+func init() {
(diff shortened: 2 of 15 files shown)


marmos91:
Ran a concurrency-focused review of the warm registry. **No HIGH issues** — data races (all WarmJob field access under the registry mutex incl. the progress callback), detached-context cancel lifecycle, cancellation classification (sampled before `cancel()`), the one-active-job-per-share check+insert atomicity, the errgroup semaphore release, `ErrDiskFull` termination, and the `RemoveShare`→`cancelForShare`→`Close` ordering all verified correct.

Two MED findings, both fixed in 758f1765:
- **Unbounded jobs map** — completed/failed/canceled jobs were never pruned. Now retains the most recent 32 terminal jobs (running jobs never evicted); added `TestWarmRegistry_RetireBoundsJobsMap`.
- **404 vs 400** — `StartWarm` returned a plain "not found" error so the handler mapped a missing share to 400. Now wraps `ErrShareNotFound` and the handler maps it to 404, reserving 400 for "share exists but has no block store".

`go vet`, `golangci-lint` (0 issues), and the shares + handlers + engine suites are green.

Issue #1228, opened 2026-06-16
SMB Windows-ACL + stable-handle fidelity hardening


marmos91:
Part of #1221 (ROI: #6 — makes the 'native, not re-exported' claim credible).

SMB Windows-ACL/permission fidelity + stable file handles are exactly where FUSE competitors fail. To win the enterprise wedge, DittoFS must demonstrably preserve them.

**Scope**
- Audit + harden Windows ACL mapping fidelity and stable/durable handle behavior across restarts.
- Interop tests vs real Windows/macOS SMB clients (extend existing smbtorture/WPTS harness).

**Verify** smbtorture ACL + durable-handle batteries; documented interop matrix.

marmos91:
Closed — all scope delivered across 4 PRs on develop:

- **#1376 (A)** stable/durable handle fidelity across restart: seed the SMB2 FileID counter past persisted durable handles so post-restart CREATEs can't collide with a reclaimable durable open (MS-SMB2 §3.3.5.9.7).
- **#1378 (B)** Windows-ACL owner/group fidelity: SET_INFO owner/group with a genuinely foreign-domain unmappable SID now returns STATUS_INVALID_OWNER/NONE_MAPPED instead of a silent no-op; well-known/BUILTIN/own-domain/current-owner re-sets accepted (smbtorture acls.SDFLAGSVSCHOWN green).
- **#1381 (C)** SACL preserve-on-write round-trip (audit ACEs), no migration; DACL path byte-for-byte unchanged.
- **#1379 (D)** documented interop matrix `docs/SMB_ACL_FIDELITY.md` (Works/Partial/Unsupported) + durable restart-survival test.

Verify gate met: smbtorture smb2.acls + durable-open/durable-v2/lease batteries green on memory + badger backends; durable handles persisted across restart. Remaining out-of-scope items (foreign AD/LDAP SID→UID mapping) tracked under the AD umbrella; SACL was implemented rather than deferred. A flaky netlogon test surfaced during this work was fixed separately (#1386, closed #1383).

PR #1376, opened 2026-06-24
fix(smb): stable FileID across restart (#1228 Fix A)

marmos91:
## What

Seed the SMB2 FileId persistent-half counter past any persisted durable handles at adapter startup, so a post-restart `CREATE` cannot re-mint a FileID still owned by a reclaimable durable open.

## Why (the bug)

`GenerateFileID` mints the FileId persistent half from `nextFileID`, an in-memory counter that resets to `1` every server start (`handler.go` `NewHandler`). Durable-handle records, however, persist FileIDs in the metadata store across restart.

So after a restart, fresh `CREATE`s re-issue persistent halves `1,2,3,…` — the same values already held by persisted durable handles. A **V1 durable reconnect (DHnC)** matches on the persistent half **alone** (no CreateGuid to disambiguate), so it could reconnect or atomically consume the **wrong file** (MS-SMB2 §3.3.5.9.7). Low probability, real correctness bug.

## Fix

New `Handler.SeedFileIDFromDurableHandles(ctx, store)`:
- Lists persisted durable handles, finds the max persistent half (little-endian `FileID[:8]`, matching `GenerateFileID`'s encoding).
- Bumps the counter to that max via an **upward-only CAS loop** (never lowers — a concurrent `CREATE` may already have advanced it). Counter holds the *last-issued* value, so storing `maxID` makes the next issued FileID `maxID+1`.
- No-ops on nil store, list error (logged), or empty store.

Called once in `pkg/adapter/smb/adapter.go` `Serve`, right after `DurableStore` is wired in (same gate as the scavenger). Reuses the existing `ListDurableHandles` — no new per-backend store method.

## Tests

`fileid_seed_test.go`: seeds past max persisted; never lowers when counter already ahead; nil/empty store no-op.

## Scope

Fix A of #1228. Fix B (SET_INFO unmapped-SID silent no-op) + interop-matrix doc to follow in a separate PR. SACL preserve-on-write tracked separately.

### DIFF
--- a/internal/adapter/smb/handlers/fileid_seed_test.go
+++ b/internal/adapter/smb/handlers/fileid_seed_test.go
@@ -66,6 +66,26 @@ func TestSeedFileIDFromDurableHandles_NeverLowers(t *testing.T) {
 	}
 }
 
+// TestSeedFileIDFromDurableHandles_MaxUint64 verifies the overflow guard: a
+// persisted FileID holding the maximum uint64 must NOT seed the counter to it,
+// which would make the next GenerateFileID wrap to 0.
+func TestSeedFileIDFromDurableHandles_MaxUint64(t *testing.T) {
+	store := newMockDurableStore()
+	_ = store.PutDurableHandle(context.Background(), durableWithFileID("max", ^uint64(0)))
+
+	h := NewHandler()
+	start := h.nextFileID.Load()
+	h.SeedFileIDFromDurableHandles(context.Background(), store)
+
+	if got := h.nextFileID.Load(); got != start {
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed Copilot review in aab0e969:
1. **Doc accuracy** — reworded the comment to drop the inaccurate `1,2,3,…` sequence (first issued persistent half is 2 due to `Add(1)` pre-increment).
2. **Overflow edge** — added an explicit guard: if a persisted handle holds the max `uint64` FileID, skip seeding (seeding to it would wrap the next `GenerateFileID` to 0) and log a warning. New test `TestSeedFileIDFromDurableHandles_MaxUint64` covers it.

Issue #1374, opened 2026-06-24
share warm + read-through inert when FileBlock.RefCount index is desynced from file metadata (audit shows ΣRefCount=0 vs Σlen(Blocks)=24427)


raffaelegiustini-create:
Follow-up to #1362. After updating to `develop` @ `549ee6d` (includes #1371/#1372/#1373), I tried to validate the read-through + `share warm` fixes on my archive deployment. The new code is present, but **completely inert on this instance** because its block-store refcount index is desynced from the file metadata. Reads still work, which masks the problem.

## Environment
- `develop` @ `549ee6d`, single-node sqlite control plane, badger metadata, fs local store, S3 (Cubbit) remote
- Share `/drive`: 11,451 files, 43.42 GiB logical, single S3 remote
- This instance has history: it survived the #1267/#1274 corrupt-upload era, recovery re-uploads, and several binary swaps. The desync is likely collateral from that, but the failure modes below are the actionable part.

## The smoking gun — `audit-refcounts`

```
$ dfsctl store block audit-refcounts drive
Total Files                  11451
Total Refs (Σ len(Blocks))   24427      <- file metadata references 24,427 blocks
Total RefCount (Σ RefCount)  0          <- block index refcounts are all zero
Delta (refs - refcount)      24427
```

File→block references exist, but every `FileBlock.RefCount` is 0. `dfsctl store block stats` correspondingly reports `Blocks Total/Local/Remote/Cached = 0`.

## Consequences (all observed)

1. **`dfsctl share warm /drive` is a silent no-op.** API returns `202 Accepted` with a job id, but nothing is fetched: `du` on the block dir is flat, `Local Disk Used` stays 0, no S3 GET traffic, no progress in logs. The warm walk enumerates the block index (empty) instead of the file metadata, so it finds nothing to materialize.

2. **`dfsctl share warm /drive --watch` → `Error: failed to poll warm job: warm job not found`.** The job id returned by the start call is not resolvable by the status endpoint — the job lifecycle/registry doesn't persist the job for polling. (Reproducible independent of the index issue.)

[...]

PR #1377, opened 2026-06-24
fix(block): warm + stats enumerate from metadata, not local store (#1374)

marmos91:
## Problem (#1374)

`dfsctl share warm` is a silent no-op and `dfsctl store block stats` reports `Blocks Total=0` (while `Local Disk Used` shows real bytes) on any share whose files have **rolled up**.

Root cause: both enumerate payloads via the **local block store** (`FSStore.ListFiles`), which only tracks payloads with a live append-log. Once a file rolls up, its CAS chunks and FileBlock metadata rows persist, but the payload drops out of `ListFiles` — so warm finds nothing to fetch and stats count nothing. This is independent of the refcount index the issue suspected.

## Fix

- Add `EnumeratePayloads(ctx, fn)` to the FileBlock store — streams every distinct payloadID from the **rollup-durable metadata** (`fb-file:` index / `file_blocks` table). Implemented for badger, memory, postgres, sqlite + conformance test.
- `WarmAll` and `populateBlockCounts`/`FileCount` now enumerate via `EnumeratePayloads` instead of `local.ListFiles()`. Per-block classification and fetch logic are unchanged. `GetStatsLite` stays O(1) and does not enumerate.

## Verification (live, SCW VM, badger + S3 remote)

After writing a 100 MiB file and rolling it up:
- **Before:** `Files=0, Blocks Total=0, Blocks Local=0` while disk held 100 MB; `warm` returned a job that fetched nothing.
- **After:** `Files=1, Blocks Total=100, Blocks Local=100`; `warm` enumerates all 100 blocks.

`go build ./...`, `go vet`, and `go test ./pkg/block/... ./pkg/metadata/...` (incl. the new `EnumeratePayloads` conformance across all 4 backends) pass.

Part of a larger #1374 cleanup (CAS-honest audit, warm job-id, legacy-refcount removal) landing in follow-up PRs.

### DIFF
--- a/cmd/dfs/commands/migrate_to_cas.go
+++ b/cmd/dfs/commands/migrate_to_cas.go
@@ -403,3 +403,6 @@ func (nopFileBlockStore) GetFileBlock(_ context.Context, _ string) (*block.FileB
 func (nopFileBlockStore) ListFileBlocks(_ context.Context, _ string) ([]*block.FileBlock, error) {
 	return nil, nil
 }
+func (nopFileBlockStore) EnumeratePayloads(_ context.Context, _ func(payloadID string) error) error {
+	return nil
+}
--- a/internal/bench/phase19_test.go
+++ b/internal/bench/phase19_test.go
@@ -24,6 +24,7 @@ import (
 	"context"
 	"math/rand"
 	"os"
+	"strings"
 	"sync"
 	"testing"
 	"time"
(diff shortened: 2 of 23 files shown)


marmos91:
Addressed the review + Copilot findings in bf386974:

- **SQLite single-connection deadlock** (Copilot): `EnumeratePayloads` no longer calls `fn` while a row cursor is open — sqlite/postgres now collect all ids, close the cursor, then call `fn` (so `fn → ListFileBlocks` on the `MaxOpenConns(1)` pool can't deadlock). Verified the sqlite conformance suite runs without hanging.
- **payloadID parsing**: payloadIDs contain slashes (`BuildPayloadID` = `share/path`), so memory/sqlite/postgres now split on the **last** `/`, not the first (badger was already correct via its `:`-separated index). Added a conformance case with a subdirectory payloadID that fails-before/passes-after.
- **warm fetch misalignment**: warm fetched via `fetchBlock(payloadID, blockIdx)`, which missed non-BlockSize-aligned FastCDC chunks; it now fetches by the `FileBlock` row in hand via the new `fetchResolvedBlock`. Added a warm regression test with a non-aligned chunk.

PR #1378, opened 2026-06-24
fix(smb): reject SET_INFO owner/group change with unmappable SID

marmos91:
**Fix B of #1228.**

## What
SMB `SET_INFO` with `SECURITY_INFORMATION` now returns an explicit NTSTATUS failure when the request asks to change the owner and/or group to a SID that cannot be mapped to a local UID/GID, instead of silently succeeding.

## Why
`ParseSecurityDescriptorWithOptions` returns a `nil` `ownerUID`/`ownerGID` for two distinct cases: the SD omitted that SID section, **and** the section was present but its SID was unmappable. `setSecurityInfo` only applied the section when the pointer was non-nil, so a present-but-unmappable owner/group SID was silently dropped and the handler returned `STATUS_SUCCESS`. Windows then believed the owner change had taken effect when nothing changed — a fidelity bug.

## Fix
- New `securityDescriptorHasOwner` helper decodes the SD header to report whether an Owner / Group SID section is *present* (independent of mappability), disambiguating the two `nil` cases.
- `setSecurityInfo`: when `OWNER`/`GROUP` is requested and the SID is present but unmappable, return `STATUS_INVALID_OWNER` (owner) / `STATUS_NONE_MAPPED` (group) rather than success.
- Added the two NTSTATUS constants (`StatusInvalidOwner` 0xC000005A, `StatusNoneMapped` 0xC0000073) + `String()` cases.

Narrow by design — DACL-only sets, absent owner/group sections, and mappable SIDs are all unchanged. Foreign AD/LDAP SID→UID mapping remains out of scope (tracked under #1231); an unmappable foreign SID correctly errors here.

## Tests
- New `set_info_security_unmapped_sid_test.go` (table-driven): owner-requested + unmappable → `STATUS_INVALID_OWNER`; owner-requested + mappable → success; DACL-only with no owner info → unaffected; group-requested + unmappable → `STATUS_NONE_MAPPED`; owner-requested but no owner SID in SD → success no-op.
- Updated the existing authz test helper `minimalDACLSDForOwner` to use a mappable owner SID (it relied on the old silent-no-op behavior).
- `go build ./...`, `go vet ./internal/adapter/smb/handlers/`, and `go test ./internal/adapter/smb/handlers/ -count=1` all pass.

### DIFF
--- a/internal/adapter/smb/handlers/security.go
+++ b/internal/adapter/smb/handlers/security.go
@@ -515,29 +515,28 @@ func ParseSecurityDescriptorWithOptions(data []byte, opts ParseSDOptions) (owner
 	return ownerUID, ownerGID, fileACL, nil
 }
 
-// securityDescriptorHasOwnerGroup reports whether a self-relative Security
-// Descriptor carries an Owner SID section (OffsetOwner != 0) and, separately,
-// a Group SID section (OffsetGroup != 0).
+// securityDescriptorOwnerGroupSIDs decodes the raw Owner and Group SIDs carried
+// by a self-relative Security Descriptor. A nil return for either means the SD
+// did not carry that SID section (OffsetOwner / OffsetGroup == 0).
 //
-// ParseSecurityDescriptorWithOptions returns a nil ownerUID/ownerGID for two
-// distinct cases that the SET_INFO Security path must tell apart: (1) the SD
-// did not carry that SID section at all, and (2) the section was present but
-// its SID could not be mapped to a local UID/GID. Callers use these flags
-// together with the parsed UID/GID to reject an unmappable owner/group change
-// instead of silently succeeding (refs #1228).
+// ParseSecurityDescriptorWithOptions collapses two cases into a nil
+// ownerUID/ownerGID that the SET_INFO Security path must tell apart: (1) the SD
+// omitted the SID section entirely, and (2) the section was present but its SID
+// could not be mapped to a local UID/GID. The caller pairs the raw SIDs from
+// here with the parsed UID/GID to reject a genuinely unmappable owner/group
+// change instead of silently succeeding (refs #1228), while still treating a
+// re-set of the file's existing owner/group SID as a no-op success.
 //
-// Presence is decided purely on "offset != 0" — it is NOT bounds-gated against
-// len(data). A non-zero offset that points out of range is still "present": it
-// is malformed input that ParseSecurityDescriptorWithOptions (which ignores an
-// out-of-range offset) leaves unmapped, so the caller's unmappable-SID gate
-// must still fire rather than mistaking it for an absent section and silently
-// succeeding (the very bug #1228 fixes). Header decode errors (this helper
-// parses only the fixed 20-byte SD header, never the Owner/Group SID bytes
-// themselves) yield "not present" so the existing StatusInvalidParameter
-// parse-error path continues to govern a truncated/garbled header.
-func securityDescriptorHasOwnerGroup(data []byte) (hasOwner, hasGroup bool) {
+// Presence is decided purely on "offset != 0"; it is NOT bounds-gated against
+// len(data). A non-zero offset is reported as present even when the SID bytes
+// fail to decode (returning a nil SID for that section) — that is malformed
+// input the unmappable-SID gate must still reject rather than mistaking for an
+// absent section. The fixed 20-byte SD header is the only part parsed here; a
+// truncated/garbled header yields (nil, nil) so the caller's existing
+// StatusInvalidParameter parse-error path continues to govern it.
+func securityDescriptorOwnerGroupSIDs(data []byte) (ownerSID, groupSID *sid.SID, hasOwner, hasGroup bool) {
 	if len(data) < sdHeaderSize {
-		return false, false
+		return nil, nil, false, false
 	}
 	r := smbenc.NewReader(data)
 	r.Skip(2) // Revision(1) + Sbz1(1)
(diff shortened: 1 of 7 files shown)


marmos91:
Addressed the 3 review comments in fedf2cd7:

1. **(substantive)** Presence detection is no longer bounds-gated. `securityDescriptorHasOwnerGroup` now returns presence purely on `offset != 0`. A non-zero-but-out-of-range owner/group offset is malformed-but-present: `ParseSecurityDescriptorWithOptions` leaves it unmapped, so the unmappable-SID gate fires (`STATUS_INVALID_OWNER` / `STATUS_NONE_MAPPED`) instead of falling through as a silent success — closing the reopened #1228 hole. Header decode errors still yield "not present" so `STATUS_INVALID_PARAMETER` governs a truncated header.
2. **(cosmetic)** Clarified the doc comment to state the helper parses only the fixed 20-byte SD header, never the Owner/Group SID bytes.
3. **(cosmetic)** Renamed `securityDescriptorHasOwner` → `securityDescriptorHasOwnerGroup` and updated the call site.

Added `TestSetInfo_Security_OwnerOffsetOutOfRange`: a SD with a non-zero OffsetOwner pointing past the buffer + OWNER requested now asserts `STATUS_INVALID_OWNER` (not a silent success).

`go build ./...`, `go vet ./internal/adapter/smb/handlers/`, and `go test ./internal/adapter/smb/handlers/ -count=1` all pass.

marmos91:
## SDFLAGSVSCHOWN conformance fix (297f67c9)

**Root cause.** `smb2.acls.SDFLAGSVSCHOWN` ("SD flags vs chown") reads a file's SD and then re-SETs the *same* owner SID, expecting `NT_STATUS_OK`. The owner SID the server emits on the read path is `mapper.UserSID(file.UID)`. For a root-owned file (UID 0) that is `BUILTIN\Administrators` (`S-1-5-32-544`) — a SID that **does not reverse** to a UID, because `UIDFromSID` only resolves S-1-5-21 domain SIDs. This asymmetry is intentional and pre-existing (`TestUserSIDRIDMapping` maps UID 0 → Administrators, while `TestUIDFromSIDRoundTrip` deliberately omits UID 0). The new #1228 gate saw `ownerUID == nil` and rejected the no-op re-set with `STATUS_INVALID_OWNER`.

**Fix (narrowed boundary).** A present-but-unmappable owner/group SID is now rejected **only when it differs from the SID the server currently mints for the file's owner/group** (`UserSID(curUID)` / `GroupSID(curGID)`). Re-setting the identical SID — the SDFLAGSVSCHOWN case — is a legitimate no-op and returns `NT_STATUS_OK`. A genuinely foreign/unresolvable domain SID still errors (`STATUS_INVALID_OWNER` / `STATUS_NONE_MAPPED`), so the original silent-no-op bug stays fixed. Foreign AD SID→UID mapping remains out of scope (#1231).

The presence helper `securityDescriptorHasOwnerGroup` was replaced by `securityDescriptorOwnerGroupSIDs`, which returns the raw decoded Owner/Group SIDs (plus presence) so the gate can compare against the current owner's minted SID. I did **not** change `UIDFromSID`/`mapper.go` — the Administrators non-reversal is a deliberate, separately-tested contract.

**Tests.** Added regression cases: owner re-set to the current owner SID (Administrators on a root-owned file) and group re-set to the current group SID both return success; foreign-SID and out-of-range-offset cases still reject. `go build ./...`, `go vet ./internal/adapter/smb/handlers/`, and `go test ./internal/adapter/smb/handlers/ ./pkg/auth/sid/ -count=1` all pass.

marmos91:
## SDFLAGSVSCHOWN: read the real test, corrected the boundary (f43fec7d)

I pulled the actual Samba test (`source4/torture/smb2/acls.c`, `test_sd_flags_vs_chown`, registered as `SDFLAGSVSCHOWN` at line 3307). The relevant section (~lines 1760-1774):

```c
owner_sd->owner_sid = &world_sid;        // global_sid_World == S-1-1-0
set.set_secdesc.in.secinfo_flags = SECINFO_OWNER;
status = smb2_setinfo_file(tree, &set);
CHECK_STATUS(status, NT_STATUS_OK);      // <-- expects OK

owner_sd->owner_sid = owner_sid;         // back to the original owner
set.set_secdesc.in.secinfo_flags = SECINFO_OWNER;
status = smb2_setinfo_file(tree, &set);
CHECK_STATUS(status, NT_STATUS_OK);      // <-- expects OK
```

**What SDFLAGSVSCHOWN actually does:** it chowns the directory's owner to the **World SID `S-1-1-0`** (and then back to the original owner), each via `SECINFO_OWNER`, asserting `NT_STATUS_OK`. My previous "accept only if the SID equals the current owner's minted SID" heuristic could not cover this — World is not the current owner, so it was still rejected with `STATUS_INVALID_OWNER`. That's why the failure persisted.

**Why Samba returns OK:** `unpack_nt_owners` maps the owner SID through `sid_to_uid`; well-known SIDs like World resolve via idmap, so the chown proceeds. Samba only returns `NT_STATUS_INVALID_OWNER` for a SID that is genuinely unresolvable (and `force_unknown_acl_user` is off).

**Corrected boundary:** reject a present-but-unmappable owner/group SID **only when it is a foreign domain account** — an `S-1-5-21-{a}-{b}-{c}-{RID}` SID whose domain differs from this machine's (the AD/LDAP case, #1231). Well-known SIDs (World, `BUILTIN\*`, `NT AUTHORITY\*`, `CREATOR\*`), our own machine-domain SIDs, and current-owner re-sets are all accepted as no-op successes. Implemented as `SIDMapper.IsForeignDomainSID`.

This **keeps the genuine #1228 fix** (a foreign-principal owner-set still errors rather than silently succeeding) while restoring `acls.SDFLAGSVSCHOWN` to `NT_STATUS_OK`. Fix B remains viable — the premise needed narrowing, not dropping.

**Tests:** unit case `owner_set_to_world_sid_succeeds` plus a dedicated `TestSetInfo_Security_SDFLAGSVSCHOWN_ChownRoundTrip` mirroring the exact World→original round-trip; `TestIsForeignDomainSID` in `pkg/auth/sid`; foreign-domain SID still rejected. `go build ./...`, `go vet`, and `go test ./internal/adapter/smb/handlers/ ./pkg/auth/sid/ -count=1` all pass.

Issue #236, opened 2026-02-28
Implement LSA named pipe (lsarpc) for SID-to-name resolution


marmos91:
## Summary

Windows Security tab shows raw SIDs (e.g., `S-1-5-21-2460966292-2413871697-2891260931-3000`) instead of friendly names (e.g., `wpts-admin`) because there is no LSA (Local Security Authority) service to resolve SID-to-name mappings.

## Current Behavior

When right-clicking a file → Properties → Security tab on a DittoFS SMB share, Windows displays:
- `S-1-5-21-...-3000` instead of the owner username
- `S-1-5-21-...-3001` instead of the group name
- `S-1-5-32-544` instead of "Administrators"

## Expected Behavior

Friendly names displayed for all principals, similar to Samba.

## Implementation

Implement the `\pipe\lsarpc` named pipe with enough LSARPC operations for Windows to resolve SIDs:
- `LsarOpenPolicy2` — open policy handle
- `LsarLookupSids2` / `LsarLookupSids3` — resolve SIDs to names
- `LsarLookupNames3` — resolve names to SIDs (optional)
- `LsarClose` — close policy handle

Reference: [MS-LSAR] and Samba source code.

## Context

Discovered during Phase 32 Windows integration testing (Test 8: Explorer file operations). The SMB SD synthesis from Phase 31 works correctly — this is a UX improvement only.

PR #1379, opened 2026-06-24
docs(smb): ACL/SD fidelity interop matrix + durable restart test

marmos91:
**Fix D of #1228** — the documented interop matrix + a durable-handle restart-survival test.

## What

- **`docs/SMB_ACL_FIDELITY.md`** — a precise, evidence-based **Works / Partial / Unsupported** matrix for the SMB Windows-ACL + Security-Descriptor path, derived by reading the actual code (not the spec). Covers:
  - Owner/Group SID read+write (machine-SID + `uid*2+1000` / `gid*2+1001` RID derivation; UID0→Administrators).
  - DACL ACE count / type (allow+deny) / mask; ACE/DACL caps.
  - Inheritance flags OI/CI/NP/IO + the `INHERITED_ACE` 0x80↔0x10 remap; inherit-time + propagation.
  - SD control flags: `SE_DACL_PROTECTED`, `SE_DACL_AUTO_INHERITED` (+ canonicalization toggle), null DACL.
  - **SACL** — unsupported: empty stub on read, dropped on write; AUDIT stored-only, ALARM dropped.
  - `GENERIC_*` mask expansion on write and at inherit time.
  - Foreign AD/LDAP SID mapping (out of scope, #1231) and unmappable-SID-on-write behavior.
  - SET_INFO Security access gating table.
  - "Client-tested" (Win11 Explorer Security tab, AD/Kerberos SMB, smbtorture; macOS Finder noted untested) and "Known limitations".
  - Cross-references #1228, #1231, #236.
  - Linked from `README.md` and `CLAUDE.md`.

- **`internal/adapter/smb/handlers/durable_restart_survival_test.go`** — restart-survival tests for durable handles (match `durable_scavenger_test.go` style via `newMockDurableStore()`):
  - `TestDurableHandleSurvivesSimulatedRestart`: persists a durable handle, simulates restart with a fresh `Handler` (FileID counter resets), asserts it stays reconnect-resolvable by FileID (persistent-half-keyed `GetDurableHandleByFileID`, the lookup `processV*Reconnect` use), and demonstrates the un-seeded persistent-half collision hazard.
  - `TestDurableHandleNoCollisionAfterReseed`: the closed-loop counterpart — persists a handle with a high FileID, runs `Handler.SeedFileIDFromDurableHandles` (Fix A / #1376, now merged), then asserts the counter seeds to the max persisted persistent half, a freshly minted FileID mints `max+1` (no collision), and the persisted handle stays reconnect-resolvable.

## Why

#1228 explicitly demands a documented interop matrix for SMB Windows-ACL / stable-handle fidelity; it did not exist. There was also no test asserting durable-handle FileID identity survives a restart.

## Verify

`go build ./...` and `go test ./internal/adapter/smb/handlers/ ./pkg/adapter/smb/ -count=1` both pass.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -18,6 +18,7 @@ Two binaries: `dfs` (server, `cmd/dfs/`) and `dfsctl` (REST client, `cmd/dfsctl/
 - `docs/CONFIGURATION.md` — config file + env vars (`DITTOFS_*`)
 - `docs/CLI.md` — full `dfs` / `dfsctl` reference (**generated** — see below)
 - `docs/NFS.md`, `docs/SMB.md` — protocol details
+- `docs/SMB_ACL_FIDELITY.md` — SMB Windows-ACL/SD interop matrix (Works/Partial/Unsupported)
 - `docs/IMPLEMENTING_STORES.md` — metadata/block store contracts
 - `docs/DEBUGGING.md` — SMB/NFS pcap-diff interop playbook
 - `docs/FAQ.md` — known limitations (ETXTBSY, POSIX gaps, single-node)
--- a/README.md
+++ b/README.md
@@ -246,6 +246,7 @@ an isolated local storage directory and its own caching tiers. See
 - [NFS](docs/NFS.md) — NFSv3/v4.0/v4.1 status, mounting, internals
 - [SMB](docs/SMB.md) — SMB2/3 status, encryption, signing, leases, durable handles
 - [ACLs](docs/ACLS.md) — the cross-protocol ACL model shared by NFS and SMB
+- [SMB ACL Fidelity](docs/SMB_ACL_FIDELITY.md) — what the SMB Security Descriptor path round-trips (Works/Partial/Unsupported)
 - [Encryption](docs/ENCRYPTION.md) — client-side per-remote envelope encryption
 - [Snapshots](docs/SNAPSHOTS.md) — reference-based share snapshots and restore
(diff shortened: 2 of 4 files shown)


marmos91:
Rebased onto latest develop. Fix A (#1376, `Handler.SeedFileIDFromDurableHandles`) is merged, so `TestDurableHandleNoCollisionAfterReseed` is now fully active (skip removed) with real assertions: after the startup reseed, the FileID counter seeds to the max persisted persistent half, a freshly minted FileID mints max+1 (no collision), and the persisted handle stays reconnect-resolvable. Both durable tests pass; build + `go test ./internal/adapter/smb/handlers/ ./pkg/adapter/smb/` green.

marmos91:
Addressed both Copilot review notes on `durable_restart_survival_test.go`:

1. **`[8]byte(id[:8])` conversion** — false positive. Slice-to-array conversion is valid Go since 1.20 (repo is on Go 1.25) and compiles cleanly. Left as-is, added a comment noting it's intentional.
2. **Volatile-half storage shape** — valid; fixed. Both tests now persist `FileID` with the volatile half (bytes 8-15) zeroed (matching `buildPersistedDurableHandle`) and keep the full original value in `OriginalFileID`. The seed/collision logic keys on the persistent half (bytes 0-7), so all assertions still hold.

`go build ./...` and `go test ./internal/adapter/smb/handlers/ -count=1` both pass.

PR #1380, opened 2026-06-24
fix(block): slash-free warm job id + warn on empty enumeration (#1374)

marmos91:
## Problem (#1374)

`dfsctl share warm <share> --watch` immediately fails with `Error: failed to poll warm job: warm job not found` (HTTP 404). The warm job id embeds the share name including its slash — for `/export` it is `warm-/export-1` — and the slash breaks the poll route `.../warm/{job_id}`.

Separately, a warm run that enumerates nothing returns `202` + a job that fetches nothing, with no signal to the operator.

## Fix

- **Opaque job id:** `warm-%d` (monotonic counter) instead of `warm-{share}-{n}`. Per-share dedup still uses the registry's `active` map; nothing parses the id to recover the share. Poll route now resolves.
- **Empty-enumeration warning:** when a warm run finds 0 enumerable blocks but the share reports nonzero stored bytes, set `WarmJob.Warning` (surfaced in the status response, apiclient, and `--watch` output) and log at Warn. Used-bytes is read from the O(1) `GetStatsLite().LocalDiskUsed` at the shares layer.

Tests: job ids are slash-free and resolvable by id (incl. a `/`-named share); warning fires on empty enumeration of a non-empty share and stays quiet otherwise.

Companion to #1377 (warm/stats metadata enumeration). Part of the #1374 cleanup.

### DIFF
--- a/cmd/dfsctl/commands/share/warm.go
+++ b/cmd/dfsctl/commands/share/warm.go
@@ -97,6 +97,9 @@ func watchWarm(client *apiclient.Client, shareName, jobID string) error {
 			fmt.Printf("\rfetched %d/%d blocks (%s) — done\n",
 				status.BlocksDone, status.BlocksTotal,
 				bytesize.ByteSize(status.BytesDone).String())
+			if status.Warning != "" {
+				fmt.Fprintf(os.Stderr, "warning: %s\n", status.Warning)
+			}
 			return nil
 		case warmStateFailed:
 			fmt.Println()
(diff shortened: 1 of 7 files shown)


marmos91:
Addressed the Copilot nit in 4b6e1cef: the empty-enumeration warning + log field (`local_disk_used_bytes`) now make clear the byte count is local-disk-tier usage, not overall share size.

Issue #583, opened 2026-05-22
Crash-consistency: badger metadata writes lost on kill -9 (FileBlock rows unsynced)


marmos91:
## Symptom

After `kill -9` of dfs, restart, files written before the crash read as zero-filled (with correct sizes preserved). CAS chunks survive on disk but reads return zeros from `engine.go:1072 clear(dest)`.

Surfaced during PR #566 (Direction-1 rollup) live smoke. Pre-existing, not introduced by #566.

## Reproduction

1. dfs running, share mounted via NFSv3.
2. `echo -n "test" > /mnt/foo.txt` ; sleep 1 (let rollup commit).
3. Verify sha256 matches via NFS read.
4. `kill -9 \`cat ~/.local/state/dittofs/dittofs.pid\``
5. Restart dfs.
6. Read foo.txt: zeros.

Graceful shutdown (`./dfs stop` → SIGTERM → badger.Close()) does NOT exhibit the bug — the crash window is the kill -9 path.

## Root cause

`pkg/metadata/store/badger/store.go:192` uses `badger.DefaultOptions(...)` which has `SyncWrites=false`. Each `badger.DB.Update` returns when the value is in the memtable + WAL buffer — NOT after the WAL is fsynced. kill -9 between flush boundaries loses every metadata write since the last sync.

Critical writes affected:
- `pkg/blockstore/engine/engine.go:180` per-chunk FileBlock rows from rollup
- `pkg/blockstore/engine/engine.go:238` per-chunk emitter rows (in-memory backend path)
- `coordinator.PersistFileBlocks` manifest + ObjectID
- `SetRollupOffset` rows
- All FileAttr / directory metadata mutations

`fbs.Put(ctx, fb)` at `engine.go:180` also **ignores the error** (`_ = fbs.Put`). Any badger-side commit failure is silently swallowed.

## Fix direction

Option 1 — **enable global `SyncWrites=true`** in the badger store config. Simplest. Performance cost: every metadata operation fsyncs. Empirically observed ~10x slowdown in similar projects; needs measurement.

Option 2 — **explicit `db.Sync()` at rollup-commit boundary**, before `advanceRollupOffset`. Amortizes one fsync per rollup pass instead of per write. Requires plumbing a `Sync()` surface on the metadata store interface.

Option 3 — **per-transaction sync for D-12 critical writes**. Most surgical; targets ObjectIDPersister + SetRollupOffset paths only. Other metadata writes (path lookups, dir entries) stay async.

[...]

PR #1382, opened 2026-06-24
fix(block): make audit-refcounts CAS-honest (#1374)

marmos91:
## Problem (#1374)

`dfsctl store block audit-refcounts <share>` reports a huge, alarming `Delta` (e.g. `delta=24427`) that looks like index corruption but is a **false positive**. It computed `TotalRefCount = Σ GetByHash(hash).RefCount`. In the CAS model `FileBlock.RefCount` is vestigial (reclamation is mark-sweep GC; eviction keys off `IsSynced`; delete/truncate reap by block ID — none consult RefCount), and `GetByHash` is remote-gated so it never resolves pending CAS blocks. So `Σ RefCount` is structurally **always 0** → `Delta = TotalRefs` always. There is no desync to repair.

## Fix

Redefine the audit to the **meaningful CAS invariant**: every `FileAttr.Blocks` manifest reference must have a backing `FileBlock` row. A manifest ref with no row is a genuine dangling reference (the silent-data-loss class, cf. #583/#789).

- New fields: `Manifest Refs (Σ len(Blocks))`, `Backed by FileBlock row`, `Dangling refs (missing row)`. `Delta == DanglingRefs`; invariant is `DanglingRefs == 0`.
- Dropped `TotalRefCount` and all `GetByHash`/RefCount use from the audit path (the store methods remain for now; a later phase removes them).
- Non-zero exit still fires on `DanglingRefs > 0`, so `audit-refcounts <share> || alert` works on real corruption. `docs/CLI.md` regenerated.

Tests: healthy share → `DanglingRefs==0` (exit 0); deleting a backing FileBlock row while leaving the manifest entry → `DanglingRefs==1` (non-zero exit).

Part of the #1374 cleanup (companion to #1377 enumeration, #1380 job-id). RefCount/dedup machinery removal lands in a follow-up phase.

### DIFF
--- a/cmd/dfsctl/commands/store/block/audit.go
+++ b/cmd/dfsctl/commands/store/block/audit.go
@@ -11,22 +11,31 @@ import (
 	"github.com/marmos91/dittofs/pkg/apiclient"
 )
 
-// auditRefcountsCmd verifies the refcount invariant (∑ FileBlock.RefCount
-// == ∑ len(FileAttr.Blocks)) for the named share. Persists last-run
-// summary at <localStore>/audit-state/last-inv02.json.
+// auditRefcountsCmd verifies the CAS manifest-consistency invariant for
+// the named share: every block referenced by a file's manifest
+// (FileAttr.Blocks) must have a backing FileBlock row in the metadata
+// store. Persists last-run summary at <localStore>/audit-state/last-inv02.json.
 var auditRefcountsCmd = &cobra.Command{
 	Use:   "audit-refcounts <share>",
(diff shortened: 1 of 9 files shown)


marmos91:
Addressed the Copilot finding in cf682d87: the audit now counts a manifest ref as *backed* only when a FileBlock row exists at that offset **and its hash matches** the ref's hash — a row with a mismatched non-zero hash is real corruption and is now counted as dangling. Added `TestAuditRefcounts_DetectsHashMismatch`.

PR #1384, opened 2026-06-24
feat(nfs): NFSv4.2 CLONE (RFC 7862 op 71) — O(1) reflink over the dedup block store

### LINKED ISSUE #1302: feat(nfs): NFSv4.2 CLONE (reflink) over dedup block refs
**NFSv4.2 (RFC 7862) — highest-ROI v4.2 op for DittoFS.** Tracks `CLONE` (and the supporting `FATTR4_CLONE_BLKSIZE`).

## Why it fits DittoFS
The block store is content-addressed with dedup. A clone is a **pure metadata operation**: the target file references the same block(s) and bumps refcounts — near-instant, zero data movement, even on S3-backed shares. `cp --reflink`, macOS Finder duplicate, and VM-image provisioning all become O(1).

## Cross-protocol unification
The SMB adapter already implements server-side copy (`FSCTL_SRV_COPYCHUNK`, `FSCTL_DUPLICATE_EXTENTS_TO_FILE`) over the same engine. NFS `CLONE` should reuse that block-clone/refcount path so both protocols share one implementation and one test surface.

## Scope
- `OP_CLONE` (op 50) handler: src stateid + dst stateid + src/dst offsets + count; range or whole-file.
- Block-layer `Clone(srcRange) -> dst refs` contract (likely already present via the SMB duplicate-extents path) added to the store contract + `blockstoretest` conformance.
- `FATTR4_CLONE_BLKSIZE` GETATTR support.
- e2e: `cp --reflink=always` round-trip; verify shared blocks (dedup gauge unchanged), then COW on write to one side.

Lineage: NFSv4.2 xattr landed in #1285. Sibling sparse-files cluster: SEEK/READ_PLUS/DEALLOCATE.

marmos91:
## What

Implements NFSv4.2 `OP_CLONE` (RFC 7862 Section 15.13) end-to-end. Closes #1302.

A whole-file clone is a **pure metadata reflink**: the destination references the same content-addressed blocks as the source and the CAS RefCount is bumped once per unique hash — O(1), zero data movement, even on S3. Copy-on-write is intrinsic to the content-addressed store (a later WRITE to either file produces new CAS blocks under a new hash, leaving the other side untouched). This enables `cp --reflink=always`, Finder duplicate, and VM-image provisioning as O(1) operations.

## Why / RFC reference

`CLONE` is the strongest NFSv4.2 feature fit for DittoFS's dedup block store. Per RFC 7862 Section 15.13:

- `CLONE4args` = `cl_src_stateid`, `cl_dst_stateid`, `cl_src_offset`, `cl_dst_offset`, `cl_count`; `CLONE4res` is status-only. Wire order verified against Linux knfsd `nfsd4_decode_clone`.
- `SAVED_FH` = source, `CURRENT_FH` = destination (the client SAVEFHs the source first, like COPY).
- `cl_count == 0` means clone from `cl_src_offset` to the end of the source file.
- `FATTR4_CLONE_BLKSIZE` = attribute **77** (uint32), reported as **1** (byte-granular alignment).

### Correction vs the issue body

The issue says "op 50" — that is **wrong** (op 50 is LAYOUTGET). CLONE is **op 71**, confirmed against the RFC 7862 operation table and knfsd. `FATTR4_CLONE_BLKSIZE` is attribute **77**, not 71.

## Reference-implementation notes

Studied Linux knfsd (`nfsd4_clone` / `nfsd4_decode_clone`), nfs-ganesha, and the Linux NFS client (`nfs42_proc_clone`, which is what `cp --reflink=always` issues). DittoFS matches the wire format, the SAVED_FH/CURRENT_FH convention, the `count==0` whole-file semantic, and the established error codes (`NFS4ERR_ISDIR`, `NFS4ERR_WRONG_TYPE`, `NFS4ERR_INVAL`, `NFS4ERR_NOFILEHANDLE`, `NFS4ERR_NOTSUPP`).

## Cross-protocol reuse

The O(1) reflink primitive is `engine.CopyPayload` (bumps RefCount per unique hash, atomic in one metadata txn) — the same content-addressed refcount path the SMB server-side-copy IOCTLs build on. NFS CLONE wraps it in `common.CloneWholeFile`, a single cross-protocol clone primitive that SMB `FSCTL_DUPLICATE_EXTENTS_TO_FILE` can adopt later without a second engine. No second clone engine was written.

## Scope

- **Whole-file** clones (src/dst offset 0, count 0 or source size — exactly what `cp --reflink` issues) are served as true O(1) reflinks.
- **Sub-range / offset** clones return `NFS4ERR_NOTSUPP` (RFC 7862 15.13 permits a server to decline clone requests it cannot satisfy). DittoFS uses content-defined FastCDC chunking, so an honest partial-range reflink would need block-boundary splicing the chunker does not give for free; rather than ship a half-correct sub-range path, this declines it cleanly.
- **Cross-share** clones return `NFS4ERR_INVAL` (per-share dedup; NFSv4 has no `NFS4ERR_XDEV`).

## Changes

- `OP_CLONE = 71` + `FATTR4_CLONE_BLKSIZE = 77` constants; registered in the v4.2 dispatch table (auto-gated to minorversion 2 — returns `NFS4ERR_NOTSUPP` on v4.0/v4.1).
- `CLONE4args` decode + status-only result encode.
- `handleClone` (protocol-only): decode, SAVED_FH/CURRENT_FH resolution, src-READ / dst-WRITE stateid validation, type and same-share checks, delegate to `common.CloneWholeFile`.
- `common.CloneWholeFile`: whole-file O(1) reflink, atomic (RefCount bumps bound to the dst PutFile txn via `metadata.WithTx`).
- `FATTR4_CLONE_BLKSIZE` GETATTR support (pseudo-fs + real-file encoders, supported-attrs bitmap).
- `docs/NFS.md` updated.

## Test evidence

- Unit: `handleClone` validation (BADXDR / NOFILEHANDLE / ROFS), `decodeCloneArgs` (wire order, overflow guards, whole-file), `cloneErr` envelope.
- Block/refcount: `CloneWholeFile` O(1) — asserts one IncrementRefCount per unique source hash, dst inherits the source BlockRef list, dst.Size matches; plus a rollback-atomicity test (mid-loop increment failure rolls back the dst PutFile and skips cache invalidation).
- Attr: `FATTR4_CLONE_BLKSIZE` advertised in supported-attrs and GETATTR-encodes to 1.
- E2E (`test/e2e/nfsv42_clone_test.go`, build-tag `e2e`): `cp --reflink=always` over a `vers=4.2` mount proves the reflink round-trip (content-identical) and copy-on-write divergence after writing one side.
- `go test -race`, `go vet`, and `golangci-lint run` clean on all changed packages.

## Review

Ran code-simplifier and code-reviewer. The reviewer caught (and this PR fixes) a dead `ErrCloneCrossShare` branch and a TOCTOU double-fetch of the source — `CloneWholeFile` now takes the caller's already-fetched source blocks/size, removing the redundant metadata read and the race window. An earlier iteration's sub-range path was removed after I found that `engine.Store.WriteAt` returns `currentBlocks` unchanged (projection happens at Flush), which would have torn the destination.

### DIFF
--- a/docs/NFS.md
+++ b/docs/NFS.md
@@ -571,20 +571,30 @@ NFSv4.1 extends v4.0 with session-based operation, backchannel callbacks, and ad
 ### NFSv4.2 Status
 
 NFSv4.2 (RFC 7862) is a collection of independent, individually-optional features. DittoFS
-implements the named **extended attributes** set (RFC 8276); the other v4.2 operations
-return `NFS4ERR_NOTSUPP`. A client negotiates v4.2 with `mount -o vers=4.2`.
+implements the named **extended attributes** set (RFC 8276), the **sparse-files** cluster,
+and **CLONE**; the remaining v4.2 operations return `NFS4ERR_NOTSUPP`. A client negotiates
+v4.2 with `mount -o vers=4.2`.
 
-The features below are **planned** and tracked, ordered by fit with DittoFS's
-content-addressed/dedup block store. `CLONE` is the strongest fit — a reflink is a pure
-metadata ref over the same dedup blocks, reusing the SMB server-side-copy engine; it is
-**not yet implemented**. `SEEK`, `READ_PLUS`, `DEALLOCATE`, and `ALLOCATE` form a
-sparse-files cluster that shares one hole-tracking foundation and **are implemented**.
+`CLONE` (RFC 7862 Section 15.13, operation 71) is the strongest fit for DittoFS's
+content-addressed/dedup block store: a whole-file reflink is a pure metadata ref over the
+same dedup blocks, bumping the CAS RefCount per unique hash with no data movement — O(1)
+even on S3. It reuses the same `engine.CopyPayload` refcount primitive the SMB
+server-side-copy IOCTLs build on (`common.CloneRange` — one clone path, both protocols).
+Copy-on-write is intrinsic: a later WRITE to either file produces new CAS blocks under a
+new hash, leaving the other side untouched. `CLONE` uses `SAVED_FH` as the source and
+`CURRENT_FH` as the destination (the client `SAVEFH`s the source first, like `COPY`).
+`FATTR4_CLONE_BLKSIZE` is reported as 1 (byte-granular alignment): DittoFS clones by
+splicing the content-addressed block list, so no offset/count alignment constraint applies.
+Source and destination must live in the same share (per-share dedup); a cross-share clone
+returns `NFS4ERR_INVAL`.
 
-The hole map is derived directly from the file's content-addressed block list (a byte
-range covered by no block ref is a hole that reads as zeros) — there is no separate hole
-bitmap. The same `pkg/block` hole-map query backs all four ops so SEEK / READ_PLUS /
-DEALLOCATE report consistent boundaries. A file whose blocks cover its whole extent
-yields a single data segment (the dense-file fallback).
+`SEEK`, `READ_PLUS`, `DEALLOCATE`, and `ALLOCATE` form a sparse-files cluster that shares
+one hole-tracking foundation and **are implemented**. The hole map is derived directly from
+the file's content-addressed block list (a byte range covered by no block ref is a hole
+that reads as zeros) — there is no separate hole bitmap. The same `pkg/block` hole-map
+query backs all four ops so SEEK / READ_PLUS / DEALLOCATE report consistent boundaries. A
+file whose blocks cover its whole extent yields a single data segment (the dense-file
+fallback).
 
 | Operation | Status | Notes |
 |-----------|--------|-------|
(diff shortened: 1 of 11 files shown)


marmos91:
Addressed both Copilot findings in 2c8e05e2:

- **Permission enforcement**: special stateids pass `ValidateStateid` without an OPEN, so they carried no POSIX/ACL or read-only-export check. Added `metaSvc.CheckPermissions(PermissionRead)` on the source and `CheckPermissions(PermissionWrite)` on the destination (the latter also rejects a write to a read-only export) — mirroring how ALLOCATE/DEALLOCATE re-validate via the Service.
- **Self-clone**: short-circuit a same-file/same-payload clone as a no-op before touching the block store, so `CopyPayload` is never invoked with `srcPayloadID == dstPayloadID` (which would inflate the shared RefCount). Guarded in both the handler (same handle) and `CloneWholeFile` (same payload, since the helper is the shared cross-protocol primitive), with a self-clone test.

PR #1386, opened 2026-06-24
fix(netlogon): don't fail logon on local channel-teardown race under concurrent reload

### LINKED ISSUE #1383: Flaky test: TestReloadCredential_ConcurrentLogons (internal/auth/netlogon)
## Summary
`TestReloadCredential_ConcurrentLogons` in `internal/auth/netlogon` fails intermittently in CI (`Integration Tests` and `Windows Build & Test` jobs), unrelated to the PR under test.

## Evidence
Observed red on two independent #1228 PRs on 2026-06-24, neither touching netlogon code:
- PR #1379 (docs + durable test) — `Windows Build & Test`: `--- FAIL: TestReloadCredential_ConcurrentLogons (0.39s)` → `FAIL internal/auth/netlogon`
- PR #1381 (SACL) — `Integration Tests (1.25.x)`: `--- FAIL: TestReloadCredential_ConcurrentLogons (0.52s)` → `FAIL internal/auth/netlogon`

Sub-second failure + concurrency in the name points to a race / ordering assumption in the hot-reload-during-concurrent-logon path (introduced around the #1325 hot-reload machine-credential work).

## Action
- Reproduce: `go test -run TestReloadCredential_ConcurrentLogons -count=200 -race ./internal/auth/netlogon`
- Fix the race or the test's timing assumption; do not paper over with a retry.

## Workaround until fixed
CI red on this test alone (diff not touching netlogon) is not a merge blocker — re-run the job or admin-merge.

marmos91:
## What

Fixes the ~50% flake in `TestReloadCredential_ConcurrentLogons` (#1383), which intermittently failed with `expected zero logon errors under concurrent reload, got 1` and blocked unrelated PRs.

This is a **production concurrency bug**, not a test-only issue — the test was correctly asserting an invariant the code violated. (No data race: all shared state is already guarded by the credential `RWMutex` and the `a.mu` channel-pointer swap, so the race detector would not have flagged it; the flake surfaces deterministically under `-count`.)

## Why (root cause)

`NetworkLogon`'s retry loop bounded **all** transient secure-channel retries by a single small budget, `maxReloadRetries = 8`. Two very different conditions shared that budget:

1. **Local teardown race** (`errChannelNotConnected`) — a concurrent `ReloadCredential` closed the cached channel *after* a logon read the channel pointer but *before/while* its RPC ran. This never reaches the DC.
2. **DC-side transient rejection** (`STATUS_ACCESS_DENIED`) — a freshly rebuilt channel's MS-NRPC credential chain is still settling.

The retry of case (1) also ran with **no backoff and no yield**, in a tight spin against 4 reload goroutines spinning equally tightly. Under that burst, a single logon could lose the purely-local race 8 times in a row, exhaust the shared budget, and return `errChannelNotConnected` to the caller — a spurious logon failure caused entirely by an internal channel swap.

## Fix

Split the two transient cases into separate budgets:

- **Local teardown race**: retried under a separate, generous live-lock guard (`maxTeardownRetries = 1024`) with a `runtime.Gosched()` yield, so an in-flight reload makes progress instead of being starved by a tight retry spin. It no longer consumes the DC-retry budget. Safe to retry freely: it never touches the DC, so it can never amplify failed logons toward AD account lockout, and it clears the instant the logon grabs the freshly rebuilt channel.
- **DC-side rejection**: unchanged behaviour — still bounded by `maxReloadRetries` with `reloadRetryBackoff` between attempts.

A credential-provider error (passthrough disabled) still fails closed, and a hard auth failure still fails fast after one call — both preserved.

## Verification

- `go test -run TestReloadCredential_ConcurrentLogons -count=500 ./internal/auth/netlogon` — clean (was ~50% failure before)
- `go test -count=5 ./internal/auth/netlogon` — clean
- `go build ./...` and `go vet ./internal/auth/netlogon/` — clean

(`-race` is unsupported on this windows/arm64 host; the flake is a deterministic assertion failure exercised by `-count`, and the underlying shared state was already correctly synchronized.)

Fixes #1383.

### DIFF
--- a/internal/auth/netlogon/authenticator.go
+++ b/internal/auth/netlogon/authenticator.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"runtime"
 	"sync"
 	"time"
 )
@@ -58,10 +59,27 @@ func isTransientChannelError(err error) bool {
 }
 
 // maxReloadRetries bounds how many times NetworkLogon rebuilds the channel and
-// retries after a transient secure-channel error (a concurrent reload tearing
-// down / rebuilding the channel), so it can never live-lock. In practice a
-// reload is a rare admin action and the first rebuild succeeds.
+// retries after a DC-side transient rejection (STATUS_ACCESS_DENIED while a
+// freshly rebuilt credential chain settles), so it can never live-lock. In
+// practice a reload is a rare admin action and the first rebuild succeeds.
+//
+// It does NOT bound the purely-local teardown race (errChannelNotConnected): see
+// maxTeardownRetries.
 const maxReloadRetries = 8
+
+// maxTeardownRetries bounds retries of the purely-local teardown race
+// (errChannelNotConnected): a concurrent reload closed the cached channel after
+// this logon read the pointer but before/while its RPC ran. Unlike a DC-side
+// rejection this never touches the DC (so it can never amplify toward account
+// lockout) and clears as soon as the logon grabs the freshly rebuilt channel.
+//
+// It therefore must NOT consume the small maxReloadRetries budget: under a burst
+// of admin reloads a single logon can legitimately lose this race many times in
+// a row, and failing the user's logon for an internal teardown race is wrong (it
+// was the flake behind #1383). The bound here exists only as a live-lock guard
+// and is generous; each losing attempt yields the P so a reload can make
+// progress instead of being starved by a tight retry spin.
+const maxTeardownRetries = 1024
 
 // reloadRetryBackoff is the pause between rebuild attempts after a DC-side
 // failure, giving a freshly rebuilt channel's MS-NRPC credential chain a moment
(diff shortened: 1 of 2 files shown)


marmos91:
Good catch — fixed in 0965efe3.

`SecureChannel.samLogon` returns the **bare** `errChannelNotConnected` for the purely-local teardown race (securechannel.go:315), but **wraps** it (`fmt.Errorf("...%w...", errChannelNotConnected, err)`) for transport-level failures including `context.Canceled`/`DeadlineExceeded` (securechannel.go:357). The teardown branch used `errors.Is`, which matched both — so a wrapped transport/cancellation error could be absorbed by the new 1024 teardown budget and spin.

Fix:
- The high-budget teardown branch now uses an identity compare (`err == errChannelNotConnected`), so only the bare local race takes that budget. Every wrapped transient (incl. cancellation) falls through to the bounded `maxReloadRetries` branch, which already `select`s on `ctx.Done()`.
- Added an explicit `if ctx.Err() != nil { return ctx.Err() }` at the top of the teardown branch so a canceled/deadline-exceeded caller returns promptly instead of spinning the teardown budget.

New regression test `TestNetworkLogon_CanceledContextStopsTeardownRetry` drives the bare-sentinel path (`fakeState.alwaysTornDown`) and asserts a canceled context returns `context.Canceled` after ≤4 `samLogon` calls (not 1024). `TestReloadCredential_ConcurrentLogons` still passes 500×; full package 5×; `go build ./...` + `go vet` clean.

Issue #1387, opened 2026-06-24
Remove legacy refcount/dedup machinery that is dead in the CAS block store


marmos91:
Follow-up to #1374 (closed by #1377 / #1380 / #1382).

While root-causing #1374 we confirmed that a chunk of the block-store metadata surface is **vestigial in the content-addressed (CAS) model**: physical dedup is intrinsic to CAS (identical content → identical hash → one object), reclamation is **mark-sweep GC** over the live block set (`EnumerateFileBlocks`), eviction keys off `SyncedHashStore.IsSynced`, and delete/truncate reap by block id. The per-hash `RefCount` index and `GetByHash` (remote-gated; never resolves pending CAS blocks) are legacy carry-over. `audit-refcounts` was already redefined off this in #1382.

This issue tracks deleting the dead machinery (per "less is more / delete legacy eagerly").

### Confirmed dead → remove
- **File-level dedup** — `pkg/block/engine/dedup.go` (`trySpeculativeFileLevelDedup`, `applyFileLevelDedupHit`, eager small-file dedup). Never hits in CAS: it increments via `GetByHash`, which misses pending CAS donors (locked in by `dedup_pending_donor_test.go`).
- **Block-level dedup LRU + AddRef** — the `dedupLRU` hit path in `rollup.go` always returns `ErrUnknownHash` (hash index never written for pending CAS) and falls back to `StoreChunk`.
- **`ListPending`** — no production caller.
- **Dead `FileBlock` helpers** — `IsRemote` / `IsFinalized` / `IsDirty` / `IsLocal` / `HasLocalFile` (zero callers).
- **`LastAccess` field** — never read/written in live code.

### Pending verification (a re-audit is in progress)
PR #1384 (NFSv4.2 CLONE) wired `engine.CopyPayload` via `common.CloneWholeFile` for CLONE + SMB server-side-copy, which bumps `RefCount` per unique hash. So **`RefCount` / `IncrementRefCount` / `DecrementRefCount` / `GetByHash` / `CopyPayload` are no longer obviously unwired** — before removing any of these we must confirm whether CLONE/SMB-copy actually depend on the RefCount bump for correctness, or whether the cloned file's manifest reference + mark-sweep GC already protects shared blocks (making the bump a no-op). If the latter, this set is still removable; if the former, it stays. Will update this issue with the re-audit verdict.

### Likely keep (needs care, not "dead")
[...]

PR #1391, opened 2026-06-24
fix(block): CopyPayload must create dst FileBlock rows — clone read-back corruption (#1384)

marmos91:
## Severity: silent data corruption (found via #1374 follow-up audit)

NFSv4.2 CLONE and SMB server-side-copy route through `engine.CopyPayload`. For any **rolled-up** source file, the clone produced a destination that **reads back as all zeros** — verified on a live VM (NFSv4.2 + badger + S3).

### Root cause (two layers)
1. **EREMOTEIO failure:** `CopyPayload` calls `IncrementRefCount` per source hash and returns the error on failure. CAS blocks are `Pending` and never resolve behind the Remote-gated `GetByHash`, so `IncrementRefCount` returns `ErrFileBlockNotFound` → clone failed loudly.
2. **The corruption:** once that's unblocked, `CopyPayload` only copied the source's `BlockRef` manifest into the destination's `FileAttr.Blocks` and explicitly skipped row creation (`_ = dstPayloadID`). But the cold-read path resolves bytes via `ListFileBlocks(dstPayloadID)` (per-payload FileBlock rows), **not** `FileAttr.Blocks`. No dst rows → zero-fill.

The #1384 unit test missed both because its fake coordinator always succeeds at `IncrementRefCount` and doesn't model the FileBlock store.

### Fix
- Create one dst `FileBlock` row per source block (same hash/offset/size, `Pending`), mirroring `ObjectIDPersister`, so clone reads resolve the shared CAS chunks.
- The row `Put` uses the **context-bound metadata txn** (`TxFromContext`) when present: a store-level `Put` re-enters the memory backend's non-reentrant mutex held by `WithTransaction` and self-deadlocks; the tx-bound `Put` also commits atomically with the caller's dst-manifest persist.
- Tolerate `ErrFileBlockNotFound` from `IncrementRefCount` (the per-hash bump is a no-op in CAS — blocks are kept alive by the manifest + GC mark-sweep, not RefCount). A genuine backend fault still aborts.

### Verification
- **Live VM:** cloned a rolled-up 16 MiB file; after `echo 3 > drop_caches`, `md5(dst) == md5(src)` (real block reads). `Files` went 1→2, dst payload now has its rows.
- Regression test `TestCopyPayload_CreatesDstFileBlockRows` (real FileBlock store) asserts dst rows exist per source block — fails before, passes after. Existing CopyPayload + `CloneWholeFile` tests stay green (the latter previously **hung** without the tx-bound Put).

Refs #1384.

### DIFF
--- a/pkg/block/engine/api_blockref_test.go
+++ b/pkg/block/engine/api_blockref_test.go
@@ -58,6 +58,96 @@ func TestCopyPayload_O1_IncrementsRefCountPerUniqueHash(t *testing.T) {
 	}
 }
 
+// TestCopyPayload_ToleratesPendingBlocks pins the #1384 fix: on a CAS share
+// every source block is Pending, so the Remote-gated GetByHash behind
+// IncrementRefCount resolves to "no FileBlock row" and returns
+// ErrFileBlockNotFound. That must NOT fail the copy — NFSv4.2 CLONE and SMB
+// server-side-copy route through CopyPayload, and the cloned blocks are kept
+// alive by the destination's manifest via the GC mark phase, not by RefCount.
+// (The pre-fix code returned the error and CLONE failed with EREMOTEIO on any
+// rolled-up source; the original test masked it with a fake that never gated.)
+func TestCopyPayload_ToleratesPendingBlocks(t *testing.T) {
+	fc := &fakeCoordinator{incAllNotFound: true}
+	bs := newTestEngineWithCoordinator(t, fc)
+	ctx := context.Background()
+
+	src := []block.BlockRef{
+		{Hash: block.ContentHash{0x01}, Offset: 0, Size: 1024},
+		{Hash: block.ContentHash{0x02}, Offset: 1024, Size: 1024},
+		{Hash: block.ContentHash{0x03}, Offset: 2048, Size: 1024},
+	}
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed the Copilot findings:

- **dst-row write gated on the wrong thing (readwrite.go:535)** — fixed in 86b2a333. The write now goes through the bound txn when present (else the store), and `CopyPayload` returns an error if there are real blocks to clone but neither a txn nor a store can persist the rows — so a nil `fileBlockStore` can never silently skip rows and re-introduce the zero-fill.

- **stale dst rows on copy-onto-existing (readwrite.go:503)** — real but deferred, and not reachable on the current paths: the only live caller is NFSv4.2 CLONE via `cp --reflink`, where the client `open(O_TRUNC)`s the destination first (Truncate already reaps the dst's prior rows before `CopyPayload` runs), and no SMB server-side-copy caller is wired yet. A correct in-txn reap needs `ListFileBlocks` on the `metadata.Transaction` interface (it currently exposes `Put` but not list; a store-level list inside `WithTransaction` self-deadlocks the memory backend). Tracking it in #1387 rather than widening the interface in this hotfix.

Issue #1266, opened 2026-06-21
Performance: SMB throughput over WAN (RTT-bound round-trips) + large-file digest/upload bottleneck


marmos91:
Tracking the **performance** observations from #1245 (which the reporter explicitly flagged as "not a bug, product-level"). The #1245 *correctness* fixes (dedup-race EIO, rollup convergence, crash-safety, cache sizing) are merged and do **not** address this throughput axis — splitting it out so it isn't lost.

## Observations (from #1245, client = macOS over WireGuard, ~38 ms RTT)

| Path | Throughput |
|---|---|
| Client uplink | ~47 MB/s (380 Mbit/s) |
| dittofs local write on server (dd, direct I/O) | ~78 MB/s |
| **SMB over the tunnel (38 ms RTT)** | **~6 MB/s** |
| `scp` over the same WAN | ~22 MB/s |
| `cp` of a large file *through* dittofs (cache + hash + S3 upload) | ~9 MB/s |

Two distinct bottlenecks:

### 1. SMB is ~3.5× slower than a plain TCP stream on the same link (latency-bound)
SMB throughput collapses under WAN RTT because the write path issues too many **synchronous round-trips** — each one pays the 38 ms latency. `scp` (a single streamed TCP flow) gets ~22 MB/s on the identical link; SMB gets ~6 MB/s.

**Likely levers:** more aggressive **write-behind / pipelining** on the SMB write path (credits / multiple outstanding `SMB2 WRITE`s in flight), large-MTU/compound handling, and avoiding per-write metadata round-trips. Worth a pcap to count round-trips per MB vs Samba/Windows.

### 2. Large files bottleneck on dittofs itself (digest + upload), ~9 MB/s
For large files the limiter becomes dittofs's own pipeline: content **hashing (BLAKE3/FastCDC) + S3 upload**, even on a fast local write path (~78 MB/s dd). Worth profiling whether chunk hashing, the syncer's upload concurrency, or serialization between hash→CAS→upload is the constraint, and whether write-behind can overlap upload with the next incoming write.

## Notes
- Loopback benchmarking will **not** reproduce bottleneck #1 (no RTT) — needs a latency-injected link (`tc netem delay 38ms`) or a real WAN to measure SMB round-trip amplification.
- Bottleneck #2 *is* measurable locally (local write vs through-dittofs on large unique/non-dedup data).

## Suggested first steps
[...]

PR #1400, opened 2026-06-24
perf(block): adaptive upload concurrency (BBR-style goodput-gated window) (#1398)

### LINKED ISSUE #1398: Adaptive upload concurrency (BBR-style goodput probing) for the mirror loop
Follow-up to #1266 / #1397.

#1397 parallelized the mirror upload loop and added a per-remote `--parallel-uploads N` cap. When that flag is **unset**, concurrency still defaults to the CPU-deduced `max(4, nCPU)` — wrong-shaped, since uploads are network-bound, not CPU-bound (measured: 5.5% CPU at saturation).

## Goal
Replace the unset default with an **adaptive limiter** that discovers the optimal in-flight upload count at runtime.

## Signal (decided)
Primary = **aggregate goodput gradient** (BBR-style bandwidth probe): ramp concurrency while total delivered MB/s keeps rising; settle at the knee where more in-flight only adds latency, not throughput. The `datapath_upload_bytes` + `uploads_inflight` metrics from #1397 already provide the inputs.

Secondary (corroborating) = **per-PUT latency inflation** (rising latency with flat goodput confirms the bandwidth knee).

`503 SlowDown` / upload errors = rare hard multiplicative back-off safety only — **not** the primary driver. (The bottleneck is uplink bandwidth, which saturates far below S3's request-rate limit, so loss-based control would never trip in the common case.)

## Design notes
- Per-**remote** limiter (one bandwidth estimate per shared uplink) so multiple shares to the same remote cooperate on one window rather than each ramping independently.
- Persistent cross-pass limiter (today's loop uses a per-pass fixed semaphore) + EWMA + hysteresis to survive bursty writes.
- `--parallel-uploads N` set → hard ceiling at N (the congestion-window cap; disables ramp above N).
- Unset → no *tuning* ceiling, only a safety rail (~256 in-flight) against FD/goroutine exhaustion.

## Measurement
Validate before/after on the Scaleway VM → fr-par S3 (single share AND multi-share competing for one uplink), comparing adaptive vs static N=16/32. Baseline from #1266: serial 2.7 MiB/s → static-16 40 MiB/s; raw-S3 ceiling N=32 ≈ 44 MiB/s.

marmos91:
## What

Replaces the CPU-deduced upload-concurrency default with a **per-remote adaptive limiter** for the CAS→remote mirror loop. Follow-up to #1266 / #1397, closes #1398.

#1397 parallelized the mirror loop but, when `--parallel-uploads` was unset, still defaulted to `max(4, nCPU)` — the wrong shape, since uploads are network-bound, not CPU-bound (measured ~5% CPU at saturation). This PR discovers the right in-flight count from the link itself.

## How it works

A hybrid controller (the algorithm decided in #1398):

- **Latency-gradient drives the window** (Vegas/Gradient2): track the fastest per-PUT latency seen (the propagation floor); grow while latency stays near it; when more in-flight only inflates latency without adding throughput, the link is saturated and growth stops.
- **Aggregate-goodput gate caps the top end**: once total delivered MB/s plateaus, the window won't grow past the point that hit peak throughput (bufferbloat guard — the gradient's sqrt headroom would otherwise probe upward forever).
- **Upload errors** (S3 `503 SlowDown`, timeouts) → multiplicative back-off, a rare safety valve only. They do not drive the steady state: the bottleneck is uplink bandwidth, which saturates well below S3's request-rate limit.

`--parallel-uploads N` set → static cap `N` (no ramp). Unset → adaptive, bounded only by a 256-in-flight safety rail.

## Design

- **Per-remote limiter**, shared by every share targeting the same remote (keyed alongside the ref-counted `sharedRemote`), so shares on one uplink cooperate on a single window instead of each ramping independently.
- Built once in `acquireRemoteStore` from the already-parsed config — which also **removes #1397's second `resolveBlockStoreConfig` DB read** at attach (the prior review wart).
- New `engine.UploadLimiter` (`Acquire`/`ReleaseSlot`/`Observe`/`Limit`), `sync.Cond`-based, injectable clock for tests. `mirrorOnce` uses it in place of the per-pass fixed semaphore; `mirrorChunk` feeds each Put's latency/result via `Observe`.
- New `dittofs_datapath_upload_window` gauge.

## Changes

- `pkg/block/engine/uploadlimiter.go` (new) + tests
- `pkg/block/engine/syncer.go`, `types.go`, `metrics.go` — wire the limiter through the mirror loop
- `pkg/controlplane/runtime/shares/service.go` — per-remote sharing + limiter injection (and the DB-read removal)
- `pkg/metrics/instruments.go` — the gauge
- `cmd/dfsctl/.../remote/{add,edit}.go` — flag help reworded (0 = adaptive)
- `docs/guide/configuration.md` + regenerated `docs/guide/cli.md`

## Verification

- `go build ./...`, `go vet`, `gofmt` clean.
- `go test -race ./pkg/block/engine/...` green; full engine + metrics + runtime/shares suites green.
- New unit tests: fixed vs adaptive, grow-when-uncongested, shrink-under-congestion, error back-off, safety-rail, goodput-gate-halts-growth, Acquire-blocks-at-window, ctx-cancel, concurrent-no-over-dispatch (`-race`).
- code-simplifier + code-reviewer passed (reviewer: 0 critical; 2 findings — stale comment + a loose test assertion — both fixed).

## Follow-up

Empirical before/after on the Scaleway VM → fr-par S3 (single share **and** multi-share competing for one uplink) vs static N=16/32, per the #1398 measurement plan — to run next.

### DIFF
--- a/cmd/dfsctl/commands/store/block/remote/add.go
+++ b/cmd/dfsctl/commands/store/block/remote/add.go
@@ -83,7 +83,7 @@ func init() {
 	addCmd.Flags().StringVar(&addAccessKey, "access-key", "", "AWS access key ID (for s3)")
 	addCmd.Flags().StringVar(&addSecretKey, "secret-key", "", "AWS secret access key (for s3)")
 	addCmd.Flags().StringVar(&addCompression, "compression", "", "Enable per-block compression: zstd, lz4 (default: off)")
-	addCmd.Flags().IntVar(&addParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = auto, scales with CPU count)")
+	addCmd.Flags().IntVar(&addParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-ramp to the link's bandwidth knee)")
 	// Encryption flags
 	addCmd.Flags().StringVar(&addEncryptionAEAD, "encryption-aead", "", "Enable client-side encryption with the given AEAD: aes-256-gcm, chacha20-poly1305, xchacha20-poly1305")
 	addCmd.Flags().StringVar(&addEncryptionKeyKind, "encryption-key-kind", "", "Key provider: local | kmip (required when --encryption-aead is set)")
--- a/cmd/dfsctl/commands/store/block/remote/edit.go
+++ b/cmd/dfsctl/commands/store/block/remote/edit.go
@@ -52,7 +52,7 @@ func init() {
 	editCmd.Flags().StringVar(&editEndpoint, "endpoint", "", "Custom S3 endpoint")
 	editCmd.Flags().StringVar(&editAccessKey, "access-key", "", "AWS access key ID (for s3)")
 	editCmd.Flags().StringVar(&editSecretKey, "secret-key", "", "AWS secret access key (for s3)")
-	editCmd.Flags().IntVar(&editParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = auto, scales with CPU count)")
+	editCmd.Flags().IntVar(&editParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-ramp to the link's bandwidth knee)")
 }
 
 func runEdit(cmd *cobra.Command, args []string) error {
@@ -111,9 +111,10 @@ func runEdit(cmd *cobra.Command, args []string) error {
 		if editSecretKey != "" {
 			currentConfig["secret_access_key"] = editSecretKey
 		}
-		// parallel-uploads is an int where 0 is meaningful ("reset to auto"),
-		// so key off Changed rather than a zero-value check: set clears to
-		// auto-deduce, a positive value pins the per-remote cap.
+		// parallel-uploads is an int where 0 is meaningful ("reset to
+		// adaptive"), so key off Changed rather than a zero-value check:
+		// clearing it re-enables the adaptive window, a positive value pins
+		// the per-remote cap.
 		if cmd.Flags().Changed("parallel-uploads") {
 			if editParallelUploads > 0 {
 				currentConfig["parallel_uploads"] = editParallelUploads
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -6069,7 +6069,7 @@ Flags:
       --encryption-kmip-key-uid string    KMIP managed symmetric key UID (kind=kmip)
       --endpoint string                   Custom S3 endpoint (for S3-compatible stores)
       --name string                       Store name (required)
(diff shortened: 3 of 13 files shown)


marmos91:
Addressed all three:

1. **Goodput gate (uploadlimiter.go) — real bug, fixed.** The gate compared `goodputEWMA` against `peakGoodput` *after* overwriting it, so the condition was always true and the window froze at the first sample. Reworked to an explicit plateau flag: while goodput improves sample-over-sample the window keeps ramping; it only clamps to `windowAtPeak` once a sample shows no improvement. Added a regression test (`TestUploadLimiter_AdaptiveRampsWhileGoodputRises`) that fails on the old logic.
2. **docs config example** — split into raise-ceiling vs `--parallel-uploads 0` (re-enable adaptive) so the `0` note no longer sits next to a `32`.
3. **blockstore_init comment** — dropped the cross-package `UploadSafetyRail` identifier; now reads "in-process safety rail of 256 in flight".

Pushed as `1a14ca4e`.

marmos91:
## On hold — blocked on #1407

Benchmarking on the scw VM against real fr-par S3 (after fixing #1405, which was preventing data from reaching the remote at all) shows the adaptive upload window does **not** improve write-path throughput, because the bottleneck is **rollup chunk production**, not upload concurrency (#1407). Across adaptive / static-16 / static-32, end-to-end throughput sits in the same ~5–8 MB/s band with in-flight ≈ 4 regardless of the cap.

Separately, in the deterministic controller simulation the gradient limiter currently **underperforms a hand-tuned static cap** (~70% throughput) with noisy convergence — it settles below the bandwidth knee and overshoots transiently during ramp.

### Decision
Fix #1407 (rollup throughput) first. Upload concurrency only becomes the bottleneck in a backlog-drain workload (offline → reconnect → drain), which isn't the common path. Do **not** merge until either rollup is fixed and a workload shows concurrency is the limit, or the control loop is tuned to reliably reach the knee.

### Kept on the branch (delicate-balance design, per review feedback)
`--parallel-uploads N` is now a **ceiling** the adaptive controller operates beneath (adapt within [1, N]) instead of a hard static pin that disabled adaptation — the flag and the controller compose. The window is also bounded by the S3 connection pool so it can't outrun available connections. Tests updated to assert only honest, measured properties; the under-the-knee gap is logged, not asserted.

marmos91:
Superseded by #1410. That PR rebuilds adaptive upload concurrency on a **goodput-driven** controller instead of the latency-gradient one here — the latency-gradient equilibrium (window = 1/(1−gradient)²) collapses the window to ~4 because it reads the per-PUT latency rise that healthy concurrency *causes* as congestion. #1410 is VM-validated: the window auto-ramps 16→64 and holds through pipeline starvation, hitting 73.6 MiB/s (matches the best hand-tuned fixed window, 4.2× the old default) with zero config. The one idea worth carrying forward from here — a shared per-remote limiter so multiple shares on one uplink cooperate — is noted as a follow-up on #1410.

Issue #1407, opened 2026-06-24
Rollup chunk production starves the S3 uploader — write→remote throughput is rollup-bound, not concurrency-bound


marmos91:
## Summary

Once #1405 lets NFS/SMB-written data actually flow to a remote, the end-to-end **write→S3 throughput is bottlenecked by rollup chunk production, not by upload concurrency.** For a single large file the uploader sits at ~4 in-flight regardless of the configured parallelism, because the per-payload rollup emits CAS chunks roughly serially and the syncer starves waiting for them.

## Evidence (bench VM → scw fr-par S3, 512 MiB single NFSv3 file, #1405-fixed)

| upload concurrency | throughput | upload window peak | steady inflight |
|---|---|---|---|
| adaptive (BBR, #1398) | 5.5 MB/s | 9 | ~4 |
| static `parallel_uploads=16` | 7.4 MB/s | 8 | ~4 |
| default (develop) | 8.0 MB/s | n/a | 1–8, dropping to 1 |

All three land in the same ~5–8 MB/s band. The upload window never sustains above ~8 even when the cap is 16/32 — there simply aren't enough rolled-up chunks ready to upload concurrently. Local write was ~45 MB/s and raw parallel S3 Put was measured at 40 MB/s in #1266, so neither disk nor the network is the limit here.

## Root cause (hypothesis to confirm)

`FSStore` rollup processes one payload's earliest stable interval per pass and emits chunks sequentially; the syncer's `pendingHashes` set therefore fills slowly for a single hot file. Candidate levers:
- Parallelize chunk emission within a single large stable interval (FastCDC + BLAKE3 are CPU-parallelizable across sub-ranges).
- Increase `rollup_workers` effectiveness for a single payload (today workers are per-payload-serialized by `rmu`).
- Pipeline rollup→mirror so the uploader drains chunks as they're emitted rather than per-tick.

## Impact on #1398 / PR #1400 (adaptive upload window)

The adaptive limiter is mechanically correct (ramps to a knee, no overshoot) but delivers **no throughput improvement on the single-file path** because concurrency is not the bottleneck. Recommend holding PR #1400 until (a) rollup throughput is improved and (b) a workload where upload concurrency is actually the limit (multi-share / many concurrent files / post-outage backlog drain) is benchmarked to show the win.

## Follow-up

Benchmark multi-file / multi-share concurrent writes to confirm the uploader can saturate parallelism once many chunks are ready — that is the scenario where BBR vs static would diverge.

[...]

PR #1410, opened 2026-06-25
perf(block): adaptive upload concurrency (#1407)

marmos91:
## Adaptive upload concurrency (#1407, redesign of #1400)

Auto-tune the number of concurrent CAS-chunk uploads to **saturate the uplink with no configuration**. When `--parallel-uploads` is unset (the new default) the syncer discovers the right concurrency itself; an explicit `--parallel-uploads N` still pins a fixed window (unchanged behaviour).

### Why

Uploads are **network-latency bound, not CPU bound** — a single PUT to a remote region sustains only a few MiB/s, so throughput scales with concurrency until the uplink saturates. The previous default derived concurrency from CPU count, which throttled fast links (an 8-core host mirrored at ~14 MiB/s while the link could do far more). v0.22.2 replaced that with a fixed default of 32; this PR removes the guesswork entirely.

#1400 tried a latency-gradient controller, but its closed-form equilibrium collapses the window to ~4: it reads the per-PUT latency rise *caused by useful concurrency* as congestion. The only honest saturation signal is **"did goodput stop rising"**, so this controller is goodput-driven.

### How

- **`goodputController`** — pure, deterministic, unit-tested. TCP-slow-start-like: ramp the window while delivered goodput keeps improving (EWMA-smoothed), settle at the knee once it plateaus, back off only on upload errors or a real goodput collapse — never on latency.
- **`windowLimited` gate (the crux)** — the controller only ramps/backs-off when the window was actually saturated. When the upstream rollup pipeline produced fewer chunks than the window could carry (app-limited), it **holds** the window. Without this, a transient pipeline stall collapses the window and throttles the drain the moment the backlog bursts — caught on the VM (see below).
- **`dynamicSemaphore`** — resizable concurrency limiter the controller drives; `TakePeak()` reports the in-flight high-water (the window-limited signal).
- A 1s control-interval goroutine samples delivered bytes → goodput → new window. `mirrorOnce` uses the shared limiter. Pinned mode keeps a fixed window and runs no controller.
- s3 HTTP pool sized so it never caps the window; new `dittofs_datapath_upload_window` gauge.

### Benchmarks (VM → Scaleway fr-par S3, 1 GiB, sar NIC TX, avg over active interval)

| config | throughput | in-flight |
|---|---|---|
| pinned 8 (old CPU-derived default) | 17.7 MiB/s | 8 |
| pinned 32 (v0.22.2 default) | 73.5 MiB/s | 32 |
| **adaptive — no flag (this PR)** | **73.6 MiB/s** | **auto 16→64** |
| raw boto3 C=32 (uplink ceiling, no CAS pipeline) | 114.9 MiB/s | 32 |

The window auto-ramps (16→24→36→54→64) and holds; adaptive **matches the best hand-tuned fixed window with zero configuration** (4.2× the old CPU-derived default). The remaining gap to raw S3 is the rollup/CAS pipeline (FastCDC + BLAKE3 + NFS read), not upload concurrency — that's tracked as the separate #1407 levers.

A first VM run exposed (and this PR fixes) the window collapsing to the floor when the rollup pipeline briefly starved the uploader — the `windowLimited` gate now holds the window so the backlog drains fast.

### Tests

- `upload_controller_test.go` — ramp / settle-at-knee / ceiling / error backoff / collapse backoff / floor / **app-limited hold** / recovery.
- `dynsem_test.go` — limit/grow/shrink/context/peak, race-clean.
- `upload_adaptive_test.go` — end-to-end tick glue (ramp, settle, **hold-on-starve**, idle-skip, error backoff).

### Follow-ups

- Rollup pipeline is now the throughput ceiling (#1407 levers 2/3: parallelize rollup chunk emission, batch fsync) — that's the remaining gap to raw S3, not upload concurrency.
- CLI footgun: `--config <json>` silently drops sibling flags incl. `--parallel-uploads`.
- A shared per-remote limiter (so two shares on one uplink cooperate on a single window) — currently each share auto-tunes independently, bounded by the shared 128-conn pool.

Part of #1407 (upload-concurrency lever). **Supersedes #1400** (latency-gradient design — its equilibrium collapses the window to ~4).

### DIFF
--- a/cmd/dfsctl/commands/store/block/remote/add.go
+++ b/cmd/dfsctl/commands/store/block/remote/add.go
@@ -83,7 +83,7 @@ func init() {
 	addCmd.Flags().StringVar(&addAccessKey, "access-key", "", "AWS access key ID (for s3)")
 	addCmd.Flags().StringVar(&addSecretKey, "secret-key", "", "AWS secret access key (for s3)")
 	addCmd.Flags().StringVar(&addCompression, "compression", "", "Enable per-block compression: zstd, lz4 (default: off)")
-	addCmd.Flags().IntVar(&addParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = use server default of 32)")
+	addCmd.Flags().IntVar(&addParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink)")
 	// Encryption flags
 	addCmd.Flags().StringVar(&addEncryptionAEAD, "encryption-aead", "", "Enable client-side encryption with the given AEAD: aes-256-gcm, chacha20-poly1305, xchacha20-poly1305")
 	addCmd.Flags().StringVar(&addEncryptionKeyKind, "encryption-key-kind", "", "Key provider: local | kmip (required when --encryption-aead is set)")
--- a/cmd/dfsctl/commands/store/block/remote/edit.go
+++ b/cmd/dfsctl/commands/store/block/remote/edit.go
@@ -52,7 +52,7 @@ func init() {
 	editCmd.Flags().StringVar(&editEndpoint, "endpoint", "", "Custom S3 endpoint")
 	editCmd.Flags().StringVar(&editAccessKey, "access-key", "", "AWS access key ID (for s3)")
 	editCmd.Flags().StringVar(&editSecretKey, "secret-key", "", "AWS secret access key (for s3)")
-	editCmd.Flags().IntVar(&editParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = use server default of 32)")
+	editCmd.Flags().IntVar(&editParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink)")
 }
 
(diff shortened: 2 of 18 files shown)


marmos91:
@copilot review

marmos91:
Addressed Copilot review in 5f70b18e:

- **dynsem `Acquire` goroutine-per-call** → added a fast path: when a slot is free (the common mirror-loop case) it's taken without parking or spawning a watcher goroutine. Only a genuinely-blocked Acquire spawns one.
- **`cond` signalled outside the lock (lost-wakeup risk)** → all signals (Acquire's cancellation watcher, `Release`, `SetLimit`) now broadcast **under `s.mu`**, serializing them against a waiter's ctx-check/`Wait` window. Kept `Broadcast` (not `Signal`) since a woken waiter may bail on a cancelled context without consuming the freed slot. Stress-verified 200× `-race`.
- **s3 pool 128 caps pinned `--parallel-uploads`** → raised `maxS3ConnsPerHost` to **256**, matching the `--parallel-uploads` validator ceiling, so the HTTP pool never caps a pinned window.

Also fixed a **pre-existing flake** (commit ce04b5ec) the CI run surfaced: the `MirrorLoop` integration tests raced the wake-driven periodic uploader (#1408). Added `SyncerConfig.ManualSync` so Flush is the sole deterministic mirror driver in those tests. Flaky on develop too; now 100× `-race` clean.

Issue #1414, opened 2026-06-25
Small-file performance: object packing (many tiny files → one CAS object)


marmos91:
> **Priority: tier-1 — flagship small-file performance feature.** Real workloads (source trees, mail dirs, photo libraries, build outputs) are small-file-dominated; this is one of the most important remaining DittoFS capabilities — we need it. Core small-file lever alongside #1416 (metadata commit batching).

---

## Motivation

Benchmarking on the demo VM (scw fr-par S3) shows the CAS pipeline is **catastrophic for many small files** and good for big files:

| workload | write | e2e drain | S3 inflight |
|---|---|---|---|
| big 64×16MiB | 143 MiB/s | 36.5 MiB/s | healthy |
| small 16384×64KiB | ~4 MiB/s (64 files/s) | ~0.5 MiB/s | **1** |

Full analysis in #1411. Two independent walls, both per-FILE:
1. **Write ~64 files/s** — per-file synchronous metadata commit (badger `SyncWrites`): CREATE + COMMIT each fsync. NFSv3 vs v4.0 measured — protocol doesn't help (v4 slower).
2. **Rollup→S3 ~12 chunks/s** — each tiny file becomes its own ~64 KiB CAS object: 8 rollup workers serialize on the shared badger commit path (~15ms × several commits/file), and the S3 side does one tiny PUT per file. The adaptive upload window (v0.23.0) is irrelevant here — supply-starved.

This is arguably the most important remaining DittoFS performance feature: real workloads (source trees, mail dirs, photo libraries, build outputs) are dominated by small files.

## Proposal: object packing

Coalesce many small CAS chunks into **one larger S3 object** (Haystack / SeaweedFS-style volume packing):
- A packed object holds N small chunks; the metadata records (objectID, offset, length) per chunk.
- Cuts S3 object count + PUT count by ~N× → fixes the per-tiny-PUT drain ceiling and S3 per-object cost / read amplification.
- Amortizes per-file overhead in the rollup→upload path.

### Design questions (own design phase)
- **Dedup granularity**: packing groups chunks into a shared object — must preserve per-chunk content-addressing/dedup (chunk hash still the key; the pack is a container). GC/refcount must work at chunk granularity within a pack (compaction when a pack is mostly dead — cf. JuiceFS slice compaction).
- **Read path**: range-GET a sub-object from a pack; cache at chunk granularity.
[...]

Issue #1411, opened 2026-06-25
Rollup→upload pipeline smoothness: close the remaining write→S3 throughput gap


marmos91:
## Context

Adaptive upload concurrency shipped in #1410 (closing #1398/#1407): the upload window auto-ramps and holds, so **upload concurrency is no longer the bottleneck**. Measured on the demo VM → Scaleway fr-par S3 (1 GiB, sar NIC TX):

| config | throughput | in-flight |
|---|---|---|
| pinned 8 (old CPU default) | 17.7 MiB/s | 8 |
| pinned 32 (v0.22.2 default) | 73.5 MiB/s | 32 |
| **adaptive (no flag, #1410)** | **73.6 MiB/s** | auto 16→64 |
| raw boto3 C=32 whole-file 16 MiB | 116 MiB/s | 32 |

The remaining gap (73.6 → ~116) is **the rollup→upload pipeline**, not concurrency: during the transfer the uploader repeatedly goes **app-limited** (in-flight drops to 1–2 while the window sits at 64) — it starves waiting for rollup to produce CAS chunks.

## What we ruled out (measured)

- **Not chunk size.** Bumping `AvgChunkSize` 4 → 16 MiB made it *worse* (60.6 vs 73.6 MiB/s). Bigger chunks = fewer/larger PUTs (S3-efficient) but coarser rollup→upload pipelining (a big chunk must fully roll up before its upload starts; fewer chunks to spread across the window). The response is non-monotonic and pipeline-coupled — 4 MiB is near the sweet spot. **Auto-tuning chunk size by file type is not worth it** (non-monotonic, complex, marginal).
- **Not raw object-size overhead alone.** Object size dominates S3 throughput (raw boto3: 64×16 MiB = 116, 256×4 MiB = 37, 1024×1 MiB = 60 MiB/s), but the small-object boto3 numbers are GIL-throttled (Python threads don't parallelize TLS) — Go (DittoFS) already beats them at the same chunk size. So we're not far off what the link gives for 4 MiB objects.
- **Not CPU.** #1266 measured chunker ~1571 MB/s, BLAKE3 ~4354 MB/s, ~5.5% CPU at upload saturation. The stalls are I/O / coordination, not compute.

## Levers (ranked)

[...]

PR #1415, opened 2026-06-25
perf(block): parallelize rollup backlog drain + single-fsync header advance (#1411)

marmos91:
Two rollup-pipeline improvements from the small-file performance investigation in #1411. Both are durability-preserving and leave the NFS write path untouched.

## Changes

**1. Parallelize rollup backlog drain** (`scanAllFiles`)
Previously every rollup worker ran the same serial scan over all dirty payloads, walking the id list in the same order and colliding on per-file `rmu` → effectively serial. Now a single scanner is elected (`scanning` CAS) and fans out `rollupFile` across the worker budget via a bounded errgroup. `rollupFileInner` is already per-payload safe (rmu serializes same-payload passes; cross-payload state is sharded/atomic/content-addressed).

**2. Single-fsync rollup_offset header advance** (`advanceRollupOffset`)
Collapsed the two-fsync (write offset+fsync, write CRC+fsync) header update into one combined `WriteAt(hdr[8:32])` + one fsync. Offset [8:16) and CRC [28:32) share the first 512-byte sector → atomic persist; a torn write yields a CRC mismatch caught by the existing validation → safe re-init. Cuts one per-rolled-up-file fsync. (Doc is explicit that this newly relies on first-sector atomicity, unlike the prior ordered two-fsync protocol.)

## Honest framing

These are **prerequisites / pipeline cleanups**, not the small-file headline fix. VM benchmarking (#1411) shows small-file throughput is bound by **per-file synchronous metadata commits** (badger `SyncWrites`: ~64 files/s write, ~12 chunks/s rollup drain), which serialize the rollup workers regardless of dispatch parallelism. Removing one filesystem fsync (change 2) doesn't move that headline. The real small-file levers are tracked separately:
- **#1414** — object packing (many tiny files → one CAS object; the SeaweedFS-style S3-object-count fix)
- batched/async metadata commit (the JuiceFS-style write-back fix for the per-file fsync wall) — to be tracked

Change 1 (parallel dispatch) is still a genuine win for fast-writer / backlog scenarios and is the prerequisite that lets a lower per-file commit cost actually fan out across workers.

## Testing
- `pkg/block/local/fs` + `pkg/block/engine` full suites green with `-race`.
- New: `TestScanAllFiles_DispatchesConcurrently` (barrier-latched concurrency proof), `TestAdvanceRollupOffset_ValidHeader` (valid header + cold-reopen durability), `TestAdvanceRollupOffset_TornWriteDetected` (mismatch → ErrLogBadHeaderCRC).
- `gofmt` / `go vet` / `golangci-lint` clean.
- Reviewed (concurrency, CRC correctness, sector-atomicity, no deadlock with the rollup worker pool).

### DIFF
--- a/pkg/block/local/fs/appendlog.go
+++ b/pkg/block/local/fs/appendlog.go
@@ -227,22 +227,28 @@ func readRecordAt(rf io.ReaderAt, logPos uint64, payloadLen uint32) (uint64, []b
 	return fileOffset, payload, nil
 }
 
-// advanceRollupOffset atomically updates the log header's rollup_offset
-// field and fsyncs. FIX-6: split into a two-phase pwrite+fsync sequence
-// so the on-disk state is always either fully old-valid, fully new-valid
-// or "new-offset present but CRC stale" (which recovery treats as a hard
-// header corruption → re-init, the same posture as any other bad-CRC
-// header). The previous single 32-byte WriteAt could yield a torn
-// rollup_offset whose lower bytes were updated but upper bytes weren't
-// while the CRC happened to land on disk first and validate against the
-// torn value — silently corrupting recovery's view of the rollup point.
+// advanceRollupOffset updates the log header's rollup_offset field and
+// fsyncs. The new rollup_offset [8..16) and the recomputed CRC [28..32) are
+// written together in a SINGLE WriteAt(hdr[8:32]) followed by ONE fsync
+// (#1411 Lever 2), collapsing the prior two-phase write-offset-fsync /
+// write-crc-fsync sequence that cost two fsyncs per rolled-up file —
+// a per-file tax that dominated the small-file rollup pipeline.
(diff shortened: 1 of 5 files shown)


marmos91:
Addressed both Copilot comments in 72b98d01:

1. **2× rollup concurrency** — added a shared `rollupSlots` semaphore sized `rollup_workers`. Both the `rollupCh` worker pool (nudge path) and the `scanAllFiles` fan-out now acquire a slot before `rollupFile`, so `rollup_workers` is a true global cap rather than per-path. Drain/graceful-stop/conformance all still green under `-race`.
2. **Test hook race** — the timeout path now waits for the `scanAllFiles` goroutine to return (`<-done`) before `t.Fatalf`, so `t.Cleanup` never nils the package-global hook while rollups still read it.

Issue #670, opened 2026-05-28
B7 (v1.0 Wave 1): client NFS hangs in uninterruptible D-state — requires server restart


marmos91:
Surfaced during Wave 0 baseline (PR #666).

When `dfsctl bench` is killed mid-rand-write at 1 GiB, the dfsctl process enters uninterruptible disk-wait (`Dl`) and cannot be killed even with SIGKILL. The NFS mount is not responsive to `umount -f` or `umount -l`. Only recovery is `systemctl restart dfs.service` on the server; client process then becomes a zombie.

NFS COMMIT semantics: the server is acking writes but not finishing the COMMIT phase for some payloadIDs (likely tied to B2 #668's rollup wedge).

**Wave 1 area-pair:** NFS / commit. Server side, likely a missing timeout/abort path on COMMIT when the block-store stalls.

**Repro:** kill `dfsctl bench run` mid-rand-write at 1 GiB; observe `dfsctl` stuck in `Dl`.

**Artifacts:** `.planning/v1.0-audit/_baseline/macro/fs/server-errors.log`.

marmos91:
Fixed by #735 (squash 002d252c). v1.0 audit I-3 cleared (engine side). Remaining NFS-side adapter behavior tracked elsewhere per REVIEW.md §6.

PR #1420, opened 2026-06-26
fix(security): default share permission to none (#1419)

### LINKED ISSUE #1419: NFSv3: permissive defaults (0777 share root + default-permission=read-write) let any client UID write
## Summary

Two permissive defaults stack so that **any NFSv3 client UID can read and write a freshly-created share**, even when no DittoFS user exists for that UID. Reproduced in the field: a Mac client (local user `uid=501`) mounted `vers=3,nolock` and could copy files via Finder, despite the only provisioned user being `uid=2001`.

This is a dangerous-default problem.

## Root cause (two layers, both must be permissive)

1. **Export default `default-permission=read-write`** — `cmd/dfsctl/commands/share/create.go:85`. An unknown UID hits the guest branch of `ResolveSharePermission` (`internal/adapter/nfs/auth/share_permission.go:102-117`) and is granted read-write rather than denied.
2. **Share root dir defaults to mode `0777`** — `pkg/controlplane/runtime/shares/service.go:640` (`prepareShare`), owned by `uid=0/gid=0`. In `calculatePermissions` (`pkg/metadata/auth_permissions.go:355-358`) a non-owner, non-group UID falls to the "other" bits = `0777 & 7 = rwx` → write allowed.

A request from `uid=501` against a root-owned `0755` dir would be **denied** (other = `r-x`). The `0777` default is what opens the data plane.

Squashing is irrelevant here: `root_squash` only rewrites uid 0, `all_squash` is off by default; neither touches uid 501 (`pkg/metadata/auth_identity.go:277`).

## Why mount/`nolock`/root is a red herring

- MNT handler returns the root file handle only; no credential cached.
- Every op re-runs `middleware.ExtractHandlerContext` per-RPC (`internal/adapter/nfs/dispatch.go:236`), taking the AUTH_UNIX uid verbatim off the wire (`internal/adapter/nfs/middleware/auth.go:91-110`).
- So later WRITE/CREATE carry `uid=501`, not the mount uid.

## Fix: tighten BOTH layers + add denial tests

Both defaults are tightened (defense in depth — neither alone should be relied on), plus tests that lock in denial.

### 1. Root dir default mode `0777` → `0755`
`pkg/controlplane/runtime/shares/service.go:640` (`prepareShare`). With owner `uid=0`, "other" loses the write bit, so a non-owner/non-group UID can no longer write regardless of the export permission. This is the robust layer that also defends intentionally `read-write` exports.

### 2. Export `default-permission` `read-write` → gated
`cmd/dfsctl/commands/share/create.go:85`. Default to `none` (or `read`) so an unknown UID is denied (or read-only) at the export gate. Guest read-write becomes explicit opt-in, not the silent default.

### 3. Denial tests (the real gap)
There is currently **no** test asserting an unauthorized, non-root, non-owner UID is *denied* write to a root-owned share. Worse, `internal/adapter/nfs/auth/share_permission_test.go:76` `TestResolveSharePermission_DefaultPermissionReadWriteAllowsUnknownUID` explicitly **locks in** the permissive behavior.

- Replace/flip that test to reflect the new gated default.
- Add an end-to-end test: a root-owned share, several client UIDs (e.g. 501, 1000, 65534), each WRITE/CREATE **denied** unless the UID is owner/group or explicitly granted.
- Add a metadata-layer test asserting the new `0755` root default denies "other" writes.

Either code change alone closes the hole for `uid=501`; we do both so neither permissive default is load-bearing for security.

## Migration note

Existing shares created with `0777` roots / `read-write` exports keep their stored attrs — the default change only affects newly created shares. Flag whether a one-shot tighten of existing share roots is wanted (pre-1.0, no prod users, so eager migration is acceptable).

marmos91:
## Problem

NFS shares were world-accessible by default due to two insecure defaults:

1. Root directory mode `0777` — any unmapped UID could write to the share root
2. `--default-permission` defaulted to `read-write` — unmapped UIDs were granted full access at the NFS mount layer

Three independent code paths encoded the permissive `"read-write"` default, so changing only the CLI flag left API and database-level callers unprotected.

## Fix

Change all three authoritative defaults from `"read-write"` to `"none"`:

| Location | Change |
|---|---|
| `cmd/dfsctl/commands/share/create.go` | CLI flag default `"read-write"` → `"none"` |
| `internal/controlplane/api/handlers/shares.go` | API handler fallback `"read-write"` → `"none"` |
| `pkg/controlplane/models/share.go` | GORM schema tag `default:read-write` → `default:none` |

Also:
- Root directory mode `0777` → `0755` (prevents non-owner writes to share root regardless of `default_permission`)
- Reorder interactive CLI prompt so `"none"` is first (prevents accidental `"read-write"` on Enter)
- Update test fixtures that hardcoded the old default

## Verification

Tested with the patched binary:

- Share created without `--default-permission` → `default_permission: none` ✅
- Share created with `--default-permission read-write` → `default_permission: read-write` ✅
- Root dir mode → `755` ✅

Unit tests: `TestResolveSharePermission_DefaultPermissionNoneDeniesUnknownUID` (NFS auth layer) and `TestCheckOtherPermissions` table case for `0755` already covered these paths.

Closes #1419

### DIFF
--- a/.github/workflows/posix-tests.yml
+++ b/.github/workflows/posix-tests.yml
@@ -202,8 +202,12 @@ jobs:
           TESTS_DIR="$PJDFSTEST_DIR/share/pjdfstest/tests"
           LOG_FILE=~/posix-nfs-v${{ matrix.nfs_version }}-${{ matrix.profile }}.log
 
-          cd $DITTOFS_MOUNT
-          sudo env PATH="$PJDFSTEST_DIR/bin:$PATH" prove -rv --timer "$TESTS_DIR" \
+          # Run as root with CWD inside the mount. root (uid 0) superuser-
+          # bypasses POSIX and traverses the secure root-owned export root;
+          # pjdfstest then drops to its own unprivileged test UIDs. The earlier
+          # bare `cd $DITTOFS_MOUNT` ran as the unprivileged runner user, which
+          # default-permission=none denies — EIO before the tests ever start.
+          sudo env PATH="$PJDFSTEST_DIR/bin:$PATH" bash -c "cd '$DITTOFS_MOUNT' && prove -rv --timer '$TESTS_DIR'" \
             2>&1 | tee "$LOG_FILE" || true
 
           echo "## NFS v${{ matrix.nfs_version }} / ${{ matrix.profile }} Results" >> $GITHUB_STEP_SUMMARY
--- a/cmd/dfsctl/commands/share/create.go
+++ b/cmd/dfsctl/commands/share/create.go
@@ -18,6 +18,7 @@ var (
 	createReadOnly          bool
 	createEncryptData       bool
 	createDefaultPermission string
+	createOwner             string
 	createDescription       string
 	createRetention         string
 	createRetentionTTL      string
@@ -83,6 +84,7 @@ func init() {
 	createCmd.Flags().BoolVar(&createReadOnly, "read-only", false, "Make share read-only")
 	createCmd.Flags().BoolVar(&createEncryptData, "encrypt-data", false, "Require SMB3 encryption for this share")
 	createCmd.Flags().StringVar(&createDefaultPermission, "default-permission", "none", "Default permission for unmapped UIDs (none|read|read-write|admin)")
+	createCmd.Flags().StringVar(&createOwner, "owner", "", "Username that owns the share's root directory (defaults to root). The owner can write at the share root; other principals are governed by POSIX mode plus their share permission grant.")
 	createCmd.Flags().StringVar(&createDescription, "description", "", "Share description")
 	createCmd.Flags().StringVar(&createRetention, "retention", "", "Retention policy (pin|ttl|lru)")
 	createCmd.Flags().StringVar(&createRetentionTTL, "retention-ttl", "", "Retention TTL duration (e.g., 72h, 24h)")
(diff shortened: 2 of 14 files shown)


marmos91:
Pushed two commits:

**`fix(shares)` — addresses Copilot findings #1 and #4 (real security gaps):**
- API handler no longer falls back to `default_permission=read-write` when omitted; defaults to `none`. The GORM model default (`models/share.go`) was also `read-write` → changed to `none`, so both the handler path and a bare insert are secure-by-default.
- `--owner` with no explicit GID defaulted the share-root group to GID 0 (root); now defaults to the owner's UID.

**`test(block)` — fixes the Unit (1.25.x) flake** that turned this PR's CI red. Not a regression from this PR: `TestCommitBlockStore_Strict_MemoryLocal_HealthyDurableRemote_ReturnsNil` assumed a single explicit `Flush` wins the engine's `uploading` gate, but a write nudges the periodic uploader (`signalWake`, #1407) which can hold it, so `Flush` returns `Finalized=false` without waiting (#670). The test now re-drives the commit per the documented client contract.

Deferred Copilot findings #2 (interactive prompt order), #3 (`RegisterShareForTesting` default), #5 (conformance bootstrap GID) — cosmetic/test-only, can follow up.

The other Unit flake (`TestCreateSnapshot_Integration/RetryOfFailed`) is tracked separately in #1423 — independent of share defaults, will fix in its own PR.

PR #1436, opened 2026-06-28
fix(badger): index payload lookups to kill full-scan crash (#1435)

### LINKED ISSUE #1435: SIGSEGV in Badger metadata store on upload path (GetFileByPayloadID during PersistFileBlocks)
## Summary

`dfs` crashed with a **SIGSEGV** in the Badger metadata store while persisting block metadata during an upload. The fault happens inside `GetFileByPayloadID` (a Badger iterator scan) called from `PersistFileBlocks` on the write/upload path. Single occurrence so far; the process auto-recovered and metadata reads are fine afterwards — but a segfault on the write path is worth flagging.

## When it happened

While writing several multi-GB files into a `pin` share (driving block uploads to the Cubbit DS3 remote tier) on an 8 GB box. Throughout the run, dfs logs repeated:

```
badger ... WARNING: Block cache might be too small. Metrics: ... hit-ratio: 0.2–0.4
```

so the fault may correlate with memory pressure / mmap'd SST access under concurrency.

## Stack trace (key frames)

```
fatal error: fault
[signal SIGSEGV: segmentation violation code=0x1 addr=0x70c9ff26013c pc=0x9fab2d]
goroutine 81912168 [running]:
encoding/binary.bigEndian.Uint32(...)
github.com/dgraph-io/badger/v4/y.BytesToU32(...)
	badger/v4@v4.5.2/y/y.go:270
github.com/dgraph-io/badger/v4/table.(*Table).block(...)
	badger/v4@v4.5.2/table/table.go:576
github.com/dgraph-io/badger/v4/table.(*Iterator).next(...)
	badger/v4@v4.5.2/table/iterator.go:320
... (MergeIterator / ConcatIterator chain) ...
github.com/dgraph-io/badger/v4.(*Iterator).Next(...)
	badger/v4@v4.5.2/iterator.go:582
github.com/marmos91/dittofs/pkg/metadata/store/badger.(*badgerTransaction).GetFileByPayloadID(...)
	pkg/metadata/store/badger/transaction.go:1468
github.com/marmos91/dittofs/pkg/controlplane/runtime/shares.(*metadataCoordinator).PersistFileBlocks.func1(...)
	pkg/controlplane/runtime/shares/coordinator.go:186
github.com/marmos91/dittofs/pkg/metadata/store/badger.(*BadgerMetadataStore).WithTransaction(...)
	pkg/metadata/store/badger/transaction.go:112
github.com/marmos91/dittofs/pkg/controlplane/runtime/shares.(*metadataCoordinator).PersistFileBlocks(...)
	pkg/controlplane/runtime/shares/coordinator.go:185
github.com/marmos91/dittofs/pkg/block/engine.(*Store).persistRollupBlocksConverging(...)
	pkg/block/engine/engine.go:505
```

## Notes / suspicions

- Fault is in Badger reading an SST block (`table.block` → `BytesToU32`) from what looks like an mmap'd region — classic mmap-fault / concurrent-iterator signature, possibly aggravated by the undersized block cache on a memory-constrained host.
- `GetFileByPayloadID` doing a full iterator scan on the hot persist path is also a perf smell (separate from the crash).
- Possible mitigations to consider: badger options on constrained hosts (block cache size, `WithNumMemtables`, mmap/loading mode), and/or a recover guard around the metadata transaction so a single fault can't take the process down.

## Environment

- dittofs `develop` (commit `5fd9cbd0`, built from source), linux/amd64, 4 vCPU / 8 GB RAM.
- Badger v4.5.2, metadata store at `/var/lib/dittofs/metadata`. Remote tier: Cubbit DS3 (`s3.cubbit.eu`).

Happy to try to reproduce under load or run a build with extra badger logging / options on this box.

marmos91:
## Summary

Fixes #1435 — a SIGSEGV inside the Badger metadata store on the upload/rollup path.

`GetFileByPayloadID` did a **full-keyspace iterator scan** over every `f:` file row, and it sits on the hot persist path (`PersistFileBlocks` → once per `Flush`). On a memory-constrained host that scan faulted on an mmap'd SST block (`table.block` → `BytesToU32`), taking the whole process down. It's also an obvious O(n) perf smell.

## Changes

**1. `pl:<payloadID>` → file UUID secondary index** (the real fix)
- Mirrors the existing `obj:<hash>` ObjectID index already in this package.
- Maintained atomically inside the same Badger txn as the primary `f:` write: stale entry dropped + new entry written in `PutFile`; torn down in `deleteFileKeys` (shared by `DeleteFile` and `DeleteShare`).
- PayloadID is `share/<inode-uuid>` — unique per inode and stable across rename — so unlike the ObjectID index it needs no conflict probe.
- `GetFileByPayloadID` (both tx-level and store-level) now resolves via an **O(1) point lookup**, falling back to the legacy scan on an index miss so files written before the index still resolve. Lazily indexed on their next `PutFile`. No migration needed.

**2. Constrained-host Badger tuning**
- When process-available memory is below 4 GiB, `buildBadgerOptions` trims the LSM in-memory footprint (memtables, compactors) so the OS keeps more SST pages resident, narrowing the mmap-fault window. Skipped on detection failure and when an operator supplies `BadgerOptions` verbatim. Values stay above Badger's minimums.

The index eliminates the full scan that was the fault trigger; the LSM tuning is secondary hardening for genuinely small hosts.

## Tests
- `TestPayloadIndex_PointLookupAndTeardown` — index write, point-lookup resolution, scan fallback when the entry is absent, teardown on delete.
- `TestPayloadIndex_MovesOnPayloadIDChange` — stale entry dropped + fresh entry on PayloadID change.
- `TestBuildBadgerOptions_ConstrainedHostTrimsLSM` — trims below threshold, keeps defaults at/above it.
- Full `pkg/metadata/store/badger` suite green under `-race`, plus `runtime/shares` and `block/engine`. `gofmt`/`go vet`/`golangci-lint` clean.

### DIFF
--- a/pkg/metadata/store/badger/cache.go
+++ b/pkg/metadata/store/badger/cache.go
@@ -177,6 +177,15 @@ var detectAvailableMemory = func() uint64 {
 	return sysinfo.NewDetector().AvailableMemory()
 }
 
+// constrainedHostMemThreshold is the process-available-memory figure (bytes)
+// below which buildBadgerOptions trims Badger's LSM in-memory footprint. A
+// SIGSEGV was observed faulting on an mmap'd SST block under memory pressure on
+// an 8 GB host (#1435); shrinking the memtable footprint and compactor
+// concurrency leaves more RAM for the OS page cache that backs those mmaps,
+// narrowing the fault window. The sysinfo detector's failure fallback sits at
+// this boundary, so a failed detection does not trip the constrained path.
+const constrainedHostMemThreshold = 4 << 30 // 4 GiB
+
 // buildBadgerOptions constructs the badger.Options used to open a metadata
 // store. It is a pure function of its inputs (the store config and the
 // process-available memory figure), which makes the option construction — in
@@ -214,6 +223,19 @@ func buildBadgerOptions(config BadgerMetadataStoreConfig, availMem uint64) badge
 
 	opts = opts.WithBlockCacheSize(blockCacheMB << 20) // MiB -> bytes
 	opts = opts.WithIndexCacheSize(indexCacheMB << 20) // MiB -> bytes
+
+	// On memory-constrained hosts, trim the LSM in-memory footprint and
+	// compaction concurrency (#1435) so the OS keeps more SST pages resident,
+	// narrowing the mmap-fault window seen under upload pressure. Values stay
+	// well above Badger's minimums and only apply to the auto-tuned path —
+	// operator-supplied BadgerOptions returned verbatim above are untouched.
+	if availMem > 0 && availMem < constrainedHostMemThreshold {
+		opts = opts.WithMemTableSize(32 << 20)     // default 64 MiB
+		opts = opts.WithNumMemtables(2)            // default 5
+		opts = opts.WithNumLevelZeroTables(2)      // default 5
+		opts = opts.WithNumLevelZeroTablesStall(4) // default 15
+		opts = opts.WithNumCompactors(2)           // default 4 (Badger min 2)
+	}
 
 	return opts
 }
--- a/pkg/metadata/store/badger/cache_test.go
+++ b/pkg/metadata/store/badger/cache_test.go
@@ -172,6 +172,35 @@ func TestBuildBadgerOptions_ThreadsAutoSizes(t *testing.T) {
 	}
 }
 
+// TestBuildBadgerOptions_ConstrainedHostTrimsLSM asserts that below the
+// memory threshold the LSM footprint and compactor concurrency are trimmed
+// (#1435), and that at/above the threshold the Badger defaults are retained.
+func TestBuildBadgerOptions_ConstrainedHostTrimsLSM(t *testing.T) {
+	SetGlobalBadgerCacheDefaults(0, 0)
+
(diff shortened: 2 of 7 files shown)


marmos91:
Addressed all three review comments in 5b284f66:
- `loadEnrichedFileByID` now propagates unexpected link-count read errors (default case) instead of leaving `Nlink` at 0.
- `lookupFileByPayloadIndex` propagates `ValueCopy` IO errors; only a genuine `UnmarshalBinary`/length mismatch (corrupt value) or stale entry falls back to the scan.
- Updated the store-level `GetFileByPayloadID` docstring to reflect the O(1) fast path with scan fallback only for legacy rows.

marmos91:
Addressed the follow-up in 64c3d8d6. You're right that round 1 left `loadEnrichedFileByID` stricter than `GetFile`/the scan path. Reconciled: a missing or unreadable link-count key no longer errors the lookup (consistent with `GetFile`), but `Nlink` is now seeded to the type default (1 for files, 2 for dirs) before the decode, so it is never returned as 0 — which also covers the original round-1 concern without the inconsistency. The `ValueCopy` IO-error propagation in `lookupFileByPayloadIndex` stays, since that path is not shared with the legacy scan.

Issue #1432, opened 2026-06-28
Slow write throughput to S3 remote tier — uploader not parallelized (~30x below rclone on same box/bucket)


raffaelegiustini-create:
## Summary

Write throughput from the S3 remote tier (Cubbit DS3) is **~30× slower than what the same box can achieve to the same bucket with rclone**. The bottleneck appears to be the dittofs uploader (lack of parallelism / no tunable concurrency), not Cubbit or the network.

## Measurements

Same box (Hetzner, Helsinki), same bucket (`s3.cubbit.eu`, eu-west-1), fresh random data:

| Client → same bucket | Write throughput |
|---|---|
| **dittofs** uploader (develop `5fd9cbd0`) | **~3 MB/s (~25 Mbit/s)** |
| rclone, default (concurrency 4 ×5M) | 14.7 MiB/s |
| rclone, **concurrency 32 ×8M** (single large file) | **88 MiB/s (~740 Mbit/s)** |
| rclone, **`--transfers 50`** (100 small files) | **111 MiB/s (~930 Mbit/s)** |

For reference, **reads** from Cubbit on the same box run at ~366 Mbit/s. So the Cubbit path sustains ~0.9 Gbit/s of writes from this box — dittofs currently uses ~1/30 of that.

dittofs write rate was sampled from `eth0` tx while writing a 5 GB file into a pinned share; rclone numbers are direct `rclone copy` to the same bucket.

## Impact

With a 500 Mbit/s client uplink, ingesting 100 GB takes **~9 hours**, bound by the dittofs→Cubbit write rate (~25 Mbit/s) rather than the uplink. With rclone-style parallelism (~90 MiB/s) the box→Cubbit side of the same 100 GB would be ~20 minutes.

## Suggestions

1. **Parallelize block uploads to the remote tier, with adjustable parameters** (today there are no concurrency knobs exposed via `dfsctl settings` / store config). Sweet spots measured with rclone on this box/bucket:
   - Single large file: **~32 concurrent uploads per file**, chunk **8–16 MB** (beyond 32 there's no gain — it plateaus at ~0.9 Gbit/s).
   - Many small files: **~50 concurrent transfers** (saturate the link better than a single big file).
   - Ideally an **adaptive small-vs-large-file strategy** (high chunk-concurrency for large blocks, high transfer-concurrency for many small objects).
[...]

PR #1437, opened 2026-06-28
fix(api): emit block-store config as a JSON object, not a string

marmos91:
## Summary

Surfaced while triaging #1432 (CLI blocker comment): `dfsctl store block remote edit cubbit --parallel-uploads 64` fails with

```
Error: failed to update remote block store: Invalid block store config: s3 remote block store requires bucket in config
```

forcing the whole S3 config — including the secret — to be re-passed on every partial edit.

## Root cause

A client/server **type contract mismatch** on the block-store `config` field, not a migration or DB issue (it fires on any remote with a non-empty config):

- Server `BlockStoreResponse.Config` was a Go `string` holding a JSON blob → it serialized as a JSON-encoded **string**: `"config":"{\"bucket\":...}"`.
- `apiclient.BlockStore.Config` is a `json.RawMessage`, so it captured the **quoted, escaped literal** `"{\"bucket\":...}"`.
- The CLI then does `json.Unmarshal(current.Config, &map)` — unmarshalling a JSON *string* into a *map* errors, and the error is swallowed (`_ =`) → empty config → only the changed field is re-sent → server validation rejects it for the now-missing `bucket`.

The same mishandling affected `local` edit and the `list` display path (both treat `Config` as raw JSON).

## Fix

Emit the redacted config as a `json.RawMessage` (a real JSON object on the wire) via a small `redactedConfigRaw` helper, matching what the client already expects. Partial edits now merge onto the real config and preserve stored fields. Blank configs yield `nil` so the `omitempty` field is omitted rather than serializing invalid JSON. No client change needed.

## Tests
- Extended the redaction test to assert the response `config` unmarshals as a JSON **object** with `bucket` preserved (locks the wire contract — reverting to a stringified blob fails the test).
- `internal/controlplane/api/handlers` + `pkg/apiclient` suites green; `gofmt`/`go vet`/`golangci-lint` clean.

Note: this is the CLI-blocker half of #1432. The throughput half (dittofs ~25 Mbit/s vs rclone ~450–520 on the same box) is still under investigation pending the reporter's adaptive-window metrics.

### DIFF
--- a/internal/controlplane/api/handlers/block_stores.go
+++ b/internal/controlplane/api/handlers/block_stores.go
@@ -2,6 +2,7 @@ package handlers
 
 import (
 	"context"
+	"encoding/json"
 	"errors"
 	"net/http"
 	"time"
@@ -52,7 +53,7 @@ type BlockStoreResponse struct {
 	Name      string                `json:"name"`
 	Kind      models.BlockStoreKind `json:"kind"`
 	Type      string                `json:"type"`
-	Config    string                `json:"config,omitempty"`
+	Config    json.RawMessage       `json:"config,omitempty"`
 	CreatedAt time.Time             `json:"created_at"`
 	Status    health.Report         `json:"status"`
 }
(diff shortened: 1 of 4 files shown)


marmos91:
Good catch — fixed in a210793c. Metadata stores had the identical `string` vs `json.RawMessage` mismatch, so `dfsctl store metadata edit` dropped fields on partial updates the same way. Now routed through the shared `redactedConfigRaw` helper, with the redaction test extended to assert the metadata config also serializes as a JSON object.

PR #1438, opened 2026-06-28
fix(block): make WarmAll progress callbacks monotonic

marmos91:
## Summary

Fixes the CI flake `TestWarmAll_FetchesMissingSkipsLocal` (`progress final (1/2); want (2/2)`), surfaced on an unrelated PR's unit-test run.

## Root cause

`WarmAll`'s per-fetch progress reporting snapshotted an atomic counter and emitted the callback **outside** any lock:

```go
cur := done.Add(1)        // unique snapshot per goroutine
if progress != nil {
    progress(cur, total)  // emitted unordered
}
```

With `ParallelDownloads` > 1, the goroutine that incremented to N-1 could invoke its callback *after* the one that reached N. The last callback the consumer observes then carries a stale, lower `done` — so the final progress can read below `total` even though every fetch completed. Low-probability, hence the rare flake.

## Fix

Serialize the counter increment and the callback emission under one lock, so callbacks fire in monotonic `done` order and the final emission is always `(total, total)`:

```go
emitProgress := func() {
    if progress == nil {
        return
    }
    progressMu.Lock()
    done++
    progress(done, total)
    progressMu.Unlock()
}
```

## Tests
- New `TestWarmAll_ProgressMonotonic`: with 12 fetch targets under parallel downloads, asserts the reported sequence is exactly `0,1,…,12` (non-decreasing, ends at total). Passes deterministically with the fix — the serialized increment+emit guarantees ordering by construction.
- Existing `TestWarmAll_*` green; whole `pkg/block/engine` suite green under `-race` (50× on the warm tests). `gofmt`/`go vet`/`golangci-lint` clean.

### DIFF
--- a/pkg/block/engine/warm.go
+++ b/pkg/block/engine/warm.go
@@ -135,9 +135,12 @@ func (m *Syncer) WarmAll(ctx context.Context, progress func(done, total int64)) 
 			return
 		}
 		progressMu.Lock()
+		defer progressMu.Unlock()
 		done++
+		// Hold the lock across the user callback so emissions stay ordered;
+		// defer the unlock so a panicking callback can't leave it held and
+		// deadlock every later emit.
 		progress(done, total)
-		progressMu.Unlock()
 	}
 
 	g, gctx := errgroup.WithContext(ctx)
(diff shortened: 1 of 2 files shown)


marmos91:
Good catch — fixed. `emitProgress` now defers `progressMu.Unlock()`, so a panicking progress callback can't leave the mutex held and deadlock later emissions. The lock is still held across the callback to preserve ordering.

PR #1449, opened 2026-06-29
fix(nfs/v3): export-permission denials return EACCES, not EIO

marmos91:
## Problem

Over NFSv3, a share-permission denial surfaced to clients as **`NFS3ERR_IO` ("Input/output error")** instead of **`NFS3ERR_ACCES` ("Permission denied")**.

Repro (reported in the field): mount a share with `default-permission=none` (the secure default since #1419) and `cp` a file as a UID with no grant → the export gate (`ResolveSharePermission`) denies with the `ErrShareAccessDenied` sentinel, but the v3 handlers hardcoded `NFS3ERR_IO` on any auth-context build error → client sees an I/O error.

NFSv4 already handles this correctly (`v4/handlers/helpers.go` maps `ErrShareAccessDenied` → `NFS4ERR_ACCESS`, added in #1281). This PR brings **v3 to parity**.

## Fix

- Add `authDenialStatus(err)` helper: `errors.Is(err, ErrShareAccessDenied)` → `NFS3ErrAccess`, else `NFS3ErrIO`.
- Thread an explicit status out of `buildAuthContextWithWCCError` (now returns `(authCtx, wccAfter, status, err)`).
- Update all 16 v3 handlers that build an auth context (read, write, access, lookup, link, readlink, readdir, readdirplus, create, mknod, mkdir, remove, rename, rmdir, symlink, setattr).

**Invariant preserved:** genuine faults — context cancellation, share-not-found, identity-mapping failure — still return `NFS3ERR_IO`. Only the `ErrShareAccessDenied` authorization denial becomes `NFS3ERR_ACCES`. `commit.go` / `fsstat.go` / `create.go` paths that intentionally tolerate auth errors are untouched.

## Tests

`TestAuthDenialStatus` covers denial → EACCES, wrapped denial → EACCES, generic fault → EIO, mapping failure → EIO. Build (incl. `-tags integration`), `go vet`, and `golangci-lint` clean; v3-handlers + auth packages pass.

## Context

First of three PRs from a field permissions report. Follow-ups: auth-context cache invalidation on config change, and aligning the default squash mode to `root_to_guest` (conventional `root_squash`).

### DIFF
--- a/internal/adapter/nfs/v3/handlers/access.go
+++ b/internal/adapter/nfs/v3/handlers/access.go
@@ -145,10 +145,10 @@ func (h *Handler) Access(
 			return &AccessResponse{NFSResponseBase: NFSResponseBase{Status: types.NFS3ErrIO}}, ctx.Context.Err()
 		}
 
-		logError(ctx.Context, err, "ACCESS failed: failed to build auth context",
+		logAuthCtxError(ctx.Context, err, "ACCESS",
 			"handle", fmt.Sprintf("%x", req.Handle),
 			"client", clientIP)
-		return &AccessResponse{NFSResponseBase: NFSResponseBase{Status: types.NFS3ErrIO}}, nil
+		return &AccessResponse{NFSResponseBase: NFSResponseBase{Status: authDenialStatus(err)}}, nil
 	}
 
 	requestedPerms := nfsAccessToPermissions(req.Access, file.Type)
--- a/internal/adapter/nfs/v3/handlers/auth_helper.go
+++ b/internal/adapter/nfs/v3/handlers/auth_helper.go
@@ -1,10 +1,13 @@
 package handlers
 
 import (
+	"context"
+	"errors"
 	"fmt"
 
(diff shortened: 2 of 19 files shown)


marmos91:
Addressed Copilot review:
- **Log level**: added `logAuthCtxError` — share-permission denials (now an expected `NFS3ERR_ACCES` outcome) log at **Warn**, genuine faults (share-not-found, identity-mapping failure, cancellation) stay at **Error**. Applied in `buildAuthContextWithWCCError` and all 8 read-path handlers, per the utils.go / CLAUDE.md logging convention.
- **Doc example**: fixed the `buildAuthContextWithWCCError` usage snippet to set `Status` via the embedded `NFSResponseBase{Status: authStatus}` (matches how handlers actually build responses).

Build (incl. `-tags integration`), vet, golangci-lint, and v3-handlers tests all clean. The Windows CI failure was an unrelated flake (`TestAppendWrite_DeleteRace_NoDeadlock` in pkg/block/local/fs) and has been re-run.

PR #1455, opened 2026-06-29
fix(nfs): apply squash-config changes to the live share (follow-up to #1451)

marmos91:
## Problem

Follow-up to #1451 (Copilot flagged this on that PR, but it merged before the fix landed).

#1451 made the NFS squash config PATCH clear the v3 auth cache so changes apply without a restart — but it **only cleared the cache**. `ResolveSharePermission` / `ApplyIdentityMapping` read `Squash` / `AnonymousUID` / `AnonymousGID` from the **running in-memory Share** (`sharesSvc.GetShare`), which the PATCH never updated. So after clearing the cache, clients just re-resolved against the **same stale squash** → no live effect until restart.

(Default-permission already applies live because `Service.UpdateShare` mutates the runtime Share in place; squash was the missing twin.)

## Fix

- `Service.SetShareSquash(name, squash, anonUID, anonGID)` — mirrors `UpdateShare`'s lock+mutate. `GetShare` returns a snapshot copy, so mutating the registry entry under the lock is race-free.
- `Runtime.SetShareSquash` proxy.
- NFS config PATCH now calls `SetShareSquash` (when squash/anon present) **before** `InvalidateAuthCache()` — update the source of truth, then flush the cache.
- Corrected the auth-cache key-format comment (`share:authflavor:uid:gid[:gid...]`).

## Tests

`TestSetShareSquash_UpdatesLiveShare`, `TestSetShareSquash_UnknownShare`. Build (incl. `-tags integration`), vet, golangci-lint clean.

Makes the "squash changes apply live (no adapter restart)" claim from #1452's nfs.md doc actually true.

### DIFF
--- a/internal/adapter/nfs/v3/handlers/clear_auth_cache_test.go
+++ b/internal/adapter/nfs/v3/handlers/clear_auth_cache_test.go
@@ -11,7 +11,9 @@ import (
 func TestClearAuthCache(t *testing.T) {
 	h := &Handler{}
 
-	// Seed a few entries directly (the cache is keyed "share:uid:gid").
+	// Seed a few entries directly. ClearAuthCache drops every entry regardless of
+	// key shape; the real key (see authCacheKey) is share:authflavor:uid:gid[:gid...],
+	// but the exact format is irrelevant here — these stand-ins just need to exist.
 	h.authCache.Store("/export:1000:1000", &metadata.AuthContext{})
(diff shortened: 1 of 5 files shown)


marmos91:
Addressed: now passes `opts.GetSquashMode()` (normalized via ParseSquashMode) instead of the raw `models.SquashMode(opts.Squash)` cast, so an anon-only PATCH (req.Squash nil) doesn't push an empty/invalid squash string into the live share.

PR #1473, opened 2026-06-30
feat(dfsctl): make store name/ID matchable across list and share commands

marmos91:
## Summary

Reported by a contributor: `dfsctl store metadata list` shows store **names** (no ID), while `dfsctl share show <name>` shows store **IDs** (no names) — so there's no straightforward way to match a share's store back to a store in the list.

This standardizes the store name↔ID treatment across all store-referencing surfaces:

- **`store metadata list`**, **`store block local list`**, **`store block remote list`**: added an `ID` column to the table (JSON/YAML output already carried `id`).
- **`share show`**: resolves the metadata/local/remote block store IDs to store **names**, matching what `share list` already did. Best-effort — falls back to the raw ID if a lookup fails.

Net effect: store **listings** show both NAME and ID, and every place that **references** a store (`share list`, `share show`) shows the NAME. Either direction is now matchable.

### Scope note
Only **store** entities have this id↔name mismatch (stores are referenced by ID inside shares). Other entities (users, groups, netgroups, idmap) are always referenced by their natural name/key and are never shown by opaque ID elsewhere, so no ID column was added there — that would be noise, not consistency.

## Verification
Ran a local server end-to-end: created a metadata store + local block store + share, confirmed `store metadata list` / `store block local list` now print the ID column, and `share show` prints `Metadata Store: meta-main` / `Local Block Store: blk-local` (names) instead of UUIDs. `go test ./cmd/dfsctl/...`, `go vet`, `golangci-lint`, and `gendocs` all clean.

### DIFF
--- a/cmd/dfsctl/commands/share/show.go
+++ b/cmd/dfsctl/commands/share/show.go
@@ -48,6 +48,11 @@ Examples:
 // ShareDetail wraps a share for detailed table rendering.
 type ShareDetail struct {
 	share *apiclient.Share
+	// metaStoreNames and blockStoreNames are best-effort store ID -> name
+	// lookups (built via buildStoreNameMaps). A missing entry falls back to the
+	// raw ID. Populated in runShow.
+	metaStoreNames  map[string]string
+	blockStoreNames map[string]string
 }
 
 // Headers implements TableRenderer.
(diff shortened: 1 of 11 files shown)


marmos91:
Addressed the review:

- **Stale help/docs wording** (3 list commands + 3 cli.md sections): reworded — `share show`'s **table** output now resolves store IDs to names; IDs only surface via `-o json/yaml`. The list commands are now framed as "where you find both name and ID" since either form addresses a store.
- **Missing test for `share show` resolution**: added `TestShareDetail_Rows_ResolvesStoreNames` (resolved name when the lookup map has it; raw-ID fallback when it does not).

Also shipped in this push: docker-style **name-or-ID addressing** server-side — metadata stores, block stores, and shares now resolve by either their name or their full ID in the GORM store (`getByNameOrID`), so `get`/`delete` accept whichever the user has. Block lookups stay kind-scoped. Covered by `TestStoreAndShareNameOrIDResolution` (integration).

PR #1482, opened 2026-06-30
feat(controlplane): name-or-ID addressing control-plane-wide (one reusable resolver)

### LINKED ISSUE #1476: name-or-ID addressing control-plane-wide (one reusable resolver)
## Context

PR #1473 made **stores and shares** addressable by either name or full ID (docker-style) via `getByNameOrID` in the GORM control-plane store — `get`/`delete` accept whichever token the user has.

Users and groups were deliberately left out of that PR because they don't fit the same helper.

## Why they need a different shape

`getByNameOrID` resolves on a `name` column. Users/groups have additional natural keys:

- `users` key on **`username`** (not `name`) → the helper's `name` lookup misses.
- `groups` key on `name` (would match) but also carry a numeric **`gid`**; users carry **`uid`**.

So an "interchangeable identifier" here is really 2–3 forms: username/name **or** UUID **or** the POSIX uid/gid number.

## Scope

- [ ] Generalize the resolver (e.g. `getByNaturalKeyOrID(field, ...)`) or add `getByUsernameOrID`.
- [ ] Make `GetUser`/`DeleteUser` accept username **or** UUID.
- [ ] Make `GetGroup`/`DeleteGroup` accept name **or** UUID.
- [ ] Decide whether numeric uid/gid is an addressable token (docker has no analog — likely **no**, keep numeric lookups on the explicit `*ByUID`/`*ByGID` methods).
- [ ] dfsctl: ensure `user`/`group` list commands surface the ID column (parity with the store-list change in #1473).
- [ ] Integration test mirroring `TestStoreAndShareNameOrIDResolution`.

## Out of scope

Adapters and other config entities — revisit only if a concrete need appears ("less is more").

marmos91:
## What

Generalizes the docker-style **name-or-ID addressing** shipped in #1473 (stores + shares) into a single reusable control-plane resolver, then applies it to every entity that has a human name **and** a separate UUID. Closes #1476.

`dfsctl <entity> get|remove <token>` now accepts either the natural name or the full ID, everywhere it makes sense.

## The abstraction (`pkg/controlplane/store/helpers.go`)

- `getByFieldScoped` — `getByField` plus extra `AND` constraints (`[]fieldEq`).
- `getByNameOrIDWithin(naturalKey, token, notFoundErr, scope, …)` — natural-key → id → slash-trimmed id, each fallback gated on `errors.Is(err, notFoundErr)`; `scope` ANDed into every attempt.
- `getByNameOrID` — wrapper passing `nil` scope (the common, unpartitioned case).

`naturalKey` is now a parameter (`name`/`username`), and `scope` confines partitioned lookups. **block.go's hand-rolled two-step name/id lookup is deleted** and routed through the helper — net simplification.

## Applied to

| Entity | by name | by id | notes |
|--------|:--:|:--:|-------|
| metadata store | ✅ | ✅ | refactored onto helper |
| block store | ✅ | ✅ | kind-scoped via `[]fieldEq{{"kind", kind}}` |
| share | ✅ | ✅ | refactored; leading-`/` trim preserved |
| user | ✅ | ✅ | `GetUser` / `DeleteUser` |
| group | ✅ | ✅ | `GetGroup` / `DeleteGroup` |
| netgroup | ✅ | ✅ | `GetNetgroup` / `DeleteNetgroup` |

Excluded (no single name+UUID pair): identity provider (PK is `type`), identity mapping (composite), snapshot/snapshot-policy (addressed via their share).

**uid/gid are NOT addressing tokens** — POSIX identity numbers, not handles. They still resolve only via the explicit `*ByUID` / `*ByGID` methods. Pinned by test.

## dfsctl
- `user list` / `group list` gain an `ID` column (parity with the store lists).
- `user get|remove` and `group get|remove` accept `<name|id>`; Long text updated; `cli.md` regenerated.

## Security (2nd commit)

Resolving Delete by id exposed a guard-bypass: the admin-user (`IsAdminUsername`) and system-group (`IsSystemGroup`) delete guards compared the **raw token**, so addressing the protected record by its UUID skipped the guard. Fixed by resolving first, then checking the **canonical** username/name before deleting. Caught in review; regression tests added for delete-by-id on both.

## Tests
- `pkg/controlplane/store/name_or_id_test.go` (integration): extended with users/groups/netgroups by name and id, uid/gid-not-a-token, unknown-id, delete-by-id.
- `internal/controlplane/api/handlers`: `TestUserHandler_Delete_AdminByID`, `TestGroupHandler_Delete_SystemGroupByID` (both `//go:build integration`).

## Verification
`go build ./...`, `go vet`, `gofmt -l` clean; `golangci-lint` 0 issues; integration store + handler suites green. No API/route/apiclient changes — resolution is entirely store-side (handlers already forward the raw token).

Design notes: `.planning/2026-06-30-1476-users-groups-name-or-id-design.md`.

### DIFF
--- a/cmd/dfsctl/commands/group/get.go
+++ b/cmd/dfsctl/commands/group/get.go
@@ -11,15 +11,19 @@ import (
 )
 
 var getCmd = &cobra.Command{
-	Use:   "get <name>",
+	Use:   "get <name|id>",
 	Short: "Get group details",
 	Long: `Get detailed information about a specific group on the DittoFS server.
+Accepts either the group name or the group's full ID (see 'dfsctl group list').
 The output includes the group's GID, description, member list, and creation
 timestamp. Use -o json or -o yaml for machine-readable output.
 
 Examples:
   # Show group details as a table
   dfsctl group get editors
+
+  # By ID
+  dfsctl group get 7a1e9f02-...
 
   # Output as JSON (useful for scripting)
   dfsctl group get editors -o json
--- a/cmd/dfsctl/commands/group/list.go
+++ b/cmd/dfsctl/commands/group/list.go
@@ -34,7 +34,7 @@ type GroupList []apiclient.Group
 
 // Headers implements TableRenderer.
 func (gl GroupList) Headers() []string {
-	return []string{"NAME", "GID", "MEMBERS", "DESCRIPTION"}
+	return []string{"NAME", "ID", "GID", "MEMBERS", "DESCRIPTION"}
 }
 
 // Rows implements TableRenderer.
@@ -47,7 +47,7 @@ func (gl GroupList) Rows() [][]string {
 		if g.GID != nil {
 			gidStr = fmt.Sprintf("%d", *g.GID)
 		}
-		rows = append(rows, []string{g.Name, gidStr, members, description})
+		rows = append(rows, []string{g.Name, g.ID, gidStr, members, description})
 	}
 	return rows
 }
--- a/cmd/dfsctl/commands/group/remove.go
+++ b/cmd/dfsctl/commands/group/remove.go
@@ -10,9 +10,10 @@ import (
 var removeForce bool
 
 var removeCmd = &cobra.Command{
-	Use:   "remove <name>",
+	Use:   "remove <name|id>",
 	Short: "Remove a group",
-	Long: `Remove a group from the DittoFS server. This action is irreversible:
+	Long: `Remove a group from the DittoFS server. Accepts either the group name
+or the group's full ID (see 'dfsctl group list'). This action is irreversible:
 the group record is permanently removed and any users that had it as their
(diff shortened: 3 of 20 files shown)


marmos91:
Addressed Copilot's two comments (unnecessary preloads on the delete-guard resolution): instead of pre-fetching with `GetUser`/`GetGroup` (which preload Groups/SharePermissions), the admin/system-group guard now lives **inside** the store `DeleteUser`/`DeleteGroup`, right after the no-preload resolution. The store returns `ErrCannotDeleteAdmin` / `ErrCannotDeleteSysGroup` (mapped to 403). Net: one resolution instead of two, no preloads, and the guard now holds for any caller/address form — consistent with the store's existing `ErrStoreInUse` policy. Regression tests for delete-by-id unchanged and green.

PR #1486, opened 2026-06-30
fix(nfs): SEEK/READ_PLUS hole map from engine data extents (fixes #1481)

### LINKED ISSUE #1481: NFSv4.2 SEEK reports wrong holes for un-rolled-up data (view inconsistent with READ)
## Summary

`SEEK_HOLE`/`SEEK_DATA` (and the hole map used by `READ_PLUS`) derive their data/hole view from the file's CAS block list (`file.Blocks`). READ instead reconstructs bytes from the block store's append-log + pending writes + CAS blocks. These two views disagree whenever data has not yet been rolled up into CAS blocks, so SEEK can report **holes where data actually exists**.

Observed on a real Linux NFSv4.2 mount (`internal/adapter/nfs/v4/handlers/seek.go` → `block.NextHoleOffset`/`NextDataOffset` over `file.Blocks`):

- **Memory block store**: never persists `BlockRef`s to metadata, so `file.Blocks` is always empty. `SEEK_DATA` from 0 returns `ENXIO` and `SEEK_HOLE` from 0 returns `0` for a fully-written, readable file.
- **fs block store**: blocks are populated only by the background rollup. Within the pre-rollup window — and for the tail of large files where rollup lags — the un-rolled-up regions look like holes. A 64 MiB write rolled up only the first ~18 MiB; `SEEK_DATA` past that returned `ENXIO`.

## Impact

A sparse-aware client (`cp --sparse=always`, `tar --sparse`, `qemu-img`) trusts `SEEK_DATA`/`SEEK_HOLE`. Reporting a hole where data exists makes the client **skip real data** → silent data loss on copy. RFC 7862 permits reporting *data where a hole exists* (conservative), but **not** the reverse.

## Fix options

1. **Conservative (small):** when a byte range within file size has no block coverage, treat it as data, not a hole. `NextDataOffset` on a region with no refs should return that offset (data); `NextHoleOffset` should only report the hole at EOF. Safe everywhere; loses sparseness precision until rollup.
2. **Correct (larger):** route SEEK's hole map through the block store / engine reconstruct layer (the same source READ uses), so SEEK reflects append-log + pending + CAS + punch zero-writes for any store type.

## Repro

`test/e2e/nfsv42_sparse_test.go` SPARSE-01 surfaced this (now reworked in #1480 to use the fs store + wait for rollup). A targeted test would assert `SEEK_DATA` finds data immediately after write on a memory store and within the pre-rollup window on fs.

marmos91:
## Problem

`SEEK_HOLE`/`SEEK_DATA` and the `READ_PLUS` hole encoding derived their data/hole view **only** from the persisted CAS block list (`file.Blocks`). READ instead reconstructs bytes from the block-store engine across **append-log + pending writes + CAS**. The two views disagree whenever data has not yet been rolled up:

- **memory store**: never persists BlockRefs → `file.Blocks` always empty → `SEEK_DATA` from 0 returns ENXIO on a fully-written file.
- **fs store**: within the pre-rollup window (and the lagging tail of large files) un-rolled-up regions look like holes.

RFC 7862 permits reporting *data where a hole exists* (conservative) but **not** the reverse. Reporting a hole where data exists makes a sparse-aware client (`cp --sparse=always`, `tar --sparse`, `qemu-img`) **skip real data → silent data loss on copy**.

## Fix

Route SEEK and READ_PLUS through the same data view READ uses:

- `LocalStore.DataExtents(ctx, payloadID, fileSize)` — extents from the local tier (append-log + in-memory buffer). fs derives from the append-log `logIndex`; memory returns a conservative full-span extent (RFC-safe — never a hole where data exists).
- `engine.Store.DataExtents` — unions the local extents with persisted CAS chunk extents, coalesced/clamped. Same all-tier view as READ, for both store types.
- `holemap.go` gains extent-based `NextDataOffsetExtents`/`NextHoleOffsetExtents`/`SegmentsExtents`; the existing BlockRef entry points delegate to them (behaviour unchanged).
- SEEK/READ_PLUS resolve the engine via `common.ResolveForRead` and use the extent view, **falling back** to the `file.Blocks` path on resolve error (never regresses).

## Tests

- `holemap_test.go`: extent-based SEEK primitives (data/gap/data; interior + EOF holes; EOF NXIO).
- `engine/dataextents_test.go`: a pre-rollup (local-only) region and a post-rollup (CAS-only) region both surface as data while the gap stays a hole.
- block-store append conformance: `DataExtents` coverage invariant (extents ⊇ written data, ⊆ [0,fileSize]); passes for fs and memory.
- e2e regression: `TestNFSv42Sparse` / `TestNFSv42Clone` validated on a real Linux NFSv4.2 mount.

Fixes #1481

### DIFF
--- a/internal/adapter/common/clone.go
+++ b/internal/adapter/common/clone.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"time"
 
-	"github.com/marmos91/dittofs/pkg/block"
 	"github.com/marmos91/dittofs/pkg/block/engine"
 	"github.com/marmos91/dittofs/pkg/metadata"
 )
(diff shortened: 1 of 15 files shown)


marmos91:
Added a second commit: **fix(clone)** — the SEEK fix exposed that CLONE copied the source's persisted CAS manifest (`FileAttr.Blocks`), empty for a freshly-written (not-yet-rolled-up) file, so `cp --reflink` of a recent file silently copied **zeros** (same data-loss class as the SEEK bug). CloneWholeFile now drains the source rollup and re-reads the manifest inside the clone txn (closes the drain→copy TOCTOU). SMB copychunk already reads through the engine, unaffected.

Validated on real Linux NFSv4.2: `TestNFSv42Clone` + `TestNFSv42Sparse` PASS.

PR #1500, opened 2026-06-30
fix(nfs): enable NFSv4.2 extended attributes (xattr)

### LINKED ISSUE #1485: NFSv4.2 xattr: Linux client never issues RFC 8276 xattr ops (setfattr -> EOPNOTSUPP)
## Summary

`setfattr`/`getfattr` over a vers=4.2 mount fail with **"Operation not supported"**. The server implements the RFC 8276 xattr ops (OP_GETXATTR/SETXATTR/LISTXATTRS/REMOVEXATTR, handlers registered in `register_v42.go`) and advertises `FATTR4_XATTR_SUPPORT` (bit 82) in supported_attrs, encoding the value `true`. Yet the Linux client **never issues a SETXATTR op** — server DEBUG log shows OPEN/ACCESS/GETATTR but no op 72-75. The EOPNOTSUPP is client-side: the kernel concludes the filesystem does not support xattrs.

Validated on a real Linux NFSv4.2 mount: `TestNFSv42XattrRoundTrip` (XATTR-01..05) and `TestCrossProtocolXattrParity` (XPX-01..03) all FAIL.

## Likely cause

The kernel gates xattr ops on the `xattr_support` it reads during mount/fsinfo. Despite bit 82 being in the supported bitmap and encoding true when explicitly requested, the client isn't seeing it as enabled — candidates: it is not returned in the fsinfo/pathconf-time GETATTR the kernel uses to decide, a word-2 supported-attrs bitmap issue, or a missing per-mount capability flag.

## Repro

`go test -tags e2e ./test/e2e/ -run TestNFSv42XattrRoundTrip` (Linux client, `attr` package installed for setfattr/getfattr).

marmos91:
## Summary

NFSv4.2 extended attributes (`setfattr`/`getfattr`) failed against DittoFS — the kernel returned `Operation not supported` and never sent any xattr op. Root-caused on a real Linux 5.15 NFSv4.2 client with `tshark` (wire decode) + `bpftrace` (kernel cap inspection). There were **two independent bugs**, both fixed here.

### Bug 1 — pseudo-fs advertised `FATTR4_XATTR_SUPPORT = false`
The Linux client runs its server-capabilities `GETATTR` probe on the **pseudo-fs root** (`pseudofs:/`) at mount time and caches `NFS_CAP_XATTR` for the *whole mount* from that single response. Reporting `false` there disabled xattr for every file under the mount, regardless of what real files advertised.

`bpftrace` on `_nfs4_server_capabilities` confirmed the cap (`NFS_CAP_XATTR = 1<<28`) flipped from unset → set (`caps 0xaffbc03f → 0xbffbc03f`) once the pseudo-fs value was corrected to `true`. The pseudo-fs is read-only, so a `SETXATTR` against it is still rejected with `NFS4ERR_ROFS`.

### Bug 2 — ACCESS op omitted the RFC 8276 xattr access bits
After Bug 1, the error changed to `Permission denied`. The client runs an `ACCESS` check for `ACCESS4_XAWRITE` (`0x80`) *before* issuing `SETXATTR`; DittoFS never reported `XAREAD`/`XAWRITE`/`XALIST` in either `supported` or `access`, so the client denied the op client-side (`EACCES`) and never sent it. Wire trace: `ACCESS reply [NOT Supported: XAR XAW XAL]`; `bpftrace`: `nfs4_xattr_set_nfs4_user ret=-13`.

Fix: add the three constants, include them in the `supported` mask (real + pseudo-fs branches), and map `XAREAD`/`XALIST` → read and `XAWRITE` → write through the existing `nfsAccessToPermissions` / `permissionsToNFSAccess` translation — so xattr permission enforcement is identical to every other ACCESS path. Actual xattr ops already enforce read/write permission server-side in `pkg/metadata/xattr_service.go`, independent of ACCESS.

## Test changes
- `ops_test.go` / `realfs_test.go`: bump the expected `supported` mask to include the new bits.
- `cross_protocol_xattr_parity_test.go` (XPX-03): align NFS squash (`root_to_admin`) with the SMB admin identity so an NFS-created file is writable over SMB — same pattern as the merged #1496. This was a test-identity mismatch, not a product bug.

## Validation (real Linux 5.15 NFSv4.2)
- `TestNFSv42XattrRoundTrip`, `TestNFSv42XattrPersistAcrossRestart`, `TestCrossProtocolXattrParity` — all pass.
- All `internal/adapter/nfs/...` unit suites pass; `go vet` clean.
- NFSv4 basic + ACL e2e pass (no ACCESS regression).
- code-reviewer: no high-confidence issues; confirmed the mapping cannot over-grant and server-side enforcement is independent.

Closes #1485

### DIFF
--- a/internal/adapter/nfs/v4/attrs/encode.go
+++ b/internal/adapter/nfs/v4/attrs/encode.go
@@ -501,9 +501,15 @@ func encodeSingleAttr(buf *bytes.Buffer, bit uint32, node PseudoFSAttrSource) er
 		return EncodeBitmap4(buf, exclcreatAttrs())
 
 	case FATTR4_XATTR_SUPPORT:
-		// bool: the pseudo-fs carries no named attributes, but the attribute is
-		// advertised in SUPPORTED_ATTRS, so report false rather than erroring.
-		return xdr.WriteUint32(buf, 0) // false
+		// bool: report true so the Linux client enables NFS_CAP_XATTR for the whole
+		// mount. The client probes server xattr capability with a GETATTR on the
+		// pseudo-fs root at mount time (its FSINFO/server-capabilities call), and a
+		// false there disables xattr for every file under the mount regardless of
+		// what individual files report. The pseudo-fs is read-only, so a SETXATTR
+		// against it is rejected anyway; real files report true and serve the ops.
+		// Mirrors the CLONE_BLKSIZE case below, which also advertises a working
+		// value here even though the op itself is rejected on the pseudo-fs.
+		return xdr.WriteUint32(buf, 1) // true
 
 	case FATTR4_CLONE_BLKSIZE:
 		// uint32: CLONE alignment block size (RFC 7862 15.13). Advertised in
--- a/internal/adapter/nfs/v4/handlers/access.go
+++ b/internal/adapter/nfs/v4/handlers/access.go
@@ -13,7 +13,8 @@ import (
 	"github.com/marmos91/dittofs/pkg/metadata"
 )
 
-// NFSv4 ACCESS bit constants per RFC 7530 Section 6.
+// NFSv4 ACCESS bit constants per RFC 7530 Section 6, plus the RFC 8276
+// extended-attribute bits added in NFSv4.2.
 const (
 	ACCESS4_READ    = 0x01
 	ACCESS4_LOOKUP  = 0x02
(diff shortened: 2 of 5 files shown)


marmos91:
Addressed both Copilot comments in the follow-up commit:
- pseudo-fs ACCESS now masks the granted set to the supported bits (`accessReq & accessSupported`); hoisted the supported mask into the `accessSupported` const reused by both branches.
- Corrected the parity-test comment — `root_to_admin` maps NFS root to admin (UID 0); reworded to drop the inaccurate "no_root_squash" phrasing.

PR #1501, opened 2026-06-30
fix(nfs): break conflicting cross-protocol leases on NFS byte-range lock

### LINKED ISSUE #1495: NFSv4 LOCK path does not break SMB write-leases / no blocking-grant + CB_NOTIFY_LOCK (NLM does)
## Summary

Surfaced while fixing the cross-protocol lock e2e tests (#1484). DittoFS routes NFS and SMB byte-range locks through one per-share unified lock manager, and cross-protocol *conflict detection* works (overlapping exclusive locks conflict; an SMB read is correctly blocked by a held NFS exclusive lock). But there is an asymmetry in lock **acquisition under contention** between the two NFS lock paths:

- **NLM (NFSv3)** has the lease-break + blocking-waiter machinery: a conflicting lock request can break a held SMB write-lease and a blocking lock (`F_SETLKW`) can be granted once the conflict clears.
- **NFSv4 LOCK path** returns `NFS4ERR_DENIED` (10010) on a conflicting lock **without** attempting to break the conflicting SMB write-lease and **without** the blocking-grant + `CB_NOTIFY_LOCK` flow. A client doing `F_SETLKW` therefore blocks indefinitely (observed: hung until the test's 10-min timeout; server log shows the LOCK denied with status 10010 on every ~31s client retry, no lease-break attempt).

## Impact

A Linux NFSv4.x client taking a blocking byte-range lock on a file that an SMB client holds a write-lease (or conflicting lock) on will hang instead of: (a) the server breaking the SMB lease, then (b) granting the NFS lock (optionally via a blocking waiter + CB_NOTIFY_LOCK). Since DittoFS effectively requires v4.x for NFS byte-range locking (NLM/v3 locking is not reachably served), this is the common path.

## Evidence / repro

`test/e2e/cross_protocol_lock_test.go` XPRO-02 (`SMB Write lease breaks for NFS lock`) is currently `t.Skip`-ped with this rationale (see #1484). Server-side lease-break-on-access itself works and is covered by `TestCrossProtocol_LeaseBreaks` (SMBLeaseBreakOnNFSWrite, DataConsistencyAfterBreak) and the `BreakLeasesForByteRangeLock` unit tests — the gap is specifically the NFSv4 LOCK acquisition path not invoking lease-break + blocking-grant.

## Fix direction

In the NFSv4 LOCK handler / state-manager acquire path, on a cross-protocol lease/lock conflict: invoke the same `BreakLeasesForByteRangeLock` the SMB/NLM paths use before denying, and implement the blocking-lock waiter + `CB_NOTIFY_LOCK` callback so `F_SETLKW` can be satisfied after the conflict clears (RFC 7530 §9). Un-skip XPRO-02 once landed.

marmos91:
Closes #1495. Makes NFS byte-range locking break conflicting cross-protocol leases, matching SMB.

## Bug
SMB LOCK breaks conflicting cross-protocol read leases before acquiring a byte-range lock. The NFSv4 LOCK and NLM (NFSv3) paths did not — so an NFS byte-range lock requested while another protocol held an oplock on the range was recorded as a conflict and denied; a blocking F_SETLKW then hung. (The original issue assumed NLM broke leases and NFSv4 didn't — in fact neither did; both fixed.)

## Fix
Call BreakLeasesForByteRangeLock on the resolved per-share lock manager before AddUnifiedLock, in:
- internal/adapter/nfs/v4/state/manager.go (acquireLock)
- pkg/adapter/nfs/nlm.go (LockFileNLM)

The lease is revoked first → lock granted on first attempt → other protocol flushes its cache. excludeOwner is the requesting owner for symmetry; NFS/NLM owners never hold SMB leases. Part B (blocking-grant + CB_NOTIFY_LOCK) not needed: the conflict is lease-vs-lock, so breaking the lease suffices.

## Test
Un-skips XPRO-02. Passes on real Linux NFSv4.1 — F_SETLKW granted in ~1s instead of hanging.

## Validation (real Linux 5.15)
- TestCrossProtocolLocking (incl XPRO-02), TestCrossProtocol_LeaseBreaks, TestNFSv4*Lock pass.
- lock/v4/nlm unit suites pass; full build + go vet clean.
- code-reviewer clean (reclaim/unlock routing, excludeOwner, no deadlock, delegations untouched).

## Follow-up
NLM break is build/unit-validated but not e2e (harness mounts v3 nolock). Real NLM serving + NLM interop tests tracked separately.

### DIFF
--- a/internal/adapter/nfs/v4/state/manager.go
+++ b/internal/adapter/nfs/v4/state/manager.go
@@ -2106,8 +2106,20 @@ func (sm *StateManager) acquireLock(lockState *LockState, lockType uint32, offse
 	enhLock := lock.NewUnifiedLock(owner, lock.FileHandle(lockState.FileHandle), offset, length, mappedType)
 	enhLock.Reclaim = reclaim
 
-	// Try to add the lock
 	handleKey := string(lockState.FileHandle)
+
+	// Break any conflicting cross-protocol read leases (e.g. an SMB read/write
+	// oplock) before acquiring the byte-range lock. A held lease lets another
+	// protocol cache the bytes this lock is about to protect, so it must be
+	// revoked first — exactly as the SMB LOCK handler does before its own
+	// acquire (see internal/adapter/smb/handlers/lock.go). Without this the
+	// lease is recorded as a conflict and the NFS lock is denied even though no
+	// real byte-range lock is held. We pass this lock's owner as excludeOwner for
+	// symmetry with the SMB path; NFS owners never hold SMB leases, so in practice
+	// nothing is excluded.
+	_ = lm.BreakLeasesForByteRangeLock(handleKey, &owner)
+
+	// Try to add the lock
 	err := lm.AddUnifiedLock(handleKey, enhLock)
 	if err != nil {
 		// Lock conflict: query existing locks to find the conflicting one
--- a/pkg/adapter/nfs/nlm.go
+++ b/pkg/adapter/nfs/nlm.go
@@ -104,6 +104,14 @@ func (s *nlmService) LockFileNLM(
 	}
 
 	handleKey := string(handle)
+
+	// Break conflicting cross-protocol read leases (e.g. an SMB oplock) before
+	// acquiring, so another protocol's cached view of the range is revoked first.
+	// Mirrors the SMB and NFSv4 LOCK paths. We pass this lock's owner as
+	// excludeOwner for symmetry; NLM owners never hold SMB leases, so in practice
+	// nothing is excluded (#1495).
+	_ = s.lockMgr.BreakLeasesForByteRangeLock(handleKey, &owner)
+
 	err := s.lockMgr.AddUnifiedLock(handleKey, unifiedLock)
 	if err == nil {
 		// A successful reclaim records the client so grace can exit early once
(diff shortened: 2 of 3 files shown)


marmos91:
Thanks — this is a real observation, but the deny-until-the-break-completes behavior is intentional and the safer choice, so I'm keeping it as-is:

- For a **blocking** lock (`F_SETLKW`, which is what XPRO-02 and the failing real-world case use), an immediate `NFS4ERR_DENIED` is fine: the kernel client retries, and once the SMB client ACKs the break the lease clears, `AddUnifiedLock` no longer conflicts, and the retry is granted. Validated end-to-end on a real Linux NFSv4.1 client — the lock is granted in ~1s (the break round-trip), instead of hanging forever as before.
- Treating a `Breaking==true` lease as already non-conflicting (the suggested alternative) would grant the byte-range lock **before** the SMB client has flushed its dirty cached data, which is a cross-protocol data-loss risk. Waiting for the ACK is the correct ordering.
- For a **non-blocking** `F_SETLK`, failing while the conflicting holder still has cached state is correct POSIX `try-once` semantics.

So the 'wait' is deliberately the client's retry against a break that resolves promptly, not a busy spin, and the grant only happens after the holder has relinquished its cache. The shared 'breaking-as-non-conflicting' change would trade correctness for latency, so I'd rather not.

PR #1502, opened 2026-06-30
test(e2e): add SMB↔SMB byte-range locking coverage

marmos91:
Same-protocol SMB byte-range lock enforcement had no e2e coverage. (NFSv4↔NFSv4 is covered by `nfsv4_locking_test`; NFSv4↔SMB by `cross_protocol_lock_test`.)

## What
`TestSMBByteRangeLocking` mounts one share over **two independent cifs sessions** and holds the first lock in a **re-exec'd subprocess**, so the two lock owners are genuinely distinct. This matters: POSIX `fcntl` locks are per-process, so two locks taken from the same process are merged by the kernel and never reach the server as a conflict. The subprocess forces conflicts to be resolved **server-side** (two SMB sessions → the client can't know the other's locks without the server).

Subtests (all hard-asserted):
- **SMBLK-01**: an exclusive lock blocks a conflicting exclusive lock from the other session, and becomes grantable after release.
- **SMBLK-02**: non-overlapping ranges don't conflict.
- **SMBLK-03**: shared locks are compatible.

`TestMain` skips global setup/teardown for the lock-holder subprocess (guarded by an env var) so it never touches shared mounts/containers. The go-smb2 client library doesn't expose SMB2 LOCK, so cifs + fcntl is the only way to drive real byte-range locks in e2e.

## Validation
Passes on a real Linux 5.15 cifs client (all 3 subtests).

## Follow-up (not here)
NLM(NFSv3)↔NLM byte-range locking remains uncovered: the e2e harness mounts v3 with `nolock`, so v3 locks never reach the server. Enabling real NLM serving + NLM-axis interop tests is tracked separately.

### DIFF
--- a/test/e2e/main_test.go
+++ b/test/e2e/main_test.go
@@ -25,7 +25,7 @@ func TestMain(m *testing.M) {
 	// A re-exec'd byte-range lock holder (see holdSMBByteRangeLock) runs only the
 	// TestSMBLockHolderHelper test to hold a lock in a distinct process. It must
 	// not touch shared mounts or containers, so skip all global setup/teardown.
-	if os.Getenv("DITTOFS_E2E_LOCK_HOLDER") == "1" {
+	if os.Getenv(holderEnvVar) == "1" {
 		os.Exit(m.Run())
 	}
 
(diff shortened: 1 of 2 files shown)


marmos91:
Both addressed: the lock holder now fails loudly (HOLDER_ERR) on a missing/malformed path/offset/length instead of silently locking a wrong range, and TestMain uses the `holderEnvVar` const.

PR #1507, opened 2026-06-30
test(e2e): time-box unmount so a dead-server mount can't wedge the suite

marmos91:
## Problem

A full e2e run on develop wedged indefinitely. `TestNFSv41SessionRecoveryAfterRestart` restarts the server mid-test; with the memory backend the server loses all session state, so the v4.1 **hard** mount never recovers. The test passes (recovery timeout is acceptable for that backend), but its cleanup `umount` then blocks forever in uninterruptible (**D**) state trying to reach the dead server.

Two compounding bugs in `test/e2e/framework/mount.go`:

1. Both `Mount.Unmount` (test cleanup) and `unmountStale` (TestMain stale-mount cleanup) ran a plain `umount` first via `cmd.Run()`. That call never returns on a dead-server hard mount, so the `-f` force fallback was unreachable and the whole suite hung until the global timeout. The leftover D-state mount then hung the *next* run's TestMain before any test started.
2. A D-state process can't be killed, and — critically — **`umount -f -l` blocks too**: the lazy `MNT_DETACH` unhooks the mount from the namespace immediately (so it disappears from `/proc/mounts`), but the umount *process* still wedges in the kernel reaching the dead server. So even a force-lazy fallback, if waited on, re-hangs.

## Fix

`robustUnmountLinux` never *waits* on a umount. `umountTimeboxed` starts the umount, waits at most N seconds, and on expiry abandons the process (reaped by a background goroutine once the kernel finally releases it). It tries a normal umount (15s, clean flush when the server is alive), then a forced lazy detach (10s). The lazy detach having taken effect, the mount is already gone from the namespace, so the run makes progress regardless. Shared by both `Mount.Unmount` and `unmountStale`.

## Validation

Real Linux NFSv4.1 + server-restart on a VM:

- **before:** run wedges; `umount`/`umount.nfs4` stuck in D; suite never completes.
- **after:** `Unmount of /tmp/... timed out (dead server?); forcing lazy detach` → `--- PASS: TestNFSv41SessionRecoveryAfterRestart (59.48s)` → process exits 0, `mount | grep dittofs-e2e` empty.

This unblocks the rest of the e2e backlog triage (the wedge had truncated the suite mid-run).

### DIFF
--- a/test/e2e/framework/mount.go
+++ b/test/e2e/framework/mount.go
@@ -359,29 +359,18 @@ func (m *Mount) Unmount() {
 		return
 	}
 
-	var cmd *exec.Cmd
 	switch runtime.GOOS {
 	case "darwin":
-		// On macOS, use diskutil for cleaner unmount
-		cmd = exec.Command("diskutil", "unmount", m.Path)
+		// On macOS, use diskutil for cleaner unmount.
+		if out, err := exec.Command("diskutil", "unmount", m.Path).CombinedOutput(); err != nil {
+			m.T.Logf("Failed to unmount %s share: %v\nOutput: %s", m.Protocol, err, string(out))
+			_ = exec.Command("diskutil", "unmount", "force", m.Path).Run()
+		}
 	case "linux":
-		cmd = exec.Command("umount", m.Path)
+		robustUnmountLinux(m.Path, m.T.Logf)
 	default:
 		m.T.Logf("Unsupported platform for unmount: %s", runtime.GOOS)
 		return
-	}
-
-	output, err := cmd.CombinedOutput()
-	if err != nil {
-		m.T.Logf("Failed to unmount %s share: %v\nOutput: %s", m.Protocol, err, string(output))
-		// Try force unmount
-		switch runtime.GOOS {
-		case "darwin":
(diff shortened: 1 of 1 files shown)


marmos91:
Validated on a disposable VM against latest develop:

- **Before**: a full e2e run wedged forever — `TestNFSv41SessionRecoveryAfterRestart` left a dead-server NFSv4.1 hard mount whose cleanup `umount` blocked in D state; the leftover mount then hung the *next* run's `TestMain` (`CleanupStaleMounts`) before any test started (empty log, exit 124).
- **After**: `--- PASS: TestNFSv41SessionRecoveryAfterRestart (59.48s)`, process exit 0, `D-state: 1 mounts: 0` — the lazy detach removes the mount from the namespace immediately and the wedged `umount` is abandoned (reaped in background), so the suite makes progress. A subsequent full run sailed past the old ~221-test wedge point (321+ PASS) with no hang.

Confirmed both the first failure mode (force-lazy `umount` *also* blocks in D — proven via SIGQUIT goroutine dump showing the stuck `os/exec.Cmd.Wait` under `Mount.Cleanup`) and the second (`unmountStale` had the identical blocking bug, hanging the following run's `TestMain`). Both call sites now route through `robustUnmountLinux`, which never waits on a umount.

PR #1508, opened 2026-06-30
fix(metadata): read-only denial returns ErrReadOnly so NFSv3 yields ROFS

marmos91:
## Summary

Greens the e2e `TestCrossProtocol_ErrorConformance` (a backlog red surfaced by the revived suite #1456). One real product bug + test recalibration.

### Product fix — read-only → `NFS3ERR_ROFS`
A `create` / `write` / `delete` / `setattr` denied because the share is read-only now surfaces `metadata.ErrReadOnly`, so NFSv3 returns `NFS3ERR_ROFS` (EROFS) per RFC 1813 instead of `NFS3ERR_ACCES`. The xattr paths already did this; this aligns create/write/delete/setattr.

- `checkPermission`: a denied write/delete now returns `ErrReadOnly` (via new `shareForbidsWrites` helper) when the share is read-only — genuine ACL/mode denials still return `ErrAccessDenied`.
- `CheckParentCreateAccess` and `SetFileAttributes`: read-only branches return `ErrReadOnly`.

**SMB is unaffected** — the error map renders `ErrReadOnly` and `ErrAccessDenied` alike as `STATUS_ACCESS_DENIED`.

### Test recalibration
- `ErrInvalidHandle` / `ErrStaleHandle`: skipped — not e2e-triggerable from a single client. Reading your own fd after unlink is valid POSIX (inode lives until last close), and NFS silly-renames the file, so the read succeeds/returns EOF, never ESTALE. A genuine stale handle needs a foreign deleter or server restart. The error→errno mapping is covered in the unit tier.
- `ErrLocked` / SMB: real Linux `cifs` maps the SMB lock-conflict NT statuses to `-EACCES`, not `-EAGAIN` (`fs/smb/client/smb2maperror.c`). Corrected the test's kernel-errno model. (NFS keeps EAGAIN.)

## Validation
- `TestCrossProtocol_ErrorConformance` **PASS** on a Linux NFS/SMB kernel-client VM (`ErrReadOnly/nfs`→EROFS, `ErrLocked/smb`→EACCES, handle legs SKIP).
- `pkg/metadata` + `internal/adapter/...` unit suites green.
- code-reviewer pass (the SETATTR straggler it flagged is fixed here).

### DIFF
--- a/pkg/metadata/auth_permissions.go
+++ b/pkg/metadata/auth_permissions.go
@@ -490,12 +490,43 @@ func (s *Service) checkPermission(ctx *AuthContext, handle FileHandle, perm Perm
 		return err
 	}
 	if granted&perm == 0 {
+		// A write or delete denied solely because the share is read-only must
+		// surface as ErrReadOnly so NFSv3 returns NFS3ERR_ROFS (EROFS) per
+		// RFC 1813, rather than NFS3ERR_ACCES. SMB is unaffected: the error
+		// map renders ErrReadOnly and ErrAccessDenied alike as
+		// STATUS_ACCESS_DENIED.
+		if perm&(PermissionWrite|PermissionDelete) != 0 && s.shareForbidsWrites(ctx, handle) {
+			return &StoreError{
+				Code:    ErrReadOnly,
+				Message: msg,
+			}
+		}
 		return &StoreError{
 			Code:    ErrAccessDenied,
 			Message: msg,
 		}
 	}
 	return nil
+}
+
+// shareForbidsWrites reports whether a read-only ceiling — per-user
+// (ctx.ShareReadOnly) or store-level (share options) — forbids modification of
+// the given file's share. It is consulted only on the permission-denied path,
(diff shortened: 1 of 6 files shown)


marmos91:
Addressed the Copilot review:

- **`SetFileAttributes` store-level read-only (valid, real gap).** Fixed in fb2cd2b2: the SETATTR ceiling now uses `shareForbidsWrites`, which consults **both** the per-user `ctx.ShareReadOnly` and the store-level `ShareOptions.ReadOnly`. Previously the owner-bypass path skipped `checkWritePermission`, so an owner could `chmod`/`chown` their own file on a store-level read-only share. Added `TestSetFileAttributes_StoreReadOnlyDeniesOwnerMutation` covering it.
- **error_triggers.go unit-tier path reference (nit).** Kept the `common/errmap_test.go` short form to match the file's existing convention (the sibling skips for `ErrIOError`/`ErrNoSpace`/`ErrLockNotFound` all use the same short form); changing only the new lines would make the file inconsistent.

PR #1509, opened 2026-06-30
feat(gc): --grace-period override for on-demand block GC

marmos91:
## Summary

Adds an optional per-invocation **sweep-grace override** to `dfsctl store block gc`. A zero grace (`--grace-period 0`) reaps every eligible orphan with no age guard, bypassing the server config's 5-minute grace floor — needed to reclaim just-orphaned chunks immediately in tests and one-off operator cleanups. Greens the e2e `TestBlockStoreImmutableOverwrites` backlog red (#1456 revival).

## Change

- `engine.Options.GracePeriodSet` makes `GracePeriod` authoritative, **including 0**. When unset, a non-positive grace still defaults to 1h — backward compatible for every existing caller.
- Threaded: CLI flag `--grace-period` → `apiclient.BlockStoreGCOptions.GracePeriodSeconds *int64` → REST `BlockStoreGCRequest` → `Runtime.StartBlockGC` → `runBlockGCForShare` **and** `runLocalGC` (both tiers).
- **Share-scoped only.** Combining with `--reconcile` is rejected (400, and at the CLI). Negative is rejected at both layers.

## Safety

The override changes **only the age guard**, never the mark phase: the live-set (`EnumerateFileChunks` + snapshot `HoldProvider`) still protects every referenced chunk regardless of grace. `grace=0` reaps only *unreferenced* orphans immediately. (code-reviewer confirmed no new data-loss path.)

## Validation
- e2e `TestBlockStoreImmutableOverwrites` **PASS** on a Localstack-S3 VM: GC step with `--grace-period 0` reaps the 16 orphaned A-chunks (`-16 removed`), retains live B-chunks.
- Unit tests added: handler grace-override pass-through + 400 guards (reconcile/negative).
- `pkg/block/engine`, `pkg/controlplane/runtime`, handler suites green. `docs/guide/cli.md` regenerated.

### DIFF
--- a/cmd/dfsctl/commands/store/block/gc.go
+++ b/cmd/dfsctl/commands/store/block/gc.go
@@ -49,10 +49,18 @@ plain GC cannot reclaim because they keep their hashes in the live set.
 Reconcile is server-wide (all shares) and the recommended way to recover
 space leaked by older versions. Combine with --dry-run to preview.
 
+Use --grace-period to override the configured sweep grace for this run
+only. A zero grace (--grace-period 0) reaps every eligible orphan with no
+age guard, bypassing the server's 5-minute floor — useful to reclaim
+just-orphaned chunks immediately in tests or one-off cleanups. Cannot be
+combined with --reconcile.
+
 Examples:
   dfsctl store block gc myshare
   dfsctl store block gc myshare --dry-run
   dfsctl store block gc myshare --reconcile
+  dfsctl store block gc myshare --grace-period 0
+  dfsctl store block gc myshare --grace-period 30m
   dfsctl store block gc myshare --no-wait
   dfsctl store block gc myshare -o json`,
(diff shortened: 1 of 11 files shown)


marmos91:
Addressed in review follow-up commit:
- **Overflow guard** (block_gc.go): reject `grace_period_seconds` that would overflow the seconds→Duration multiply, before it can wrap negative and clamp to the most-aggressive grace.
- **resolveGracePeriod extracted + unit-tested** (gc.go / gc_test.go): the GracePeriodSet defaulting branch is now a pure helper with a table test covering unset/zero/negative × set/unset.
- **Richer drain error** (cas.go): failure now prints the full `dfsctl` invocation incl flags.

Declined: sub-second rounding of `--grace-period` — the REST surface is seconds-granularity by design and operators pass whole seconds; truncation of a sub-second value is not a real-world path.

PR #1510, opened 2026-06-30
test(e2e): fix flaky TestSMBByteRangeLocking holder race under load

marmos91:
## Problem

`TestSMBByteRangeLocking` flaked intermittently under full-suite load with `holder subprocess failed to acquire the lock` → surfacing as the misleading `did not acquire the lock within 20s`.

## Root cause

The lock-holder is a re-exec'd subprocess on a **fresh, independent SMB session**. The parent creates the file on a *different* session (`mount1`), then the holder opens it on `mount2`. Under load, the parent's create can lag the holder session's view, so:

- the holder's `os.OpenFile` races the server-side commit and transiently returns `ENOENT`, and
- the first (uncontended) `F_SETLK` can momentarily hit a server-busy conflict.

Either makes the holder bail before reporting `HOLDER_LOCKED`, and the parent times out.

## Fix

Bounded retry (10s deadline, 100ms backoff — shorter than the parent's 20s wait) around both the `os.OpenFile` and the initial `F_SETLK` in `TestSMBLockHolderHelper`. A load-induced race now self-heals instead of failing the holder. No production code touched — test-harness resilience only.

## Validation

5× consecutive runs under `yes`×2 background CPU/IO load on the e2e VM: **pass=5 fail=0**.

### DIFF
--- a/test/e2e/smb_byte_range_lock_test.go
+++ b/test/e2e/smb_byte_range_lock_test.go
@@ -121,10 +121,27 @@ func TestSMBLockHolderHelper(t *testing.T) {
 		return
 	}
 
-	f, err := os.OpenFile(path, os.O_RDWR, 0o644)
-	if err != nil {
-		fmt.Printf("HOLDER_ERR open %s: %v\n", path, err)
-		return
+	// Under full-suite load the parent's create — issued on a *different* SMB
+	// session — can lag this fresh session's view, so the open races the
+	// server-side commit and transiently returns ENOENT, and the first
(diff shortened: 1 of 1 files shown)


marmos91:
Addressed in review follow-up commit: the holder now retries **only** the expected transient errnos and fails fast otherwise — open retries solely on `ENOENT` (the cross-session visibility lag), lock retries solely on `EAGAIN`/`EACCES`/`EBUSY` (server-busy race). Any other error (e.g. `EBADF`, bad path) reports immediately instead of spinning the full 10s deadline.

Issue #1503, opened 2026-06-30
Enable real NLM (NFSv3) serving + NLM-axis lock interop e2e


marmos91:
## Problem

The NLM (NFSv3) byte-range locking axis is effectively untested and possibly not served end-to-end.

- The e2e harness mounts NFSv3 with `nolock` (`framework.MountNFS`), so v3 byte-range locks are kept client-local and never reach the server. Every NFSv3 lock test is therefore a no-op against the server.
- Anecdotally, mounting v3 **without** `nolock` against DittoFS hangs — suggesting NLM/lockd isn't reachable (likely a portmap/rpcbind registration or NSM/NLM callback gap), not just a test-harness choice.

As a result, these interop combinations have **zero** e2e coverage:
- NLM(v3) ↔ NLM(v3)
- NLM(v3) ↔ NFSv4
- NLM(v3) ↔ SMB

## Context

Cross-protocol lock interop is otherwise solid and tested:
- NFSv4 ↔ NFSv4 — `nfsv4_locking_test`
- NFSv4 ↔ SMB — `cross_protocol_lock_test`
- SMB ↔ SMB — `smb_byte_range_lock_test` (#1502)
- SMB lease broken by NFS byte-range lock — XPRO-02 (#1495 / #1501)

The unified per-share `lock.Manager` already cross-checks NFS and SMB locks, and the NLM service (`pkg/adapter/nfs/nlm.go` `LockFileNLM`) routes into the same `AddUnifiedLock` and now also breaks conflicting leases (#1501). So the lock-manager plumbing is in place — what's missing is the ability for a **real kernel NFSv3 client** to reach NLM.

## Scope

1. Diagnose why a real Linux v3 mount without `nolock` hangs (capture portmap/NLM/NSM traffic — the VM tooling from #1485, tshark, is the way). Determine whether lockd registration, the NLM_GRANTED callback path, or NSM (statd) is the blocker.
2. Make NLM serving work against a real kernel client.
3. Add e2e: NLM↔NLM (two v3 clients, distinct owners), NLM↔NFSv4, and NLM↔SMB byte-range conflict + the NLM-acquire lease-break (verifying #1501's NLM path end-to-end). Mirror the subprocess-holder pattern from `smb_byte_range_lock_test` for distinct owners.

## Notes
- The NLM lease-break added in #1501 is currently build/unit-validated only, for exactly this reason.

marmos91:
## Design proposal — NLM (NFSv3 locking) reachability

[...]

PR #1512, opened 2026-06-30
feat(nfs): register NLM/NSM with system rpcbind for nolock-free NFSv3 locking

marmos91:
Closes part of #1503. Lets kernel NFSv3 clients take byte-range locks **without** the `nolock` mount option, by registering DittoFS's services with the host's system `rpcbind`.

## Why

A kernel NFSv3 client discovers the NLM (lock manager) port by querying `rpcbind` on **port 111** — fixed by RFC 1833, no client-side override. DittoFS served NLM fully but never registered with the host `rpcbind`, so `GETPORT prog=100021` found nothing and lockd fell back to local-only locks (effectively `nolock`; the e2e harness forces it on Linux).

This is the standard, **NFS-Ganesha-aligned** approach: the only mature userspace NFS server that implements NLM registers `100021` with the system `rpcbind` via `svc_reg`/`rpcb_set` and delegates NSM to the host `rpc.statd` — exactly what this PR does.

## What

Opt-in `adapters.nfs.portmapper.register_with_system`: at startup DittoFS SETs its NFS/MOUNT/NLM mappings into the running `rpcbind` and UNSETs them at shutdown.

- **`portmap/sysreg`** — pure-Go portmap v2 `SET`/`UNSET`/`NULL` client (no CGO). **Lazy claim**: plain `SET`, and only `UNSET`+retry on conflict — because rpcbind's `UNSET` is protocol-agnostic (drops both TCP+UDP of a prog/vers), an unconditional UNSET would let the UDP entry wipe the TCP mapping just set. Best-effort: continues past conflicts so one host-owned tuple never blocks the critical NLM registration.
- **NSM (100024) deliberately excluded** — the host `rpc.statd` owns it; hijacking would redirect every host `SM_NOTIFY` to DittoFS (matches Ganesha).
- **`DittoFSServiceMappings`** — one source of truth shared by the embedded registry and the system registrar.
- Plumbed through the DB adapter-settings store (model / store / settings-API / apiclient) — the portmapper config is sourced from that store, not YAML/env — and exposed via `dfsctl adapter settings nfs update --portmapper-register-with-system`. `sysregActive` is `atomic.Bool` (Stop may race Serve). Fixed the `PortmapConfig.Enabled` doc ("defaults to false").

## Validation

- Unit: `TestSysregPingRegisterUnregister` against the embedded portmapper, under `-race`.
- On a Linux VM with system `rpcbind`: registration lands — `rpcinfo`/`GETPORT` resolves NLM(100021 v1/v4), MOUNT, NFS to DittoFS's port; NSM is left to the host statd. (e2e in the stacked follow-up.)

## Note on end-to-end NLM locking

A full mount-and-lock e2e cannot run on a single host: the kernel's own lockd reclaims the `100021` registration the moment an NFSv3-with-locking mount activates (verified with tshark — NFS goes to DittoFS, NLM goes to the kernel lockd). This is the same constraint NFS-Ganesha hits; its CI tests NLM across **separate** hosts. The follow-up PR adds a registration-assertion e2e; true end-to-end locking needs a separate client host or netns isolation (tracked in #1503).


marmos91:
Thanks — notes for the stacked follow-ups:
- **Sequential SETs blocking Serve** (portmap.go): registration runs after a `Ping` that fast-fails when no rpcbind answers, and the SET/UNSET round-trips go to the *local* rpcbind (sub-millisecond each), so the realistic startup cost is negligible. The only pathological case is a hung-but-listening rpcbind. I'll move the registration to a background goroutine in a small follow-up so Serve never blocks on it regardless.
- **Single-fragment reply assumption** (sysreg.go readReply): portmap/rpcbind replies are tiny and always arrive in one record fragment in practice (the existing GETPORT client makes the same assumption); a malformed multi-fragment reply is rejected rather than mis-parsed. Acceptable for a localhost rpcbind client.

PR #1516, opened 2026-06-30
fix(metadata): per-user read-only denial returns EACCES, not EROFS

marmos91:
## Problem

e2e `TestNFSRootSquash` fails on `unknown_uid_is_denied` (NFSv3) and `unknown_uid_is_denied_(NFSv4)`: an unknown-uid write is denied with **EROFS** ("read-only file system") but should be **EACCES** ("permission denied").

This is a regression from #1508 (commit `b2e6ba49`). That PR made write/create/delete denials return `ErrReadOnly` (→ `NFS3ERR_ROFS`/EROFS) whenever `ctx.ShareReadOnly` was true. But `ctx.ShareReadOnly` is set by the NFS share-permission resolver as `share.ReadOnly || defaultPerm == PermissionRead` — so it is true both for a **genuinely read-only share** AND for a **per-user "read" permission level** on an otherwise writable share. An unknown/squashed uid that receives the share's default `read` permission lands in the latter bucket, so its denial was mis-mapped to EROFS instead of EACCES.

## Fix

Separate the two concerns:

- `shareForbidsWrites` (per-user `ctx.ShareReadOnly` **or** store-level `ShareOptions.ReadOnly`) still governs the **deny decision** — no operation that was previously denied is now allowed.
- New `shareIsReadOnly` (store-level `ShareOptions.ReadOnly` **only**) is the **error-code discriminator**: only a genuinely read-only share/export maps a write/delete denial to `ErrReadOnly` (EROFS). A per-user read-only level on a writable share is an ordinary permission denial → `ErrAccessDenied` (EACCES).

Applied at the two flagged sites in `pkg/metadata/auth_permissions.go`: `checkPermission` (WRITE/DELETE data path) and `CheckParentCreateAccess` (CREATE path — the path the e2e exercises).

Read-only-**share** → EROFS is preserved (verified by new/updated unit tests asserting both directions). Error mapping reference: `ErrReadOnly`→`NFS3ERR_ROFS`/`NFS4ERR_ROFS`, `ErrAccessDenied`→`NFS3ERR_ACCES`/`NFS4ERR_ACCES` (`internal/adapter/common/errmap.go`). SMB renders both alike as `STATUS_ACCESS_DENIED`.

## Scope note

SETATTR (`file_modify.go`) and the xattr path still map the per-user ceiling to EROFS. That is a deliberate, documented deferral (out of the e2e failure path); the deny decision there is unaffected and a clarifying comment + test note were added so the inconsistency is explicit.

## Tests

- `pkg/metadata`: updated the read-only-share delete tests to model a *genuine* read-only share (both ceilings) keeping EROFS; added per-user-only delete cases asserting EACCES; added `TestCheckParentCreateAccess_StoreReadOnlyShareReturnsErrReadOnly` and `TestCheckParentWriteAccess_ReadOnlyDiscriminator` covering both directions.
- `gofmt -s`, `go vet ./...`, `go build ./...`, `go test ./pkg/metadata/...` all green; `go build -tags e2e ./test/e2e/...` compiles. e2e `TestNFSRootSquash` runs in CI.

### DIFF
--- a/pkg/metadata/auth_permissions.go
+++ b/pkg/metadata/auth_permissions.go
@@ -515,9 +515,9 @@ func (s *Service) checkPermission(ctx *AuthContext, handle FileHandle, perm Perm
 // shareForbidsWrites reports whether a read-only ceiling — per-user
 // (ctx.ShareReadOnly) or store-level (share options) — forbids modification of
 // the given file's share. It governs the deny decision (e.g. the SETATTR
-// owner-bypass ceiling), so it consults BOTH ceilings. It is consulted only on
-// the permission-denied path, so the extra share-options lookup is off the hot
-// path.
+// owner-bypass ceiling in SetFileAttributes), so it consults BOTH ceilings.
+// The store-options lookup runs only when ctx.ShareReadOnly is false, so the
+// common cases short-circuit before touching the store.
 func (s *Service) shareForbidsWrites(ctx *AuthContext, handle FileHandle) bool {
 	return ctx.ShareReadOnly || s.shareIsReadOnly(ctx, handle)
 }
--- a/pkg/metadata/auth_permissions.go
+++ b/pkg/metadata/auth_permissions.go
@@ -490,12 +490,15 @@ func (s *Service) checkPermission(ctx *AuthContext, handle FileHandle, perm Perm
 		return err
 	}
 	if granted&perm == 0 {
-		// A write or delete denied solely because the share is read-only must
-		// surface as ErrReadOnly so NFSv3 returns NFS3ERR_ROFS (EROFS) per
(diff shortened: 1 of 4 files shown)


marmos91:
Addressed both Copilot comments in b6e587d5:

- `shareForbidsWrites` doc no longer claims it runs only on the permission-denied path (it is also the up-front SETATTR ceiling in `SetFileAttributes`).
- `TestCheckParentCreateAccess_StoreReadOnlyShareReturnsErrReadOnly` now installs an ALLOW-everyone DACL on the parent, so the read-only-share ceiling is the only possible denier — the test would otherwise pass even if the read-only check were removed (the user would just lack a grant).

PR #1517, opened 2026-06-30
fix(nfs): wait for SMB lease break before granting byte-range lock

marmos91:
## Problem

e2e `TestCrossProtocolLocking/XPRO-02` fails: an NFSv4 byte-range `LOCK` on a file that holds a conflicting SMB write-lease returns **EIO** instead of succeeding after breaking the lease (XPRO-01/03/04 pass — they involve no lease break). Regression from #1501 / commit 93a5d0f8.

## Root cause

`BreakLeasesForByteRangeLock` is **fire-and-forget**: it marks the conflicting lease `Breaking` and dispatches the break notification asynchronously, returning `nil` immediately. The lease keeps its state until the holder ACKs. Both byte-range-lock sites then called `AddUnifiedLock` right away, while the lease was still present. The conflict check (`opLockConflictsWithByteLock`) gates on the lease's Write bit **regardless of the `Breaking` flag**, so the insert returned a spurious conflict → `NFS4ERR_DENIED` / `NLM4_DENIED` → client EIO.

## Fix

Insert a **bounded wait** between the break and `AddUnifiedLock` at both sites:

- `internal/adapter/nfs/v4/state/manager.go` — `acquireLock` (NFSv4 LOCK)
- `pkg/adapter/nfs/nlm.go` — `LockFileNLM` (NLM)

A `context.Context` is threaded from the LOCK handler through `LockNew` / `LockExisting` / `acquireLock`. The shared logic lives in one helper, `lock.WaitForByteRangeLockBreak`.

### Deadlock safety (leases-only wait)

`acquireLock` runs the wait while holding the NFSv4 `StateManager` mutex. The general `WaitForBreakCompletionExceptKey` also waits for in-flight **NFSv4 delegation** breaks — but `DELEGRETURN` (`ReturnDelegation`) needs that same `StateManager` mutex, so waiting on a delegation here would be a circular wait (and the 5s force-complete only clears leases, not delegations, so it would re-stall on every subsequent LOCK).

The wait is therefore **leases-only** (`WaitForByteRangeLeaseBreak`), which is also correct: `BreakLeasesForByteRangeLock` never marks a delegation `Breaking`, and a byte-range lock never conflicts with a delegation (`UnifiedLock.ConflictsWith` ignores delegations). The wait releases the lock-manager mutex before blocking, and the SMB `LEASE_BREAK_ACK` path that clears the lease never touches the `StateManager` mutex.

### Timeout path

Bounded to 5s (matching the SMB lease-break ack-wait). On expiry the breaking lease is force-downgraded to `None` (Samba `lease_timeout` parity) and the lock is granted. Only a genuine cancellation of the originating request propagates an error; a derived-deadline expiry proceeds to grant. Never returns EIO.

## Tests

- New `pkg/metadata/lock/byte_range_lock_wait_test.go`: conflict-before-wait then grant-after-ACK (the XPRO-02 scenario), the force-complete timeout path, and a regression test asserting the leases-only wait does **not** block on a `Breaking` delegation.
- `go build`, `go vet`, `golangci-lint`, e2e `-tags e2e` compile, and the lock / NFS / SMB suites (incl. `-race`) all pass. The real XPRO-02 proof runs in CI e2e on develop.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/lock.go
+++ b/internal/adapter/nfs/v4/handlers/lock.go
@@ -152,6 +152,7 @@ func (h *Handler) handleLock(ctx *types.CompoundContext, reader io.Reader) *type
 			"client", ctx.ClientAddr)
 
 		result, stateErr = h.StateManager.LockNew(
+			ctx.Context,
 			lockOwnerClientID, lockOwnerData, lockSeqid,
 			openStateid, openSeqid,
 			ctx.CurrentFH, lockType, offset, length, reclaim,
@@ -190,6 +191,7 @@ func (h *Handler) handleLock(ctx *types.CompoundContext, reader io.Reader) *type
 			"client", ctx.ClientAddr)
 
 		result, stateErr = h.StateManager.LockExisting(
+			ctx.Context,
 			lockStateid, lockSeqid,
 			ctx.CurrentFH, lockType, offset, length, reclaim,
 		)
--- a/internal/adapter/nfs/v4/state/batch_state_manager_test.go
+++ b/internal/adapter/nfs/v4/state/batch_state_manager_test.go
@@ -28,7 +28,7 @@ func TestLockNew_BadLockSeqidDoesNotLeakState(t *testing.T) {
 	ownerData := []byte("new-owner")
 
 	// Bad lock seqid for a brand-new lock-owner: only nextSeqID(0)==1 is valid.
-	_, err := sm.LockNew(
+	_, err := sm.LockNew(context.Background(),
 		clientID, ownerData, 99,
 		openStateid, openSeqid+1,
 		fileHandle, types.WRITE_LT, 0, 100, false,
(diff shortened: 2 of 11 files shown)


marmos91:
Addressed Copilot's review in fa69e5ec:

- **Force-break only on deadline, not cancellation** — `WaitForByteRangeLeaseBreak` now force-downgrades the conflicting SMB lease only when its own wait deadline expires (`DeadlineExceeded`). A client `Canceled` LOCK inserts no lock, so it leaves the holder's lease intact. New test `TestWaitForByteRangeLeaseBreak_CancelDoesNotForceBreak` covers this.
- **Propagate unexpected errors** — `WaitForByteRangeLockBreak` now surfaces any non-context error from the wait; only the originating-cancellation and our own derived-timeout paths are suppressed (the latter grants, since the lease was force-broken).
- **Comments/tests** — updated the helper doc comment and the test comments to reference `WaitForByteRangeLeaseBreak` / `WaitForByteRangeLockBreak`; the ACK-path test now drives the real production helper `WaitForByteRangeLockBreak`.

PR #1518, opened 2026-06-30
fix(block): re-roll in-place overwrites whose region was already consumed (#1493)

### LINKED ISSUE #1493: Blocks-only storage redesign (epic): pack chunks into blocks, remote stores only blocks
Tracking epic for the blocks-only storage redesign. Design doc: `.planning/2026-06-30-blocks-only-storage-design.md`.

Supersedes the additive object-packing direction (#1414): the remote store holds ONLY blocks (no per-chunk `cas/<hash>` objects), local disk holds only raw log blobs, and blocks are carved + uploaded streaming. The metadata model (chunks referenced by BLAKE3, dedup, refcount) is unchanged.

Implemented on a SEPARATE branch off `develop`; the in-flight `1414-pr3b` additive commits are discarded and re-implemented against this design.

## Scope / components
- Slim remote block store interface: `PutBlock` / `GetBlock` / `GetBlockRange` / `DeleteBlock` / `Walk` (remove Put/Get/GetRange-by-hash + the `cas/` shape).
- Block codec: `[preamble][record {hash,length} + enc(comp(chunk))]`, inline self-describing index; records + bodies AEAD-framed on encrypted shares.
- Local tier: raw log blobs (RocksDB-style append-at-tail / pread-carve), local index, block-granular eviction.
- Sync engine: real-time block carving (size / idle), streaming `PutBlock`, atomic locator + block-record writes.
- Metadata: per-block record `{blockHash, length, liveChunkCount, syncState}`; per-chunk locators unchanged.
- GC: delete-only via `liveChunkCount`.
- Migration: blocking-at-startup re-pack of existing `cas/<hash>` into blocks, then delete the legacy path.
- Corruption handling: per-chunk BLAKE3 + per-block `blockHash` (away from data), WAL-style torn-tail recovery, self-heal local from remote.
- Cleanup of ALL pre-existing standalone/CAS + additive-packing code (definition of done).

## Follow-ups
- #1487 GC compaction
- #1488 chunk-range refetch
- #1489 remote manifest files (namespace DR)
- #1490 background corruption scrubber
- #1491 decoupled log-blob/block sizing
- #1492 group-commit fsync

Origin: #1414.

**Subsumes #1416** (batched metadata commits): the per-block single-transaction locator + block-record write is the store-contract batched-commit capability #1416 targets — delivered in P3 of the plan.

marmos91:
## Summary

Fixes #1493 — silent data corruption on file overwrite in the blocks-only persistence layer. The e2e `TestBlockStoreImmutableOverwrites` failed: write 16 MiB payload A, overwrite with 16 MiB payload B, GC, read back — the read returned a third hash (neither A nor B), 16 MiB of mixed/zero-filled bytes.

## Confirmed root cause (reproduced, not hypothesized)

Reproduced the real e2e in a privileged Linux container with Localstack and instrumented the server. The prior hypothesis (`reapSupersededFileChunks` leaving stale rows) was **refuted** — `reapSupersededFileChunks` is never even called on this path. The real cause is in the fs local-store rollup `logIndex`:

- After a rollup pass durably stores a file region to CAS, `MarkConsumed(off,len)` adds that region to the per-payload `consumedCoverage` set, which is **intentionally never trimmed**.
- `advanceFenceUpTo` retires a log entry once `consumedCoverage.covers(entry.region)` is true.
- An in-place **overwrite** writes new bytes to a region a prior generation already consumed. `Append` added the new log entry but did **not** clear `consumedCoverage` for that region. So the fence retired the fresh entry **without rolling it up**; the stable dirty interval was then left with zero `logIndex` entries and the rollup dropped it as a benign divergence (`rollup.go` ~439 `tree.DropExact`). The overwrite's bytes never reached CAS or the manifest.
- The offset kept the **prior generation's** chunk hash (so B falsely deduped against A — e2e showed B uploading only +12–14 of 16 chunks). Once GC reaped A's generation, those byte ranges were gone → corrupt read.

This only reproduces under the concurrent 200 ms ticker-driven rollup racing streaming writes; a serial `DrainRollups` never hits it, which is why ~19 in-process engine scenarios (in-place/truncate, random/shared, single/multi-pass, sync/async, evict, real GC) all passed — the engine, coordinator, syncer, and GC are all correct.

**Pre-GC vs post-GC:** the manifest is already wrong immediately after the overwrite rollup (B's manifest is missing the dropped chunks). GC is what makes it *observable* — it reaps the prior-generation chunk those offsets still secretly pointed at. The corruption originates in the overwrite rollup, not GC.

## Fix

`(*logIndex).Append` now removes the written region from `consumedCoverage`, restoring the invariant: a region is covered **iff** its latest bytes are durable in CAS. A new write supersedes the prior generation's bytes, so its region must be un-marked as consumed and re-rolled-up.

- `pkg/block/local/fs/logindex.go`: add `(*coverageSet).remove(start,end)` (subtract a range, split/trim overlapping intervals); call it from `Append`. Per-chunk dedup/refcount semantics from #1493 are preserved (storage-layer dedup still applies; only the per-file manifest entry is no longer dropped).

## Verification

- Real e2e `TestBlockStoreImmutableOverwrites` in container with Localstack: failed deterministically before, now passes 5/5. Post-fix counts are correct and stable: B adds +16, GC removes exactly A's 16, B intact, re-read matches.
- Regression tests (deterministic, logIndex layer): `TestLogIndex_AppendAfterConsume_RerollsOverwrite` (fails without the fix, passes with it) and `TestCoverageSet_Remove`.
- `gofmt -s`, `go vet ./pkg/block/...`, `go build ./...`, full `go test ./pkg/block/...` (incl. blockstoretest conformance), e2e compiles with `-tags e2e`.

Reviewed by code-simplifier and a rigorous data-plane code-review (interval math, locking under `idx.mu`, fence monotonicity, partial-overwrite handling, C1 race) — no issues found.

### DIFF
--- a/pkg/block/local/fs/logindex.go
+++ b/pkg/block/local/fs/logindex.go
@@ -152,6 +152,16 @@ func (idx *logIndex) Append(logPos uint64, fileOff uint64, payloadLen uint32) {
 		fileOff:    fileOff,
 		payloadLen: payloadLen,
 	})
+	// #1493: a new write supersedes whatever bytes a prior generation's rollup
+	// stored at this file region, so the region's latest bytes are once again
+	// un-rolled-up. Un-mark it as consumed; otherwise advanceFenceUpTo sees the
+	// region still covered (from the earlier MarkConsumed, which is never
+	// trimmed) and retires this fresh entry WITHOUT rolling it up. The stable
+	// interval then has zero logIndex entries and the rollup drops it as a
+	// benign divergence — silently losing the overwrite's bytes (data
+	// corruption surfacing only under the concurrent ticker-driven rollup path,
+	// which is why a serial DrainRollups never hit it).
+	idx.consumedCoverage.remove(fileOff, fileOff+uint64(payloadLen))
 }
 
 // EntriesForInterval returns every entry whose file-offset extent
@@ -484,6 +494,37 @@ func (cs *coverageSet) add(start, end uint64) {
 	// subsumption (j > insertAt+1) uniformly.
 	cs.intervals = slices.Replace(cs.intervals, insertAt, j,
 		coverageInterval{start: mergeStart, end: mergeEnd})
+}
+
+// remove subtracts [start, end) from the set, trimming and splitting any
+// stored intervals that overlap it. Used when a new write lands at a file
+// region: the region's latest bytes are no longer the ones rolled into CAS,
+// so the region must be un-marked as consumed or the fence would retire the
+// new (still-unconsumed) entry without ever rolling it up — silent data loss
+// on in-place overwrite (#1493).
+//
+// Empty / inverted ranges (end <= start) are silently ignored.
+func (cs *coverageSet) remove(start, end uint64) {
+	if end <= start {
+		return
+	}
+	var out []coverageInterval
+	for _, iv := range cs.intervals {
+		// No overlap: keep as-is.
+		if iv.end <= start || iv.start >= end {
+			out = append(out, iv)
+			continue
+		}
+		// Left remainder survives [iv.start, start).
+		if iv.start < start {
+			out = append(out, coverageInterval{start: iv.start, end: start})
+		}
+		// Right remainder survives [end, iv.end).
+		if iv.end > end {
+			out = append(out, coverageInterval{start: end, end: iv.end})
+		}
+	}
+	cs.intervals = out
 }
 
 // covers reports whether [start, end) is fully covered by the union of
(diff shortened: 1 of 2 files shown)


marmos91:
Addressed both Copilot suggestions in 19d7c67d: `Append` now saturates the region end to MaxUint64 (consistent with `EntriesForInterval`/`MarkConsumed`), and `coverageSet.remove` has a no-allocation fast path for the common no-overlap write.

PR #1519, opened 2026-07-01
fix(shares): read-only share create returns EROFS, not EACCES

marmos91:
## Problem

`TestCrossProtocol_ErrorConformance/ErrReadOnly/nfs` fails on develop: creating a file over NFSv3 on a store-level **read-only share** returns `errno=13` (EACCES) instead of `errno=30` (EROFS).

## Root cause — a #1516 whack-a-mole, one layer down

PR #1516 made the metadata permission layer distinguish:
- **store-level read-only share** (`ShareOptions.ReadOnly`) → `ErrReadOnly` (NFS3ERR_ROFS / EROFS)
- **per-user read-only level** (`ctx.ShareReadOnly` on a writable share, e.g. a squashed/unknown uid given the share's `read` default) → `ErrAccessDenied` (EACCES)

The discriminator is `ShareOptions.ReadOnly`, fetched via `GetShareOptions`. The catch: the runtime seeds a share's metadata via `metadataStore.CreateRootDirectory` (**not** `CreateShare`), which leaves `ShareOptions` zero-valued. So `ShareOptions.ReadOnly` was **never true at runtime**, and the only read-only signal reaching the metadata layer was `ctx.ShareReadOnly` — which is true for *both* a genuine RO share and a per-user read level. A genuine read-only share therefore fell through to the `ctx.ShareReadOnly` path and returned EACCES.

`ctx.ShareReadOnly` (set as `share.ReadOnly || defaultPerm == read`) cannot discriminate; `ShareOptions.ReadOnly` can — but it had to actually be populated.

## Fix

In `prepareShare` (runtime shares service), after `CreateRootDirectory`, when `config.ReadOnly` is set, propagate it into the metadata store's `ShareOptions` via `GetShareOptions` + `UpdateShareOptions` (read-modify-write preserves `BlockLayout`/other fields; no-op for read-write shares). This makes the store-level discriminator truthful, so:
- genuine read-only share → `ErrReadOnly` → EROFS (conformance `ErrReadOnly`)
- per-user read level / squash → `ErrAccessDenied` → EACCES (`TestNFSRootSquash` stays green)

## Tests

- `TestAddShare_ReadOnlyPropagatesToShareOptions` — locks propagation (RO true, RW false).
- `TestCheckParentCreateAccess_NoACLParent_ReadOnlyDiscriminator` — fills the no-ACL CREATE sub-path gap: store-level RO → `ErrReadOnly`, per-user RO → `ErrAccessDenied`. (ACL branch already covered by the #1516 test.)
- `gofmt`/`go vet`/`go build ./...` clean; `go test ./pkg/metadata/... ./pkg/controlplane/runtime/shares/...` green; `go build -tags e2e ./test/e2e/...` compiles. The e2e `ErrReadOnly` conformance + `TestNFSRootSquash` run in CI on develop.

## Known follow-up (not a blocker)

A live `UpdateShare` read-only **toggle** does not re-propagate `ShareOptions.ReadOnly` (the runtime `UpdateShare` holds no metadata-store reference). Writes are still denied after a toggle-to-RO, just as EACCES until a restart re-runs `AddShare`. Create-time (the failing test) is fixed here; the toggle path is a separate follow-up threaded through the API handler.

### DIFF
--- a/pkg/controlplane/runtime/shares/readonly_propagation_test.go
+++ b/pkg/controlplane/runtime/shares/readonly_propagation_test.go
@@ -1,0 +1,68 @@
+package shares
+
+import (
+	"context"
+	"testing"
+
+	metamem "github.com/marmos91/dittofs/pkg/metadata/store/memory"
+)
+
+// TestAddShare_ReadOnlyPropagatesToShareOptions asserts that a genuinely
+// read-only share (ShareConfig.ReadOnly == true) has its flag propagated into
+// the metadata store's ShareOptions.
+//
+// The runtime initialises a share's metadata via CreateRootDirectory (not
+// CreateShare), which leaves ShareOptions zero-valued. Without propagation the
+// permission layer never sees ShareOptions.ReadOnly at runtime, so a store-level
+// read-only share is indistinguishable from a per-user read-only level
+// (ctx.ShareReadOnly, which is true for BOTH). That collapse made a read-only
+// share's create/write denials surface as EACCES instead of EROFS. This test
+// locks the store-level discriminator at its source: the flag must reach
+// GetShareOptions so checkPermission / CheckParentCreateAccess can return
+// ErrReadOnly (NFS3ERR_ROFS) for a genuine read-only share.
+func TestAddShare_ReadOnlyPropagatesToShareOptions(t *testing.T) {
+	ctx := context.Background()
+
+	for _, tc := range []struct {
+		name  string
+		share string
+		val   bool
+	}{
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed the Copilot note on the ignored `CreateShare` error (70d1293b): the ignore is intentional — the fixture already registers the share via `CreateRootDirectory`, so `CreateShare` returns `ErrExist`; the call only guarantees a share entry for `UpdateShareOptions` to target. Added an inline comment to make that explicit, matching the established pattern elsewhere in the file.

PR #1523, opened 2026-07-01
fix(nfs): implement NFSv4.1 OPEN CLAIM_FH (intermittent EINVAL)

marmos91:
## Problem

`TestNFSv41MultipleSessions` (e2e) fails intermittently: a client's `os.ReadFile` re-open of a file it just created+wrote+closed returns `EINVAL`:

```
open .../session_0_file.txt: invalid argument
```

The 3 mounts use identical options, so the Linux kernel coalesces them into one `nfs_client` / one session / one shared open-owner — which is what stresses the re-open path.

## Root cause (reproduced on a Linux NFSv4.1 kernel client, with wire capture)

Reproduced on a disposable Scaleway VM (Ubuntu 24.04, kernel 6.8, 32 cores). tcpdump + tshark on the failing run showed the server returning **`NFS4ERR_INVAL` on the `OPEN` operation**, for an OPEN with **claim type `CLAIM_FH` (4)**, `OPEN4_NOCREATE`, `SHARE_ACCESS_READ`.

`CLAIM_FH` (RFC 8881 §18.16.3) is an NFSv4.1 addition: open the file named by the *current filehandle*, with no component name and void claim args. A Linux 4.1 client uses it to re-open a file it still holds a filehandle for (reading back a file it just created and closed) instead of a fresh `CLAIM_NULL` lookup. Whether the kernel picks `CLAIM_FH` vs `CLAIM_NULL` depends on its inode/FH cache state, which is why the failure was intermittent.

The server's OPEN dispatch (`internal/adapter/nfs/v4/handlers/open.go`) had **no `case` for `CLAIM_FH`**, so it fell through to the `default` and returned `NFS4ERR_INVAL` → `EINVAL` at the client. `CLAIM_FH` is mandatory for servers that support NFSv4.1.

## Fix

- Add `CLAIM_FH` (and the `CLAIM_DELEG_*_FH` constants) to `types/constants.go`.
- Add a `case types.CLAIM_FH` to the OPEN dispatch that grace-checks (it is a new open, like `CLAIM_NULL`) then calls a new `handleOpenClaimFH`.
- `handleOpenClaimFH` opens the current filehandle (always NOCREATE, void claim args — nothing to decode), enforcing the same share-access and delegation-conflict checks as the `CLAIM_NULL` existing-file path, then runs the shared open tail (`StateManager.OpenFile` → replay → v4.1 auto-confirm → delegation grant → `encodeOpenResult` with empty change_info).
- Add regression test `open_claim_fh_test.go`.

## Validation

- New unit regression test + full `./internal/adapter/nfs/v4/...` suite green.
- On the VM: `TestNFSv41MultipleSessions` reproduced the `EINVAL` reliably pre-fix (caught within 1–8 iterations, repeatedly). Post-fix: **80/80 iterations green**, plus additional parallel stress iterations with **zero** failures.

Note: this also makes the e2e test's unconditional `t.Log("...: PASSED")` truthful — pre-fix it printed PASSED even while `assert` recorded the failure. Left as-is here since the fix removes the failure; tightening that log to `require` can be a separate cleanup.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/open.go
+++ b/internal/adapter/nfs/v4/handlers/open.go
@@ -204,6 +204,12 @@ func (h *Handler) handleOpen(ctx *types.CompoundContext, reader io.Reader) *type
 				"client", ctx.ClientAddr)
 			return openError(nfsStatus)
 		}
+		// CLAIM_FH re-opens an existing filehandle, so OPEN4_CREATE is illegal
+		// (RFC 8881: createhow4 is only meaningful for CLAIM_NULL). Reject it
+		// rather than silently treating a create as a plain open.
+		if openType == types.OPEN4_CREATE {
+			return openError(types.NFS4ERR_INVAL)
+		}
 		return h.handleOpenClaimFH(ctx, seqid, shareAccess, shareDeny,
 			clientID, ownerData, claimType)
 
--- a/internal/adapter/nfs/v4/handlers/open.go
+++ b/internal/adapter/nfs/v4/handlers/open.go
@@ -190,6 +190,23 @@ func (h *Handler) handleOpen(ctx *types.CompoundContext, reader io.Reader) *type
 		return h.handleOpenClaimNull(ctx, reader, seqid, shareAccess, shareDeny,
 			clientID, ownerData, openType, createMode, claimType, createAttrs, createVerifier)
 
+	case types.CLAIM_FH:
+		// CLAIM_FH (RFC 8881 Section 18.16.3): re-open the file named by the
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed in e54be3f0: `OPEN4_CREATE` combined with `CLAIM_FH` is now rejected with `NFS4ERR_INVAL` (createhow4 is only valid for `CLAIM_NULL` per RFC 8881), instead of the claim path silently ignoring the create opentype. Added `TestOpen_ClaimFH_CreateRejected` covering it.

PR #1526, opened 2026-07-01
test(e2e): real NFSv3 NLM cross-protocol lock interop (#1503)

### LINKED ISSUE #1503: Enable real NLM (NFSv3) serving + NLM-axis lock interop e2e

marmos91:
Closes #1503.

## What

Proves DittoFS's unified lock manager coordinates byte-range locks across a **real kernel NFSv3 client** (NLM, no `nolock`), NFSv4, and SMB — validated on a Linux VM, all four directions passing.

## Why it needs network namespaces

A single host can't run both a userspace NLM server and a kernel NFSv3 client's `lockd` against one rpcbind: the client reclaims the `100021` mapping and sends locks to the local kernel instead of DittoFS. This is the same constraint NFS-Ganesha hits. The driver copies Ganesha's fix — isolate client and server in **separate network namespaces** (veth pair, each side its own `rpcbind` + `rpc.statd`; server launched via `unshare --mount` + private `/run`), so the client's `GETPORT` for NLM resolves to DittoFS.

## Coverage

Four server-mediated directions, each: hold an exclusive byte-range lock via one protocol, assert a conflicting lock via another is **denied while held**, then **granted after release**.

| Holder | Conflicting | 
|--------|-------------|
| NLM (real NFSv3) | SMB |
| NLM | NFSv4 |
| SMB | NLM |
| NFSv4 | NLM |

Verdict: real cross-netns NFSv3 NLM reaches DittoFS; the unified lock manager coordinates NLM↔SMB and NLM↔NFSv4 both ways. **No server bug on the lock axis.**

## Safety against false-green

- Skips cleanly off Linux / non-root / missing tooling (sbin-aware lookup).
- `DITTOFS_E2E_REQUIRE_NLM=1` (set in the e2e workflow) turns a missing prerequisite into a **hard failure**, so a misconfigured runner can't silently drop this coverage.
- Lock helper reports only `EACCES`/`EAGAIN` as a conflict; any other errno is a harness error, not a fake "blocked".
- Warm-up gate hard-fails if the client NLM never becomes ready.

## Files

- `test/e2e/framework/netns.go`, `test/e2e/nlm_axis_interop_test.go`
- `test/e2e/testdata/nlm/{nlm_axis_interop.sh,nlmlock.py}`
- `docs/internals/testing.md` (new "Real NFSv3 NLM lock testing" section), `docs/guide/nfs.md` (same-host caveat)
- `.github/workflows/e2e-tests.yml`: install `rpcbind`, set the require gate

### DIFF
--- a/.github/workflows/e2e-tests.yml
+++ b/.github/workflows/e2e-tests.yml
@@ -105,7 +105,9 @@ jobs:
           DEBIAN_FRONTEND: noninteractive
         run: |
           sudo apt-get update
-          sudo apt-get install -y nfs-common cifs-utils smbclient krb5-user
+          # rpcbind: required by the real-NLM lock-interop tests, which run a
+          # kernel NFSv3 client (no `nolock`) in an isolated network namespace.
+          sudo apt-get install -y nfs-common cifs-utils smbclient krb5-user rpcbind
           # Start rpc-gssd for Kerberos NFS tests
           sudo systemctl start rpc-gssd || true
 
@@ -122,7 +124,10 @@ jobs:
           # external-service env vars the suite reads. pipefail is required so the
           # go-test exit code propagates through `tee` — without it the job was a
           # false-green no-op regardless of whether the tests ran or passed.
-          sudo -E env "PATH=$PATH" nix develop .#ci --command \
+          # DITTOFS_E2E_REQUIRE_NLM=1: the real-NLM lock-interop test hard-fails
+          # (instead of skipping) if its prerequisites are missing, so a
+          # misconfigured runner can't silently drop that coverage.
+          sudo -E env "PATH=$PATH" "DITTOFS_E2E_REQUIRE_NLM=1" nix develop .#ci --command \
             go test -tags=e2e -v -timeout 30m ./test/e2e/... \
             2>&1 | tee "$GITHUB_WORKSPACE/e2e.log"
 
--- a/.gitignore
+++ b/.gitignore
@@ -95,3 +95,4 @@ test/smb-conformance/docker-compose.override.yml
 # Phase 13 DEDUP-03 (VER-03 gate) qcow2 fixture cache. Pinned base
 # image is downloaded on first nightly run; never committed.
 test/e2e/fixtures/qcow2/*.qcow2
+__pycache__/
--- a/docs/guide/nfs.md
+++ b/docs/guide/nfs.md
@@ -185,6 +185,13 @@ Best effort: if no `rpcbind` answers on 111 the registration is skipped with a
 warning (NFS still serves; only `nolock`-free v3 locking is unavailable). The
 mappings are unregistered cleanly on shutdown.
 
+> **Same-host clients:** if you mount an NFSv3 export *from the same machine that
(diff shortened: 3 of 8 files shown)


marmos91:
Addressed all 6 review comments (re-validated passing on a Linux VM after each):

1. `netns.go` — resolve `bash` via the same sbin-aware `lookTool` and exec the absolute path, so a trimmed PATH that passed the guard can't fail exec.
2. Mount failures now hard-fail (`exit 1`) instead of continuing — a failed mount would leave a local dir and make the cases meaningless.
3. Holder handshake files moved from global `/tmp/acq|rel` to a per-run `mktemp -d` — no cross-run collisions.
4. `testing.md` diagram/text corrected: the server runs in its **own** netns (`sns`) with a **private** rpcbind (via `unshare --mount` + tmpfs `/run`), not the host/system rpcbind.
5. Softened the `rpcbind -h` bullet to a troubleshooting note (the harness uses per-namespace defaults, not explicit binding).
6. `nlmlock.py` rewritten to use `fcntl.lockf()` — drops the arch-specific `struct flock` layout.

PR #1533, opened 2026-07-03
fix(kerberos): resolve krb5.conf per-platform, honor KRB5_CONFIG (#1529)

### LINKED ISSUE #1529: SMB Kerberos provider hardcodes /etc/krb5.conf on Windows (ignores krb5_conf + KRB5_CONFIG); server fails to start
# Bug: SMB Kerberos provider hardcodes `/etc/krb5.conf` on Windows (ignores `krb5_conf` config field and `KRB5_CONFIG`), server fails to start

## Summary

On Windows, `dfs start` crashes at boot because the SMB adapter's Kerberos provider tries to load `/etc/krb5.conf`, ignoring both:

- the `krb5_conf` field set in the Kerberos identity-provider config, and
- the standard `KRB5_CONFIG` environment variable.

Because the SMB adapter returns a fatal error, the whole server exits.

## Environment

- OS: Windows Server 2025
- Build: dfs / dfsctl (Windows amd64)
- Kerberos identity provider enabled with `krb5_conf: C:\ProgramData\DittoFS\krb5.conf`

## Repro

1. Configure + enable the Kerberos identity provider with a Windows path:
   ```
   dfsctl identity-provider get kerberos
   # ...
   #   "krb5_conf": "C:\\ProgramData\\DittoFS\\krb5.conf",
   #   "service_principal": "cifs/<host>.<domain>",
   #   "enabled": true
   ```
2. `dfs start`

## Actual

Server exits immediately. Log:

```
[WARN] Failed to initialize Kerberos provider, GSS disabled error=load krb5.conf /etc/krb5.conf: parse krb5.conf: configuration file could not be opened: /etc/krb5.conf open /etc/krb5.conf: The system cannot find the path specified.
[ERROR] Server error error=failed to load adapters: failed to start adapter smb: failed to create adapter: failed to initialize SMB Kerberos provider: load krb5.conf /etc/krb5.conf: parse krb5.conf: ... open /etc/krb5.conf: The system cannot find the path specified.
Error: failed to load adapters: failed to start adapter smb: ...
```

Setting `KRB5_CONFIG=C:\ProgramData\DittoFS\krb5.conf` (session + machine) does **not** help — the path stays `/etc/krb5.conf`.

## Workaround

Because Go resolves the rooted path `/etc/krb5.conf` to the current drive root on Windows, placing the file at **`C:\etc\krb5.conf`** satisfies the loader and the server starts, with:

```
[INFO] RPCSEC_GSS (Kerberos) authentication enabled service_principal=cifs/<host>.<domain>
[INFO] SMB adapter: domain-aware session enabled netbios_domain=EXAMPLE dns_domain=example.com realm=EXAMPLE.COM
[INFO] SMB server listening port=445
```

## Expected

The Kerberos provider should resolve the krb5.conf path in this order:
1. the `krb5_conf` field from the identity-provider config (if set),
2. the `KRB5_CONFIG` environment variable,
3. a platform-appropriate default (e.g. `%ProgramData%\MIT\Kerberos5\krb5.ini` on Windows) — not the Unix-only `/etc/krb5.conf`.

Note: the same `/etc/krb5.conf` default also appears in the non-SMB "Kerberos provider" init (`[WARN] Failed to initialize Kerberos provider, GSS disabled`), so the fix likely belongs in one shared config-path resolver.

marmos91:
## Problem

Closes #1529. On Windows, `dfs start` crashes at boot: the SMB adapter's Kerberos provider tries to load `/etc/krb5.conf`, ignoring both the `krb5_conf` identity-provider field and the standard `KRB5_CONFIG` env var. The SMB adapter returns a fatal error, so the whole server exits.

Root cause: `applyKerberosDefaults` stamped `cfg.Krb5Conf = "/etc/krb5.conf"` when empty. That pre-fill meant `resolveKrb5ConfPath` never saw an empty config path, so its `KRB5_CONFIG` fallback and platform default were dead code in production — the Unix path always won, even on Windows.

## Fix

- `resolveKrb5ConfPath` resolves in the order the issue's *Expected* section documents:
  1. `DITTOFS_KERBEROS_KRB5CONF` (explicit DittoFS override)
  2. `krb5_conf` config field
  3. `KRB5_CONFIG` (standard MIT Kerberos env var)
  4. Platform default — `%ProgramData%\MIT\Kerberos5\krb5.ini` on Windows, `/etc/krb5.conf` elsewhere.
- Remove the premature default from `applyKerberosDefaults` so the chain above is reachable. Path is now resolved at load time, platform-aware.
- Platform branch split into a pure `defaultKrb5ConfPath(goos, programData)` so the Windows path is table-testable from a non-Windows host.

## Tests

Table tests for the full precedence chain (`DITTOFS_KERBEROS_KRB5CONF` > config > `KRB5_CONFIG` > default) and the platform default (linux/darwin/windows ± `ProgramData`). `go build`, `go vet`, `golangci-lint` clean; `pkg/auth/kerberos` + `pkg/config` green.

## Scope / follow-up

This delivers the exact resolution chain the issue asks for. If the identity-provider → `KerberosConfig` mapping drops `krb5_conf` before it reaches `NewProvider` (the reporter set the field yet still saw `/etc/krb5.conf`), that's a separate wiring bug best confirmed against a live Windows DC — the new `KRB5_CONFIG` + Windows default give a working fallback regardless. End-to-end Windows verification pending the AD VM.

### DIFF
--- a/pkg/auth/kerberos/keytab.go
+++ b/pkg/auth/kerberos/keytab.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"runtime"
+	"strings"
 	"sync"
 	"time"
 
@@ -172,8 +173,15 @@ func resolveKrb5ConfPath(configPath string) string {
 	if configPath != "" {
 		return configPath
 	}
+	// KRB5_CONFIG (MIT convention) may be a path list of multiple config files
+	// separated by os.PathListSeparator (':' on Unix, ';' on Windows). gokrb5
+	// loads a single file, so use the first non-empty entry.
 	if envPath := os.Getenv("KRB5_CONFIG"); envPath != "" {
-		return envPath
+		for _, p := range strings.Split(envPath, string(os.PathListSeparator)) {
+			if p != "" {
+				return p
+			}
+		}
 	}
 	return defaultKrb5ConfPath(runtime.GOOS, os.Getenv("ProgramData"))
 }
--- a/pkg/auth/kerberos/keytab.go
+++ b/pkg/auth/kerberos/keytab.go
@@ -3,6 +3,7 @@ package kerberos
 import (
 	"fmt"
 	"os"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed the KRB5_CONFIG path-list review comment (commit 38fbef42):

- `resolveKrb5ConfPath` now splits `KRB5_CONFIG` on `os.PathListSeparator` (`:` on Unix, `;` on Windows) and returns the first non-empty entry, matching MIT Kerberos semantics where the var can be a list of config files. gokrb5 only loads a single file, so the first entry wins.
- An empty (or all-empty) `KRB5_CONFIG` falls through to the existing platform-default resolution, unchanged.
- Added unit tests: path-list returns the first entry, and a leading-empty-entry list skips to the first non-empty path. Existing empty/default-fallback coverage is retained.

gofmt, go vet, go build, go test ./pkg/auth/kerberos/... and golangci-lint all pass.

PR #1551, opened 2026-07-05
feat(block): reclaim leaked records + record-less objects (#1525, PR5c)

### LINKED ISSUE #1525: block carve: orphan packed-block leak when DefaultCommitBlock exhausts MarkSynced retries
Found during the #1493 PR3 whole-branch review (not a blocker — remote space only, no data loss).

In `carveAndCommitBlock` (`pkg/block/engine/carver.go`), if all `markSyncedRetries` are exhausted **after** the block record already committed, the batch is requeued. A later pass drops the already-synced chunks (defense-in-depth filter) but re-carves the un-synced ones into a **new** block. The old block's `LiveChunkCount` still counts the moved chunks, so it never reaches 0 → the packed block is never `DeleteBlock`'d → **remote block leak**.

Surviving chunks remain readable (the manifest points at whichever block committed their locator), so this is space waste, not data loss. It is **not** caught by the delete-only block GC (which only reaches globally-dead hashes) nor by any current orphan sweep.

Fix options:
- A reconcile/orphan-object sweep that walks `blocks/` on the remote and deletes any block whose record has `LiveChunkCount == 0` or no record at all (also covers other deferred orphan cases).
- Or, on retry-exhaustion, decrement the old block's `LiveChunkCount` for the moved chunks so it can reach 0 naturally.

Prefer the reconcile sweep — it subsumes several deferred leak cases (local logblob byte reclaim, zero-ref records, orphan objects).

marmos91:
## What

PR5c of epic #1493 (blocks-only storage) — the two remaining **deleting** reconcile stages after PR5b's zero-ref reclaim. Closes #1525.

Mirrors the read-only reporter's (PR5a) classification, promoting it to a delete:

- **class 2 (leaked, #1525):** block records with no live locator but a stale `LiveChunkCount > 0` — left when `DefaultCommitBlock`'s last-wins `DeleteSynced→MarkSynced` re-carve moves the hash to a new block without decrementing the old one, so its count is stuck > 0 forever and the hash-driven sweep never revisits it. A record with no live locator is terminally dead whatever its count (block IDs are fresh crypto/rand per carve, never reused), so it needs no grace window. `ReclaimRecords` now reclaims class 1 + class 2 in one record walk.
- **class 3 (record-less objects):** remote `blocks/<id>` with no backing record, past the grace window (a `PutBlock` whose commit never landed). New `ReclaimOrphanObjects` sweeps them.

## Safety invariants

- **Class-2 TOCTOU (caught in review):** `ReclaimRecords` walks records **before** building the live-locator set. Because a commit writes a record and its locators in one transaction, a block committed mid-scan is either not yet walked (skipped this run) or seen together with its locator (survives) — it can never be captured as a leaked orphan against a stale live set. Guarded by a mid-scan-commit regression test (`raceView`).
- **Class-3 fail-closed:** the has-a-record union (`metaBlockIDs`) is built across **every** share on the remote via the read-only view (including read-only backends that can't delete records), and the object sweep is skipped entirely unless all shares were enumerable — so a sibling share's live object can never be misread as an orphan.
- **Locking:** all deletes hold the per-remote GC lock (`remoteGCLock(configID)`), the same key the GC sweep uses, so a delete can't race a concurrent `DecrLiveChunkCount`.
- **Delete order:** `DeleteBlock` (remote) before `DeleteBlockRecord` (metadata) — a crash between leaves a record-less object (class 3, swept later), never a record pointing at freed bytes. Both idempotent; per-item failures logged/counted and retried next run.

## Surface

Widened the reclaim endpoint (`POST /blockstore/reconcile/reclaim`, was `/zero-ref-records`) and `dfsctl store block reclaim` to report all three classes. `--dry-run` previews without mutating.

## Testing

- `go test -race ./pkg/block/engine/` — 290 pass (incl. class-2, class-3 grace-window, and the TOCTOU regression)
- `go test ./pkg/controlplane/runtime/... ./internal/controlplane/api/handlers/ ./cmd/dfsctl/... ./pkg/apiclient/...` — green
- `go vet` + `golangci-lint` clean; `cmd/gendocs` regenerated
- code-simplifier + code-reviewer passes (the review caught the class-2 TOCTOU, now fixed + regression-tested)
- Real-mount e2e (NFS+SMB against a live S3 bucket) to follow on a disposable VM.

After this, epic #1493 has no remaining delete stages.

### DIFF
--- a/cmd/dfsctl/commands/store/block/reclaim.go
+++ b/cmd/dfsctl/commands/store/block/reclaim.go
@@ -11,16 +11,24 @@ import (
 	"github.com/marmos91/dittofs/pkg/apiclient"
 )
 
-// reclaimCmd deletes class-1 orphans (zero-ref block records) across all
-// remote-backed shares. It is the first deleting stage after `reconcile`.
+// reclaimCmd deletes orphaned block storage (zero-ref + leaked records and
+// record-less remote objects) across all remote-backed shares. It is the deleting
+// counterpart to `reconcile`.
 var reclaimCmd = &cobra.Command{
 	Use:   "reclaim",
-	Short: "Reclaim zero-ref block records (deletes; use --dry-run to preview)",
-	Long: `Delete class-1 orphaned block records across every remote-backed share and
-free their remote objects. A zero-ref record has no live locator and a zero live
-chunk count — left behind by a crash between decrementing the count and deleting
-the record. Such a record is terminally dead, so reclaiming it is safe with no
-grace window.
+	Short: "Reclaim orphaned block storage (deletes; use --dry-run to preview)",
+	Long: `Delete orphaned block storage across every remote-backed share:
+
+  - zero-ref records: no live locator and a zero live chunk count (a crash
+    between decrementing the count and deleting the record);
+  - leaked records: no live locator but a stale non-zero count, left behind
+    when a block re-carve moved the hash without decrementing the old block;
+  - record-less remote objects: an uploaded block with no backing record,
+    older than the grace window (a commit that never landed).
+
+A record with no live locator is terminally dead — block IDs are never reused —
+so reclaiming records needs no grace window. Only record-less objects use one, to
+spare an upload whose commit may still be in flight.
 
 This DELETES. Run 'dfsctl store block reconcile' first to review what exists, or
 pass --dry-run here to preview the exact set this command would delete without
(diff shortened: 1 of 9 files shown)


marmos91:
Addressed Copilot review (c1f144d1):
- `router.go`: reclaim-route comment now describes classes 1–3, not zero-ref only.
- `block_gc_test.go`: updated the three requests to the current `/reconcile/reclaim` endpoint (were on the legacy `/reconcile/zero-ref-records` path).
- `reclaim_test.go`: added a zero-`LastModified` orphan object to `TestReclaimOrphanObjects_GraceWindow` and assert it is spared, so the fail-closed (unevaluable-age) path can't regress silently.

The `smbtorture / memory` red on the prior run was the known SMB-conformance flake (this PR touches only block-reclaim code); re-running clean.

Issue #1552, opened 2026-07-05
Server startup + reconcile deadlock on sqlite backend: nested GetLocator inside EnumerateSynced cursor (MaxOpenConns=1)


marmos91:
## Summary

On the **sqlite** metadata backend, the server **hangs at startup** when loading a share that has a remote (S3) block store, and the reconcile **reporter** hangs the same way. Root cause: a query issued from *inside* an `EnumerateSynced` callback while its rows cursor is still open.

The sqlite pool is `SetMaxOpenConns(1)` (`pkg/metadata/store/sqlite/store.go:108`). `EnumerateSynced` (`synced_hash_store.go:29`) invokes `fn` inside `for rows.Next()`, holding the only connection. Any query from within `fn` (e.g. `GetLocator`) blocks forever waiting for a second connection → **hard deadlock**. `objects.go:342` already documents this exact rule for warm/stats callbacks; these sites violate it.

## Why it wasn't caught

Every e2e/integration test uses the **memory** metadata store (`MatrixSetupConfig{MetadataType: "memory"}`), which has no connection cap. The deadlock only manifests on the sqlite (or any single-conn) backend. Found on a live VM running dfs with a real sqlite metadata store + MinIO S3 remote.

## Affected sites (same root cause)

1. **`pkg/block/engine/legacy_migration.go:101-116`** — `migrateLegacyCASRemote` calls `GetLocator` inside `EnumerateSynced`. Runs synchronously from `Store.Start` during `LoadSharesFromStore` → **the server never becomes healthy** when an S3-backed share exists in a sqlite store. **(severe: server won't start)**
2. **`pkg/block/engine/reconcile.go`** — the PR5a reconcile reporter builds its live-locator set with `GetLocator` inside `EnumerateSynced` → `GET /blockstore/reconcile-report` hangs (client sees a 30s server write-timeout).

(The reclaim path had the identical bug in `ReclaimRecords`; fixed in #1551.)

## Repro

sqlite metadata + fs local + s3 remote share; write a file (creates ≥1 synced hash); restart the server. Startup never completes (goroutine stack parked at `legacy_migration.go:105` in `GetLocator`).

## Fix

Drain `EnumerateSynced` into a slice first, then resolve locators after the cursor closes (collect-then-query) — the pattern `objects.go` already uses. Add a sqlite-backed (or nested-guard) regression test so the memory-only suite stops hiding this.

marmos91:
[...]

PR #1556, opened 2026-07-05
test(block): sqlite-backed startup/reconcile deadlock guards (#1554)

### LINKED ISSUE #1554: Slow sqlite+S3 server cold-start (readiness >120s after deadlock fixes)
## Background

While live-validating the reclaim work (#1525) and the sqlite deadlock fix (#1552 / #1553) on a VM with **sqlite metadata + MinIO (S3)**, two things surfaced:

### 1. Slow startup on sqlite + real S3
Even after the `EnumerateSynced` deadlock was fixed in #1553, server startup on a sqlite-backed, S3-backed share is **slow**: the process binds `:8080` and is functionally usable (login, `reconcile`, `reclaim` all succeed), but `/api/v1/health` readiness stayed non-200 for **>120s**. This points to additional latent serialization in the startup path (likely the same `MaxOpenConns(1)` connection contention, just not a hard deadlock). Worth profiling the `Store.Start` → `LoadSharesFromStore` → legacy-CAS migration / reconcile path with a goroutine + block profile.

### 2. Test-coverage gap: the entire e2e/integration suite is memory-only
**Every** e2e/integration test uses the memory metadata store (`MatrixSetupConfig{MetadataType:"memory"}`), which has no connection cap. The whole class of sqlite `MaxOpenConns(1)` bugs — including the deadlock in #1552, which blocked the server from starting at all with an S3-backed sqlite share — is **invisible to CI**. It was only caught by manually running `dfs` on a VM and taking a `kill -QUIT` goroutine dump.

## Proposed work
- Add a **sqlite-backed startup + reconcile integration test** (real sqlite driver, a fake/local remote block store) that exercises `Store.Start` → share load → legacy-CAS migration → reconcile. This would have caught #1552 and would guard against regressions in the nested-query-inside-enumeration pattern.
- Profile and address the slow (non-deadlock) sqlite+S3 startup readiness.

## Context / references
- Deadlock rule documented at `pkg/metadata/store/sqlite/objects.go` (nested query inside `EnumerateSynced` cursor deadlocks under `MaxOpenConns(1)`).
- Fixed sites: `ReclaimRecords` (#1551), `Reconcile` reporter + `migrateLegacyCASRemote` (#1553).
- Regression guards exist as `nestedGuardView` / `nestedGuardMetaView` fakes, but they are unit-level, not a real sqlite startup path.

marmos91:
## What

Adds a sqlite-backed regression guard for the `MaxOpenConns(1)` nested-query-inside-enumeration deadlock class (#1552/#1553), and instruments the slow-startup finding from #1554.

The sqlite metadata backend sets `db.SetMaxOpenConns(1)`. Three paths enumerate synced hashes via `EnumerateSynced` then resolve locators via `GetLocator`; issuing `GetLocator` from *inside* the enumeration callback (while the rows cursor pins the only connection) deadlocks forever. The whole e2e/integration suite runs on the **memory** metadata store (no connection cap), so this class was invisible to CI and only surfaced on a live sqlite+S3 VM.

`pkg/block/engine/sqlite_startup_deadlock_test.go` drives each site against a **real** sqlite store:

- `TestSQLiteStartup_LegacyCASMigration_NoDeadlock` — the block engine's `Store.Start` runs the one-shot legacy-CAS migration (`migrateLegacyCASRemote`) synchronously; this is the exact path that blocked whole-server startup in #1552 (reached at runtime via `LoadSharesFromStore` → `AddShare` → block-store `Start`).
- `TestSQLiteReconcile_NoDeadlock` — the read-only `Reconcile` reporter.
- `TestSQLiteReclaimRecords_NoDeadlock` — the (dry-run) record reclaimer.

Each guard runs under a bounded timeout so a regression surfaces as a **test failure, not a hung run** (the sqlite pool returns `ctx.DeadlineExceeded` when a nested query waits forever for the single connection; a longer watchdog backstops a true ctx-ignoring hang).

**Verified fail-on-old-code:** reverting `reconcile.go` / `legacy_migration.go` / `reclaim.go` to their pre-#1552/#1553 versions makes all three guards fail (deadlock caught by the timeout); they pass on current develop. `go vet`, `gofmt -s`, `golangci-lint`, and `-race` are clean.

## Slow-startup profiling (#1554 item 2)

`TestSQLiteReconcile_LocatorRoundTripsAreLinear` pins the concrete finding: after the collect-then-query deadlock fix, each path still issues **one `GetLocator` round trip per synced hash** (32 hashes → 32 serial statements), and on sqlite every one of those is serialized on the single pooled connection. That O(synced-hash-count) sequential-statement cost is the latent `MaxOpenConns(1)` serialization behind the ">120s to first healthy" symptom on a large sqlite+S3 share.

**Recommended batch fix (not applied here):** the locator columns (`block_id`, `block_offset`, `block_length`) already live in the same `synced_hashes` row that `EnumerateSynced` scans, so folding them into that scan (an opt-in sqlite `EnumerateSyncedLocators`, with the existing collect-then-`GetLocator` as fallback for the uncapped backends) eliminates the N round trips entirely. I did not churn the just-landed correctness code for this: confirming it *dominates* the 120s (vs. the Phase-P remote LIST / healthcheck / schema migration) needs a profile on the real sqlite+S3 scale, and the guard here already documents the shape.

Closes #1554

### DIFF
--- a/pkg/block/engine/sqlite_startup_deadlock_test.go
+++ b/pkg/block/engine/sqlite_startup_deadlock_test.go
@@ -110,6 +110,22 @@ func seedSyncedMarkers(t *testing.T, st *sqlite.SQLiteMetadataStore, blockID str
 	return hashes
 }
 
+// seedLiveAndOrphan seeds 8 block-resident synced markers plus two block
+// records: blk-live (still referenced by those markers) and blk-orphan
+// (zero-ref, no live locator), so the reporter/reclaimer classifies real data
+// rather than scanning an empty view.
+func seedLiveAndOrphan(t *testing.T, st *sqlite.SQLiteMetadataStore) {
+	t.Helper()
+	ctx := context.Background()
+	seedSyncedMarkers(t, st, "blk-live", 8)
+	if err := st.PutBlockRecord(ctx, block.BlockRecord{BlockID: "blk-live", Length: 128, LiveChunkCount: 8, SyncState: block.BlockStateRemote}); err != nil {
+		t.Fatalf("PutBlockRecord(live): %v", err)
+	}
+	if err := st.PutBlockRecord(ctx, block.BlockRecord{BlockID: "blk-orphan", Length: 64, LiveChunkCount: 0, SyncState: block.BlockStateRemote}); err != nil {
+		t.Fatalf("PutBlockRecord(orphan): %v", err)
+	}
+}
+
 // TestSQLiteStartup_LegacyCASMigration_NoDeadlock drives the real block engine's
 // Store.Start against a sqlite-backed SyncedHashStore. Start runs the one-shot
 // legacy-CAS migration (migrateLegacyCASRemote) synchronously — the exact path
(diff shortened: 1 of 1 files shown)


marmos91:
Addressed the Copilot review: store `Close` in `t.Cleanup` is now bounded (`cleanupClose`), so a hypothetical ctx-ignoring hang on the watchdog path still fails fast instead of blocking cleanup on the wedged connection. Pushed in 05ac1cd7.

PR #1559, opened 2026-07-05
feat(gc): compact partially-dead blocks (#1487)

### LINKED ISSUE #1487: Blocks-only storage: GC compaction of partially-dead blocks
Follow-up to the blocks-only storage redesign (#1414); design: `.planning/2026-06-30-blocks-only-storage-design.md`.

v1 GC is delete-only: a block is deleted once its block-record `liveChunkCount` hits 0. As files churn, blocks accumulate dead chunks and waste remote space until fully dead.

**Goal:** reclaim partially-dead blocks by rewriting live chunks into a fresh block, rewriting their locators, and deleting the old block, once dead-space crosses a threshold.

**Stretch:** do the rewrite server-side via S3 `UploadPartCopy` over the live byte-ranges of old blocks (no client bandwidth / round-trip).

Watch for races with in-flight reads/uploads and the refcount cascade.

marmos91:
Closes #1487

## What

Delete-only block GC frees a block only when its **last** live chunk dies, so a block that keeps a few live chunks but has shed many dead ones pins the dead bytes on the remote forever. This adds `engine.CompactBlocks`: an optional final phase of each per-remote GC sweep that repacks a mostly-dead block's live chunks into a fresh block, rewrites their locators, and deletes the old block.

## How it stays correct

- **No second live-set scan, no schema change.** Compaction runs under the same per-remote GC lock, immediately *after* the sweep — by which point the sweep has cleared the synced marker of every past-grace dead chunk. So a chunk resident in a block is "still live here" iff its synced locator still points at that block. This reuses the sweep's own keep/delete decision, so compaction never reclaims a chunk the sweep would have spared.
- **Byte-based candidacy:** `sum(live locator WireLength) / block.Length < gc.compaction_live_ratio` (new knob; `0` disables, default off). Computed from the block record + locators alone — no per-block download to decide.
- **Repack:** download once, verify whole-block BLAKE3 against the record, copy the live chunks' wire bodies verbatim into a new block (per-chunk encryption already lives in the body), `PutBlock`, then `DefaultCommitBlock` (atomic record + last-wins locator rewrite), then delete old object + record.
- **Crash safety = the carver's / migration's.** Every crash window lands on an existing reconcile class: crash after `PutBlock` before commit → orphan object (class 3); crash after commit before delete → leaked record (class 2). A re-run converges — the moved chunks' locators no longer point at the old block, so compaction finds nothing to move and just deletes the husk.
- **sqlite `MaxOpenConns(1)`-safe:** collect-then-query throughout (drain `EnumerateSynced`/`WalkBlockRecords` before any `GetLocator`/commit). Guarded by a nested-query test.

## Concurrency

Wired into both `runBlockGCSweep` and `runBlockGCForShare`, inside the existing `remoteGCLock(configID)` closure — no new lock scope. Compaction can't race a concurrent sweep's `DecrLiveChunkCount`/`DeleteBlock` on the same remote.

## Tests

`pkg/block/engine/compaction_test.go`: partially-dead block compacted (live chunk survives byte-identical, dead space reclaimed, old block object+record gone, idempotent second pass); healthy block skipped; disabled/nil-remote no-op; collect-then-query nested-query guard. Block-store conformance + metadata storetest still pass.

Docs: `architecture.md` GC section + `configuration.md` knob.

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -225,8 +225,9 @@ func runStart(cmd *cobra.Command, args []string) error {
 	// the validated gc.* config in pkg/config/config.go is silently dropped
 	// and the engine falls back to hardcoded defaults.
 	rt.SetGCDefaults(&runtime.GCDefaults{
-		GracePeriod:      cfg.GC.GracePeriod,
-		DryRunSampleSize: cfg.GC.DryRunSampleSize,
+		GracePeriod:         cfg.GC.GracePeriod,
+		DryRunSampleSize:    cfg.GC.DryRunSampleSize,
+		CompactionLiveRatio: cfg.GC.CompactionLiveRatio,
 	})
 
 	// Thread the operator-configured lock-manager grace period into the
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -487,6 +487,14 @@ gc:
                               # the snapshot.
   dry_run_sample_size: 1000   # Maximum candidate keys reported in
                               # --dry-run mode. Default 1000.
+  compaction_live_ratio: 0    # Reclaim dead bytes from partially-dead
+                              # blocks. After each sweep, a block whose
+                              # live bytes / object size is below this
+                              # ratio is repacked (live chunks only) and
+                              # the old block deleted. Must be in [0, 1];
+                              # 0 (default) disables compaction. A value
+                              # like 0.5 compacts a block once it is more
+                              # than half dead.
   auto_enabled: true          # Run background GC automatically so you
                               # don't have to invoke the CLI. Default
                               # true. Set false to require manual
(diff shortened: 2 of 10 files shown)


marmos91:
## Copilot review comments — resolved

**1. `compaction.go` — candidate filter requires `lb > 0` (skips blocks with no live locators):** working as intended, not a gap. Compaction only relocates *live* chunks out of a *partially*-dead block — a block with zero live locators has nothing to relocate. Those are owned by the reclaim path, not compaction: per `reclaim.go`, "a record with no live locator is terminally dead whatever its count" and `ReclaimRecords` deletes it (class-1 zero-ref if count==0, class-2 leaked/#1525 if stale count>0). A husk left by a failed compaction-delete is exactly that case — its locators were already rebased to the new block, so it has no live locator and the next reclaim sweep frees it. No leak.

**2. `blockgc.go` — on `CompactBlocks` error, logs "aborted" then could still log "complete":** real bug, fixed in `c9a03713` — added `return` after the abort warn so an errored run no longer double-logs or folds partial totals.

Issue #1491, opened 2026-06-30
Blocks-only storage: decoupled log-blob / block sizing


marmos91:
Follow-up to #1414; design: `.planning/2026-06-30-blocks-only-storage-design.md`.

v1 fixes log-blob cap ~1 GB and block carve target ~16 MiB. Revisit whether the two need independent per-workload tuning (e.g. smaller blobs for fsync/eviction granularity, different block target for the object store).

marmos91:
Closing per the #1555 profiling gate (PR #1562). The decoupled log-blob/block sizing knob is **not justified**: profiling shows the streamer's memory cost is transient `bytes.Buffer` doubling (~346 MB per 64 MiB, ~5× write-amp), not a fixed-sizing mismatch. That win is capturable by a one-shot carve-buffer pre-size — a few lines, no new config — which respects minimize-surface far better than a tunable knob. Reopen only if a workload shows the sizing itself (not buffer growth) is the bottleneck.

Issue #1488, opened 2026-06-30
Blocks-only storage: chunk-range refetch on read miss (eviction granularity)


marmos91:
Follow-up to #1414; design: `.planning/2026-06-30-blocks-only-storage-design.md`.

v1 eviction is block-granular: a read miss on an evicted chunk GETs the whole (~16 MiB) remote block and re-stages it. For random reads that is heavy.

**Goal:** on a miss, GET only the needed chunk byte-range (`GetBlockRange`) and cache it locally keyed by hash, instead of refetching the whole block. The local tier then holds whole log blobs plus per-chunk cache entries.

marmos91:
Closing per the #1555 profiling gate (PR #1562). The write path is CPU-bound on the FastCDC boundary scan (0.63 ms/MiB, ~3× the hash), and the warm-read path already avoids WAN — read-miss eviction granularity is not a measured bottleneck. Chunk-range refetch adds read-path complexity for no demonstrated win at this scale. Reopen if a dedicated read-path/WAN-cold profile shows whole-block refetch dominates.

PR #1562, opened 2026-07-05
perf(block): per-stage CPU+mem profiling of the write data path (#1555)

### LINKED ISSUE #1555: Per-stage CPU + memory profiling of the blocks-only storage flow
Profile and benchmark each stage of the new blocks-only storage flow **in isolation** to attribute CPU and memory bottlenecks per stage. Feeds the rclone-parity tracker (#1466) and upload-perf work (#1432).

## Reuse existing infra — do NOT build a new harness
A component-profiling tool already exists: \`cmd/bench/\` (\`blockstore.go\`, \`remote.go\`, \`metadata.go\`, \`parity.go\`) with pprof capture in \`profiles.go\` (cpu/heap/mutex/block). Existing microbenchmarks: \`chunker_bench_test.go\`, \`appendwrite_group_commit_bench_test.go\`, \`hash_bench_test.go\`, \`read_cache_bench_test.go\`. **Step 0 is to audit what \`cmd/bench\` + these already cover against the five stages below, then fill only the gaps.**

## Stages (code seam → coverage to verify)
- [ ] **Append log** — \`pkg/block/local/fs/appendwrite.go\` (+\`groupcommit.go\`). Bench exists; verify CPU + alloc coverage.
- [ ] **FastCDC chunker** — \`pkg/block/chunker/chunker.go\`. Bench exists; add memory-invariant assertion (below).
- [ ] **Streamer (carve → codec → PutBlock)** — \`pkg/block/engine/carver.go:carveAndCommitBlock\` + \`pkg/block/blockcodec/codec.go\`. Likely a gap — add CPU+mem bench with a memory remote.
- [ ] **Remote syncer** — \`pkg/block/engine/syncer.go\`, \`sync_queue.go\`, \`dispatcher.go\`, \`upload_controller.go\`, \`dynsem.go\`. See Refinement 1.
- [ ] **GC + compaction** — \`pkg/block/engine/\` GC path + compaction (#1487). Do **after** #1487 lands.

## Refinement 1 — the remote syncer can't be microbenched honestly
The other four stages isolate cleanly against in-memory adjacents (attributes CPU+alloc). The syncer is latency/concurrency-bound, not CPU-bound — a memory-fake bench shows it "fast" while the real bottleneck is single-stream WAN (that *is* #1432). Profile it against a **latency-injecting fake remote** or **live pprof against real S3/MinIO** on the bench VM, not a memory fake.

## Refinement 2 — one concrete memory pass/fail signal
For the data-plane stages, \`-benchmem\` **bytes/op must stay flat as input size grows** — allocations track *chunk size*, not *file size* (the streaming / \`sync.Pool\` invariant). Growing bytes/op = a streaming leak. Assert this in the benches; it catches the real memory regression better than eyeballing heap profiles.

## Tooling
Pure stdlib: \`go test -bench -benchmem -cpuprofile -memprofile\` + \`pprof\`, driven through the existing \`cmd/bench\` profiles harness. No new framework.

## Sequencing
After the two in-flight PRs land (#1554 sqlite test/profiling, #1487 GC compaction) — #1487 reshapes the GC/compaction stage.

Tracked under #1466.

marmos91:
Closes #1555.

Per-stage CPU + memory profiling for the blocks-only write data path, reusing plain `testing.B` + `runtime/pprof` — no new harness, no new deps. Test/tooling + docs only; zero production data-path changes.

## What's added

One isolated `Benchmark*` per write-path stage so a `-cpuprofile` / `-memprofile` run attributes cost to a single seam:

| Stage | Bench | Code seam |
| ----- | ----- | --------- |
| Append log | `BenchmarkAppendWrite_GroupCommit` (existing) | `appendwrite.go` + `groupcommit.go` |
| FastCDC chunker | `BenchmarkChunker_Throughput_64MiB` (+`ReportAllocs`) | `chunker.go` |
| BLAKE3 hash | `BenchmarkBLAKE3_256MiB` (existing) | content-hash layer |
| Streamer (carve→codec→PutBlock) | `BenchmarkStreamer_CarveCodecPutBlock` (**new**) | `carver.go:carveAndCommitBlock` + `blockcodec/codec.go` |

Plus two hard memory invariants (fail the build, not just report):

- `TestChunker_ConstantMemory` — the FastCDC scan is allocation-free regardless of input size.
- `TestStreamer_AllocationTracksBlockCount` — streamer allocation grows linearly with block count, never super-linearly with file size (Refinement 2 pass/fail signal).

`docs/BENCHMARKS.md` gains a "Per-stage write-path profiling" section: the table above, `pprof` how-to, and the baseline reading.

## Baseline reading (Apple M1 Max, in-memory, `-benchtime=5x`)

| Stage | ns/op | Throughput | B/op | allocs/op |
| ----- | ----: | ---------: | ---: | --------: |
| FastCDC chunker (64 MiB) | 42.2 ms | 1,592 MB/s | 0 | 0 |
| BLAKE3 hash (256 MiB) | 53.8 ms | 4,994 MB/s | 3.7 MB | 35,713 |
| Streamer (64 MiB) | 19.6 ms | 3,426 MB/s | 346 MB | 753 |
| Append log (coalesced 64 B) | 4.6 ms | — | 353 | 6 |

Normalized per MiB of write-path data:

- **CPU is dominated by the FastCDC chunker** — 0.63 ms/MiB, ~3× the hash (0.20) and ~2× the streamer (0.29). The boundary scan, not framing or upload, is the write-path CPU floor.
- **Allocations are dominated by the streamer** — ~346 MB per 64 MiB carved (~5× write amplification) from `bytes.Buffer` doubling growth + the remote body copy. The chunker is allocation-free.

## Gate call for the deferred perf issues this was meant to decide

- **#1491 (decoupled log-blob / block sizing) — not justified by this data.** The streamer's allocation cost is transient buffer growth, not sizing granularity; a one-line carve-buffer pre-size captures the win without decoupling the two sizes.
- **#1488 (chunk-range refetch granularity) — read-path, not a write-path bottleneck here.** The warm read path already avoids the WAN round-trip (`BenchmarkReadThroughCache_ColdVsWarm`), so there is no measured pressure to build finer read-miss granularity yet.

Runnable check: `go test -run 'TestChunker_ConstantMemory|TestStreamer_AllocationTracksBlockCount' -bench BenchmarkStreamer_CarveCodecPutBlock -benchtime=1x ./pkg/block/engine/ ./pkg/block/chunker/`.

### DIFF
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -338,6 +338,80 @@ To re-baseline on a new machine class: capture several runs, take the lowest
 ops/sec, multiply by 0.90, and update the floor constant — a deliberate
 calibration event that must be reviewed in PR.
 
+### Per-stage write-path profiling (#1555)
+
+The blocks-only write path splits into stages that each own a CPU and allocation
+budget. One `Benchmark*` per stage measures it *in isolation* against in-process
+memory stores, so a `-cpuprofile` / `-memprofile` run attributes cost to a single
+seam instead of a tangled end-to-end flow. All four are plain `testing.B` +
+`runtime/pprof` (no harness, no deps): standard
+`go test -benchmem -cpuprofile -memprofile` wiring is all that's needed.
+
+| Stage | Bench (package) | Code seam |
+| ----- | --------------- | --------- |
+| Append log | `BenchmarkAppendWrite_GroupCommit` (`pkg/block/local/fs/`) | `appendwrite.go` + `groupcommit.go` |
+| FastCDC chunker | `BenchmarkChunker_Throughput_64MiB` (`pkg/block/chunker/`) | `chunker.go` |
+| BLAKE3 hash | `BenchmarkBLAKE3_256MiB` (`pkg/block/`) | content-hash layer |
+| Streamer (carve→codec→PutBlock) | `BenchmarkStreamer_CarveCodecPutBlock` (`pkg/block/engine/`) | `carver.go:carveAndCommitBlock` + `blockcodec/codec.go` |
+
+```bash
+# Profile any single stage — CPU + heap attributed to that seam only:
+go test -bench=BenchmarkStreamer_CarveCodecPutBlock -benchmem -run=^$ \
+    -cpuprofile=cpu.out -memprofile=mem.out ./pkg/block/engine/
+go tool pprof -top cpu.out      # where the CPU went
+go tool pprof -top -alloc_space mem.out   # where the bytes were allocated
+```
+
+Two memory invariants are pinned as hard tests (fail the build, not just report):
+
+- `TestChunker_ConstantMemory` (`pkg/block/chunker/`) — the FastCDC scan is
+  allocation-free regardless of input size; `Next` must report offsets into the
+  caller's slice, never copy chunk bytes.
+- `TestStreamer_AllocationTracksBlockCount` (`pkg/block/engine/`) — the streamer's
+  allocation grows linearly with block count, not super-linearly with file size;
+  it must carve one block at a time, never buffer the whole file.
+
+GC + compaction is intentionally **not** benched here: #1487 reshaped that path
+and it is latency/sweep-bound rather than a CPU/alloc hot spot on the write path —
+a dedicated bench is deferred until there is a measured reason to add one.
+
+#### Baseline reading (Apple M1 Max, in-memory, `-benchtime=5x`)
+
+| Stage | ns/op | Throughput | B/op | allocs/op |
+| ----- | ----: | ---------: | ---: | --------: |
+| Append log (coalesced 64 B) | 4.6 ms | — | 353 | 6 |
+| FastCDC chunker (64 MiB) | 42.2 ms | 1,592 MB/s | **0** | **0** |
+| BLAKE3 hash (256 MiB) | 53.8 ms | 4,994 MB/s | 3.7 MB | 35,713 |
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed the Copilot review:

- **docs/BENCHMARKS.md inline code span across a newline** — real bug, fixed: the `go test -benchmem -cpuprofile -memprofile` span is now on a single line so it renders as code.
- **`block.ContentHash(blake3.Sum256(c))` conversion** — applied for consistency with the codebase convention (`engine/fetch.go`, `engine/read_internal.go`). Note the prior direct assignment also compiled (`blake3.Sum256` returns an unnamed `[32]byte`, assignable to the named `ContentHash`), but the explicit conversion matches surrounding code.
- **`for j := range streamerPerBlock` / `range blockCount`** — intentional and valid: range-over-int is Go 1.22+, this module is `go 1.25.0`, and CI Build Binaries + Unit Tests compile it. Kept as-is.

marmos91:
Copilot review dispositions (all three non-actionable, verified against green CI on Go 1.25):

1. `streamer_bench_test.go:62` range over int constant — valid since Go 1.22 (this repo builds on 1.25); Build/Unit/Integration/Windows all green. No change.
2. `streamer_bench_test.go:44` blake3.Sum256 → block.ContentHash — already explicitly converted: `block.ContentHash(blake3.Sum256(c))`. No change.
3. `docs/BENCHMARKS.md:348` broken inline code span — the span is self-contained on line 348 (line 347 ends at "standard", no open backtick); renders correctly. No change.

PR #1563, opened 2026-07-06
perf(block): pre-size carve block buffer to eliminate grow-and-copy

marmos91:
## What

The block streamer framed each carved block into a zero-value `bytes.Buffer` that doubled-and-copied its backing array as chunks were added. The #1555 per-stage profiling flagged this as the streamer's **dominant allocation** (~346 MB per 64 MiB of data). This is the cheap follow-up that profiling identified when #1491 was closed (the win is a pre-size, not a sizing knob).

`claimCarveBatch` already computed the exact body size of the claimed batch and threw it away. This threads that total out and pre-grows the buffer to `body + per-chunk headroom` so each block is written into a single backing array.

## Result

`BenchmarkStreamer_CarveCodecPutBlock` (64 MiB working set, M1 Max):

| | Before | After |
| --- | --- | --- |
| B/op | 345.6 MB | **294.5 MB** (−14.8%) |
| allocs/op | 752 | 704 |
| throughput | 4666 MB/s | **5530 MB/s** (+18.5%) |

## Sizing correctness

- `batchBytes` is the summed **raw** claimed size. The sync-filter in `carveAndCommitBlock` only ever *drops* chunks, so the packed body is always `≤ batchBytes` — `Grow` over-allocates, never under-allocates, for the identity sealer.
- The codec frames the **sealed** wire bytes. A decorated sealer (compression/encryption) can inflate the wire past the raw size, so the per-chunk headroom is **256 B** — covering the codec record (~38 B) plus encryption's magic + wrapped-key + nonce + AEAD tag (~80 B) with margin. A pathologically large wrapped key would cost at most one regrow — still correct, just less optimal. Documented inline.

## Tests

- `carver_claimbatch_test.go` asserts the threaded byte totals for all three claim behaviors (under-target no-op, target-reached prefix, drainAll skipping stale).
- `TestStreamer_AllocationTracksBlockCount` and the benchmark guard the streaming allocation invariant.
- `go test -race ./pkg/block/engine/` green (303 tests); lint clean.

Reviewed by code-simplifier (no change needed) and code-reviewer (one LOW finding — decorated-sealer headroom — addressed by the 256 B bump above).

### DIFF
--- a/pkg/block/engine/carver.go
+++ b/pkg/block/engine/carver.go
@@ -118,11 +118,11 @@ func (m *Syncer) carveFlush(ctx context.Context, drainAll bool) error {
 		if err := ctx.Err(); err != nil {
 			return err
 		}
-		batch := m.claimCarveBatch(target, drainAll)
+		batch, batchBytes := m.claimCarveBatch(target, drainAll)
 		if len(batch) == 0 {
 			return nil
 		}
-		if err := m.carveAndCommitBlock(ctx, batch); err != nil {
+		if err := m.carveAndCommitBlock(ctx, batch, batchBytes); err != nil {
 			// Return the batch to the queue for a later retry and surface the
 			// error so an explicit Flush reports non-durability.
 			m.requeueCarveBatch(batch)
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed the `Grow(int(...))` overflow concern (commit f9bd78f8): the pre-grow is now computed in int64 and range-checked — `if grow > 0 && grow <= math.MaxInt` — so an absurd block-size config (or a 32-bit build past MaxInt32) skips the best-effort pre-grow instead of passing a negative int to `bytes.Buffer.Grow`. No behavior change on the normal MB-scale path.

Issue #1466, opened 2026-06-30
🎯 rclone throughput parity — roadmap tracker


marmos91:
# 🎯 rclone throughput parity — roadmap tracker

Single-pane tracker for closing the throughput gap vs rclone across the four workload quadrants (upload/download × large/small file). Design: `.planning/2026-06-30-rclone-parity-roadmap.md`.

## Baseline (measured 2026-06-29, SCW VM fr-par → Cubbit, conc64)
| Quadrant | dittofs | rclone | ratio |
|---|---|---|---|
| Large-file UPLOAD | 337 Mbit/s (4MiB PUTs) | 402 (16MiB parts) | ~84% |
| Small-file UPLOAD (16K×64KiB) | ~0.5 MiB/s, inflight=1 | — | ✗ |
| DOWNLOAD (both) | not measured | — | ? |

Root cause: single-stream-over-WAN (1 stream = 25 Mbit; scales ~linearly with concurrency).

## Phases
- [x] **Concurrency** (large-file upload) — sustain inflight≈window. → #1432 (PR1 #1461, peak 1→24/24 proven)
- [x] **Drain timeout** — inactivity-bounded drain-uploads. → #1432 (PR2 #1460)
- [ ] **Phase 0 — parity benchmark harness + scorecard** (measurement first) → #1467
- [ ] **Unified object packing** — 16MiB packs close BOTH large-file (rclone-parity upload unit, keeps 4MiB dedup) AND small-file (coalescing). → #1414
- [ ] **Small-file write wall** — batched/async per-file commit (the ~64 files/s fsync ceiling; packing alone doesn't fix the write side). → #1416
- [ ] **Phase 4 — download/read parity** (unmeasured) → #1468
- [ ] *(conditional)* multipart upload for large objects — only if unified packing leaves a large-file gap.

## Honest expectation
PR1 → large-file ~84% of measured ceiling. Unified packing → large-file ~par AND small-file drain. + write-wall fix → small-file end-to-end par. Download: unknown until Phase 0.

marmos91:
The object-packing (#1414) and write-wall / batched-commit (#1416) levers in this roadmap are now delivered by the **blocks-only storage redesign** epic #1493 (supersedes #1414, subsumes #1416 via the single-transaction per-block commit). Tracking those two rows there; this roadmap stays the parity umbrella, and the Phase 0 harness (#1467) is how we validate #1493's write-path win.

marmos91:
[...]

PR #1579, opened 2026-07-07
perf(metadata): collapse redundant per-op GetFile on the read/write hot path

marmos91:
## Problem

Profiling the #1466 read benchmark (POP2-HC-32C, badger metadata) showed the NFS/SMB data path spends ~20–26% of read CPU in badger `GetFile` — not the KV get, but DittoFS's per-call `derivePath` (parent-edge walk, 2 badger gets per directory level) + full-struct `encoding/json` decode. It multiplied because each read op resolved the **same inode 2–3×**:

- `getFileOrError` (handler load)
- `PrepareRead` (permission gate, whose return v3 discarded)
- the permission path's own `store.GetFile`

## Change

Resolve the inode once per op, and skip work the hot path never uses:

- **By-`*File` permission checks** (`CheckReadPermissionFile`, `checkFilePermissionsFile`, `permissionResult`, `checkWritePermissionFile`): gate on the already-loaded file instead of re-fetching by handle. The by-handle `checkFilePermissions` now fetches once and delegates — logic is otherwise byte-for-byte identical.
- **`GetFileForRead`**: optional badger fast path that skips `derivePath` (NFS/SMB are handle-addressed; the read/getattr/write hot paths don't read `File.Path`). Other backends fall back to `GetFile`. Pending-write merge is shared (`mergePendingWrites`).
- **NFSv3 READ / SMB READ** drop `PrepareRead`'s redundant fetch and gate on the loaded file.
- **NFSv4** (READ / READ_PLUS / SEEK / COMMIT / VERIFY) use `GetFileForRead` where `File.Path` is unused.
- **WRITE**: `PrepareWrite` reuses its freshly-fetched file for the permission check **on a cache miss**; on a cache hit it keeps the fresh re-read — the pending-write file cache can be stale vs a concurrent `chmod`, and permission is a security boundary (unlike the explicitly-soft size/quota checks). Covers v3/v4/SMB since all writes funnel through `PrepareWrite`.

## Measured (read pprof, NFSv3 READ subtree — composition)

| cost | before | after |
|---|---|---|
| `PrepareRead` re-fetch | 46% | 0% |
| `derivePath` walk | 28% | ~0% |
| permission re-fetch | 33% | 3.5% |
| inode loads per read | ~3× | 1× |

The remaining single-load cost is the `encoding/json` decode of that one load — a natural follow-up (lean `File` decoder).

## Tests

`go test -race ./pkg/metadata/... ./internal/adapter/nfs/... ./internal/adapter/smb/...` — all green (45 packages). Storetest conformance passes (`GetFileForRead` falls back to `GetFile` on memory/sqlite/postgres).

## Follow-ups (not in this PR)

- NFSv4 GETATTR/ACCESS still use `GetFile` (they log `file.Path`); a fast-path pass there is a clean follow-up.
- Lean `File` decoder to cut the residual `encoding/json` cost.

### DIFF
--- a/internal/adapter/nfs/v3/handlers/doc.go
+++ b/internal/adapter/nfs/v3/handlers/doc.go
@@ -182,7 +182,9 @@ func (h *Handler) getFileOrError(
 	clientIP := xdr.ExtractClientIP(ctx.ClientAddr)
 	metaSvc := h.Registry.GetMetadataService()
 
-	file, err := metaSvc.GetFile(ctx.Context, fileHandle)
+	// GetFileForRead: NFS is handle-addressed, so no v3 handler reads
+	// File.Path — skip the derivePath parent-edge walk on this hot path.
+	file, err := metaSvc.GetFileForRead(ctx.Context, fileHandle)
 	if err != nil {
 		// Check if the error is due to context cancellation
 		if ctx.Context.Err() != nil {
--- a/internal/adapter/nfs/v3/handlers/fsinfo.go
+++ b/internal/adapter/nfs/v3/handlers/fsinfo.go
@@ -150,7 +150,7 @@ func (h *Handler) FsInfo(
 		return &FsInfoResponse{NFSResponseBase: NFSResponseBase{Status: status}}, err
 	}
 
-	logger.DebugCtx(ctx.Context, "FSINFO", "share", ctx.Share, "path", file.Path)
+	logger.DebugCtx(ctx.Context, "FSINFO", "share", ctx.Share)
 
 	// Retrieve filesystem capabilities from the store
 	// All business logic about filesystem limits is handled by the store
--- a/internal/adapter/nfs/v3/handlers/fsstat.go
+++ b/internal/adapter/nfs/v3/handlers/fsstat.go
@@ -130,7 +130,7 @@ func (h *Handler) FsStat(
 		return &FsStatResponse{NFSResponseBase: NFSResponseBase{Status: status}}, err
 	}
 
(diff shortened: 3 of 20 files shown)


marmos91:
Thanks — addressed in \`d111bf9f\`:

1. **Non-regular-file READ (SMB)** — real regression, fixed. Restored the regular-file guard on the already-loaded \`*File\` in \`smb/handlers/read.go\`, mapping through the same \`ErrIsDirectory\`/\`ErrInvalidArgument\` → status path \`PrepareRead\` used, so directory/FIFO/socket/device reads fail identically instead of returning empty EOF. (NFSv3 already rejected these above the permission gate via \`NFS3ErrIsDir\`, so only SMB needed it.)

2. **\`GetFileForRead\` test coverage** — added \`TestMetadataService_GetFileForRead\` covering the fallback branch (memory store doesn't implement \`fileForReadStore\`): asserts it matches \`GetFile\` including \`Path\` derivation, and that an unknown handle surfaces the store error rather than panicking. The pending-write overlay is the *same* \`mergePendingWrites(handle, file)\` call in both \`GetFile\` and \`GetFileForRead\`, so it's not re-tested.

3. **Extra \`storeForHandle\` in \`checkFilePermissionsFile\`** — leaving as-is. The fast path (\`CheckReadPermissionFile\` → \`checkPermissionFile\` → \`checkFilePermissionsFile\`) decodes the handle exactly once; only the legacy \`checkFilePermissions\` wrapper calls it twice. \`DecodeFileHandle\` is a UUID parse + map lookup that never appeared in the read profile (the cost was \`derivePath\` + reflection-JSON, both eliminated here), so routing via \`file.ShareName\` would trade real share-routing surface for no measurable gain.

Issue #1572, opened 2026-07-06
Read path enumerates + Sscanf-parses the whole FileChunk manifest per read → O(N²), CPU-bound reads


marmos91:
**Re-scoped after a live pprof disproved the original "sparse hole" hypothesis (see correction at bottom).**

## Problem

Every NFS/SMB read enumerates and `fmt.Sscanf`-parses the **entire per-payload FileChunk manifest** — twice. Reads are O(N) in the file's chunk count, so a large or randomly-written file makes reads O(N²) overall and **CPU-bound**, with zero network traffic (the data is local; the cost is *proving* it's local).

## Evidence (SCW POP2-HC-32C-64G, develop `0b2361b8`, live CPU profile under NFSv3 random-read load)

A 512 MiB file written by 30 s of random 4 KiB writes, then read back:

- **randread 2 IOPS, p50 625 ms; seqread 1 IOPS** — and **NIC rx = 0** (nothing fetched from S3).
- CPU profile (25 s, dfs server):

```
81.2%  BadgerMetadataStore.ListFileChunks
75.1%  Syncer.listFileChunksSnapshot
74.7%  Syncer.resolveFileChunk
36.6%  Store.ensureAndReadFromLocal → EnsureAvailableAndRead
36.1%  Syncer.enqueuePrefetch → blockIsLocal      (ListFileChunks AGAIN, per read)
36.2%  parseBlockIdx
31.9%  fmt.Fscanf                                 (Sscanf of the block index)
```

## Root cause

`readAtInternal` (`pkg/block/engine/read_internal.go`) resolves each read through the syncer, and **both** `resolveFileChunk` and the prefetch probe `enqueuePrefetch → blockIsLocal` call `ListFileChunks(payloadID)`, which:

1. enumerates the whole per-payload manifest from badger (`objects.go:listFileChunksTxn`) — O(N) per read, and
2. parses every row's block index with `fmt.Sscanf(idx, "%d", &v)` (`objects.go:717 parseBlockIdx`) — reflection-based, ~100× slower than `strconv`.

N reads × O(N) enumeration = O(N²); large N (large or scattered-write files) → reads go CPU-bound. This is the true cause of the benchmark's random-read collapse (dittofs 7–11 IOPS vs rclone 2583 / local ~200k).

## Fix (two independent wins)

[...]

PR #1580, opened 2026-07-07
perf(block): indexed covering lookup in fillFromCASManifest read hot path

marmos91:
## What

`fillFromCASManifest` (the packed-blocks read enumeration) listed **and sorted the entire per-payload FileChunk manifest on every read** — O(N) per read, N reads ⇒ O(N²), CPU-bound at high fragmentation. The #1466 profile attributed ~46% of read CPU to this walk on a densely-randwritten file.

This resolves the 1–2 chunks that actually cover a read via the badger offset index (`GetFileChunkAtOffset`) instead of scanning the whole manifest.

## How

- Fast path: type-assert the FileChunk store to the `chunkAtOffsetResolver` optional interface (badger implements it) → indexed covering lookup.
- Fallback: holes, evicted chunks, and non-badger backends (sqlite/postgres/memory) fall back to the original full walk, now named `fillFromCASManifestScan`. Behaviour is byte-identical on the fallback.

## Measured (POP2-HC-32C, badger, heavy-fragmentation regime)

Reproduced #1466: 128 MiB file under 180 s of dense 4 KiB random-write, then random-read.

| | rand-read IOPS | `ListFileChunks` share of read CPU |
|---|---|---|
| before | ~75 | 57.6% (CPU-bound, 166%) |
| after  | ~170 (**~2.3×**) | fast path engaged |

At *moderate* fragmentation the walk is only ~1.25% of read CPU (reads are I/O-bound there), so this is a heavy-fragmentation win specifically.

## Known limitation (documented, not a regression)

The fast path **bails to the full scan on a hole**, and sparse random-write files are full of holes — so data-block reads take the fast path while hole reads still scan (`fillFromCASManifestScan` was still ~29% in the heavy run). A complete fix needs a "next chunk at offset ≥ X" successor lookup; not in this PR. Data correctness is unaffected — holes zero-fill exactly as before.

Ref #1572.

### DIFF
--- a/pkg/block/local/fs/drain_reset_testhelpers_test.go
+++ b/pkg/block/local/fs/drain_reset_testhelpers_test.go
@@ -64,6 +64,31 @@ func (m *memFBS) ListFileChunks(_ context.Context, payloadID string) ([]*block.F
 	return out, nil
 }
 
+// GetFileChunkAtOffset mirrors the badger offset index: it returns the row
+// whose absolute byte range [chunkOffset, chunkOffset+DataSize) covers off,
+// or (nil, nil) for a hole. Implementing it here makes every post-rollup
+// read test drive fillFromCASManifest's indexed fast path rather than the
+// scan fallback.
+func (m *memFBS) GetFileChunkAtOffset(_ context.Context, payloadID string, off uint64) (*block.FileChunk, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	var best *block.FileChunk
+	var bestOff uint64
+	for _, fb := range m.rows[payloadID] {
+		o, ok := block.ParseChunkOffset(fb.ID)
+		if !ok || o > off {
+			continue
+		}
+		if best == nil || o > bestOff {
+			best, bestOff = fb, o
+		}
+	}
+	if best == nil || off >= bestOff+uint64(best.DataSize) {
+		return nil, nil
+	}
+	return best, nil
+}
+
 func (m *memFBS) EnumeratePayloads(ctx context.Context, fn func(payloadID string) error) error {
 	m.mu.Lock()
 	ids := make([]string, 0, len(m.rows))
--- a/pkg/block/local/fs/readpayload.go
+++ b/pkg/block/local/fs/readpayload.go
@@ -231,18 +231,110 @@ func copyRecordIntoDest(recOff uint64, payload, dest []byte, reqStart, reqEnd ui
 	}
 }
 
-// fillFromCASManifest walks the FileChunk manifest for payloadID and
-// fills any still-uncovered bytes of dest from the corresponding CAS
-// chunks. This is the post-rollup read path — bytes that the rollup has
-// already moved out of the append log into CAS storage. The manifest is
-// the authoritative ordering once a chunk is committed.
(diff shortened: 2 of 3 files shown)


marmos91:
Closing after a clean re-benchmark overturned the 2.3x this PR claimed.

A same-file A/B (develop vs this branch, POP2 badger) showed randread **814→853 single-stream** and **2583→2656 at 16 streams** — both within noise — and \`GetFileChunkAtOffset\` never appeared in any read profile. An idle profile (zero client reads) showed why: the server burns ~48% CPU on a **background rollup** (\`rollupFile → scanAllFiles → ListFileChunks\`), which is where the 41–57% \`ListFileChunks\` cost actually came from. This PR only rewires the read-path \`fillFromCASManifest\`, not rollup. Once rollup drains, the steady-state read path is \`ReadPayloadAt\` + disk \`Pread\` — it no longer touches the manifest scan at all.

So \`fillFromCASManifest\` is only hot in the transient pre-rollup window or a rollup-starved (many-small-files) workload — not steady-state reads. Not worth the hole-fallback complexity on the strength of a confounded number. Branch \`fix/1572c-fillcas-covering\` kept locally if the rollup-starved regime ever needs it.

marmos91:
## Why this was reopened

I closed this last session on a **"rollup confound"** verdict — I'd measured `ListFileChunks` at 41–57% CPU, then found a background rollup sweep (`rollupFile → scanAllFiles → ListFileChunks`) also burning CPU, and concluded the read-path cost I'd attributed to `fillFromCASManifest` was actually the rollup, which this PR doesn't touch.

**That verdict was wrong**, and a clean re-measurement on a dedicated rig (SCW POP2-HC-32C-64G) proves it:

- **Rollup fully idle** — drained before capture, `dfs` CPU 0.0%. Not the rollup sweep.
- **Pure cache-hit, zero S3/RTT** — `dittofs_datapath_block_range_reads_total` delta = **0** during the run.
- Under those controls, `handleNFSRead → ReadFromBlockStore → engine.Store.ReadAt → FSStore.ReadPayloadAt → fillFromCASManifest → ListFileChunks` is **75–82% of read CPU** — and it sits under `handleNFSRead`, not `rollupFile`.
- Confirmed at **both** 32MiB and 10GB cache (11756 IOPS/48MB/s large-cache; 2771 IOPS tiny) → not a cache-size artifact.

The earlier "confound" measurement had hit a **non-rolled-up** file (direct append-log fast path, no manifest fill). **Steady state = rolled-up (packed) files**, which route every read through `fillFromCASManifest → full ListFileChunks` = O(chunks) JSON decode per read. This PR's indexed covering lookup is the real, rollup-independent fix.

Being revived with a **successor ("next chunk ≥ offset") lookup** so sparse/holed files don't bail back to the full scan (the original commit's known limitation). Evidence: `results/read-baseline/READ-BASELINE.md`.

marmos91:
Superseded by #1587, which carries this same covering-lookup fix **plus** the successor ("next chunk ≥ offset") lookup so sparse/holed files no longer bail to the full scan — the one limitation this branch had. The reopen rationale above (clean-rig re-measurement overturning the "rollup confound" verdict) applies to #1587. Closing in favor of #1587.

PR #1581, opened 2026-07-07
perf(block): access-pattern-aware remote readahead

marmos91:
## What

Remote block prefetch fired a **fixed `PrefetchBlocks` (default 64) after every remote read, regardless of access pattern**. On a random read that means 64 speculative S3 GETs for blocks the reader will never touch — wasted WAN bandwidth and object-store cost, and download-queue contention that can slow the demanded fetches. (`doc.go` already described a "1→2→4→8 adaptive" readahead, but the code never implemented it.)

This adds the access-pattern awareness the doc promised: a per-payload read frontier drives an adaptive prefetch window.

## How

- `planReadahead(payloadID, start, end)` (new `readahead.go`) tracks each payload's last read frontier.
  - **Sequential** run (a read continuing at/just past the frontier) ramps the window `1→2→4→…→PrefetchBlocks`, the Linux readahead pattern.
  - **Random** jump resets the window to **0** — no speculative GETs.
- `EnsureAvailableAndRead` replaces the fixed `for i := range PrefetchBlocks` loop with `for i := 0; i < planReadahead(...); i++`.
- Per-payload state is a bounded map (`maxReadaheadEntries`, arbitrary eviction — readahead state is disposable, a dropped entry just re-ramps).

## Tests

5 unit tests in `readahead_test.go` pin the logic: cold start (depth 0), sequential ramp `1,2,4,…,64` capped, random reset to 0, re-read tolerance, disabled-when-`PrefetchBlocks=0`, and the map bound. Full `pkg/block/engine` suite (328 tests) green, lint clean.

## Measurement status (follow-up)

WAN throughput/GET-count A/B is **not yet included** — the bench rig couldn't be held in an S3-only state (rollup logblob + `pin` retention keep data local), so the remote-fetch path wouldn't engage reproducibly. One cold-window sample did show the waste this fixes: a **baseline 8 MB random read pulled 20 MB from S3** (2.5× prefetch amplification). A clean throughput number needs a purpose-built harness (evict-retention + tiny local cache so blocks are S3-only by construction); tracked as a follow-up.

### DIFF
--- a/pkg/block/engine/fetch.go
+++ b/pkg/block/engine/fetch.go
@@ -383,7 +383,12 @@ func (m *Syncer) EnsureAvailableAndRead(ctx context.Context, payloadID string, o
 		}
 	}
 
-	for i := range m.config.PrefetchBlocks {
+	// Prefetch ahead only as far as the payload's access pattern justifies:
+	// a sequential run ramps the window up to PrefetchBlocks, a random read
+	// prefetches nothing (planReadahead returns 0) so we never spend remote
+	// GETs on blocks a random reader will not touch.
+	depth := m.planReadahead(payloadID, startBlockIdx, endBlockIdx)
+	for i := 0; i < depth; i++ {
 		m.enqueuePrefetch(payloadID, endBlockIdx+1+uint64(i))
 	}
 
--- a/pkg/block/engine/readahead.go
+++ b/pkg/block/engine/readahead.go
@@ -1,0 +1,64 @@
+package engine
+
+// maxReadaheadEntries bounds the per-payload readahead state map. Readahead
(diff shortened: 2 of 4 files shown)


marmos91:
Addressed Copilot review:

- **readahead.go:5 comment** — fixed. Corrected the off-by-one: an evicted/cold entry re-ramps from a *cold start* (the re-creating read returns 0; the next sequential read reaches depth 1), not "from depth 1".
- **readahead.go:35 nil-map** — declined. `NewSyncer` (syncer.go:325) is the sole constructor and always inits `readahead: make(...)` (syncer.go:331). No struct-literal construction path exists in non-test code, so a lazy-init guard would be defense for an unreachable state. Left as-is per YAGNI.

PR #1583, opened 2026-07-07
docs: clarify first-run admin password is interactive-only

marmos91:
## Problem

The docs say the auto-generated first-run admin password is "printed once to the log" / "lands in `docker logs`". That's wrong. In `cmd/dfs/commands/start.go`, `emitAdminPassword` prints the password **only when stdout is an interactive terminal** — i.e. `dfs start --foreground` in a real TTY. In the default background daemon, and under Docker/systemd/CI (stdout is a pipe), the password is **never written anywhere and cannot be recovered**.

Worse, the daemon-mode fallback pointed operators at `dfsctl user passwd admin` to recover — but `dfsctl user password` requires an authenticated admin session (`cmdutil.GetAuthenticatedClient()`), which you can't have without the password you never saw. Dead end → operator locked out.

## Fix

Documentation only, plus one runtime message correction:

- **README / getting-started / install / configuration** — make `DITTOFS_ADMIN_INITIAL_PASSWORD` (or `admin.password_hash`) the recommended first-run step, and state plainly that the generated password is interactive-`--foreground`-only and unrecoverable in background/Docker/systemd.
- **`start.go` daemon-mode warning** — no longer points at the unusable `dfsctl user passwd admin`; tells operators to set the env var / config hash and restart, or use `--foreground`.

## DX note

Considered surfacing the generated password on the parent terminal before the PID lines (background mode). Rejected: it only helps *interactive* `dfs start` (already served by `--foreground`), needs child→parent password plumbing, and stays hidden under Docker/systemd where stdout is a pipe. The env var is the strictly better path. The real bug was the misleading recovery advice, fixed here.

## Verification

`go build ./cmd/dfs/...` clean; `TestEmitAdminPassword_*` and `TestAdminBootstrap_PasswordNotLoggedToStructuredLogger` pass. No behavior change to password handling — only the fallback warning text.


marmos91:
Good catch, all four addressed in 944d1d4f. Confirmed against `EnsureAdminUser` (pkg/controlplane/store/users.go:235): it returns early once the admin user exists, so `DITTOFS_ADMIN_INITIAL_PASSWORD`/`admin.password_hash` are read only while bootstrapping a *new* admin. Reworded the recovery guidance in all four docs, the daemon-mode warning, and the `emitAdminPassword` comment to state that recovery requires removing the admin user (resetting the control-plane database) and re-bootstrapping with the password pre-set — restart alone won't change it.

PR #1587, opened 2026-07-07
perf(block): route fillFromCASManifest through indexed covering+successor lookup (#1580)

marmos91:
## What

Finishes the read-path fix for #1580 (reopened). Every read of a rolled-up file walked the **entire** per-payload FileChunk manifest — `ListFileChunks` + a JSON decode of every chunk — just to find the single chunk covering the requested offset. That is O(chunks) per read (O(N²) over a random scan) and dominates read CPU.

`FSStore.fillFromCASManifest` (`pkg/block/local/fs/readpayload.go`) now resolves each touched chunk through the badger secondary index instead:

- **Covering lookup** — `GetFileChunkAtOffset(off)` returns the single chunk whose range `[off, off+DataSize)` contains the offset (added on develop by #1577; this PR routes the block-store CAS-fill through it).
- **Successor lookup** — new `GetFileChunkAtOrAfterOffset(off)` returns the next chunk starting at/after `off`. On a sparse hole the loop jumps straight to the next data boundary in one indexed keys-only scan, so a read never re-enumerates the manifest and never probes byte-by-byte.
- Non-badger backends (memory/sqlite/postgres) keep the full-manifest walk, now named `fillFromCASManifestScan`.

## Why it's backed by clean rig evidence

From `results/read-baseline/READ-BASELINE.md` — validated on a clean SCW POP2-HC-32C rig, proven pure cache-hit (`dittofs_datapath_block_range_reads_total` delta = 0), rollup drained (dfs CPU 0%), S3 untouched:

```
handleNFSRead → ReadPayloadAt (82.8%) → fillFromCASManifest (82.7%)
   → ListFileChunks (80.1%) → JSON-decode all chunks (29.6%)
```

- rand-read-4k, 32MiB cache: **2771 IOPS**, fillFromCASManifest = **82%** of read CPU.
- rand-read-4k, default 10GB cache: **11756 IOPS**, fillFromCASManifest = **75.7%**, ListFileChunks = **57.2%**.

Same mechanism at both cache sizes → **not** a cache-size / rollup / S3 artifact. This corrects the earlier "rollup confound" dismissal (which had measured a fresh, unrolled file on the append-log fast path — steady state is rolled-up files, which route through `fillFromCASManifest`).

## Correctness (no wrong bytes on a hole)

The covering guard lives inside the resolver: `GetFileChunkAtOffset` only returns a chunk whose range actually contains the offset (guard keyed on the row's own start offset, so an inconsistent index can't serve a neighbour chunk's bytes). On a hole it returns nil, and the loop advances via the successor — the hole stays uncovered and `ReadPayloadAt` surfaces it as `ErrFileChunkNotFound` (caller zero-fills / fetches remote). A hole can never be served a neighbouring chunk's bytes.

New tests:
- `pkg/metadata/storetest/file_block_ops.go` — successor conformance cases (exact-boundary, holes between chunks, EOF, multi-chunk, single-byte chunk), each cross-checked against an independent `ListFileChunks` oracle. Single-byte case added to the covering test too.
- `pkg/block/local/fs/readpayload_covering_test.go` — drives the real fast path (`memFBS` now implements the resolver): multi-chunk covering reads (aligned/unaligned/spanning) return exact bytes; a hole read and a hole-spanning read both miss rather than serve wrong bytes.

## Gates

- `pkg/metadata/storetest` + `pkg/block/blockstoretest` conformance green on **all** backends (badger, sqlite, memory, postgres all runnable here).
- `go test -race ./pkg/block/local/fs/... ./pkg/metadata/store/badger/... ./pkg/block/engine/...` green.
- `gofmt -s`, `go vet`, `golangci-lint run` clean.

Baseline to beat (measured on the live rig by the maintainer): fillFromCASManifest/ListFileChunks should collapse off the read profile; local cache-hit IOPS ceiling up several-fold.

Closes #1580.

### DIFF
--- a/pkg/block/local/fs/drain_reset_testhelpers_test.go
+++ b/pkg/block/local/fs/drain_reset_testhelpers_test.go
@@ -82,6 +82,42 @@ func (m *memFBS) EnumeratePayloads(ctx context.Context, fn func(payloadID string
 	return nil
 }
 
+// GetFileChunkAtOffset + GetFileChunkAtOrAfterOffset make memFBS satisfy the
+// coveringChunkResolver fast path so ReadPayloadAt drives fillFromCASManifest's
+// indexed covering + successor loop (not the ListFileChunks scan fallback).
+// Both mirror the badger backend's contract, which pkg/metadata/storetest pins.
+func (m *memFBS) GetFileChunkAtOffset(_ context.Context, payloadID string, off uint64) (*block.FileChunk, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	for _, fb := range m.rows[payloadID] {
+		abs, ok := block.ParseChunkOffset(fb.ID)
+		if !ok {
+			continue
+		}
+		if off >= abs && off-abs < uint64(fb.DataSize) {
+			return fb, nil // covering guard: off is within [abs, abs+DataSize)
+		}
+	}
+	return nil, nil
+}
+
+func (m *memFBS) GetFileChunkAtOrAfterOffset(_ context.Context, payloadID string, off uint64) (*block.FileChunk, error) {
+	m.mu.Lock()
+	defer m.mu.Unlock()
+	var best *block.FileChunk
+	var bestAbs uint64
+	for _, fb := range m.rows[payloadID] {
+		abs, ok := block.ParseChunkOffset(fb.ID)
+		if !ok {
+			continue
+		}
+		if abs >= off && (best == nil || abs < bestAbs) {
+			best, bestAbs = fb, abs
+		}
+	}
+	return best, nil
+}
+
 func (m *memFBS) GetFileChunk(_ context.Context, blockID string) (*block.FileChunk, error) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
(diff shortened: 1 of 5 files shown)


marmos91:
## Validated on the live rig (SCW POP2-HC-32C-64G) ✅

Back-to-back A/B, **same 512MB rolled-up local file, same session**, rollup idle, `dittofs_datapath_block_range_reads_total` delta = **0** both runs (proven pure local cache-hit — no S3/RTT). fio randread bs=4k direct=1 iodepth=32 numjobs=4 30s / NFSv3.

| metric | develop | this PR | delta |
|---|---|---|---|
| IOPS | 2254 | **10139** | **4.5×** |
| bandwidth | 8.8 MB/s | 39.6 MB/s | 4.5× |
| p50 latency | 38.0 ms | 10.8 ms | 3.5× lower |
| S3 range-reads | 0 | 0 | cache-hit |

### Structural confirmation (20s CPU pprof, cum%)
| read CPU | develop | this PR |
|---|---|---|
| `fillFromCASManifest` | **81.82%** | gone from top-20 (collapsed) |
| `ListFileChunks` (full manifest JSON scan) | **78.87%** | gone |
| new #1 cost | — | `GetFileForRead` 58.76% (File-object decode) |

The O(chunks) per-read manifest enumeration is eliminated exactly as designed. The new read ceiling is `GetFileForRead` File-object JSON decode (the #1579 area) — a separate follow-up, not the manifest scan.

Issue #1602, opened 2026-07-08
Consolidate benchmarking into one fio-based competitor harness (dfsbench)


marmos91:
## Goal

Collapse the three overlapping benchmark harnesses (plus a byte-for-byte duplicated infra tree) into **one** pure-Go, fio-based, single-disposable-SCW-VM competitor comparison harness — `dfsbench`. Standardize on fio across every system so results are comparable, and measure DittoFS against real competitors and pure-FS controls.

Full plan: `.planning/2026-07-08-bench-harness-consolidation.md`.

## The mess today

- **Harness A** (mounted-FS competitor): `pkg/bench/` runs its own Go re-implementation of fio (never calls fio), no s3fs, rclone uses `serve nfs` not `mount`, no crash-resume.
- **Harness B** (in-process blockstore orchestrator): `bench/orchestrator/` + `bench/remote/` — obviated by `go test -bench ./pkg/...`.
- **Harness C** (in-process engine parity): `bench/parity/` — dittofs-vs-rclone, no mount.
- Repo-root `infra/` byte-duplicates `bench/infra/`; competitor lists drifted between `run-all.sh` and `bench/infra/systems.go`.

## Target

- **`dfsbench`** (pure Go, `cmd/bench`): Backend × Protocol matrix — dittofs-s3, juicefs, s3ql, rclone-mount (FUSE), s3fs, plus local-disk controls (kernel-nfs, ganesha, samba) — each over **nfs3 / nfs4.1 / smb3** (FUSE backends re-exported via knfsd/Samba; capability map native/re-export/N/A).
- **fio** is the single load generator (behind a `LoadGen` interface; elbencho as future swap-in). **MinIO warp** for the raw-S3 ceiling; fio for the local-disk ceiling.
- Single disposable SCW VM (scw CLI); lifecycle `setup` / `run` / `baseline` / `report` / `teardown`; `--resume` (per-cell result JSON); selective `--systems/--protocols/--workloads/--sizes`; YAML `--config`.
- Uniform **server↔S3 byte metering** at the OS network layer (works for every tool).
- Genuine **cold-from-S3** via per-tool cache eviction (`--evict-cache`).
- `bench/` becomes **fio-only**: component microbenchmarks are audited and relocated into their home packages (`pkg/block/...`, `pkg/metadata`) as `Benchmark*` funcs.

## Hypothesis under test

[...]

PR #1606, opened 2026-07-08
feat(bench): backend×protocol matrix + SCW remote runner (#1602 PR3+PR4)

marmos91:
Part of #1602 (benchmark harness consolidation). Combines PR3 (competitors × protocols) and PR4 (single-disposable-VM remote runner) — packaged together because neither is useful without the other, and both were validated in the same live run.

## What this adds

**PR3 — Backend × Protocol matrix.** Every backend is measured across `nfs3` / `nfs4` / `smb3` through one shared re-export layer:
- `dittofs-s3` — the native subject (userspace NFS/SMB, no FUSE)
- `local-disk` — plain-dir control (the protocol-overhead ceiling)
- `rclone` / `s3fs` / `juicefs` / `s3ql` — FUSE competitors, re-exported over knfsd (NFS) or Samba (SMB) and mounted back over loopback

Adding a competitor is one registry entry + Setup/Evict recipes (`newSrcBackend`); the re-export layer gives it all three protocols for free. The managed run drives one backend at a time (Setup → Mount → warm fio → Evict + drop OS cache → cold-read fio → Unmount → Teardown) and **keeps going past a failed backend**, logging FAIL and moving on.

**PR4 — Remote runner.** `run --remote` provisions a disposable SCW VM, cross-builds + pushes `dfs`/`dfsctl`/`dfsbench` (all CGO-free), installs prereqs, launches the run **detached** (survives ssh drops via a DONE sentinel + poll), pulls results, and tears the VM down. State persists in `.bench-vm.json` so teardown always works.

## Live validation (real Scaleway S3)

Ran the full 4-backend × 3-protocol matrix end to end. 22/24 cells landed; the harness surfaced (and this PR fixes) three recipe bugs and correctly logged two competitor×protocol limitations as FAIL without wedging the run.

Signal already visible — **seq-write over NFS** separates the field exactly as the FUSE-tax thesis predicts:

| backend | nfs3 write MB/s | why |
| --- | --- | --- |
| local-disk (control) | ~470 | protocol ceiling |
| rclone | ~2050 | buffers locally (vfs-cache-mode=writes) |
| juicefs | ~8 | synchronous to S3 |
| s3fs | ~3 | synchronous to S3 |

(`dittofs-s3` native is wired but not yet in this run — it needs a config/bootstrap step on a bare VM; tracked as follow-up.)

## Fixes found by running it

- **Stale FUSE mounts** from an aborted run blocked rclone/s3fs remount ("directory already mounted") → `cleanMount` force-unmounts before every mount (also hardens teardown)
- **Daemonized-mount race** — rclone `--daemon` / juicefs `-d` return before serving, so the re-export saw an empty mount (the juicefs-nfs3 failure) → `waitMounted` polls `mountpoint -q` before re-exporting
- **Dropped `allow_other`** — Samba runs `force user = root` and nfsd is kernel, so it was never needed; sheds the `/etc/fuse.conf` dependency
- **Stale-cell masquerade** — a failed backend re-pulled a prior run's cells → the remote results dir is now wiped each run
- Core prereqs (fio, nfs/samba) must-succeed; competitor tools best-effort so one missing apt package (s3ql on noble) doesn't abort the rest

## Known follow-ups (not blockers)

- `dittofs-s3` native run needs the dfs config + dfsctl bootstrap baked into VM setup
- `rclone-nfs3` read and `s3fs-smb3` write hit `fio exit 4` — genuine competitor×protocol quirks (async-cache visibility over `actimeo=0`; s3fs sync-write over cifs), correctly logged as FAIL
- `s3ql` absent from Ubuntu noble repos — needs pip install or drop

Gates: gofmt clean, `go vet` clean, golangci-lint clean, 23 unit tests pass.


marmos91:
Pushed three follow-up commits:

**Restructure** — split the flat `cmd/bench` (`package main`) into small cohesive packages under `internal/dfsbench/` (`exec`, `config`, `fio`, `report`, `backend`, `cloud`, `run`), each owning its cobra command; `cmd/bench/main.go` just wires them. Every file is now ≤211 lines, acyclic DAG, behavior unchanged.

**Provider seam** — the VM runner no longer hard-assumes SCW: a `cloud.Provider` interface (`Provision`/`Terminate`) with `scwProvider` as the sole implementation today, selected via `--provider` (default `scw`). Adding another cloud is a new impl + one switch case; the remote-run mechanics were already provider-agnostic. VM arch is now detected over ssh (`uname -m`) and cross-built for, instead of hardcoding amd64.

**Copilot review** — addressed:
- `scwJSON` now captures stderr in the error
- `fusermount3` fallback in `cleanMount` for fuse3-only distros
- `run --remote` forwards the fully-merged config (engine/fio_bin/…), not just bucket/endpoint
- shell-quote forwarded `--systems`/`--workloads`/`--sizes`
- reject `--remote` with `--smoke`/`--local`/`--target`
- re-export `srcDir` 0755 not 0777
- refreshed the stale run help text
- `.bench-vm.json` + committed run logs untracked and gitignored

One **false positive**: the `loadVM` "returns zero-value VM" comment — Go evaluates the `Unmarshal` call before reading `vm` for the result, so `return vm, json.Unmarshal(data, &vm)` returns the populated struct (verified empirically; the round-trip test passes). Rewrote it to the explicit form anyway so it doesn't trip future readers.

PR #1615, opened 2026-07-09
feat: advertise SMB/NFS on the LAN via WS-Discovery + mDNS (#1609)

### LINKED ISSUE #1609: Advertise the service for network discovery - WS-Discovery (Windows/SMB) + mDNS/Bonjour (macOS/Linux, _smb._tcp + _nfs._tcp)
## Summary

DittoFS's SMB service does not appear in the **Windows Explorer → Network** tab or in **macOS Finder → Network**, because it doesn't advertise itself on the LAN via any discovery protocol. Enhancement request: implement network discovery — **WS-Discovery** (Windows, SMB only) and **mDNS/Bonjour** (macOS Finder + Linux/Avahi, advertising both `_smb._tcp` and `_nfs._tcp`) — so admins see the server and its SMB shares / NFS exports like a normal Samba/NFS file server.

Mounting already works fully by name/IP (`\\vm2.cubbit.local\ditto`, `\\192.168.100.50\ditto`); this is a discoverability/convenience gap, not a functional blocker.

## Investigation (current state)

- The **only** UDP listeners in the codebase are for NFS: `pkg/adapter/nfs/udp.go` and `internal/adapter/nfs/portmap/server.go`.
- The SMB adapter has **no** WS-Discovery responder (UDP `239.255.255.250:3702`), **no** NetBIOS Name Service (UDP 137/138), and **no** mDNS/`_smb._tcp` advertiser. A targeted search for `3702`, `_smb._tcp`, `ProbeMatch`/`ResolveMatch`, `WS-Discovery`, NBNS, etc. found nothing in the SMB path.
- DittoFS also **disables Windows `LanmanServer`** to bind port 445 directly (see the AD setup docs), so it cannot piggyback on Windows' own WSD/announcer either.

Net: nothing announces the service, so Explorer's Network tab never lists it.

## How Windows discovery works (for context)

- **Modern (Win10/11) Network tab** → Function Discovery via **WS-Discovery (WSD)**: the host answers multicast `Probe`/`Resolve` with `ProbeMatch`/`ResolveMatch` and sends `Hello`/`Bye` on start/stop, advertising a `Device`/`Computer`. This is what Samba deployments add via the external `wsdd`/`wsdd2` daemon — Samba itself does not implement WSD.
- **Legacy** browsing (Computer Browser / Browser protocol over NetBIOS + SMBv1) is deprecated and removed on current Windows — not worth implementing.
- **mDNS** (`_smb._tcp`) covers macOS Finder and some Linux file managers.

## Proposed enhancement

Cover **both** major desktop platforms with their native discovery protocol:

- **Windows (WS-Discovery)** — built-in, opt-in WSD responder in the SMB adapter:
  - Join `239.255.255.250:3702`, respond to `Probe`/`Resolve` with `ProbeMatch`/`ResolveMatch` advertising the server's hostname + `Computer` type, and emit `Hello` on startup / `Bye` on shutdown.
  - Config-gated (e.g. `smb.wsdiscovery.enabled`), advertising the configured NetBIOS/DNS name.
- **macOS / Linux (mDNS / Bonjour)** — first-class, so the server shows up in **Finder → Locations/Network** and the sidebar (and in GNOME/KDE via Avahi). A single in-process mDNS responder (multicast DNS on `224.0.0.251:5353`) advertises whichever adapters are enabled:
  - **SMB** → advertise **`_smb._tcp`** on port 445, instance name = configured server name.
  - **NFS** → advertise **`_nfs._tcp`**, with the SRV record carrying DittoFS's actual NFS port (**12049**, not the default 2049 — clients that assume 2049 would otherwise fail). A `path=` TXT record can name the export (e.g. `/ditto`) so Finder mounts the right export.
  - Also advertise **`_device-info._tcp`** with a `model=` TXT record (e.g. `RackMac`/`Xserve`) so Finder shows a sensible server icon instead of a generic one.
  - This is the same mechanism Samba/NFS deployments get via Avahi (`avahi-daemon` publishing `_smb._tcp` / `_nfs._tcp`); doing it in-process avoids the external dependency.
  - Config-gated per service (e.g. `smb.mdns.enabled`, `nfs.mdns.enabled`), each advertising only when that adapter is running.
- WS-Discovery stays **SMB-only** — Windows does not browse NFS servers, so there is no Windows Network-tab path for NFS. NFS discoverability comes solely from the mDNS `_nfs._tcp` record (plus the existing `showmount -e` / portmap query-a-known-host path).
- Skip legacy NetBIOS browsing (deprecated).
- Keep it host-level (advertises the server; SMB shares are still enumerated via `srvsvc`, NFS exports via MOUNT EXPORT, once the client connects).

## Notes / considerations

- WSD requires a UDP multicast listener alongside the TCP 445 listener; ensure it binds per-interface and respects the adapter's bind config.
- Purely additive and cosmetic — no impact on auth or access control.

## Environment

- Build: `develop` @ `ee4beb41`
- Server: `vm2.cubbit.local` domain member, SMB 3.1.1, share `/ditto`; `LanmanServer` stopped so dfs owns port 445.

marmos91:
## Summary

Implements **#1609** — DittoFS now advertises itself on the LAN so it appears in **Windows Explorer → Network** and **macOS Finder → Network**, the same job Samba/NFS deployments hand to the external `wsdd`/`wsdd2` and `avahi-daemon` daemons. Done **in-process, from scratch** (no reusable Go WS-Discovery *host* library exists; `golang.org/x/net/dns/dnsmessage` is already a transitive dep, so mDNS needs no new dependency), gated per service and toggleable at runtime via `dfsctl`/REST.

Mounting by name/IP already worked; this closes the discoverability gap only — no change to auth or access control.

## What's included

**A unifying `auxsvc.Service` abstraction** (`pkg/adapter/auxsvc`) — a small `Service` interface + `Group` for the companion services that run alongside an adapter. The **entire NFS auxiliary-service family** (embedded portmapper, system-rpcbind registration, UDP lock-manager transport, NSM startup) is folded into it, and the new advertisers join the same lifecycle. Behavior-preserving refactor.

**mDNS / DNS-SD** (`pkg/discovery/{hostinfo,mdns}`) — macOS Finder + Linux/Avahi. NFS advertises `_nfs._tcp` (port **12049**, `path=/<export>`); SMB advertises `_smb._tcp` + `_device-info._tcp` (`model=RackMac`). One shared, refcounted responder owns the single `224.0.0.251:5353` socket; each adapter registers only its own records.

**WS-Discovery** (`pkg/discovery/wsd`, SMB-only) — UDP `239.255.255.250:3702` SOAP responder (Hello/Bye/Probe→ProbeMatch/Resolve→ResolveMatch) **plus the HTTP `5357` metadata endpoint** whose `pub:Computer` relationship is what makes Explorer render the host as a *Computer*. AD members are labelled `Domain:`.

**Live toggle via dfsctl/API** — `MDNSEnabled` (NFS+SMB) and `WSDiscoveryEnabled` (SMB) plumbed end-to-end (DB models → REST PATCH/PUT/GET/reset → apiclient → `dfsctl adapter settings … update --mdns-enabled / --wsdiscovery-enabled`). Toggling starts/stops the advertiser **live**, no restart.

**Ports** — `Dockerfile` EXPOSE + operator Service/container ports (`5353/udp`, `3702/udp`, `5357/tcp`), with the caveat documented in code that multicast does not traverse standard K8s CNIs (hostNetwork only).

## Commits

1. `auxsvc.Service` abstraction
2. `pkg/discovery` mDNS + WS-Discovery responders
3. adapter wiring + live settings + Docker/operator ports + docs
4. `/simplify` cleanups (collapse duplicated reconcile logic, etc.)
5. `/code-review` fixes (see below)

## Review

Ran `/simplify` (4 agents) and a `/code-review` (correctness finders + verification). Fixes applied in commit 5, each with a regression test:

- **critical** — `wsd.Responder.Start` self-deadlock (sent Hello via `send()` while holding `r.mu`, which `send()` re-locks) — would have wedged the SMB sidecar group on first enable. Now covered by a Start/Stop lifecycle test.
- **high** — WS-Discovery SOAP templates didn't XML-escape substituted values; the inbound `MessageID` echoed into `RelatesTo` is LAN-controllable, so an unescaped `&`/`<` produced non-well-formed replies that clients discard.
- **high** — per-field settings reset rejected the new toggles (`dfsctl adapter settings … reset --setting mdns_enabled` → 400).
- medium/low — mDNS `stop()` racing a late `Register`; unlocked `loopCtx` read in the read loops; `auxsvc.StopAll` not clearing `baseCtx` (reconcile-vs-shutdown leak); spurious "already running" warn; non-deterministic NFS `path=`; AD Domain/Workgroup mislabel.

## Testing

- Unit tests throughout: mDNS record encode/decode + query responder (pure), WS-Discovery SOAP builders + parser + metadata endpoint (`httptest`), `auxsvc.Group` lifecycle, real-socket Start/Stop lifecycle tests (skip when the env can't bind multicast).
- `go build ./...`, `go vet ./...` clean (main module **and** the operator module); all touched suites pass; operator port-builder unit tests updated.
- **Not yet done:** on-the-LAN confirmation (Finder/Avahi listing; Explorer showing a *Computer*) — needs real macOS/Windows clients. Planned against the AD testbed.

image

## Known limitations (documented in code)

- Single default-interface multicast bind (multi-NIC join is a follow-up).
- No mDNS conflict-probing (v1).
- NFS `path=` TXT is not rebuilt on share add/rename/remove.
- Multicast discovery is LAN/hostNetwork-scoped; it does not traverse standard K8s overlays.

Closes #1609

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -62,7 +62,12 @@ USER 65532:65532
 # - 12049: NFS server (default)
 # - 12445: SMB server (default)
 # - 8080: REST API (health checks, management)
-EXPOSE 12049/tcp 12445/tcp 8080/tcp
+# - 5353/udp: mDNS / DNS-SD network discovery (macOS Finder, Linux Avahi) — issue #1609
+# - 3702/udp: WS-Discovery (Windows Explorer Network) — issue #1609
+# - 5357/tcp: WS-Discovery device metadata (HTTP) — issue #1609
+# Note: multicast discovery (5353/3702) only works on the host LAN / hostNetwork;
+# it does not traverse standard container/K8s overlay networks.
+EXPOSE 12049/tcp 12445/tcp 8080/tcp 5353/udp 3702/udp 5357/tcp
 
 # Volume mounts:
 # - /data/metadata: Metadata store (BadgerDB, etc.)
--- a/cmd/dfsctl/commands/adapter/settings.go
+++ b/cmd/dfsctl/commands/adapter/settings.go
@@ -216,18 +216,21 @@ var (
 	settingsPortmapperPort             int
 	settingsPortmapperRegisterWithSys  bool
 	settingsUDPEnabled                 bool
+	settingsMDNSEnabled                bool
 	settingsV4MinMinorVersion          int
 	settingsV4MaxMinorVersion          int
 	settingsV4MaxConnectionsPerSession int
 
 	// SMB update flags
-	settingsMinDialect         string
-	settingsMaxDialect         string
-	settingsSessionTimeout     int
-	settingsOplockBreakTimeout int
-	settingsMaxSessions        int
-	settingsEnableEncryption   bool
-	settingsSigning            string
+	settingsMinDialect            string
+	settingsMaxDialect            string
+	settingsSessionTimeout        int
+	settingsOplockBreakTimeout    int
+	settingsMaxSessions           int
+	settingsEnableEncryption      bool
+	settingsSigning               string
+	settingsSMBMDNSEnabled        bool
+	settingsSMBWSDiscoveryEnabled bool
 
 	// Common flags
 	settingsDryRun       bool
@@ -293,6 +296,7 @@ func registerNFSUpdateFlags(cmd *cobra.Command) {
 	cmd.Flags().IntVar(&settingsPortmapperPort, "portmapper-port", 0, "Portmapper listen port")
 	cmd.Flags().BoolVar(&settingsPortmapperRegisterWithSys, "portmapper-register-with-system", false, "Register NFS/MOUNT/NLM services with the host's system rpcbind on port 111, so kernel NFSv3 clients can lock without 'nolock' (restart to apply)")
 	cmd.Flags().BoolVar(&settingsUDPEnabled, "udp-enabled", false, "Serve NLM/NSM/MOUNT over UDP (needed for NFSv3 locking from macOS/BSD; restart to apply)")
+	cmd.Flags().BoolVar(&settingsMDNSEnabled, "mdns-enabled", false, "Advertise the NFS export over mDNS/DNS-SD (_nfs._tcp) for macOS Finder / Linux Avahi (applied immediately)")
 	cmd.Flags().IntVar(&settingsV4MinMinorVersion, "v4-min-minor-version", 0, "Minimum NFSv4 minor version (0=v4.0, 1=v4.1)")
 	cmd.Flags().IntVar(&settingsV4MaxMinorVersion, "v4-max-minor-version", 0, "Maximum NFSv4 minor version (0=v4.0, 1=v4.1)")
 	cmd.Flags().IntVar(&settingsV4MaxConnectionsPerSession, "v4-max-connections-per-session", 0, "Maximum connections per NFSv4.1 session (0=unlimited)")
@@ -310,6 +314,8 @@ func registerSMBUpdateFlags(cmd *cobra.Command) {
 	cmd.Flags().BoolVar(&settingsEnableEncryption, "enable-encryption", false, "Enable SMB encryption")
 	cmd.Flags().StringVar(&settingsSigning, "signing", "", "SMB message signing mode: disabled|enabled|required")
 	cmd.Flags().StringVar(&settingsBlockedOperations, "blocked-operations", "", "Comma-separated list of blocked operations")
+	cmd.Flags().BoolVar(&settingsSMBMDNSEnabled, "mdns-enabled", false, "Advertise the SMB service over mDNS/DNS-SD (_smb._tcp) for macOS Finder / Linux Avahi (applied immediately)")
+	cmd.Flags().BoolVar(&settingsSMBWSDiscoveryEnabled, "wsdiscovery-enabled", false, "Advertise the host over WS-Discovery so it appears in the Windows Explorer Network view (applied immediately)")
 	cmd.Flags().BoolVar(&smbSettingsDryRun, "dry-run", false, "Validate without applying changes")
 	cmd.Flags().BoolVar(&smbSettingsForce, "force", false, "Bypass range validation")
 }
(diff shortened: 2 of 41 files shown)


marmos91:
Addressed the CI failures and the Copilot review in `6136f5cc`:

**CI**
- **Data race** (`-race` unit-test job) — `wsd.Responder`'s HTTP `Serve` goroutine read `r.httpSrv` unlocked while `Stop` nils it under the lock. Now captured in a local for the goroutine, matching the existing `conn`/`loopCtx` captures. (The race was surfaced by the new Start/Stop lifecycle test — working as intended.)
- **golangci-lint** — checked `resp.Body.Close()` in `metadata_test.go`; reworked the `MessageID` uniqueness assertion to avoid the `SA4000` identical-expression warning.

**Copilot comments**
1. `auxsvc.Group.Reconcile` — now swallows the `StopOne` error (matches the doc; a benign stop no longer surfaces as a "failed to start" warning).
2. WS-Discovery `AppSequence` `InstanceId` — sourced from a per-adapter, strictly-increasing counter seeded at process start, instead of `time.Now().Unix()` per build, so a live toggle can never reuse an `InstanceId` with a rewound `MessageNumber`.
3. Base-context lifetime — the mDNS and WS-Discovery advertisers now tear down when the base context is cancelled without an explicit `Stop`, matching the ctx-driven NFS auxiliary services (the interface contract).

Each has a regression test.

marmos91:
Addressed the Copilot review comments — all three are already handled in the current branch:

1. `pkg/adapter/auxsvc/auxsvc.go` **Reconcile** — the `!want && running` branch now swallows the stop error (`_ = g.StopOne(name)`), matching the documented contract (StopOne already debug-logs it). No benign shutdown error is surfaced as a start failure.
2. `pkg/adapter/smb/discovery.go` **AppSequence InstanceId** — no longer `time.Now().Unix()`; each responder build takes a strictly-increasing value from the adapter's `wsdInstanceID` atomic counter. A live off→on toggle intentionally bumps the InstanceId (WS-Discovery treats it as a new service instance), and it never rewinds.
3. `pkg/discovery/wsd/responder.go` **Start ctx** — the base context is now honored: a watcher goroutine (`<-ctx.Done() → r.Stop(...)`) tears the responder down if the owning adapter's Serve context is cancelled without an explicit Stop, matching the ctx-driven NFS auxiliary services.

PR #1618, opened 2026-07-09
feat(smb): emit file owner/group as real AD SID in idmap:rid (#1617)

### LINKED ISSUE #1617: feat(smb): emit file owner/group as real AD SID (not the machine-domain SID) in idmap:rid
## Context

Follow-up from PR #1610 (#1607/#1608 — Windows Security tab). On an AD-joined share in `idmap:rid`, a file owned by an AD user (e.g. `admin.txt`, owner uid 500 = the domain Administrator) carries the **algorithmic machine-domain SID** (`S-1-5-21-<machine>-<uid>`) in its security descriptor, even though the same identity is granted/authenticated via its **real AD SID** (`S-1-5-21-<addomain>-<rid>`).

The reverse LSARPC resolution masks this cosmetically (owner still displays as `CUBBIT\Administrator`), and the referenced-domain list is now deduped by SID (PR #1610, commit d06fb7ff) so the earlier "two CUBBIT domains → raw SIDs → Explorer crash" cannot recur. But the owner's *underlying* SID is still the machine SID rather than its real AD SID — an internal inconsistency worth removing at the source.

## Proposal

Emit the file **owner** and **group** SIDs as their real directory (AD) SID when the uid/gid maps to a directory account, so the descriptor's owner/group uses the SAME SID the account's PAC and share grants carry. Local accounts (uid 0 / control-plane users) keep the machine-domain SID.

Both directions must be handled:
- **Emit** (`internal/adapter/smb/handlers/security.go` — `BuildSecurityDescriptorWithGrants`, `principalToSID`): owner/group SID = AD SID when the id resolves to a directory account, else machine SID.
- **Parse** (`ParseSecurityDescriptor` / `securityDescriptorOwnerGroupSIDs` + the `SET_INFO` owner path): map an AD SID back to its uid (rid mode: uid = RID) so a client that writes the SD back does not get its owner change rejected as unmappable.

Source the AD domain SID from the LDAP provider's discovered/cached domain SID (added in PR #1610, `pkg/identity/ldap/provider.go`). **Gate on the AD domain SID being known** so non-AD / local-only / smbtorture deployments are a no-op (machine SID, unchanged behavior).

## Acceptance

- File owned by an AD user shows owner `DOMAIN\name` backed by the **real AD SID**.
- Owner round-trips (read SD → write SD back) do not fail for AD-owned files.
- No change / no regression for local-only shares and the SMB conformance suite.

marmos91:
Follow-up to #1610. Closes #1617.

Emits a file's owner/group SID as its **real directory (AD) SID** when the uid/gid maps to a directory account, instead of the algorithmic machine-domain SID — so an AD-owned file's owner uses the same SID its PAC and share grants carry. This also removes the last source of the multi-domain LSARPC crash class (`d06fb7ff` fixed it minimally by deduping referenced domains by SID; emitting the owner as its true AD SID means the owner and the AD grants share one referenced domain to begin with).

Local accounts keep the machine SID; the bridge only hits when the directory actually holds the account, so **non-AD / local-only / smbtorture deployments are a byte-for-byte no-op**.

### Design
Machine SIDs use the Samba algorithmic RID (`uid*2+1000`); AD SIDs in `idmap:rid` use `RID = uid` against the AD domain SID — a separate path. A package-level `DirectorySIDBridge` hook in `handlers` couples uid/gid ↔ AD SID in both directions:

- **Emit** (`ownerSIDFor`/`groupSIDFor`): owner/group SD fields and the `OWNER@`/`GROUP@` special ACEs advertise the AD SID when the bridge maps it, else fall back to the machine SID.
- **Parse** (`directoryUIDForSID`/`directoryGIDForSID`): when Windows echoes the AD-SID owner back through SET_INFO, recover the same uid/gid so the change round-trips instead of tripping the `#1228` unmappable-owner gate.

The bridge (`pkg/adapter/smb/directory_sid_bridge.go`) wraps `identity.Resolver` (`SIDForUID/SIDForGID` emit, `Resolve` by SID for parse) and is wired in `wireForeignSIDResolver` next to the LSARPC foreign-SID resolver — same resolver chain, same config-change re-run.

### Changes
- [x] `ldap.Provider.SIDForUID/SIDForGID` (uid/gid → directory SID; reverse lookup returns objectSid)
- [x] `identity.Resolver.SIDForUID/SIDForGID` (cached, provider-chain delegation)
- [x] `handlers` `DirectorySIDBridge` hook + emit/parse integration in `security.go`
- [x] `pkg/adapter/smb` bridge over `identity.Resolver` + wiring
- [x] Unit tests (emit AD SID, machine-SID fallback no-op, AD-owner round-trip, OWNER@ ACE)
- [x] AD testbed validation — mounted as alice; `admin.txt` owner shows `CUBBITAdministrator` (real AD SID), all rows resolve `CUBBIT...`, no crash, no raw SIDs

### DIFF
--- a/internal/adapter/smb/handlers/security.go
+++ b/internal/adapter/smb/handlers/security.go
@@ -238,6 +238,30 @@ func directoryGIDForSID(s *sid.SID) (uint32, bool) {
 	return 0, false
 }
 
+// sameSID reports whether two SIDs are identical. A nil operand never matches.
+// Used by the SET_INFO owner/group gate to recognize a re-set of the file's
+// current owner/group SID as a no-op (#1617).
+func sameSID(a, b *sid.SID) bool {
+	if a == nil || b == nil {
+		return false
+	}
+	return sid.FormatSID(a) == sid.FormatSID(b)
+}
+
+// isCurrentOwnerSID reports whether reqSID equals the SID the server currently
+// emits for a file owned by uid — the machine-domain SID, or the real directory
+// SID when the bridge maps the uid (#1617). The SET_INFO gate uses this to
+// distinguish a no-op owner re-set (accept) from a genuine unmappable chown
+// (reject, #1228), independent of whether the reverse directory lookup is up.
+func isCurrentOwnerSID(reqSID *sid.SID, uid uint32) bool {
+	return sameSID(reqSID, ownerSIDFor(_defaultSIDMapper.Load(), uid))
+}
+
+// isCurrentGroupSID mirrors isCurrentOwnerSID for a file's group GID.
+func isCurrentGroupSID(reqSID *sid.SID, gid uint32) bool {
+	return sameSID(reqSID, groupSIDFor(_defaultSIDMapper.Load(), gid))
+}
+
 // ============================================================================
 // Security Descriptor Building
 // ============================================================================
--- a/internal/adapter/smb/handlers/security.go
+++ b/internal/adapter/smb/handlers/security.go
@@ -137,6 +137,108 @@ func GetSIDMapper() *sid.SIDMapper {
 }
 
 // ============================================================================
+// Directory SID Bridge (#1617)
+// ============================================================================
+
+// DirectorySIDBridge couples a POSIX UID/GID to its real directory (AD) SID in
+// both directions, backing #1617. Emit (SIDForUID/SIDForGID) advertises a file
+// owner/group as the account's actual AD SID instead of the algorithmic
+// machine-domain SID from the SIDMapper; parse (UIDForSID/GIDForSID) recovers
+// the UID/GID when Windows echoes that AD SID back through SET_INFO so an owner
+// emitted this way round-trips instead of tripping the unmappable-owner gate.
+//
+// Every method returns ok=false on a miss (never an error) and the caller falls
+// back to the SIDMapper's machine-domain mapping. Because the bridge only hits
+// on accounts the directory actually holds, a local-only or conformance
+// deployment (no AD provider) is a transparent no-op: SIDForUID/GID always miss
+// and the descriptor stays byte-for-byte what it was before #1617.
+type DirectorySIDBridge interface {
+	// SIDForUID resolves a POSIX UID to its directory user SID string.
+	SIDForUID(uid uint32) (sidStr string, ok bool)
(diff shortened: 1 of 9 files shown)


marmos91:
Addressed the Copilot comments plus a code-simplifier / code-reviewer pass. Commit `ae413792`.

**Copilot**
- `provider.go` `SIDForUID`/`SIDForGID` now return `ok=false` when the object's `objectSid` is missing/undecodable (empty string is a miss, not a hit).
- `reverseLookup` logs an `objectSid` decode failure / absence at Debug, so an objectSid read regression is diagnosable instead of a silent machine-SID fallback.
- Added `Resolver.SIDForUID/SIDForGID` unit tests: `_Chain` (first SID-capable provider wins, plain provider skipped), `_Cached` (hit+miss cached, matching `LookupUID`), `_EmptyIsMiss` (provider returning `("", true)` is treated as a miss).

**Code-reviewer (2 HIGH, both made live by this change — fixed)**
- **rfc2307 group GID round-trip:** a group object carries `gidNumber` but no `uidNumber`, so `deriveUIDGID` fell through to the RID path and `GIDForSID` recovered the RID — a Group SD round-trip would silently rewrite a file's gid. Now honors a group's stamped `gidNumber` directly. Regression test `TestDeriveUIDGID_RFC2307_GroupUsesGidNumber`.
- **foreign-domain owner gate:** emitting the owner as a foreign AD SID meant a transient reverse-lookup miss turned a Windows no-op owner re-set into `StatusInvalidOwner`. The gate now accepts a re-set of the file's *current* owner/group SID as a no-op independent of directory reachability (only a genuine change to an unmappable foreign SID is rejected). Test `TestSetInfo_DirectoryBridge_CurrentOwnerGroupSIDIsNoOp`.
- *Negative-cache poisoning (LOW):* largely neutralized by the no-op fix above — a no-op re-set no longer depends on the reverse path being up. Distinguishing "no such object" from "directory unreachable" on the parse cache is left as a pre-existing follow-up (same pattern as `reverseLookup`).
- *Local/AD uid namespace collision (LOW):* inherent to sharing one POSIX id space under RFC2307; noted, not gated.

**Code-simplifier**
- Extracted the shared `resolveSID` helper in the bridge (deduped the `Resolve`+timeout body of `UIDForSID`/`GIDForSID`).

PR #1623, opened 2026-07-10
perf(block): don't block FILE_SYNC writes on a synchronous S3 drain

### LINKED ISSUE #1621: perf(write): O_DIRECT / FILE_SYNC NFS writes stall at multi-second per-op latency
## Summary

The `dfsbench` fair-run work (#1619) surfaced a real pathology in the **synchronous** NFS write path (O_DIRECT / per-write FILE_SYNC), distinct from the buffered/UNSTABLE path that #1584 optimised.

## What we measured

`seq-write.fio` = `rw=write bs=1m direct=1 fsync_on_close=1` (4 jobs, iodepth 1) over native NFSv3 to an S3-backed share, 256 MiB:

- Throughput: **~1–2.6 MB/s** (varies run to run)
- **p50 latency: 25 ms – 4.5 s per 1 MB write** (highly variable, bimodal)
- CPU: **~3%** — the server is *stalling*, not working

Arithmetic check: 4 jobs × (1 MB ÷ 4.5 s) ≈ 0.9 MB/s ≈ observed. So the bottleneck is the multi-second per-write latency, not CPU or bandwidth.

## Why it matters

`direct=1` over NFS forces each block through a durability barrier the server **still fsyncs on** — the commit message for #1584 explicitly lists FILE_SYNC as "still fsyncs before ack". So this is the path #1584 opts *out* of. But an NVMe append-log fsync is single-digit ms; **4.5 s is not explainable by fsync alone.** Candidates to investigate:

- a **synchronous S3 PUT** in the FILE_SYNC hot path (should be async/local-first)
- the per-store `syncLeader` stalling when *every* write demands its own barrier (coalescing adds latency with no batching benefit)
- SSI-retry churn on metadata under per-write commits
- O_DIRECT forcing read-modify-write on chunk boundaries

Real clients live on this path: O_DIRECT databases, `sync` mounts, SMB write-through.

## Suggested next step

`#1555` already added per-stage write-path CPU/mem profiling — run it under `direct=1` seq-write and find where the multi-second stall is. Compare against the buffered (`seq-write-buffered`) path which should be fast.

## Context / caveat

This is **not** the "DittoFS is 1900× slower than rclone" framing from the first bench pass — that number conflated O_DIRECT durability (which DittoFS honours and rclone's writeback skips) with throughput. On the realistic buffered path DittoFS is expected to be fine (48–510 MB/s historically). This issue is specifically about the synchronous path's pathological *latency*.

marmos91:
## Problem

`dfsbench` fair-run work (#1619) surfaced a real pathology in the **synchronous** NFS write path (O_DIRECT / per-write FILE_SYNC), distinct from the buffered/UNSTABLE path #1584 optimised. `seq-write.fio` with `direct=1` over native NFSv3 to an S3-backed share stalled at **~1–2.6 MB/s, p50 25 ms – 4.5 s per 1 MB write, ~3% CPU** — the server was *blocked*, not busy. The buffered path ran at 242 MB/s.

## Root cause

A `FILE_SYNC`/`DATA_SYNC` WRITE reaches the shared commit seam per write:

`flushStableWrite` → `common.CommitBlockStore` → `engine.Store.Flush` → `syncer.Flush` → `carveFlush(drainAll=true)` → `wg.Wait()` on an inline `s3.PutObject`.

The local append-log fsync (`SyncPayload`) already makes the payload crash-safe (`localDurable=true` for the fs local store). The subsequent synchronous carve/upload to S3 is what the WRITE reply waited on — a network round-trip per 1 MB write.

`common.CommitBlockStore` already **documents** that in the default policy "the remote mirror stays fully asynchronous" — but `engine.Flush` ran the drain synchronously before that decision, contradicting the contract.

## Fix

In `engine.Store.Flush`, once the local durability barrier is paid, return without the syncer carve drain when `LocalDurable() && !RequireDurableCommit()`. The background carve loop mirrors the data asynchronously (exactly as the fast UNSTABLE path already relies on). A strict share (`require_durable_commit=true`) still drains inline; a volatile local store (memory) still drains inline.

Ruled out: the per-store fsync leader (lone `direct=1` writer hits the inline bypass, no added latency), rollup ticker / upload semaphore (only in the background idle loop), and O_DIRECT read-modify-write (append-log never fetches back).

## Test

`TestEngine_Flush_DurableLocalDefault_NoSyncRemote` — deterministic discriminator with the background carve loop unstarted: a strict flush drains to the remote synchronously (non-empty), the default flush does not (empty). Fails against the unfixed code.

Verified: full `pkg/block/...` (1213) + `internal/adapter/...` (4305) suites pass; `go vet` + `golangci-lint` clean.

Fixes #1621

### DIFF
--- a/pkg/block/engine/durability_test.go
+++ b/pkg/block/engine/durability_test.go
@@ -1,11 +1,59 @@
 package engine
 
 import (
+	"context"
 	"testing"
 
 	localmemory "github.com/marmos91/dittofs/pkg/block/local/memory"
 	remotememory "github.com/marmos91/dittofs/pkg/block/remote/memory"
 )
+
+// TestEngine_Flush_DurableLocalDefault_NoSyncRemote proves the #1621 fix: on a
+// durable local store under the default (async-remote) policy, Flush satisfies
+// durability with the local fsync alone and does NOT block the ack on a
+// synchronous remote carve/upload — that inline S3 PutObject-per-FILE_SYNC-WRITE
+// was the multi-second write stall. A strict share still drains inline.
+//
+// Discriminator: the store's background carve loop is never Start()ed, so the
+// only path to the remote is the synchronous carve inside Flush itself. Rollup
+// is forced first so there are carve-ready chunks; a non-empty remote after
+// Flush therefore means the drain ran synchronously on the ack path.
+func TestEngine_Flush_DurableLocalDefault_NoSyncRemote(t *testing.T) {
+	writeAndFlush := func(t *testing.T, strict bool) int {
+		t.Helper()
+		ctx := context.Background()
+		mem := remotememory.New()
+		mem.SetDurable(true)
+		fx := newCarveFixture(t, mem, DefaultBlockCarveBytes) // durable fs local, ManualSync, wired carve
+		bs, err := New(BlockStoreConfig{
+			Local:           fx.local,
+			Remote:          mem,
+			Syncer:          fx.syncer,
+			FileChunkStore:  fx.ms,
+			SyncedHashStore: fx.ms,
+		})
+		if err != nil {
+			t.Fatalf("New: %v", err)
+		}
+		t.Cleanup(func() { _ = bs.Close() })
+		bs.SetRequireDurableCommit(strict)
+
+		// Register a carve-ready CAS chunk exactly as the rollup's onChunkComplete
+		// hook would — the state a FILE_SYNC WRITE reaches by the time it flushes.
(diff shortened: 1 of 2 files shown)


marmos91:
Fixed: the local-durable/default branch now calls `bs.local.SyncFileChunksForFile(ctx, payloadID)` — exactly what `syncer.Flush` step 1 does — before returning. Only the synchronous remote carve drain (the inline S3 PutObject that caused the #1621 stall) is skipped; strict shares still drain inline. Pushed as a follow-up commit.

PR #1624, opened 2026-07-10
fix(block): run cas→blocks repack in background, serve standalone chunks via fallback

### LINKED ISSUE #1622: Startup hang: repackStandaloneChunks migration blocks server start on a synchronous S3 PutBlock (dd65d16b)
## `develop` (dd65d16b) hangs on startup — synchronous `repackStandaloneChunks` migration blocks on an S3 `PutBlock`

Upgrading a real deployment from `8e376d36` → `dd65d16b` (which includes the #1567 parallel-upload fix), **dfs never finishes starting**: control-plane `:8090` never comes up, NFS `:12049` / SMB `:1445` never bind, `/health` returns nothing. Only the config-load banner is logged (even with `DITTOFS_LOGGING_LEVEL=DEBUG`), then silence. The process sits at **0% CPU and 0 disk I/O** indefinitely.

### Root cause (goroutine dump via SIGQUIT)
`goroutine 1` (main / startup) is blocked in an **S3 `PutObject`** waiting on the HTTP roundtrip, driven by a **startup migration**:

```
net/http ... (*Transport).RoundTrip                     # blocked in select, awaiting S3 response
github.com/aws/aws-sdk-go-v2/service/s3 (PutObject middleware chain)
github.com/marmos91/dittofs/pkg/block/remote/s3.(*Store).PutBlock
github.com/marmos91/dittofs/pkg/block/engine.(*Syncer).packAndCommitMigrated
github.com/marmos91/dittofs/pkg/block/engine.(*Syncer).repackStandaloneChunks.func1
```

So `repackStandaloneChunks` runs **synchronously on the main startup path** and issues S3 `PutBlock`s to repack standalone chunks into the new carve format. One of those uploads stalls, and because it's on the critical startup path with no async/timeout/progress, **the whole server never starts**.

### Why it bites here
- Remote is Cubbit DS3; box uplink to Cubbit is modest and the local disk is near-full (97%). A migration that re-reads + re-uploads the store's standalone chunks is both slow and (on a ~full disk) may not have room to stage repacked blocks — either way it wedges startup with no operator recourse (you can't reach the control plane to intervene).
- The prior build (`8e376d36`) has no such migration and starts cleanly in seconds on the exact same store.

### Recovery
Rolled back the binary to `8e376d36` → dfs starts healthy immediately, all adapters bind, data intact. So this is specific to the startup repack migration in the newer build.

### Suggestions
- Run `repackStandaloneChunks` **asynchronously in the background** after the server is serving (like async GC in #1479), not on the blocking startup path.
- Make it **resumable + progress-logged**, and **bounded/timed** per PutBlock so a slow or stalled remote degrades gracefully instead of hanging the server forever.
- Guard against **near-full disk** (fail fast with a clear error rather than blocking), since repacking needs staging headroom.

### Env
- dittofs `develop` `dd65d16b`, linux/amd64, 8 GB RAM, disk 97% full (~57 GiB live data, `pin` share), Cubbit DS3 remote. Reproduces every restart on this store.

Happy to grab the full goroutine dump or run a patched build with the migration gated off.

marmos91:
## Problem

Fixes #1622. Upgrading a real deployment (`8e376d36` → `dd65d16b`) hangs on startup: the control plane never binds, NFS/SMB never bind, `/health` returns nothing, process sits at 0% CPU. A SIGQUIT dump shows `goroutine 1` (startup) blocked in an S3 `PutObject` driven by `repackStandaloneChunks` → the cas→blocks legacy migration ran **synchronously on the critical startup path**, and one stalled upload wedged the whole server with no operator recourse.

## Fix — online (background) migration with a read fallback

This is the standard online-migration pattern (dual-read + migrate in the background), the same shape #1479 used for async GC:

1. **Async migration** (`engine.go`) — `migrateLegacyCAS` now runs in a background goroutine spawned in `Store.Start` instead of blocking. `Store.Close` cancels its (detached) context and waits on `migrateDone` before tearing down the local/remote/syncer stores it uses. A slow/stalled remote or near-full disk no longer wedges startup — the pass is idempotent/resumable and simply retries next start.

2. **Read-path fallback** (`fetch.go`, `legacy_migration.go`) — the read path previously *refused* a synced chunk with an empty `BlockID` ("post-migration drift"), which is why the migration had to be a blocking startup gate. Making it async requires the reader to still understand the legacy layout, so `resolveAndReadChunk` now serves standalone chunks via new `readStandaloneChunk` (local-first BLAKE3-verified, then `legacyCAS().ReadLegacyChunkVerified`). A chunk resident **nowhere** still fails closed with `ErrChunkNotFound` (no silent zeros). The pre-existing single re-resolve-on-`ErrChunkNotFound` retry in `dispatchRemoteFetch` absorbs the migration's commit-then-delete window.

Net effect: the share serves immediately; the repack (and `cas/` purge) happen lazily in the background.

## Testing

- `go build ./...`, `go vet ./pkg/block/...`, `gofmt -s` — clean
- `go test -race ./pkg/block/engine/` and full `./pkg/block/...` — green
- Updated the tests that encoded the old blocking/refuse invariants; added `TestReadPath_StandaloneLocatorServedViaFallback` + `…MissingEverywhere` (serve vs. fail-closed) and the dispatch-layer equivalents.

## Notes

- Read semantics reversal is deliberate: the legacy-CAS read path (removed in #1493 PR4) is re-added as the fallback — inherent to any correct online migration (you can't lazily migrate a format you can't still read).
- Per-`PutBlock` timeout (issue suggestion #2) intentionally skipped: the async move already removes the "hangs forever" failure mode. Easy follow-up if we want the repack itself to give up faster.

### DIFF
--- a/pkg/block/engine/engine.go
+++ b/pkg/block/engine/engine.go
@@ -138,6 +138,13 @@ type Store struct {
 	closed   bool  // guarded by closeMu; true once teardown has run
 	closeErr error // memoized result of the first Close (idempotent)
 
+	// migrateCancel/migrateDone govern the background cas→blocks migration
+	// goroutine spawned by Start. Close cancels the context and waits on the
+	// channel so the goroutine's in-flight remote I/O never races the store
+	// teardown below it. Both nil/zero until Start runs.
+	migrateCancel context.CancelFunc
+	migrateDone   chan struct{}
+
 	// requireDurableCommit gates the strict honest-CLOSE/COMMIT rule (#1274).
 	// When false (the default), CommitBlockStore acks once engine.Flush
 	// succeeds regardless of local/remote durability — the remote mirror
@@ -570,15 +577,24 @@ func (bs *Store) Start(ctx context.Context) error {
 	// Use background context so these outlive the calling request context.
 	bs.local.Start(context.Background())
 
-	// One-shot cas→blocks migration (#1493 PR4): import pre-flip local
-	// per-chunk files into the log-blob substrate and re-pack standalone
-	// remote objects into packed blocks. Blocking by design — the share must
-	// not serve until the legacy layout is gone — and a no-op on migrated
-	// (or fresh) stores. Runs BEFORE the syncer starts so no mirror/carve
-	// activity races the rewrite.
-	if err := bs.migrateLegacyCAS(ctx); err != nil {
-		return fmt.Errorf("cas→blocks migration: %w", err)
-	}
+	// One-shot cas→blocks migration: import pre-flip local per-chunk files
+	// into the log-blob substrate and re-pack standalone remote objects into
+	// packed blocks. Runs in the background so a slow/stalled remote or a
+	// near-full disk can't wedge startup — the share serves immediately and any
+	// not-yet-repacked standalone chunk is read through the legacy fallback
+	// (resolveAndReadChunk). Idempotent and resumable: a failed or cancelled
+	// pass is a no-op retry on the next start. Uses a detached context so it
+	// outlives Start; Close cancels it and waits on migrateDone before tearing
+	// down the stores it uses.
+	migrateCtx, cancel := context.WithCancel(context.Background())
+	bs.migrateCancel = cancel
+	bs.migrateDone = make(chan struct{})
+	go func() {
+		defer close(bs.migrateDone)
+		if err := bs.migrateLegacyCAS(migrateCtx); err != nil && migrateCtx.Err() == nil {
+			logger.Warn("cas→blocks migration: background pass failed; will retry next start", "error", err)
+		}
+	}()
 
 	// Wire the health callback BEFORE starting the syncer. The health monitor
 	// captures the callback at Start time (startHealthMonitor reads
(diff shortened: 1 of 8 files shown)


marmos91:
## Real end-to-end migration test (SCW VM + real S3)

Validated on a fresh Scaleway **PRO2-XXS (8 GiB RAM — matches the reporter's env)** against **real SCW Object Storage (S3)**. Flow: build the reporter's pre-upgrade commit **8e376d36** (`dfs-old`) and this branch (`dfs-new`), have the old binary create a pre-flip `cas/` store over NFS, then start the fixed binary on the same store and exercise it.

### Result — reporter's real setup (durable `fs` local + S3), cold-cache verified

All reads done after `drop_caches` (served from the server, not the NFS client cache); each read path is BLAKE3-verified:

| Check | Result |
|---|---|
| Control plane `/health/ready` with 48 standalone chunks pending repack | **0.79 s** (pre-fix: blocks forever) |
| cas→blocks migration runs in background to completion | **PASS** — `cas/` 48→0, `repacked_chunks=48` |
| Standalone files served byte-identical **while** repacking (read fallback) | **PASS** |
| Pre-existing files byte-identical **after** migration | **PASS** |
| Files written **during** migration read back intact | **PASS** |

So: the server binds immediately instead of hanging, serves reads/writes throughout, and the background migration completes with data byte-identical before/during/after.

### Note on an initial confounded run

A first pass using a **volatile memory** local store showed a checksum mismatch. I isolated it: the **8e376d36 binary does not round-trip its own `cas/` objects** across a restart — with *no migration involved*, writing `0fd9932b…` and reading it back from S3 after a restart yields `0529397c…`. That's a defect in the old binary's standalone-remote read path, not this change (and the after-migration reads that also mismatched go through the untouched block path, not the new fallback). With a **durable local** store — the reporter's actual configuration — every check passes, because the migration reads clean bytes from the local tier.

VM + bucket torn down.

PR #1625, opened 2026-07-10
perf(block): parallelize the cold-read demand fetch

marmos91:
## Problem

A cold read — one that needs CAS blocks not in the local cache — fetched them from the S3 remote **one blocking GET at a time** in the caller's goroutine (`Syncer.EnsureAvailableAndRead`). Throughput pinned at `blockSize / latency`: one round-trip per 8 MiB block ≈ **159 MB/s** at ~50 ms RTT (measured, and the arithmetic matches exactly). Competitors saturate S3 with many parallel GETs.

## Fix

Fan the missing-block fetches out with bounded concurrency. Rather than add a *fourth* copy of the errgroup+semaphore pattern (WarmAll, the download pool, the prefetch pool, the upload limiter already exist), extract one shared helper:

```go
func (m *Syncer) fetchGroup(ctx) (*errgroup.Group, context.Context) // bounded by ParallelDownloads
```

Both fan-out-and-wait fetchers — the cold-read demand loop **and** `WarmAll` — use it, so there's a single place that decides download concurrency. This also **replaces WarmAll's hand-rolled semaphore**, making the change a net simplification (warm.go shrinks). The background queue-drain pools and the *resizable* upload limiter keep their own primitives on purpose — they have a different shape.

Safety: each block writes a **disjoint** region of `dest`, and `inlineFetchOrWait`'s in-flight map already dedups concurrent callers, so the fan-out is race-free (verified under `-race`).

## Test

`TestColdRead_DemandFetchIsConcurrent` asserts the **peak in-flight GET count** structurally via a latency-injecting fake remote — deterministic, no wall-clock timing (which the bench rig has proven too noisy to trust). Serial code ⇒ maxInFlight=1 (the test fails); the fix ⇒ concurrent GETs. Uses distinct per-block content so CAS dedup can't mask the loop behavior.

Verified: `pkg/block/...` (1211) + `internal/adapter/...` (4305) pass, `-race` clean, `go vet` + `golangci-lint` clean.

## Follow-ups (not in this PR)
- Async bounded cache write-back (the inline 8 MiB `local.Put` per block) — separate issue.
- The geometric prefetch ramp / doubled per-block metadata resolve.

### DIFF
--- a/pkg/block/engine/cold_read_demand_concurrency_test.go
+++ b/pkg/block/engine/cold_read_demand_concurrency_test.go
@@ -1,0 +1,59 @@
+package engine
+
+import (
+	"context"
+	"testing"
+	"time"
+
+	memorylocal "github.com/marmos91/dittofs/pkg/block/local/memory"
+	metastore "github.com/marmos91/dittofs/pkg/metadata/store/memory"
+)
+
+// TestColdRead_DemandFetchIsConcurrent pins the cold-read fix: a single read
+// spanning many not-local blocks must fetch them from the remote CONCURRENTLY,
+// not one blocking S3 GET per block. Before the fix the demand loop in
+// EnsureAvailableAndRead was serial, which pinned cold-read throughput at
+// blockSize/latency (one GET per round-trip — the ~159 MB/s wall).
+//
+// The assertion is STRUCTURAL, not wall-clock timing: the latency-injecting
+// remote records the peak number of GETs in flight at once. Serial fetching
+// pins that at 1; the bounded parallel fan-out drives it up to nBlocks. This
+// keeps the test deterministic and free of timing flakes.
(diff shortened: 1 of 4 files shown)


marmos91:
Holding this as a draft. A same-VM e2e A/B showed it **flat** for NFS cold seq-read: the Linux NFS client caps rsize at 1 MB and SMB MaxReadSize is 1 MB, both below the 8 MiB block size, so a single server read never spans multiple blocks and the demand loop is always ~1 iteration — nothing to parallelize on the normal path.

The change is correct and race-tested (and the warm.go dedup is a genuine simplification), but per 'don't ship latent complexity' it shouldn't merge as a standalone 'cold-read fix.' It becomes live only once reads span multiple blocks (read size > block size). Folding into the prefetch-pipelining work — the actual cold-read lever — where fetchGroup is reused. Tracked in the new prefetch issue.

Issue #1626, opened 2026-07-10
perf(block): async cache write-back on the cold-read path


marmos91:
## Context

Follow-up from #1625 (cold-read demand-fetch parallelization). While tracing the cold-read path, the second-biggest structural stall after serial fetching is the **inline cache write-back**.

## Problem

In `inlineFetchOrWait` → the fetched block is persisted to the local CAS tier with a **synchronous** `local.Put(ctx, hash, data)` (`pkg/block/engine/fetch.go`, ~line 487) *before* the bytes are returned to the reader. So every demand block pays an **8 MiB local-disk write on the read's critical path**, on top of a full blake3 re-verify.

With #1625 the S3 GETs now run in parallel, which makes this inline write-back a larger relative share of cold-read latency.

## Why it's not a trivial revert

Background write-back was tried before and reverted (#1362): the naive version spawned a goroutine per block, each pinning 8 MiB of data, and OOM'd under load. So the correct form is **not** one goroutine per block — it's a **bounded write-back queue**: a fixed worker count with backpressure, so at most N×8 MiB is in flight, and a full queue blocks (or falls back to inline) rather than growing unboundedly.

## Proposed approach

- Return the verified bytes to the caller first, enqueue the `local.Put` to a bounded write-back pool.
- Bound memory explicitly (fixed workers × block size); apply backpressure when full.
- Preserve the #1362 invariant: a read that observes a block as "cached" must not race a not-yet-persisted write-back into re-fetching (keep the in-flight/synced markers correct across the async boundary).
- Deterministic test: assert the read returns before the `local.Put` completes (inject latency into the local store's Put), and that total in-flight write-back memory stays bounded.

## Value

Removes an 8 MiB disk write from every cold demand block's critical path. Complements #1625; together they target the full cold-read wall.

marmos91:
Closing as obsolete. The cost model this issue targeted was invalidated by the journal switchover.

- The `local.Put` on the "local CAS tier" (an 8 MiB disk write + BLAKE3 re-verify on the read's critical path) no longer exists. The write-back at `fetch.go` is now `Hydrate` → buffered journal append (`appendRecord`, `synced=true`) which **never fsyncs per call** — fsync is amortized at segment seal/rotation. The expensive thing to defer is gone; what remains is a memcpy + buffered `write()` to page cache under a shard lock.
[...]

PR #1636, opened 2026-07-10
perf(block): offset-based sliding-window readahead for cold-from-S3 reads

marmos91:
## Summary

Raises cold-from-S3 sequential-read throughput by keeping the local read-serving tier populated **ahead** of the reader, instead of demand-fetching each block on the S3-GET critical path.

**Measured on a real VM (Scaleway POP2-8C-32G, fr-par S3, seq-read large, genuinely-cold):**

| dfs | cold MB/s | p50 | p99 |
|---|---|---|---|
| develop | 74 | 3.6 ms | **566 ms** |
| this branch | **115** | 4.0 ms | **13 ms** |

**+55% throughput, and the p99 tail collapses 566 ms → 13 ms (43×)** — the reader no longer stalls on the ~566 ms S3-GET tail because the window has already fetched the block.

> Note: the cold measurement is only trustworthy with the eviction fix in #1635 (which makes the "cold" pass actually fetch from S3 instead of the local block volume). Merge that first for reproducibility.

## Root cause

The old read-prefetch only advanced on a local **miss** (inside `EnsureAvailableAndRead`). Once a sequential reader started serving blocks from the local tier, `EnsureAvailableAndRead` was no longer reached, so the window stopped sliding exactly when it needed to stay ahead. The geometric 1→2→4 ramp also never caught up (the reader drains a local block faster than one S3 fetch completes).

## What changed

- **Drive readahead from `Store.ReadAt` on EVERY read** via an offset/frontier-based sliding window (`Syncer.scheduleReadahead` + `planWindow`). Keeps a full fixed window (`PrefetchBlocks`) of blocks fetched ahead of the read frontier, each scheduled exactly once (`scheduledUpTo`); random access schedules nothing.
- **Route the SyncQueue prefetch worker (`fetchBlock`) through `inlineFetchOrWait`** so prefetch registers in the in-flight dedup map: a demand read for an in-flight block piggybacks instead of issuing a duplicate S3 GET. `fetchBlock` also skips already-local blocks.
- **Retire the dead read-prefetch paths**: the geometric on-miss `enqueuePrefetch`, the never-firing `cache.OnRead` read hint (the RAM cache is off the read path), and the now-orphaned `planReadahead`/`enqueuePrefetch`/`blockIsLocal`/`computeFileSize`.

## Tests

Deterministic guards for `planWindow` (schedule-once, slides, random-reset) and `scheduleReadahead` (fires on every read incl. local hits — fails on develop). 334 engine + 1275 block/adapter tests green under `-race`; `go vet` + `golangci-lint` clean.

## Follow-up

~115 MB/s still trails the ~688 MB/s 16-wide S3 ceiling. The next lever is removing the synchronous `local.Put` staging from the fetch path (async bounded write-back / RAM L0, #1626) so S3 bandwidth reaches the reader directly.

### DIFF
--- a/pkg/block/engine/engine_test.go
+++ b/pkg/block/engine/engine_test.go
@@ -1,6 +1,7 @@
 package engine
 
 import (
+	"bytes"
 	"context"
 	"strings"
 	"sync"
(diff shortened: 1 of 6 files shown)


marmos91:
Addressed Copilot review:
- **fetch.go comment** — reworded; each block is scheduled at most once (via `scheduledUpTo`), the worker just skips a block that a demand read staged locally in the meantime.
- **readahead.go dropped-enqueue** — documented as intentional rather than adding rollback logic. A dropped best-effort prefetch is **not** a window hole: the reader's own demand fetch (`EnsureAvailableAndRead`) still serves the block, just without the head start. Rolling `scheduledUpTo` back on a drop would cause re-enqueue churn every read under a saturated queue, deepening the saturation — degrading prefetch→demand is the correct backpressure. Comment added at the enqueue loop.

Issue #1629, opened 2026-07-10
SMB: NTLM authentication for AD users via netlogon pass-through (enables Explorer double-click browsing of a discovered server)


marmos91:
## Summary

With LAN service discovery (#1609), a DittoFS SMB server now appears in **Windows
Explorer → Network**. **Double-clicking** the discovered computer, however, fails
to authenticate **AD/domain users**: dfs cannot validate an NTLM logon against
Active Directory because it has no **netlogon secure channel** (machine account +
NTLM pass-through) to the DC.

Mounting by the Kerberos name (`\\vm2.cubbit.local\ditto`) works today — this
issue is specifically the **NTLM ↔ AD** path that Explorer double-click depends on.

## Why double-click uses NTLM (not Kerberos)

When you double-click the discovered entry, Windows connects by the **discovery
name / IP**, which is **not** a Kerberos SPN. With no SPN to request a ticket
for, the SMB client falls back to **NTLM**. So the discovered-browse experience
inherently requires the server to accept NTLM.

## Root cause

dfs's SMB NTLM path validates only against the **local UserStore**. For an AD
user it finds no local account and returns `STATUS_LOGON_FAILURE`. There is no
secure channel to the DC to pass the NTLM response through for validation
(`NetrLogonSamLogon` / schannel).

Observed on the AD testbed (double-click as `CUBBIT\alice`):

```
NTLM AUTHENTICATE message parsed username=alice domain=CUBBIT workstation=VM1 isAnonymous=false
User not found in UserStore username=alice error=user not found
Sent SMB2 response command=SESSION_SETUP status=STATUS_LOGON_FAILURE
```

Authorization is **not** the problem: `CUBBIT\alice` is granted on the share
(resolved to her AD SID via the LDAP idp), and the same user authenticates and
mounts fine over **Kerberos**.

## What works today

- **Kerberos** SMB via the SPN FQDN (`\\vm2.cubbit.local\ditto`): keytab-validated,
  PAC SID → LDAP → share grant (#1528). Terminal `net use` and Explorer address-bar
  both succeed.

## Proposed work

Implement NTLM pass-through authentication for domain users:

1. Provision/join a **machine account** for the DittoFS server (secret already has
   partial scaffolding — see `cmd/dfs/commands/netlogon.go`, the `machine_sid`
   setting, `wireNetlogonReload`).
2. Establish a **netlogon secure channel** to the DC and validate the client's
[...]

Issue #1325, opened 2026-06-22
AD: hot-reload NETLOGON machine credential (no restart)


marmos91:
## Context
Follow-up to #1314 (SMB NTLM passthrough via NETLOGON / MS-NRPC). Part of the AD umbrella #1231.

The first NETLOGON cut applies machine-account credential changes **on restart** (same posture as the kerberos keytab today). #1311 already added `NotifyIdentityProviderConfigChange`, which LDAP uses to hot-reload live.

## Ask
Hot-reload the NETLOGON machine credential / DC binding without a restart:
- On an identity-provider config change for the machine-account sub-block, tear down and rebuild the cached `SecureChannel` atomically (no in-flight logon disruption).
- Mirror the LDAP hot-reload path wired in #1311.

## Notes
- Pure operational nicety; no protocol change.
- Must be safe under concurrent logons (the secure channel is mutex-guarded + sequence-chained).

marmos91:
Blocked by #1314. The base NETLOGON/MS-NRPC implementation does not exist yet — there is no `netlogon` package, `MachineCredentialProvider` interface, or `SecureChannel` in the tree, and #1314 is still open. This issue extends that foundation ("the first NETLOGON cut"), so it cannot start until #1314 lands. Marking the dependency so the ordering is explicit.

marmos91:
Fixed by #1369 (merged to develop 2026-06-23): NETLOGON machine credential now hot-reloads without a server restart.

PR #1639, opened 2026-07-10
feat(netlogon): dfsctl netlogon status + rotate (#1631)

### LINKED ISSUE #1631: SMB/NETLOGON: dfsctl netlogon status + rotate for the running server
## Summary

Follow-up to #1629 (NTLM↔AD pass-through). That PR added a self-contained,
config-scoped `dfs netlogon test` probe. This tracks the richer **operator surface
against the running server**: `dfsctl netlogon status` and `dfsctl netlogon rotate`.

## Motivation

Today the machine-account join/rotation is invisible at runtime: the online-join
fires lazily on the first domain logon and the password rotates on a timer, with
no way to inspect state or force a rotation without reading the server log.

## Proposed commands

- `dfsctl netlogon status` — report: enabled, provider (offline/online-join),
  account name, DC/realm/NetBIOS domain, joined?, secure-channel health, last/next
  rotation.
- `dfsctl netlogon rotate` — force a machine-password rotation now.

## Why it's non-trivial (wiring notes from the #1629 investigation)

There is currently **no path from the control-plane API/Runtime to the
`*netlogon.Authenticator`** — it's a local var in `runStart`, injected into the SMB
adapter (`SetNetlogonAuthenticator`) and held in a private field with no getter.
Implementing this needs:

1. **New introspection surface** on `internal/auth/netlogon`: getters for provider
   type / account / joined-state / channel-connected / last-rotation on
   `Authenticator` / `onlineProvider` / `RotationManager` (none exist today).
2. **Expose the authenticator to the Runtime**, e.g. `rt.SetAdapterProvider("netlogon", handle)`
   in `SetNetlogonAuthenticator` (mirrors the NFS client provider pattern).
3. **API handler + router route** under `RequireAdmin()` (mirror `/identity-providers`),
   plus `apiclient` methods + DTOs.
4. **A correct rotate entrypoint**: `Authenticator.RotatePassword` alone only sets the
   password on the DC — it does NOT persist/switch the secret. The complete rotation
   is the unexported `onlineProvider.rotate` (PasswordSet2 → update in-memory →
   persist). A `rotate` endpoint must call that (e.g. a new `RotationManager.RotateNow()`),
   or it will desync the persisted secret from the DC.

## Acceptance

`dfsctl netlogon status` shows live machine-account/secure-channel state, and
`dfsctl netlogon rotate` performs a correct (persist+switch) rotation on the
running server.

## Related
- #1629 — NTLM↔AD pass-through (the feature this manages)

marmos91:
## Summary

Closes #1631. Follow-up to #1629. Adds the operator surface for the NETLOGON machine account against the **running** server: `dfsctl netlogon status` and `dfsctl netlogon rotate`. Previously the machine-account join/rotation was invisible at runtime (online-join fires lazily on the first domain logon; the password rotates on a timer) with no way to inspect state or force a rotation without reading the server log.

## Commands

```
$ dfsctl netlogon status          # enabled, provider (offline/online-join), account,
                                  # realm/NetBIOS/DC, joined?, channel health, last/next rotation
$ dfsctl netlogon rotate          # force a correct (PasswordSet2 + persist) machine-password rotation
```

`status` is **non-intrusive** — it reads cached provider state and never calls `Credential()`, so it cannot trigger the lazy AD join. `rotate` is valid only for the online-join provider (which owns the password lifecycle); the offline/static provider returns a clear error (HTTP 409).

## Wiring (full stack)

- **netlogon introspection** (`internal/auth/netlogon`): `Authenticator.Connected()`, side-effect-free `onlineProvider.snapshot()` / `MutableProvider.snapshot()`, `RotationManager.nextRotation()`, and a `Controller` (status.go) bundling auth + provider + rotation.
- `online.go`: `rotate()` records `lastRotation` under a new `rotateMu` so a manual rotate can't race the scheduled timer into a DC/persist desync.
- **Runtime exposure**: `buildNetlogonAuthenticator` returns a `*Controller`; `start.go` registers it via `rt.SetNetlogonController(...)` → `SetAdapterProvider("netlogon", ...)` (mirrors the NFS-client-provider pattern).
- **Admin API**: `GET /api/v1/netlogon/status`, `POST /api/v1/netlogon/rotate` under `RequireAdmin()` (mounted only when a controller is registered); apiclient DTOs; dfsctl commands. CLI docs regenerated.

## Rotate correctness

`Controller.Rotate` routes through `onlineProvider.rotate` — the COMPLETE rotation (PasswordSet2 on the DC → switch the in-memory credential → persist) — never the bare `Authenticator.RotatePassword` (which would set the DC password without persisting, desyncing the stored secret). Per code review, the rotation now also runs under a **detached context** (`context.WithoutCancel` + a 2-minute timeout, mirroring the scheduler) so a client-side cancellation (dropped connection / the apiclient 30s deadline) can't abort it between the DC set and the persist and leave the stored secret out of sync.

## Tests

`status_test.go` (offline/online status, the no-join guarantee, complete-rotation persist+switch, next-rotation schedule) and `handlers/netlogon_test.go` (nil-controller 404, offline status, offline rotate → 409). `go build ./...`, `go vet ./...`, and the netlogon / api-handler / dfsctl / apiclient test suites all green.

## Validation

Live-validated on a real Windows AD DC (cubbit.local, online-join provider): `dfsctl netlogon status` showed `provider=online-join`, `Joined=no` **before** any logon (proving the non-intrusive read), and `dfsctl netlogon rotate` performed a real online-join (`DITTOFS-OJ$` provisioned + persisted) followed by `NetrServerPasswordSet2` — HTTP 200, `Last rotation` stamped, repeatable across restarts.

### DIFF
--- a/cmd/dfs/commands/netlogon.go
+++ b/cmd/dfs/commands/netlogon.go
@@ -45,6 +45,11 @@ func (m *machineSecretStore) SetMachineSecret(ctx context.Context, secret string
 // It also returns a *netlogon.RotationManager when the online-join provider is
 // active (nil otherwise); the caller starts it and stops it on shutdown.
 //
+// The third return is a *netlogon.Controller (nil when passthrough is disabled):
+// the runtime handle the caller registers with the Runtime so the admin API can
+// report `netlogon status` and force `netlogon rotate` against this running
+// authenticator (#1631).
+//
 // The offline path is backed by a netlogon.MutableProvider so the machine
 // credential / DC binding can be hot-reloaded over the API without a restart
 // (#1325). The online-join path (#1323) owns its own credential lifecycle via
@@ -58,9 +63,9 @@ func (m *machineSecretStore) SetMachineSecret(ctx context.Context, secret string
 // ReloadCredential/Close methods for the #1325 hot-reload and shutdown. The
 // disabled path returns a nil *Authenticator, so callers can test `auth == nil`
 // directly.
-func buildNetlogonAuthenticator(k config.KerberosConfig, secret netlogon.SecretStore) (*netlogon.Authenticator, *netlogon.RotationManager) {
+func buildNetlogonAuthenticator(k config.KerberosConfig, secret netlogon.SecretStore) (*netlogon.Authenticator, *netlogon.RotationManager, *netlogon.Controller) {
 	if !k.MachineAccount.Enabled {
-		return nil, nil
+		return nil, nil, nil
 	}
 	ma := k.MachineAccount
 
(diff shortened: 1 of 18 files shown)


marmos91:
Addressed the Copilot review comment: `netlogonStatusToDTO` no longer hard-codes `enabled: true` — it's now derived from a non-empty account name, so a #1325 hot-reload that clears the offline credential correctly surfaces as `enabled: false` (commit 1aa17fac).

Issue #1573, opened 2026-07-06
Metadata create/commit not batched — per-file fsync wall (~168 ops/s)


marmos91:
## Problem

Pure metadata throughput (create/stat/delete of small files) is walled at **~168 ops/s** — measured on the #1466 bench (SCW POP2-HC-32C-64G, develop `0b2361b8`, NFSv3 single-thread). Competitors on the same box: rclone **1384 ops/s**, ganesha **1154**, kernel-nfs **588**.

Root cause: each file CREATE and each COMMIT is a **separate synchronous badger commit** (`SyncWrites=true`, ~15 ms fsync on this disk). Two fsyncs per file → ~64–170 files/s regardless of parallelism.

## Why #1416 did not fix this

#1416 (batched metadata commits) is CLOSED/COMPLETED, subsumed by #1493 PR2 (#1511): `metadata.DefaultCommitBlock` folds a block record + all its chunk-index entries into **one fsync per block**. That removed the *rollup/drain* (data-path) fsync wall — not the *file-create* metadata wall. The CREATE + COMMIT metadata transactions are still per-file, unbatched. So the 168 ops/s number is untouched by #1416.

## Proposed fix: group-commit for concurrent metadata ops

A backend-agnostic store-contract capability that coalesces N logically-independent metadata transactions into **one durable commit** (one fsync), preserving durability (group-commit, not async write-back — the batched commit is durable before NFS COMMIT returns):

- **badger** — join concurrent `writeCh` entries into one value-log fsync (badger group-commits already; investigate why 16-way parallel CREATEs still serialize — is the serializer the per-fd append-log fsync rather than badger?).
- **postgres** — pipelined / multi-statement transaction.
- **sqlite** — WAL + single grouped transaction.
- **memory** — no-op.

Two call sites: the NFS CREATE path and the COMMIT/FlushPendingWrite path, so concurrent operations share a commit group.

## Open question (worth measuring first)

badger already group-commits concurrent writers, yet 16-way parallel CREATEs still measured ~64 files/s. Before building the contract, confirm the actual serializer: per-file append-log fsync (separate per-fd, not badger) vs. per-op metadata fsync. That measurement decides whether the lever is in `pkg/metadata` (group-commit) or `pkg/block` (append-log fsync coalescing).

## Verify

[...]

Issue #1571, opened 2026-07-06
SMB: mkdir returns EDEADLK (resource deadlock avoided) under concurrent small-file creation


marmos91:
Found during the #1466 benchmark run (SCW VM, DittoFS develop @ `0b2361b8`, Linux cifs client, SMB 2.1).

## Symptom

The bench small-files workload (create a dir, then create many files in it) fails on a **DittoFS SMB** mount:

```
workload small-files: create small files dir:
mkdir /mnt/bench/_dfsctl_bench/small_files: resource deadlock avoided
```

`resource deadlock avoided` = `EDEADLK` (errno 35) returned to the client for a **`mkdir`**. It aborts the whole bench (no partial result), which is why dittofs-badger-s3-smb2 produced no matrix JSON.

## Why it's a DittoFS SMB-server bug (not the client)

- **Samba** (real smbd) over the *same* cifs client + *same* workload **succeeds**. So it is not a cifs-client or mount-option issue.
- Only the DittoFS SMB adapter returns the status that the Linux client maps to `EDEADLK` (likely `STATUS_FILE_LOCK_CONFLICT` / a lease/oplock or metadata-lock self-conflict surfaced on a create/mkdir under concurrency).

## Repro

Mount a DittoFS share over SMB2.1 and run the bench small-files lane (or `mkdir` + rapid file creates in a fresh dir):
```
dfsctl bench run /mnt/smb --workload small-files --small-file-count 1000
```
NFSv3/NFSv4 on the same share do **not** hit this (their small-files lane runs).

## Suspected area

Directory-create path in the SMB adapter / metadata lock ordering during create — a self-deadlock-avoidance (`EDEADLK`) being returned instead of granting/queuing. Needs a look at how the SMB create/mkdir path acquires metadata locks vs. the NFS path (which is fine).

Scope: independent of the #1466 throughput work; blocks SMB small-file/metadata workloads.

marmos91:
## Investigation: hypothesized mechanism does not exist on current develop

Traced the full SMB create/mkdir path against current `origin/develop` (verified independently). The ticket's suspected chain — *SMB create → metadata lock manager → wait-for-graph → ErrDeadlock → EDEADLK* — is **broken at every link**:

[...]

PR #1647, opened 2026-07-10
perf(metadata): coalesce parent-inode mtime out of rmdir/rename txns (#1643)

### LINKED ISSUE #1643: perf(metadata): coalesce parent-inode mtime write out of rmdir/rename transactions (like #1573 did for create)
## Background

#1573 removed the parent-directory **mtime** read-modify-write from the *create* transaction (coalesced via `DirTimesTracker` / `recordDirTimes`), because that shared parent-inode write made concurrent same-parent creates serialize on one key and abort under BadgerDB SSI. #1571 (PR #1640) then removed the last shared key on the create path — the parent **link-count** RMW — by serializing it per-parent.

But the same parent-inode mtime write is **still done inline** in the two *namespace-removal/move* paths, so they retain the pre-#1573 contention profile:

- **rmdir** — `pkg/metadata/directory.go:259-263`: inside `withRelaxedTransaction`, `parent.Mtime = now` then `tx.PutFile(parent)`.
- **cross-parent rename (`Move`)** — `pkg/metadata/file_modify.go:894-905`: `srcDir.Mtime = now` + `PutFile(srcDir)`, and `dstDir.Mtime = now` + `PutFile(dstDir)`.

## Impact

Lower severity than #1571 — rmdir/rename are rarer than create, and PR #1640's per-parent link-count mutex already incidentally serializes the concurrent-rmdir and rmdir-vs-mkdir cases (so they no longer SSI-abort). But the parent-inode write itself is still an in-transaction RMW of a shared key, so:

- Concurrent same-parent rmdirs (and rmdir racing a create) touch the parent inode key inside the txn — extra serialization / SSI-retry pressure that #1573's coalescing was designed to avoid.
- The two paths are inconsistent with the create path, which reads parent mtime through the `DirTimesTracker` overlay and never writes the inode in the txn.

## Proposed fix

Route the rmdir and `Move` parent-mtime bump through the same `recordDirTimes` coalescing the create path already uses, and drop the inline `parent.Mtime = now; PutFile(parent)` from those transactions — so the removal/move transaction touches only disjoint namespace keys. Mirror the existing create-path pattern (`recordDirTimes` after commit + `mergeDirTimes` overlay on read). For `Move`, both source and destination parents get a coalesced bump.

Add contention regression tests analogous to `TestParentMtimeContention` / `TestConcurrentDirRenameNoParentLinkConflict` for the rmdir and rename paths.

## Notes

- Same relaxed-durability semantics as #1573 (a crash can lose a coalesced mtime bump; children/namespace remain the durable truth).
- Confirm the WCC/pre-op-attr contract for rmdir/rename still holds when the parent mtime comes from the overlay rather than a fresh in-txn write.

Follow-up to #1571 / #1640.

marmos91:
Closes #1643.

## What

`rmdir` and cross-parent `rename` (Move) still wrote the shared parent-inode
key **inside** their transaction (`parent.Mtime = now; tx.PutFile(parent)`),
retaining the pre-#1573 contention profile: concurrent same-parent namespace
ops touch that one key inside the txn and serialize / SSI-retry on it under
BadgerDB.

This routes both paths through the same `DirTimesTracker` coalescing that
`create` and `unlink` already use:

- The parent mtime/ctime/atime bump is recorded **after commit** via
  `recordDirTimes`, not written inside the txn.
- Reads still see the latest timestamp through the `mergeDirTimes` overlay.
- The removal/move transaction now touches only disjoint namespace keys
  (child map, link-count — the latter already serialized per-parent by #1571).

Sibling of #1571 (parent link-count) and #1573 (create parent-mtime).

## Scope / expected impact

This closes a **concurrency/scalability** gap on the rmdir/rename axis. It is
*not* expected to move the `dfsbench` `metadata` workload number, which is
create+small-write only and already benefits from #1573's create-path
coalescing — flagged here so the benchmark delta isn't misread as a regression.

## Tests

- `go test ./pkg/metadata/...` — 2665 pass.
- Existing `TestParentMtimeContention` (create) and
  `TestConcurrentDirRenameNoParentLinkConflict` (rename link-count) stay green;
  WCC pre/post-op attrs for rmdir/rename are unchanged (verified in the
  directory/rename suites).

### DIFF
--- a/pkg/metadata/directory.go
+++ b/pkg/metadata/directory.go
@@ -227,6 +227,7 @@ func (s *Service) RemoveDirectory(ctx *AuthContext, parentHandle FileHandle, nam
 	// rmdir decrements the parent's link-count key (dropping the ".." ref, below).
 	// Serialize that per-parent against concurrent mkdir/rmdir on the same parent
 	// so the shared counter key is never a BadgerDB SSI conflict source (#1571).
+	now := time.Now()
 	defer s.lockParentLink(parentHandle)()
 	txErr := withRelaxedTransaction(store, ctx.Context, func(tx Transaction) error {
 		// Re-read the parent inside the transaction so the pre-op snapshot and
@@ -234,6 +235,10 @@ func (s *Service) RemoveDirectory(ctx *AuthContext, parentHandle FileHandle, nam
 		if txParent, pErr := tx.GetFile(ctx.Context, parentHandle); pErr == nil && txParent != nil {
 			parent = txParent
 		}
+		// Overlay any pending coalesced bump so Before reflects the same mtime a
+		// concurrent GETATTR would see (and that the prior op's After returned),
+		// keeping WCC continuity — the tx read only sees durable state (#1573).
+		s.mergeDirTimes(parentHandle, &parent.FileAttr)
 		wcc.Before = CopyFileAttr(&parent.FileAttr)
 
 		// Remove directory entry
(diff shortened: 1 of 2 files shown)


marmos91:
Folded in a WCC-continuity fix caught in review: rmdir/rename now call `s.mergeDirTimes(...)` before capturing `wcc.Before`, matching the create/unlink siblings. Without it, the in-txn `tx.GetFile` sees only durable state and `Before` would under-report mtime whenever a coalesced parent bump is still pending (within the 2s flush window) — breaking WCC continuity in exactly the rapid same-dir mutation case this feature targets. `go test ./pkg/metadata/...` still green (2665).

Issue #1225, opened 2026-06-16
fs local cache: index + eviction refactor (the real 'faster than fs')


marmos91:
Part of #1221 (ROI: #3). NOT a new backend — JuiceFS/Alluxio both store bulk blobs as raw files; the win is the index+eviction, not mmap/LSM/Redis/io_uring.

**Scope** (modify `pkg/block/local/fs/`)
- Name CAS files by BLAKE3 hash (atomic create-if-absent + dedup; immutable → no invalidation).
- Move bookkeeping (size/atime/refcount) into a fast embedded index (Badger, or Pebble *as index only*).
- Replace directory-scan eviction with 2-random/LRU.
- `posix_fadvise(DONTNEED)` after serving + optional write-side `O_DIRECT` to kill double-buffering.
- Keep `local.LocalStore` contract (`local.go:28-177`) + LSL-08 invariant (eviction must not consult FileBlockStore — `fs/eviction.go`).

**Verify** `BlockStoreConformance` + `BlockStoreAppendConformance`; before/after `dfsbench` numbers to substantiate 'faster'.

marmos91:
## Design pass — and the issue's premise no longer matches the tree

Design only; no code, no branch. Read `pkg/block/local/`, `pkg/block/local/fs/` and
`pkg/block/journal/` at `6d2afd6d` (develop) before writing this.

**The headline: this issue was written against the pre-journal architecture and three of its
four bullets describe code that no longer exists — two because they were already built, one
because it was deliberately deleted.** Checked before designing anything, per the
"test the issue's premise" rule:

| Issue says | Tree at `6d2afd6d` |
|---|---|
| `local.go:28-177` | `pkg/block/local/local.go` is **141 lines** |
| `fs/eviction.go` | **does not exist** |
| "LSL-08 invariant" | **0 matches repo-wide** (`rg LSL-08`, 2643 files) |
| eviction "must not consult FileBlockStore" | no eviction code in `fs/` to consult anything |
| `BlockStoreAppendConformance` | **does not exist** (only `BlockStoreConformance`, `RemoteBlockStoreConformance`) |
| scope = `pkg/block/local/fs/` | that package is now a **321-line adapter** holding no naming, no bookkeeping, no eviction |

Since the Wall-A switchover (#1692 / #1714 / #1716), `fs.FSStore` embeds `*journal.Store` and
does nothing but bridge `string`↔`journal.FileID`. All the state this issue wants to refactor
[...]

PR #1648, opened 2026-07-10
perf(block): verify-once cache for warm random reads (~3x, skip per-read blake3)

marmos91:
## Problem

Profiling the warm 4 KiB random-read hot path (`engine.Store.ReadAt` → `ReadPayloadAt` → `fillFromCASManifest`) showed **~518 µs per read**, dominated by two halves of one root cause — the engine materializes and BLAKE3-verifies an entire ~1 MiB FastCDC chunk to serve 4 KiB:

1. **BLAKE3 whole-chunk verify — ~116 µs/read**, and its parallel worker-pool fan-out drives **69% of per-read allocations** and most of the goroutine-sync churn. Added by #1644 as the warm-read integrity gate.
2. Full-chunk pread + 1 MiB buffer alloc to serve 4 KiB (256× amplification) — the remaining cost, addressed separately.

This matches the competitor benchmark shape: dittofs rand-read-4k on **large** files (4010 IOPS) is worse than **medium** (10981) precisely because larger files have larger chunks → more per-read hashing.

## Fix

CAS chunks are content-addressed and **immutable** — a chunk that BLAKE3-verifies once stays valid for the identical bytes, so re-hashing it on every read is pure waste. This adds a bounded per-store **verified-chunk set** (`verified_chunks.go`): the read path verifies each chunk once, records the hash, and skips re-hashing on subsequent reads. Two-generation eviction bounds memory (~a few MiB/store) with O(1) ops.

Integrity is preserved where it matters: corrupt bytes still fail the **first** verify and route to the remote-verified heal path; an evicted chunk is simply re-verified next read; and this keeps a per-chunk first-touch gate that page-cache-serving competitors (JuiceFS/s3ql/s3fs) don't have at all. The documented tradeoff: post-verify on-disk bit-rot of a still-tracked chunk goes undetected until eviction — but after first read the chunk is in the OS page cache, so later reads come from RAM, not the platter.

## Measured (engine microbenchmark, 16 MiB payload, 4 KiB random reads, warm local CAS)

| | before | after |
|---|--:|--:|
| ns/op | 437 µs | **148 µs** (~3.0×) |
| allocs/op | 183 | **32** (−83%) |
| MB/s | 9.4 | **27.7** |

`BenchmarkWarmRandRead_FS` is included as a regression guard.

## Tests

- New `verified_chunks_test.go` — two-generation eviction + verify-once/reject-corrupt semantics.
- `go test ./pkg/block/...` — 1228 pass; `-race` on `fs`+`engine` clean; corrupt-chunk integrity tests (`readpayload_verify_test.go`) still green.

## Follow-up

The remaining ~1 MiB read+alloc per 4 KiB (cost #2) needs a ranged sub-chunk read on a verified hit — separate PR. Refs #1225.

### DIFF
--- a/pkg/block/engine/warm_randread_bench_test.go
+++ b/pkg/block/engine/warm_randread_bench_test.go
@@ -1,0 +1,162 @@
+package engine_test
+
+// Warm local random-read benchmark + regression guard for the verified-chunk
+// cache (skip re-hashing immutable CAS chunks on every 4 KiB read).
+//
+// Unlike BenchmarkRandRead_Phase12 (memory local store, no disk, no blake3),
+// this drives the REAL fs local store: write a multi-MiB payload, DrainRollups
+// so the bytes live as rolled-up CAS chunks (WARM local read, no remote), then
+// do 4 KiB reads at random offsets through engine.Store.ReadAt — the exact
+// production warm path (ReadPayloadAt -> fillFromCASManifest -> ReadChunk ->
+// blake3 verify). Remote is nil so scheduleReadahead early-returns.
+//
+// Run:
+//   go test -run=XXX -bench=BenchmarkWarmRandRead_FS -benchtime=3s \
+//     -cpuprofile=/tmp/rr.prof -memprofile=/tmp/rr.mem ./pkg/block/engine/
+
+import (
+	"context"
+	"math/rand"
+	"testing"
+
+	"github.com/marmos91/dittofs/internal/logger"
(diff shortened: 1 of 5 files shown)


marmos91:
Added nil-receiver guards to the verified-set (defensive, per review Q5): a bare `&FSStore{}` literal built outside `newFSStore` now falls back to full per-read verification instead of panicking. Reviewer confirmed no production path bypasses `newFSStore`; this just hardens direct test literals.

Issue #1173, opened 2026-06-15
perf(metadata): decide IF a server-side read cache is needed — gated on scaling + k8s scale-out study


marmos91:
Follow-up to #1169 (measurement) / #1172 (bench tooling + decision doc). **Status: decision deferred — do NOT build yet.** Query-reduction (the cheap, unconditional win) is split to #1176.

## Where we landed

Measurement showed postgres metadata reads cost ~160µs/op loopback and ~44% of server CPU under a *forced cold* NFS crawl (#1172). But:

- **160µs/op is acceptable** in absolute terms. The 44% figure came from a `drop_caches`-forced crawl (worst case); real NFS/SMB clients absorb most repeat reads in their own attr/dentry cache, so the server only pays on cold / large-working-set crawls.
- A cache is the *bigger hammer* and carries an invalidation tax. Before committing to it we need to understand whether the problem is real **at scale**, not just in a synthetic cold crawl.

## Open questions to answer before deciding

1. **Concurrency scaling.** How does postgres metadata-read latency/throughput hold up as concurrent NFS/SMB clients grow? Sweep `--workers` in `dfsbench metadata` and a real multi-client mount. Does p99 stay flat or collapse?
2. **Dataset scaling.** Does per-op cost stay ~constant as the tree grows to 1M+ / 10M+ entries? (Point lookups should, but confirm — index depth, table bloat, autovacuum.)
3. **k8s scale-out.** This is the decisive one. DittoFS on k8s = **multiple `dfs` replicas sharing one postgres**. That **breaks the single-node single-writer assumption** that made cache invalidation tractable:
   - A per-pod in-process cache needs **cross-pod invalidation** — pod A mutates an inode, pods B/C must drop their cached copy. That needs postgres `LISTEN/NOTIFY`, a pub/sub bus, or a short global TTL (accepting bounded staleness). All materially harder/riskier than the single-writer design.
   - So in the world we actually care about (scale-out), the cache is **not** obviously worth it, and #1176's query-reduction may be the better lever.

## Decision rule

Build the cache only if the scaling/scale-out study shows postgres metadata reads remain a **real** bottleneck after #1176, AND a cross-pod-safe invalidation design is justified. Otherwise close as won't-do and rely on client caching + query-reduction.

## If built (design constraints carried over)

[...]

PR #1654, opened 2026-07-11
perf(metadata): read cache for badger GetFileForRead (~5700x on hit)

marmos91:
Follow-up to the warm-read fast path (#1648/#1651). A **VM server-pprof** under warm random-read (badger, S3) pinpointed the real remaining bottleneck: **`GetFileForRead` = ~47% of server CPU, almost entirely JSON-decoding the File blob**, and badger read transactions as the **top mutex contender**. The File blob inlines the full `[]ChunkRef` (~1024 refs for a 1 GiB file, ~800 µs to decode), so a random-read fleet on one file re-decoded it on **every 4 KiB read**. This is why the engine microbench showed 20× but the rig only ~2× — the microbench used the in-memory metadata store and never touched badger.

## Rejected alternatives (benchmarked first, per request)

| attempt | result |
|---|--|
| **easyjson** codec for File/FileChunk | **~0%** (835 µs vs 789 µs) — the cost is the per-ChunkRef `ContentHash.UnmarshalJSON` (hex decode), which both paths dispatch to identically, not reflection. Reverted. |
| **blocks-skipping decode** on the read path | ~23% + 99% fewer allocs, but json still *scans* the 270 KB blob, and READ_PLUS/SEEK legitimately need `Blocks` (can't blanket-skip without growing the store interface). |

## This change

A lock-free (`sync.Map`) read cache of decoded `File` records on the badger store, keyed by fileID, so `GetFileForRead` skips **both** the View transaction and the JSON decode on a hot file.

**Correctness** (single-node badger is single-writer):
- **Invalidate after commit**: for every file a transaction mutated (`PutFile`/`DeleteFile`/`SetLinkCount`, tracked via `tx.dirtyFiles`), `withTransaction` deletes the entry and bumps a generation *after* the badger commit.
- **Generation-guarded populate**: a read snapshots the generation before the backing read and only caches if it's unchanged, so a write racing the read can never leave a stale value cached (a lost populate is a miss, never a stale hit).
- Returned Files are **copies** (caller-mutation-safe).
- Bounded (soft cap + best-effort prune).

Scoped to badger; the #1173 deferral was specifically about multi-replica **postgres** scale-out invalidation, which this does not touch.

## Measured

- `BenchmarkGetFileForRead_Cached`: **~800 µs (uncached decode) → 139 ns/op** on a cache hit (~5700×), and it also skips the contended badger View txn.
- `TestReadCache_NoStaleReadAfterMutation` guards invalidation + caller-copy-safety.
- `go test -race ./pkg/metadata/store/badger/` — **522 pass** (incl. storetest conformance).

## Follow-up

End-to-end VM validation of the rig random-read number (recommended) — this should close most of the remaining gap to the page-cache re-exports. The readahead lock-free fix (#1653) is a separate, minor concurrency improvement (pprof showed it wasn't the bottleneck).

### DIFF
--- a/pkg/metadata/clone.go
+++ b/pkg/metadata/clone.go
@@ -1,0 +1,55 @@
+package metadata
+
+import (
+	"github.com/marmos91/dittofs/pkg/block"
+	"github.com/marmos91/dittofs/pkg/metadata/acl"
+)
+
+// CloneBlocks returns a deep copy of a []block.ChunkRef. ChunkRef is a value
+// type (Hash is a [32]byte array, Offset and Size are scalars), so a flat
+// element-wise copy fully detaches the clone. Returns nil for a nil/empty
+// input so a round-trip preserves the omitempty wire form.
+func CloneBlocks(in []block.ChunkRef) []block.ChunkRef {
+	if len(in) == 0 {
+		return nil
+	}
+	out := make([]block.ChunkRef, len(in))
+	copy(out, in)
+	return out
+}
+
+// CloneACL returns a deep copy of a *acl.ACL, including fresh copies of its ACEs
+// (DACL) and SACL slices. acl.ACE is a flat value type (uint32 fields + a string
+// Who), so an element-wise slice copy fully detaches the clone. Returns nil for
+// a nil input so a round-trip preserves the "no ACL set" semantics.
+func CloneACL(in *acl.ACL) *acl.ACL {
+	if in == nil {
+		return nil
+	}
+	out := *in
+	if in.ACEs != nil {
+		out.ACEs = make([]acl.ACE, len(in.ACEs))
+		copy(out.ACEs, in.ACEs)
+	}
+	if in.SACL != nil {
+		out.SACL = make([]acl.ACE, len(in.SACL))
+		copy(out.SACL, in.SACL)
+	}
+	return &out
+}
+
+// CloneEAs returns a deep copy of a FileAttr.EAs map, copying each value slice
+// so neither side can mutate the other's EA bytes in place. Returns nil for a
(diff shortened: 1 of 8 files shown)


marmos91:
Addressed Copilot's review (thanks):

**2. Shallow copy shares Blocks/ACL/EAs (files.go:58) — fixed** (df28846). This was a real regression of badger's no-alias invariant: before the cache, every `GetFileForRead` JSON-decoded a fresh `File`, so callers never aliased stored state. Now the cache return deep-copies the reference-bearing fields via `metadata.{CloneBlocks,CloneACL,CloneEAs}` (promoted from the memory store's helpers, which now delegate). Added a Blocks-aliasing assertion to `TestReadCache_NoStaleReadAfterMutation`. The clones are cheap next to the ~800 µs decode the cache skips.

**1. invalidate() bumps gen before delete (read_cache.go:71) — working as intended, not changed.** The suggested reorder (delete then bump) would *introduce* a stale-populate: a reader that snapshotted the old generation and is mid-`store()` still sees the old gen in the window after the delete, so it re-inserts the pre-write value *after* the delete, and the subsequent bump can't remove it. Bumping gen first rejects that reader's `store()` the instant the generation moves. The get() path is intentionally ungated: `invalidate()` runs inside `withTransaction`, so a write's `PutFile` returns only once the entry is gone — any read ordered-after the write misses; a read merely concurrent with the uncommitted write has no ordering guarantee. Expanded the doc comment to capture this.

**3. dirtyFiles dedup / invalidateMany (transaction.go:161) — harmless, left as-is.** Redundant `LoadAndDelete` on an already-removed key is a no-op and extra gen bumps are monotonic (a racing reader is rejected regardless of count). Not worth the extra set-alloc on the commit path.

PR #1657, opened 2026-07-11
perf(metadata): read cache for badger GetShareOptions

marmos91:
## Problem

A fresh server-pprof of warm NFS random-read (dittofs-s3, badger metadata) shows every NFS read runs a permission check that badger-`View`s and JSON-decodes the share-config record:

- `CheckReadPermissionFile -> GetShareOptions -> decodeShareData` = **17.4% of server CPU**
- The badger read-transaction (`(*DB).View`/`readTs`) is also the **top mutex contender at 14.5%**

Share config changes only on a rare admin reconfigure, so caching the decoded `*metadata.ShareOptions` per share eliminates this entirely. Mirrors the just-shipped `GetFileForRead` read cache.

## Change

- New `shareReadCache` (`sync.Map` + `atomic.Uint64` generation), keyed by share name. Shares are few and rarely written, so no prune/cap logic is needed.
- `GetShareOptions` fast path: return a **deep copy** of the cached options on a hit; on miss, snapshot the generation, do the existing `db.View` + `decodeShareData`, generation-guarded `store`, return a deep copy.
- Deep copy clones all three string slices (`AllowedClients`, `DeniedClients`, `AllowedAuthMethods`) **and** the `IdentityMapping` pointee (its `*uint32` / `*uint32` / `*string`), so no caller or concurrent reader ever aliases the cached object.

## Security: invalidation completeness

A stale cache entry is a **wrong permission decision**, so every share-record write site invalidates after the write commits. All 8 sites (grep-verified: `Set(keyShare` / `Delete(keyShare`):

Direct `s.db.Update` sites (`shares.go`, invalidate immediately after `Update` returns nil):
1. `shares.go` CreateShare
2. `shares.go` UpdateShareOptions
3. `shares.go` DeleteShare
4. `shares.go` CreateRootDirectory (covers `createNewRoot`'s `Set(keyShare)`)

Transaction sites (`transaction.go` `*badgerTransaction` methods -> `dirtyShares` -> post-commit loop in `withTransaction`, mirroring `dirtyFiles`):
5. `transaction.go` CreateShare
6. `transaction.go` UpdateShareOptions
7. `transaction.go` DeleteShare
8. `transaction.go` CreateRootDirectory `Set(keyShare)`

`dirtyShares` is captured inside `db.Update` (reset per attempt like `pendingDelta`) so a conflict-retry cannot leak a phantom invalidation; invalidation runs only after commit succeeds. Generation guard: `invalidate` bumps `gen` before deleting, and `store` writes only if `gen` is unchanged since the snapshot taken before the backing read, so a racing populate can never re-insert a stale value.

## Verification

- `go build ./...`, `go vet ./pkg/metadata/...`: clean
- `gofmt -s -l` on changed files: empty
- `golangci-lint run ./pkg/metadata/...`: 0 issues
- `go test -race ./pkg/metadata/store/badger/...`: 525 passed (incl. store conformance)
- New tests: no-stale-permission-after-update (ReadOnly + slice), invalidate-on-delete, deep-copy safety (mutate + append), tx-path (`dirtyShares`) invalidation
- `BenchmarkGetShareOptions`: **cached 83 ns/op (2 allocs)** vs **uncached 6072 ns/op (26 allocs)** — ~73x

### DIFF
--- a/pkg/metadata/store/badger/read_cache.go
+++ b/pkg/metadata/store/badger/read_cache.go
@@ -59,6 +59,15 @@ func (c *fileReadCache) store(key string, file *metadata.File, genAtRead uint64)
 			c.pruneToHalf()
 		}
 	}
+	// The guard above and the Swap are not atomic: a write could commit and
+	// invalidate() (bump gen + delete) in between, leaving our now-stale entry
+	// live. Re-check the generation; if it moved, drop our entry — but only if a
+	// newer reader hasn't already replaced it (CompareAndDelete on our pointer).
+	if c.gen.Load() != genAtRead {
+		if c.m.CompareAndDelete(key, file) {
+			c.n.Add(-1)
+		}
+	}
 }
 
 // invalidate drops key and advances the generation so any in-flight populate for
--- a/pkg/metadata/store/badger/share_read_cache.go
+++ b/pkg/metadata/store/badger/share_read_cache.go
@@ -1,0 +1,113 @@
+package badger
+
+import (
+	"slices"
+	"sync"
+	"sync/atomic"
+
+	"github.com/marmos91/dittofs/pkg/metadata"
+)
+
+// shareReadCache is a lock-free cache of decoded ShareOptions keyed by share
+// name, so the permission hot path (CheckReadPermissionFile → GetShareOptions)
+// skips BOTH the BadgerDB read transaction AND the JSON decode of the share
+// record on every NFS read. Server pprof of warm random-read showed
+// GetShareOptions → decodeShareData at 17.4% of server CPU, and the badger
+// read-transaction the top mutex contender at 14.5% — every read re-decodes a
+// config record that changes only on a rare admin reconfigure.
+//
+// Shares are FEW and rarely written, so unlike fileReadCache this needs no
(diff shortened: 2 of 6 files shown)


marmos91:
Addressed Copilot's review + rebased onto develop (now includes #1654):

**1. store() populate race (share_read_cache.go) — real, FIXED.** Good catch — the generation guard and the map `Store` are not atomic, so a write committing + `invalidate()` (bump gen, delete) between them could leave a now-stale entry live. For a permission cache that's a wrong-decision bug. Fixed with a post-publish re-check: after `Store`, re-read the generation and `CompareAndDelete` our own entry if a write raced (compare-on-pointer so a newer reader's entry is never evicted). **The just-merged file read cache (#1654) had the identical race** — fixed it too in this branch (second commit), with the count bookkeeping preserved.

**2. dirtyShares dedup — harmless, left as-is.** A duplicate name causes a redundant `invalidate()`: the `Delete` is a no-op on an absent key and extra generation bumps are monotonic (a racing populate is rejected regardless of count). Not worth a per-commit set allocation on the write path.

Rebased cleanly (kept both `dirtyFiles` and `dirtyShares` invalidation loops in `withTransaction`). 527 badger tests pass `-race`, lint 0.

Issue #432, opened 2026-04-23
SMB: durable handles V2 — reconnect, lease coordination, app-instance


marmos91:
## Context

Durable handle V2 is partially implemented but fails on 33 \`smb2.durable-v2-open.*\` subtests and 1 \`smb2.durable-v2-delay.*\` subtest (KNOWN_FAILURES.md "Durable Handles V2 (Fix Candidate)"). Broad areas: DH2 reopen, disconnected-handle preservation/purge rules, app-instance ID, DH2 lock interactions, persistent handles.

## Failing tests

**Create + reopen:**
- \`create-blob\` (DH2Q context validation)
- \`open-oplock\`, \`open-lease\`
- \`reopen1\`, \`reopen1a\`, \`reopen1a-lease\`, \`reopen2\`, \`reopen2b\`, \`reopen2-lease\`, \`reopen2-lease-v2\`
- \`durable-v2-setinfo\`

**Locks interacting with DH2:**
- \`lock-oplock\`, \`lock-lease\`, \`lock-noW-lease\`

**Lease state + DH2:**
- \`stat-and-lease\`, \`nonstat-and-lease\`, \`statRH-and-lease\`
- \`two-same-lease\`, \`two-different-lease\`

**Disconnected-handle preservation rules (MS-SMB2 §3.3.4.18):**
- \`keep-disconnected-rh-with-stat-open\`
- \`keep-disconnected-rh-with-rh-open\`
- \`keep-disconnected-rh-with-rwh-open\`
- \`keep-disconnected-rwh-with-stat-open\`

**Disconnected-handle purge rules:**
- \`purge-disconnected-rwh-with-rwh-open\`
- \`purge-disconnected-rwh-with-rh-open\`
- \`purge-disconnected-rh-with-share-none-open\`
- \`purge-disconnected-rh-with-write\`
- \`purge-disconnected-rh-with-rename\`

**App instance + reconnect delay:**
- \`app-instance\`
- \`smb2.durable-v2-delay.durable_v2_reconnect_delay\`

**Persistent handles (separate feature, not implemented):**
- \`persistent-open-oplock\`, \`persistent-open-lease\`

## References

- MS-SMB2 §2.2.13.2.11 (DH2Q), §2.2.13.2.12 (DH2C), §3.3.5.9.10–12 (V2 durable), §3.3.4.18 (disconnected handle preservation/purge state machine)
[...]

PR #1664, opened 2026-07-13
fix(metadata): SQL backends corrupt extreme timestamps + drop durable ClientGUID (#1663)

marmos91:
## Problem

develop CI is red (#1663). The smbtorture SMB conformance suite fails **5 tests on the sqlite and postgres metadata backends** that pass on memory/badger. These are pre-existing backend gaps newly *exposed* by #1659 (which added sqlite/postgres to the postsubmit matrix), not regressions.

| Failing test | Root cause |
|---|---|
| `smb2.timestamps.time_t_0` | epoch (1970) collided with the zero-time sentinel |
| `smb2.timestamps.time_t_10000000000` (yr 2286) | `UnixNano()` overflows int64 past ~2262 |
| `smb2.timestamps.time_t_15032385535` (yr 2446) | same overflow |
| `smb2.durable-open.reopen1a-lease` | durable reconnect returned `OK` instead of `OBJECT_NAME_NOT_FOUND` |
| `smb2.durable-v2-open.reopen1a-lease` | same |

## Cluster A — timestamps

sqlite/postgres stored file timestamps as `int64` unix-nanoseconds. That window is only ~1678–2262, and the epoch maps to `0`, colliding with the "unset" zero-time sentinel. memory/badger store `time.Time` (JSON, full range) and pass.

Fix: store **Windows FILETIME** (signed 100ns ticks since 1601) in the existing `INTEGER`/`BIGINT` columns — spans years 1601–~30828, epoch is a distinct non-zero value. This is exactly SMB FILETIME granularity and matches every value the storetest suite asserts (all 100ns-multiples). Migrations \`000037\` (pg) / \`000004\` (sqlite) re-encode existing rows; the arithmetic is provably identical to the Go encoder.

## Cluster B — durable reconnect

The \`durable_handles\` table never got a \`client_guid\` column when ClientGUID (#432) landed, so a scanned \`ClientGUID\` was always zero. That short-circuits \`leaseReconnectClientGUIDMismatch\`, wrongly allowing a lease-backed durable reconnect from a *different* client. Add the column (migrations \`000038\` pg / \`000005\` sqlite); no handler change needed.

## Tests

\`storetest\` (the shared conformance suite, the deterministic CI replacement for these smbtorture asserts) gains:
- extreme-timestamp cases (epoch / 2286 / 2446) in \`testTimestampPrecision\`
- a \`ClientGUID\` round-trip assertion in the durable-handle fixture

Both fail before the fix and pass after. Full conformance verified locally on all four backends: memory (297), badger (527), sqlite (240), postgres (all).

### DIFF
--- a/pkg/metadata/store/postgres/durable_handles.go
+++ b/pkg/metadata/store/postgres/durable_handles.go
@@ -28,12 +28,13 @@ const durableHandleColumns = `
 	username, session_key_hash, is_v2, created_at, disconnected_at,
 	timeout_ms, server_start_time,
 	delete_pending, parent_handle, file_name, is_directory,
-	position_info, original_file_id, requested_alloc_size, is_persistent
+	position_info, original_file_id, requested_alloc_size, is_persistent,
+	client_guid
 `
 
 func scanDurableHandle(row pgx.Row) (*lock.PersistedDurableHandle, error) {
 	var h lock.PersistedDurableHandle
-	var fileIDBytes, leaseKeyBytes, createGuidBytes, appInstanceIdBytes, sessionKeyHashBytes, originalFileIDBytes []byte
+	var fileIDBytes, leaseKeyBytes, createGuidBytes, appInstanceIdBytes, sessionKeyHashBytes, originalFileIDBytes, clientGuidBytes []byte
 	var positionInfoSigned, requestedAllocSizeSigned int64
 	var leaseEpochSigned int32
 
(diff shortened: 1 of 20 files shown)


marmos91:
Duplicate of #1662, which already implements this exact fix (FILETIME timestamps + durable ClientGUID) and is green/mergeable. Arrived here via an independent reimplementation before spotting #1662 — the identical design independently validates #1662's approach. Closing in favor of #1662.

Issue #1667, opened 2026-07-13
dittofs+sqlite: end_fsync EIO on buffered 4×1g flush (bench seq-write-buffered)


marmos91:
Found while running the `dfsbench` matrix (2026-07-13). The `dittofs-sqlite-s3` backend fails the `seq-write-buffered` workload:

```
FAIL dittofs-sqlite-s3-nfs3: run fio (seq-write-buffered): exit status 4
fio: end_fsync failed for file /mnt/bench-client/seq-write-buffered.N.0
```

`seq-write-buffered` writes 4 jobs × 1 GiB buffered, then a single `fsync` at close. On the **badger** metadata backend this passes; on **sqlite** the close-time COMMIT returns EIO — the deferred-fsync fast path (#1584) not sustaining a 4×1g buffered flush + single COMMIT (append-log / write-path backpressure surfaced as a non-success).

The workload is fair (badger handles it). This is a sqlite metadata write-path limitation worth fixing or documenting as a known sqlite-backend constraint.

Repro: `dfsbench run --remote --systems dittofs-sqlite-s3-nfs3 --sizes large` (see internal/dfsbench/CLAUDE.md for the harness).

marmos91:
Fixed in #1675 (merged to develop): sqlite DSN now sets `synchronous=NORMAL` under WAL, dropping the per-commit fsync that made the write-heavy buffered flush fsync-bound. Verify on the next bench matrix rerun.

marmos91:
Reopening — the 2026-07-14 dfsbench re-run (develop @ 38679f1c, *after* #1675 `synchronous=NORMAL`) reproduces this as a **hard wedge**, not just an EIO:

- Cell: `dittofs-sqlite-s3-nfs3__seq-write-buffered__large` (4×1 GiB buffered write).
- `dfs` server RSS ballooned to **3.1 GB** and stopped draining.
- `fio` workers stuck in uninterruptible **`D` state for 12+ min**; `df` on the NFS mount hung (mount unresponsive).
- No self-recovery — had to kill `dfs` (EIO the mount) to unblock, then lazy-unmount.

`synchronous=NORMAL` (#1675) did **not** fix it. The buffered large-write flush path (append-log backpressure) still deadlocks/leaks under the 4×1g close-time flush. This wedged the whole benchmark matrix (sqlite runs before the competitors), so sqlite is currently unbenchmarkable on buffered-large.

[...]

PR #1683, opened 2026-07-14
docs(bench): 2026-07-14 warm re-run results

marmos91:
Refreshes `docs/BENCHMARKS.md` with the 2026-07-14 warm benchmark re-run — the first clean full matrix since the harness fixes (#1666/#1670/#1676).

## What changed
- Warm large + medium throughput/IOPS/latency tables for 6 systems (DittoFS-S3, ZeroFS, JuiceFS, rclone, s3fs, local-disk).
- Coverage caveats documented: warm-only (cold pass dropped to dodge a `drain-uploads` stall), s3ql not run (SCW eventual-consistency flake), sqlite/postgres excluded (buffered-large wedge, #1667).
- Analysis rewritten around the fair comparison (DittoFS vs ZeroFS).

## Headline
- **Metadata is the one deficit, unmoved**: 259/239 ops/s ≈ the 239/219 baseline. Last of the durable field, loses even to ZeroFS (354). fsync-bound (low CPU, high ctxsw). Lever = metadata group-commit (#1573, unlanded).
- **DittoFS now leads the durable native-S3 cohort** vs ZeroFS: rand-read 3.6×, rand-write 13.7×, seq-read 57× — the warm-read fast path (#1648/#1651) + lock-free readahead (#1653) paid off.
- Warm-read gap to the FUSE re-exports is per-read server RPC/metadata CPU (kernel page-cache free ride), not cold-S3 or the block store.
- seq-write 148 MB/s is the secondary gap; likely shares the append-log fsync serializer with metadata create.

### DIFF
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -57,7 +57,7 @@ ops/s). **Bold** = DittoFS. Best *real competitor* per row is marked ✦
 | mixed-rw (IOPS) | **1409** | 1983 ✦ | 125 | 5531 | 4455 | 41634 |
 
 ‡ JuiceFS large rand-read (88 IOPS) is an outlier — its 1 GiB working set thrashes
-its local cache on this box; its medium rand-read (42 380) is representative.
+its local cache on this box; its medium rand-read (42 380 IOPS) is representative.
 
 ## Random & metadata — medium files (1 MiB), warm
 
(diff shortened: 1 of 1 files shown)


marmos91:
Addressed in 2c8a441:
- **Missing unit (line 60):** medium rand-read now reads `42 380 IOPS`.
- **bench.yaml (line 146):** the repro snippet now generates `bench.yaml` inline via heredoc, so it runs verbatim.
- **JuiceFS rand-read latency (line 94):** not an error — the latency table header is *large files, warm*, so that row is JuiceFS's **large**-file rand-read, i.e. the 88 IOPS `‡` thrashing outlier already flagged above, **not** the medium 42 380 IOPS figure the comment compared against. At 88 IOPS aggregate every read misses the thrashed 1 GiB cache and pays an S3 round trip, so a ~1 s p50 is consistent. Added a one-line pointer under the table so readers don't repeat the cross-table comparison; numbers unchanged.

Issue #1416, opened 2026-06-25
Batched metadata commits: abstract store-contract capability to remove the per-file fsync wall


marmos91:
> **Priority: tier-1 — highest-leverage small-file fix.** The per-file synchronous metadata fsync is the shared root of BOTH the write wall (~64 files/s) and the rollup→S3 drain wall (~12 chunks/s), so this single capability moves small-file performance the most. Core small-file lever alongside #1414 (object packing).

---

## Problem

Small-file throughput is walled by **per-file synchronous metadata commits**, measured on the demo VM (#1411):
- **Write ~64 files/s** — CREATE (1 badger commit) + COMMIT/FlushPendingWrite (1 badger commit), each a synchronous fsync (`SyncWrites=true`, ~15 ms on this disk).
- **Rollup→S3 drain ~12 chunks/s** — FileBlock.Put + persistRollupBlocks + SetRollupOffset = 3 synchronous metadata commits per rolled-up file, serializing the 8 rollup workers on the single store.

DittoFS's data path is already JuiceFS-like (local append-log → background rollup → **async** S3; NFS COMMIT does not block on S3). The gap vs JuiceFS is purely this per-file synchronous metadata fsync. JuiceFS avoids it with a fast external metadata engine (Redis/TiKV) that batches and doesn't fsync per op. The data path being ours means we can close this in-tree.

## Proposal: an abstract batching capability on the metadata store contract

Make commit-batching a **first-class, backend-agnostic capability of the metadata store interface** — NOT a badger-specific hack. Each backend implements it with its own native primitive:
- **badger** — its `doWrites` group-commit / `WriteBatch` (coalesce many txns into one value-log fsync).
- **postgres** — batched multi-statement transaction / pipelined commits.
- **sqlite** — WAL + a single grouped transaction (or `PRAGMA synchronous` tier per durability mode).
- **memory** — no-op (already in-RAM).

### Shape (to be designed)
- A contract method/mode that lets callers coalesce N logically-independent operations into **one durable commit** (one fsync), e.g. a `BatchCommit`/group-commit handle, or a store-level "join the current commit group" primitive. The two concrete call sites that need it:
[...]

PR #1684, opened 2026-07-14
perf(metadata): store-agnostic group-commit for durable flush (#1573)

### LINKED ISSUE #1573: Metadata create/commit not batched — per-file fsync wall (~168 ops/s)

marmos91:
Closes #1573 (reframed — see below).

## The reframe
#1573's stated premise (per-create fsync wall) is already fixed: `relaxed_durability: true` is the shipped default, so a plain CREATE does zero hot-path fsyncs. The *live* metadata wall (dfsbench: **259 ops/s, last of the durable field, loses to ZeroFS 354**) is elsewhere:

- The dfsbench `metadata` workload is create→write-4k→**close** per file (`numjobs=4`). Each close → NFS COMMIT → `flushPendingWrite` → a **raw, uncoalesced `db.Sync()`**. 4 threads never coalesce (259/4 ≈ 65 ops/s/thread ≈ one 15 ms fsync each).
- The block side already solved this with a group-commit `syncLeader` (#1416/PR3). The metadata side never got it.

## The fix
- **`pkg/metadata/commit_leader.go`** — store-agnostic coalescing coordinator (sibling of `syncLeader`). N concurrent durability barriers collapse onto one backend sync per drain batch. **Strict group-commit**: the fsync completes before the caller returns → durability unchanged.
- **`SyncDurable(ctx) error`** on the `Store` contract (no new capability type): badger `db.Sync()`; sqlite `wal_checkpoint(PASSIVE)`; postgres synchronous WAL flush when relaxed; memory no-op.
- **`flushPendingWrite`** stages via the relaxed (no-fsync) txn, then the coalesced barrier.

## Durability
Identical to before — coalesce fsyncs, don't defer them. Reviewed for the postgres relaxed-staging gap (fixed: synchronous WAL flush via `txid_current()`) and the sqlite PASSIVE-under-readers gap (safe: single-connection pool; documented).

## Gates
`go build`, `go vet`, `gofmt -s` clean; **2691 tests pass** incl. storetest conformance for every backend; `-race` clean.

## Pending before un-drafting
- [ ] **Measured before/after on a Linux VM** (macOS fsync isn't a full barrier — the wall only shows on Linux). Expect ~2–4× (bounded by 4-thread coalescing) — past ZeroFS, not to JuiceFS's faster engine.
- [ ] Postgres `SyncDurable` is CI-verified only (needs a live postgres); logically sound, untested locally.

Follow-up (separate): seq-write 148 MB/s (bandwidth-bound, needs a profiling pass) and sqlite buffered-large wedge (#1667).

### DIFF
--- a/pkg/metadata/commit_leader.go
+++ b/pkg/metadata/commit_leader.go
@@ -1,0 +1,108 @@
+package metadata
+
+import (
+	"context"
+	"sync"
+	"sync/atomic"
+)
+
+// commitLeader coalesces concurrent metadata durability barriers into as few
+// backend Sync calls as possible. It is the store-agnostic sibling of the block
+// store's syncLeader (pkg/block/local/fs/sync_leader.go).
+//
+// The one structural difference from syncLeader: syncLeader batches per-fd fsync
+// closures (N distinct fds → N back-to-back fsyncs the FS journal coalesces),
+// whereas a metadata store exposes ONE barrier that makes ALL committed-but-
+// unsynced writes durable at once (Store.SyncDurable → badger db.Sync / sqlite
+// WAL checkpoint). So the leader calls a single shared barrier once per drain
+// batch and hands its result to every waiter in that batch.
+//
+// Why (#1573): every NFS COMMIT / stable SMB flush runs flushPendingWrite, which
+// stages the write and then forces durability. Before this, each concurrent
+// commit issued its own uncoalesced store.SyncDurable, so N files committing at
+// once paid N independent fsyncs — the metadata wall. Routing them through one
+// leader collapses a concurrent burst onto one barrier.
+//
+// Durability contract: Sync blocks until a barrier that ran AFTER this caller
+// enqueued has completed, and returns that barrier's result. The barrier always
+// completes before Sync returns — this is STRICT group-commit, not async write-
+// back. NFS COMMIT depends on the synchronous ack. (Correctness of the "ran
+// after enqueue" guarantee: a caller that enqueues while a leader is mid-barrier
+// lands in the NEXT batch, which the leader always drains — it only exits after
+// locking and finding the queue empty, and the enqueue-under-lock happens-before
+// that check. So its write, committed before enqueue, is covered by a fresh
+// barrier.)
+//
+// Adaptive bypass: a lone caller arriving at an idle leader runs the barrier
+// inline (one drain pass) with no goroutine hand-off, mirroring syncLeader.
+type commitLeader struct {
+	// barrier makes every write committed so far on the owning store durable.
+	barrier func() error
+
+	mu      sync.Mutex
+	pending []chan error
+	running bool
+
+	// drainPasses counts non-empty barrier batches. A burst that coalesces onto
+	// one leader run shows far fewer passes than Sync calls; read by the
+	// coalescing tests.
+	drainPasses atomic.Int64
+}
+
+func newCommitLeader(barrier func() error) *commitLeader {
+	return &commitLeader{barrier: barrier}
+}
+
+// Sync makes this caller's already-staged write durable, coalescing with any
+// concurrent callers, and returns the shared barrier's result. On ctx
+// cancellation the caller observes ctx.Err() but the barrier still runs
+// (durability is never abandoned); the done channel is buffered so the leader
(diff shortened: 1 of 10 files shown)


marmos91:
## ⚠️ Measured A/B: this fix is INEFFECTIVE — do not merge as-is

Same-VM metadata A/B (SCW fr-par-1, dittofs-s3-nfs3, `--workloads metadata --runtime 30`, binaries swapped in place, sha-verified distinct):

| | baseline (develop `2e24035`) | treatment (this branch `70c37cf`) |
|---|--:|--:|
| metadata large | 256 ops/s | 250 |
| metadata medium | 234 | 224 |

**Zero improvement** (treatment marginally lower, within run noise). The uncoalesced per-COMMIT `db.Sync()` is **not** the metadata wall — the diagnosis behind this PR was wrong. Since fsync coalescing doesn't move the number, the sync→async escalation wouldn't help either.

Clues for the real wall (from the run): CPU ~10%, but **DiskWr ~20 MB/s at 250 ops/s ≈ 80 KB written per 4 KB file** — heavy per-op write amplification + something synchronous per file that isn't the metadata sync. Likely candidates to profile: badger's own value-log commit, or the **block-side per-file append-log/logblob fsync** (the metadata workload is create+**write-4k**+close, so it hits the block path too — `syncLeader` may not be coalescing across the 4 staggered threads).

**Next step: profile before any further code** — enable `/debug/pprof`, capture CPU + block + mutex profiles under the metadata workload, find the true serializer. Holding this branch as a documented negative result; the code itself is correct + durability-preserving (reviewed) and could still help higher-thread-count workloads, but it does not close the benchmarked gap and should not merge on those grounds.

marmos91:
Closing: measured ineffective (see A/B above — 256→250 ops/s, zero improvement). The diagnosis was wrong; the metadata wall is not the uncoalesced COMMIT db.Sync. Reverting the branch and re-approaching via profiling. Code preserved in git history if the coordinator is ever wanted for a high-thread-count workload.

PR #1686, opened 2026-07-14
ci(health): retry conformance run once before opening develop-red

marmos91:
## Problem

`develop-red` P0 issues (e.g. #1685) are opened for **transient conformance flakes**, not real breakage. In the run that triggered #1685, the exact same heavy smbtorture tests (`maxfid`, `lease`, `oplock`, `notify`, `replay`) hit `smbtorture infrastructure failure (exit code 124)` — a per-suite **timeout** — across *every* profile (badger-fs, sqlite, memory, memory-fs, postgres). The blamed commit (#1682, snapshot decryptability) doesn't touch SMB. Classic slow-runner flake, but it still paged.

## Fix

Retry-before-alarm in `ci-health.yml`:

- **Attempt 1 red** → re-run the failed jobs once (`gh run rerun --failed`), open **no** issue.
- **Attempt ≥ 2 red** → open/append the `develop-red` P0 as before.

A genuine break fails attempt 1 **and** 2, so it still pages. A flake passes on attempt 2 and self-heals silently. Bounded to a single retry (attempt ≥ 2 never re-runs). No test-grading changes — real assertion failures are unaffected.

Adds `actions: write` (needed for the rerun) and gates all alarm steps on `SHOULD_ALARM` (attempt > 1); attempt 1 hits the new `SHOULD_RETRY` step. Close-on-green path is untouched.

### DIFF
--- a/.github/workflows/ci-health.yml
+++ b/.github/workflows/ci-health.yml
@@ -34,6 +34,8 @@ permissions:
   contents: read
   issues: write
   pull-requests: write
+  # Needed for the retry-before-alarm step to re-run the failed conformance jobs.
+  actions: write
 
 jobs:
   # ---------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


marmos91:
Both addressed in 06238060:
- `actions: write` moved off the workflow level and scoped to the `alarm` job (job-level perms re-list contents/issues/actions); `warn-on-pr` no longer gets it.
- `gh run rerun` failure now promotes to the alarm path via `SHOULD_ALARM=true` in `$GITHUB_ENV` instead of failing the step and swallowing the signal.

Issue #1701, opened 2026-07-15
Flaky: smb2.dirlease.unlink_*_initial_and_close double-break (lease_break_info.count==2) in smbtorture conformance


marmos91:
Splitting a persistent smbtorture flake out of the auto-closing CI-health alarm (#1689), so it survives develop going green.

## Symptom
`smbtorture / memory` (and `badger-fs`) intermittently red on develop:

```
smb2.dirlease.unlink_same_initial_and_close       FAIL
smb2.dirlease.unlink_different_initial_and_close   FAIL
(../../source4/torture/smb2/lease.c:7653): wrong value for lease_break_info.count got 0x2 - should be 0x1
```

Rotating set — one sibling of the `unlink_{same,different}_{initial,set}_and_close` family trips per run, different backend job each time. Recent hits: `1090a549` (memory), `3ceb4f99` (badger-fs).

## This is NOT
- **Not a regression from #1688** (`1090a549`): that commit is a pure local block-store refactor (delete legacy per-chunk CAS) and cannot touch SMB dir-lease dispatch.
- **Not #1658**: that flake is `NT_STATUS_SHARING_VIOLATION` (share-mode gate false-positive). This one is `lease_break_info.count got 0x2` — the client observes **two** wire LEASE_BREAK notifications for one logical lease. Overlapping test names, distinct root cause.

## Root cause (narrowed, from prior fixes)
This is the residual of the #981 double-break race (see #951, #743). smbtorture reuses the `DLEASE` lease-key constant across every dirlease subtest on a single ClientGUID. An orphaned ack-to-None UnifiedLock record left under the same handleKey by an unclean disconnect can coexist with the live record under the same lease key. #981 added per-lease-key dedup in `pkg/metadata/lock/directory.go` (the guard is present and explicitly names this flake at `directory.go:165`), but it does not fully close the intermittent cross-test-state window — `count==2` still surfaces ~occasionally.

## Not reproducible locally
Upstream smbtorture test (`lease.c:7653`); needs CI-runner timing/contention. Consistent with the sibling flake #1658 which also only reproduces in CI.

## Next step
[...]

Issue #1658, opened 2026-07-11
Flaky: intermittent SMB rename/dirlease SHARING_VIOLATION in smbtorture conformance


marmos91:
Splitting the persistent flake out of the auto-closed CI-health alarm #1652 (which reopens/closes with develop's red/green state).

## Symptom
`smbtorture / memory` (and occasionally `memory-fs`/`badger-fs`) intermittently red on develop with `NT_STATUS_SHARING_VIOLATION` where `NT_STATUS_OK` is expected. Rotating set: `smb2.rename.rename_dir_bench`, `smb2.rename.close-full-information`, `smb2.dirlease.unlink_*`. ~10-20% of runs.

## Root cause (narrowed)
- **Not a regression** — same pair failed pre-suspect-commit (`9af0beda`).
- **Not metadata** — no metadata error maps to `SharingViolation` (conflicts → `FileLockConflict`).
- Must be a **share-mode gate false-positive** on a stale/foreign handle: `close-full-information` opens all handles full-share (incl. `FILE_SHARE_DELETE`) with no lease, so neither `checkShareDeleteConflict` (src) nor `checkParentDirRenameConflict` (dst-parent) can match its own handles. Returns from `set_info.go:705` / `:957`.
- **Not reproducible locally** — dozens of full-suite + isolated runs under CPU-throttle and host-stress: zero flakes. Needs the CI-runner timing/contention.

## Diagnostic in place
#1655 (merged, `f4ad808d`) logs the offending handle at the moment a rename gate trips: `SET_INFO rename conflict holder gate=... holderPath=... holderSession=... holderShareAccess=... holderDesiredAccess=... holderIsDurable=... holderDeletePending=...`. The harness captures server logs to the CI artifact (`results/smbtorture-*/dittofs.log`).

## Next step
On the next develop CI flake, download the run artifact, grep `dittofs.log` for `SET_INFO rename conflict holder` **at the failing test's timestamp** (the line fires on legitimate conflicts too — must correlate with an actual failure). The holder's session/tree/path names the leaked/stale handle → targeted verified fix.

Refs #1652, #1655.

marmos91:
### Status (2026-07-13): still in wait-for-evidence mode — no recurrence

[...]

PR #1705, opened 2026-07-15
ci(health): silence develop-red email + issue alarm, keep PR warning

marmos91:
## Why
The CI-health alarm opened a P0 `develop-red` issue **and** emailed on every red conformance run on develop. Flaky smbtorture subtests — the `smb2.dirlease` double-break (#1701), the rename `SHARING_VIOLATION` (#1658) — trip it constantly, so the mailbox and issue tracker fill with alerts for non-regressions. The signal is buried; the alarm is net-negative.

## What
- **Remove all alerting**: email send, `develop-red` issue open/update/close, label creation, the failure-collection step, and the now-dead `.github/scripts/ci_health_email.py` (−231 LOC, referenced only here).
- **Keep the flake self-heal**: the attempt-1 auto-rerun of failed jobs stays, so a transient flake goes green on attempt 2 with no human touch. (Renamed the job `alarm` → `retry`; dropped its `issues: write`.)
- **Rewire `warn-on-pr`**: it used to key off the existence of the `develop-red` issue. It now computes red directly from the four monitored conformance workflows' latest completed develop-run conclusions, and clears the PR warning on green.

Net **−379 LOC**. `contents: read` at workflow level; each job re-declares only what it needs (`actions: write` for retry, `actions: read` + `pull-requests: write` for the PR warning).

## Kept on purpose (say the word to drop)
The auto-rerun job. It's not an alert — it's what makes flaky conformance self-heal to green, which is what clears the PR warning without anyone acting. If you'd rather have zero machinery beyond the PR comment, I'll delete it too.

## Note
With the issue machinery gone, existing open `develop-red` issues (#1700, and #1689) no longer auto-close — I'll close them manually on merge. The persistent flake itself is tracked in #1701.


marmos91:
Addressed both Copilot points in `10c13101`:
1. **issues:write** — `warn-on-pr` posts via the `/issues/{pr}/comments` endpoint, so it now carries `issues: write` (the scope the original workflow relied on) instead of `pull-requests: write`.
2. **Never-block** — every `gh` call in the warn-only job is now best-effort (`|| echo …`), so a GitHub API/CLI hiccup can't turn this warn-only check red.

Issue #1692, opened 2026-07-15
perf: maximize NFS/SMB read/write — three-wall plan (segstore redesign + adapter wins)


marmos91:
Tracking issue for the NFS/SMB end-user read/write performance effort. Phase 0 (#1688) and Phase 1 (#1690) shipped; this is the follow-on.

**Design docs (on develop):**
- `.planning/perf/2026-07-15-nfs-smb-three-walls-PLAN.md`
- `.planning/perf/2026-07-15-wall-a-segstore-BLUEPRINT.md`

## Three walls

**Wall A — data plane: `journal` block store redesign (chunk-at-carve, write-once).** _(package renamed `segstore` → `journal`.)_ Collapse append-log + logblob into one append-only log-structured write-back cache over S3; FastCDC+BLAKE3+dedup move to carve; sidecar `.idx`; self-describing versioned records; lazy pressure-gated eviction; standalone benchmarkable library. 12-PR sequence.

- [x] PR0 — standalone bench harness + skeleton (#1697)
- [x] PR1 — segment format + append primitive (#1702)
- [x] PR2 — interval index + DataExtents (#1707)
- [x] PR3 — recovery (tail-scan + rebuild) (#1708)
- [x] PR4 — carve (FastCDC+BLAKE3+dedup + in-place synced flip) (#1709)
- [x] PR5 — pressure-gated eviction (#1710)
- [x] PR6 — GC/repack (#1711)
- [ ] PR7 — engine wiring (riskiest; carve-seam rework)
- [ ] PR8 — FSStore adapter rewrite + delete legacy
- [ ] PR9 — share wiring + docs refresh + maintainer ADR
- [ ] PR10 — drain→discard→re-hydrate migration
- [ ] PR11 — O_DIRECT + io_uring (deferred)

**Wall C — RPC/adapter/auth per-op quick wins (DONE except C8).**

- [x] C1 — global activity-lock → per-client atomic (#1695)
- [x] C2 — cache per-share store + decode handle once (#1698)
- [x] C3 — DRC lock shard (#1693)
- [x] C4 — SMB per-request buffer pool (#1694)
- [x] C5 — SMB auth-context cache (#1696)
- [x] C6 — NFS auth-cache TTL + user/group-change invalidation (correctness) (#1703)
- [x] C7 — hot-path micro-locks + per-op alloc cleanup (#1704)
- [ ] C8 — rsize/wsize buffer-pool ceiling (measure-gated)

[...]

Issue #953, opened 2026-06-01
blockstore: rollup leaves stale overlapping FileBlock rows on in-place overwrite → cold read can return stale bytes


marmos91:
## Summary
Surfaced while attempting #1 H5 (index-seek ReadPayloadAt, PR #952, deferred). The CAS manifest (`ListFileBlocks`) accumulates **stale, overlapping FileBlock rows from multiple rollup generations** after an in-place overwrite of a rolled-up extent, because the rollup persister never deletes superseded rows when the FastCDC chunk boundaries differ between the original write and the overwrite.

## Mechanism
On overwrite with different FastCDC boundaries, the rollup writes new FileBlock rows (keyed `payloadID/offset`) but does not delete the superseded rows from the prior generation. `ListFileBlocks` then returns overlapping rows, e.g. for a 2 MiB file:
- `off=0 size=1072729`        → [0, 1072729)
- `off=1050553 size=1046599`  → [1050553, 2097152)  (overlaps)
- `off=1072729 size=1024423`  → [1072729, 2097152)  (overlaps)

An offset-sorted "fill uncovered bytes" walk over these mixes chunks from different write generations → wrong/stale bytes.

## Why it's latent on develop today
The warm read path (`replayLogIntoDest`) does a full sequential scan of the on-disk append log, which still physically contains the overwrite records until physical compaction (threshold `maxLogBytes/divisor`, rarely hit in tests). It covers the whole window from the log and **never consults the manifest**, so the stale rows are masked.

## The exposed risk (cold read)
The cold-read path `readLocalByHash` / `findRowCoveringOffset` consults the manifest directly. After physical compaction (or local-state eviction/restart) drops the log records, a cold read of an in-place-overwritten rolled-up extent could resolve to a stale-generation chunk and **return stale bytes** — a data-correctness bug.

## Repro lead
`pkg/controlplane/runtime` `TestSnapshotByteVerify_Matrix` + `TestRestoreRollback_QuiescesRollupBeforeReset` fail the moment the warm-read masking is removed (PR #952 branch): `post-mutate read returns OLD bytes (want 0xf2 got 0x67)`. A direct cold-read repro (overwrite a rolled-up extent with differing FastCDC boundaries → force compaction/eviction → cold read) is the next step to confirm exploitability on develop.

## Fix direction
[...]

PR #1716, opened 2026-07-15
perf(block): journal switchover — journal is the live local store (Wall A S2)

marmos91:
Stage 2 of the Wall-A block-store redesign (#1692): the atomic switch that makes `pkg/block/journal` the live local block store and deletes all legacy local-store code. Builds on S1 (#1714).

## What changed

**Narrowed `local.LocalStore` (Option A — kept the interface + name).** Dropped every hash-keyed method (`Get`/`Put`/`Has`/`Head`/`Delete(hash)`/`Walk`/`AppendWrite`/`ListUnsynced`) and the `ChunkLifecycleHooks`/`localBlobReader` capabilities. Journal-native surface: `WriteAt`/`ReadAt`/`Hydrate`/`Commit`/`Delete`/`Truncate`/`FileSize`/`DataExtents`/`ListFiles`/`Carve`/`Evict`/`SetCarveTargets`/`SetEvictionEnabled`/`UnsyncedBytes`/`SetRetentionPolicy`/`Start`/`Close`/`Stats`/`Healthcheck`. Two impls: journal-backed `*fs.FSStore` (thin adapter embedding `*journal.Store`) and reimplemented `*memory.MemoryStore`.

**Engine rewired onto the journal:** carve dispatcher (`carve_dispatch.go`) drives `journal.Carve` via the `engineDeduper`/`engineBlockSink` targets; the BlockSink writes the per-file FileChunk manifest rows inside the `DefaultCommitBlock` txn. Writes → `WriteAt`; Flush → `Commit`; DrainRollups → `Carve(Force)`; ResetLocalState → `ListFiles`+`Delete`. Deleted `carver.go` (pending-carve set, claimBatch, chunk-at-ingest machinery) and the rollup/ChunkLifecycleHooks wiring.

**Read path:** warm reads resolve via the journal interval index; cold/evicted reads fetch the covering block from the remote, **verify BLAKE3 before use**, and `Hydrate(id, offset, data)` back into the journal, then re-read.

**Deleted:** the 8 `pkg/block/local/fs/{rollup,compaction,appendlog,appendwrite,logindex,readpayload,eviction,recovery}.go` files + the rest of the legacy fs package (chunkstore/logshard/interval_tree/…), `pkg/block/local/logblob/`, the hash-keyed `fs`/`logblob` conformance suites, `pkg/block/engine/carver.go`, and the tests for the retired machinery. `engine/legacy_migration.go` (cas→blocks migration) preserved; only its hash-keyed local-first read branches dropped.

## Deliberate behavior changes / caveats (please read)

- **Warm-read self-heal is DROPPED, not merely ported.** The journal read path reads raw segment bytes by offset and has no per-read content hash to detect warm-tier bit-rot against. Integrity now rests on (a) the journal's recovery-time CRC / torn-tail truncation and (b) **mandatory BLAKE3 verification of every remote-fetched chunk before Hydrate**. Follow-up if per-offset warm self-heal is wanted.
- **#953 superseded-row reap not carried over → stale cold-read after in-place overwrite.** The rollup persister that reaped superseded FileChunk rows is gone and carve does not reap. **Warm reads are always correct** (journal's versioned interval index), but a file **evicted after an in-place overwrite** may resolve an older overlapping FileChunk row on cold-read. Recommend a follow-up: have `engineBlockSink` reap superseded rows, or add row generations.
- **Risk #2 (cross-file local dedup dropped):** carve skips re-uploading an already-durable chunk but each file keeps its own FileChunk rows, so two files sharing content cold-fetch independently (more S3 GETs on heavily-deduped datasets). Locked D1 behavior.
- **Risk #6 (GC stays on the engine):** FileChunk reap + refcount stay on `engine/gc_block.go`; the carve loop never triggers `journal.GC`. The obsolete hash-namespace local GC no-ops (`LocalStore()` returns nil → the local-tier sweep iterates an empty share list) — zero controlplane churn.
- **D2 consequence:** carve writes per-`(file,offset)` FileChunk rows but not `FileAttr.Blocks`/ObjectID (rollup persister gone); that projection is now a metadata-layer concern validated by e2e.

## Verification (local, green)

- `go build ./...` and `GOOS=windows go build ./...`
- `gofmt -s -l` clean, `go vet ./...` clean, `golangci-lint run` **0 issues**
- `go test -race ./pkg/block/... ./pkg/metadata/...` — **27 packages, 3548 tests pass, no data races** — including BlockStoreConformance through the real Syncer with a journal-backed local store, the carve-seam acceptance test, and durability #1274.

Full e2e/NFS/SMB + smbtorture are CI's job.


marmos91:
## ⛔ Merge blocked — confirmed CRITICAL: silent cold-read data corruption (#953 reap regression)

Adversarial review confirmed the author-flagged #953 gap is a real, silent data-corruption regression. Marking draft until fixed.

**Root cause:** FileChunk manifest rows (`<payloadID>/<fileOffset>`) are now written *only* by carve (`engine/blocksink.go` → `DefaultCommitBlock`), which upserts rows for the re-carved dirty range but **never reaps rows superseded by a re-chunk**. The old path did (`origin/develop:engine/engine.go` `reapSupersededFileChunks`); it's gone with no replacement. Carve only covers `!synced && !cold` intervals, so a partial overwrite of an **evicted** chunk re-carves the dirty sub-range and orphans the cold remainder's manifest coverage.

**Two silent failure sequences** (file evicted, then in-place overwrite, then cold read):
1. **Shrink → zeros.** Overwrite re-chunks the first chunk smaller; the cold remainder is covered by no row → cold read returns **zero-fill, no error** (`journal ReadAt` cold path + `read_internal.go` ignoring the 2nd read's cold flag).
2. **Boundary shift → stale bytes.** A new chunk boundary leaves a pre-existing higher-offset row surviving; `GetFileChunkAtOffset` (badger objects.go:456) picks the **stale** row → fetches the OLD chunk → returns **overwritten-away bytes**.

**Why CI is green anyway:** the regression test `pkg/block/engine/cold_read_overwrite_test.go` (257 lines) was **deleted with no replacement**; local conformance/e2e never exercise the evict→remote→cold-read cross-generation cycle.

**Fix required before merge:** restore superseded-row reap in the carve/commit path (handling the partial-overwrite-of-cold-chunk coverage correctly, not a mechanical port), and re-add a cold-read-after-overwrite regression test covering BOTH sequences. In progress.

### Non-blocking follow-ups (track, don't block this PR)
- **MEDIUM — warm-read self-heal dropped.** Every *remote* fetch is BLAKE3-verified (no hole there), but at-rest warm-segment bit-rot is served with no per-read hash check until eviction+refetch. Bounded; acceptable follow-up.
- **Perf — carve uploads are now serial** (`carveRun` commits one block at a time; `uploadLimiter`/`ParallelUploads` #1407/#1432 now bound nothing on the carve path). Likely WAN throughput regression — needs a before/after benchmark + tracked issue.
- **Doc residue** — `readwrite.go` `WriteAt` still documents the deleted `fs/rollup.go` model; scrub.

PR #1719, opened 2026-07-16
test(nfs): harden READDIR verifier test against coarse Windows clock

marmos91:
## Root cause

The READDIR/READDIRPLUS cookie verifier is a pure function of the directory mtime:

```go
func directoryMtimeVerifier(mtime time.Time) uint64 { return uint64(mtime.UnixNano()) }
```

`TestReadDir_StaleVerifierBadCookie` and `TestReadDirPlus_StaleVerifierBadCookie` read a directory (saving the verifier), then modify it and re-read, asserting `NFS3ERR_BAD_COOKIE` because the verifier changed. They relied on the two rapid mutations landing on distinct mtimes.

On Windows the clock resolution is ~15ms, so two rapid directory mutations can share an **identical** mtime → identical verifier → no `BAD_COOKIE` → the assert fails. On Linux the ns clock always advances, so it's green. This surfaced as a flake on #1716's Windows CI, but is pre-existing latent fragility — that PR does not touch this code, and the rerun passed on identical code, confirming nondeterminism.

## Fix

Test/fixture layer only; production code untouched.

- Add a `SetMtime(handle, time)` helper to the handler test fixture (wraps `SetFileAttributes` with an explicit `Mtime`).
- In both stale-verifier tests, after the directory mutation, advance the directory mtime by a known delta (`dir.Mtime.Add(time.Second)`) so the verifier provably differs regardless of clock granularity. No `time.Sleep`.

The sibling `ZeroCookieIgnoresVerifierMismatch` tests use junk/zero verifiers on the bypass path and do not depend on an mtime change — left unchanged.

## Verification

- `go test ./internal/adapter/nfs/v3/handlers/...` green, run 5x for determinism
- `go build ./...`, `GOOS=windows go build ./...`
- `gofmt -s`, `go vet`, `golangci-lint run` on touched packages — clean

### DIFF
--- a/internal/adapter/nfs/v3/handlers/readdir_test.go
+++ b/internal/adapter/nfs/v3/handlers/readdir_test.go
@@ -2,6 +2,7 @@ package handlers_test
 
 import (
 	"testing"
+	"time"
 
 	"github.com/marmos91/dittofs/internal/adapter/nfs/types"
 	"github.com/marmos91/dittofs/internal/adapter/nfs/v3/handlers"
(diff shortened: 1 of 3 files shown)


marmos91:
Reviewed — correct, minimal, and the right layer.

**Root cause confirmed.** `directoryMtimeVerifier` is `uint64(mtime.UnixNano())` (readdir.go:17), and the handler reads `dirFile.Mtime` live at readdir.go:146. The two stale-verifier tests mutate the directory then assert the verifier changed → `BAD_COOKIE`. On a coarse clock (Windows ~15ms) the pre- and post-mutation mtimes can collide, the verifier is unchanged, and the assert fails. Linux's ns clock always advances, so it was green there. This is pre-existing latent fragility — it surfaced on #1716's Windows CI, but #1716 doesn't touch this package (empty diff for `internal/adapter/nfs/`, readdir.go byte-identical to develop), and the rerun passed on identical code.

**Why the fix is right:**
- `SetMtime` forces the dir mtime strictly forward (`dir.Mtime.Add(time.Second)`). Since mtime never decreases in these tests, +1s is provably greater → the verifier provably changes on any clock resolution. No `time.Sleep`, so it's deterministic, not just less-flaky.
- Both READDIR and READDIRPLUS are hardened (they share the mtime verifier). The two `*ZeroCookieIgnoresVerifierMismatch` siblings use junk/zero verifiers on the bypass path and don't depend on an mtime change — correctly left alone.
- Fixture/test layer only; no production `directoryMtimeVerifier` or dir-mtime path touched.

Determinism was checked 5× + full package green. LGTM once CI is green.

PR #1727, opened 2026-07-16
feat(journal): local-only snapshot restore via segment-pinning + re-materialize (#1718)

### LINKED ISSUE #1718: Local-only snapshot-restore under the journal: segment-pinning + version-aware index rebuild
## Context

Part of the journal switchover (#1716, tracker #1692). The journal replaces the legacy CAS/logblob local store.

## Gap

Local-only (remoteless) snapshot → overwrite → restore currently reads **zeros** under the journal store. On develop this worked incidentally: content-addressed CAS chunks survived an overwrite (new-hash chunks written, old chunks kept at GC-lifetime, refcounted by the manifest), so restoring an older manifest resolved the still-present local chunks — no remote needed.

The journal (D1) is deliberately **not** hash-keyed and uses write-once segments with whole-segment eviction. An overwritten region becomes dead bytes eligible for eviction/GC, and `RestoreSnapshot` → `ResetLocalState` deletes journal segments unconditionally, so a restored file has no intervals to resolve and zero-fills. The #953 reap (landed in the switchover) compounds this by removing the pre-overwrite manifest rows.

**Remote restore is unaffected** — the old blocks live on S3 and are re-seeded as cold intervals from the restored `File.Blocks` (shipped in the switchover). This issue is local-only.

## What's needed (a real journal feature, not a bounded fix)

1. **Snapshot segment-pinning** — a snapshot must pin the journal segments/records its manifest references so eviction/GC cannot reclaim them while the snapshot exists.
2. **Version-aware restore index rebuild** — `RestoreSnapshot` must rebuild the interval index to the snapshot's version (or via a hash→location map) so reads resolve the pinned old blocks locally instead of deleting them.

## Interim state

The switchover ships with the 5 local-only restore acceptance tests `t.Skip`-ed referencing this issue (`TestSnapshotByteVerify_Matrix`, `TestSnapshotByteVerify_EncryptedRemote_Matrix` local-only cases, `TestRestoreCrashRecovery_AcrossRealReopen`, `TestRestoreRollback_*`). Remote restore is covered by a new remote-backed test. Un-skip these as this lands.

marmos91:
Fixes #1718.

## Root cause
In a remote-less (local-only) share the journal is the ONLY durable copy of bytes. snapshot@V → overwrite → restore read ZEROS: `RestoreSnapshot`→`ResetLocalState` deleted every journal file, nothing re-seeded the V-view, `seedColdFromManifest` was `if remoteVerify`-guarded (skipped local-only), and GC/repack could physically drop the pre-overwrite records (a superseded ≤V record is dead at head, so a repack victim would drop it). Remote restore already worked (cold-seed from the manifest + remote hydrate).

## Approach: version-watermark + re-materialize
The journal is a global-LSN newest-wins log; records carry a durable, monotone `version`. A snapshot == a version watermark.

- **Pin** (`segment.go`/`store.go`/`reclaim.go`): each segment tracks `minVersion` (its lowest record version, set on first append + recovery scan + repack). A per-store `pinVersion` (DERIVED = max `JournalVersion` over live snapshots) makes `pickVictim`/`claimColdestEvictable` skip any segment with `minVersion <= pinVersion`. Whole-segment granularity, `repackSegment` unchanged. The predicate only ever makes GC MORE conservative.
- **RestoreToVersion(ctx, V)** (`store.go`): (1) ceiling replay — scan every on-disk record, rebuild each file's coverage as of V (data ≤V newest-wins, tombstones/truncates ≤V honored, >V ignored); (2) re-materialize — read the V-view bytes from their pinned source records and re-append them as fresh DIRTY records at the head (a tombstone first to bury the current head, then the V-view data; a file present at head but absent at V is tombstoned away). Fresh versions exceed everything, so `recover()` rebuilds V on reopen while the ≤V and (V,head] records stay for their snapshots' rollback. A plain rewind is NOT restart-safe and the >V records CANNOT be deleted (the pre-restore safety snapshot pins them).
- **Orchestration** (`runtime/snapshot.go`): capture `JournalVersion()` after the Step-0 drain; persist it via `MarkSnapshotReady`; raise the pin before a local-only snapshot is marked ready and recompute (lower) it after a delete commits and at share start from the durable snapshot list. Restore branches on `bs.HasRemoteStore()` — remote keeps `ResetLocalState`+`seedColdFromManifest`; local-only uses `RestoreToVersion(snap.JournalVersion)` + `DrainRollups` instead of `ResetLocalState`.
- **Passthroughs** (`engine/flush.go`): `RestoreToVersion`/`SetPinVersion`/`JournalVersion` via the existing SeedCold interface-assertion pattern (FSStore embeds `*journal.Store`, so they promote; the memory store no-ops).
- **Column** (`models/snapshot.go` + `store/snapshots.go` + `store/interface.go`): `Snapshot.JournalVersion uint64` (GORM AutoMigrate covers sqlite/postgres control-plane DBs); `MarkSnapshotReady` gains the `journalVersion` param.

## Invariants preserved
- Pin only ever makes GC more conservative (raise before ready, lower only after delete commits); reads are a single atomic load under the existing shard locks.
- Reap/refcount stays on `gc_block.go` — untouched. This change is journal repack (local byte relocation) only.
- Re-materialize is idempotent: FastCDC content-defined chunking reproduces the same hashes as the restored manifest, so the #953 reap finds no straddle and `FileAttr.Blocks` is not churned.
- Pin is applied ONLY to local-only shares — a remote share needs no pin (bytes also live remotely) and pinning would strand local disk.
- Crash-safety: existing restore-marker + `recoverInterruptedRestores` rollback path is unchanged; a mid-restore crash rolls back to the pinned, intact safety snapshot.
- No on-disk format change beyond the snapshot `JournalVersion` column; re-materialized records write `.idx` like any append.

## Tests un-skipped (all green, incl. `-race` and `-tags integration` badger)
- `TestSnapshotByteVerify_Matrix` — local-only in-memory rewind (overwrite/delete/create → restore byte-identical).
- `TestRestoreCrashRecovery_AcrossRealReopen` — re-materialize survives a real metadata-store reopen; rollback reconstructs the safety-snapshot state.
- `TestRestoreRollback_ProceedsWhenShareEnabledAtBoot`
- `TestRestoreRollback_QuiescesRollupBeforeReset`

Verification: `go build ./...` + `-tags integration`; `go test ./pkg/block/journal/... ./pkg/block/engine/... ./pkg/controlplane/{runtime,models,store}/...` (918 passed); `-race` journal+runtime (478 passed); gofmt/vet clean; golangci-lint 0 issues.

### DIFF
--- a/pkg/block/engine/flush.go
+++ b/pkg/block/engine/flush.go
@@ -125,6 +125,48 @@ func (bs *Store) SeedCold(ctx context.Context, payloadID string, offset, length 
 	return cs.SeedCold(ctx, journal.FileID(payloadID), offset, length)
 }
 
+// RestoreToVersion rewinds the local journal to a snapshot's version watermark
+// and re-materializes that point-in-time view durably at the log head. It is the
+// local-only snapshot-restore primitive the runtime calls instead of
+// ResetLocalState when the share has no remote store (the journal is the only
+// durable copy of the bytes). No-op when the local store is not journal-backed
+// (e.g. the in-memory test store), which the caller only reaches off this path.
+func (bs *Store) RestoreToVersion(ctx context.Context, v uint64) error {
+	if err := bs.enter(); err != nil {
+		return err
+	}
+	defer bs.closeMu.RUnlock()
+	type restorer interface {
+		RestoreToVersion(ctx context.Context, v uint64) error
+	}
+	r, ok := bs.local.(restorer)
+	if !ok {
+		return nil
+	}
+	return r.RestoreToVersion(ctx, v)
+}
+
+// SetPinVersion sets the local journal's snapshot pin watermark so GC/eviction
+// keep the bytes of every at-or-below-watermark record (the durable copy for a
+// live local-only snapshot). No-op when the local store is not journal-backed.
+func (bs *Store) SetPinVersion(v uint64) {
+	type pinner interface{ SetPinVersion(v uint64) }
+	if p, ok := bs.local.(pinner); ok {
+		p.SetPinVersion(v)
+	}
+}
+
+// JournalVersion returns the local journal's current LSN watermark, captured by
+// snapshot create (after DrainRollups) to record the snapshot's version. Returns
+// 0 when the local store is not journal-backed.
+func (bs *Store) JournalVersion() uint64 {
+	type versioner interface{ JournalVersion() uint64 }
+	if j, ok := bs.local.(versioner); ok {
+		return j.JournalVersion()
+	}
+	return 0
+}
+
 func (bs *Store) ResetLocalState(ctx context.Context) error {
 	if err := bs.enter(); err != nil {
 		return err
--- a/pkg/block/journal/reclaim.go
+++ b/pkg/block/journal/reclaim.go
@@ -334,7 +334,7 @@ func (s *Store) pinned(seg *segmentMeta) bool {
 		return false
 	}
 	mv := seg.minVersion.Load()
-	return mv != 0 && uint64(mv) <= pv
+	return mv != 0 && mv <= pv
 }
 
 // pickVictim returns the sealed segment with the highest dead fraction, or nil
--- a/pkg/block/journal/reclaim.go
+++ b/pkg/block/journal/reclaim.go
@@ -95,6 +95,12 @@ func (s *Store) claimColdestEvictable() (*segmentMeta, *shard) {
 				if !evictable(seg) {
 					continue
 				}
+				if s.pinned(seg) {
+					// Pinned by a live snapshot: its bytes are the only durable copy
+					// of an at-or-below-watermark record (local-only) — never evict
+					// (#1718).
+					continue
+				}
 				if la := seg.lastAccess.Load(); best == nil || la < bestAccess {
 					best, bestShard, bestAccess = seg, sh, la
 				}
@@ -317,6 +323,20 @@ func (s *Store) gcShard(ctx context.Context, sh *shard, opts GCOptions) (reclaim
 	}
 }
 
+// pinned reports whether a live snapshot's watermark protects any record in seg:
+// its lowest record Version is at or below pinVersion. Whole-segment granularity —
+// segments fill sequentially, so a non-empty segment's records span a contiguous
+// version range and minVersion alone decides. pinVersion 0 (no live snapshot) or a
+// still-empty segment (minVersion 0) pins nothing (#1718).
+func (s *Store) pinned(seg *segmentMeta) bool {
+	pv := s.pinVersion.Load()
+	if pv == 0 {
+		return false
+	}
+	mv := seg.minVersion.Load()
+	return mv != 0 && uint64(mv) <= pv
+}
+
 // pickVictim returns the sealed segment with the highest dead fraction, or nil
 // if none qualifies. A segment carrying no dead bytes is never a victim (there
 // is nothing to reclaim), which is also what keeps a tombstone-only segment from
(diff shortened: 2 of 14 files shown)


marmos91:
## Orchestrator adversarial review (pre-un-draft)

Reviewed the full diff independently (the automated reviewer stalled on the watchdog; this is a manual pass). Verified against the locked design doc. **No blockers found.** Confirmed:

- `go build ./...` + `go build -tags integration ./...` clean; `go vet` clean (no internal-import boundary violation — the transient LSP "undefined Store/Runtime" wall is a single-file-isolation false positive, `go build` is authoritative).
- The re-materialize / `minVersion` pin-skip / crash-reopen paths are genuinely exercised: the byteverify + crash-reopen fixtures use the `Type: "fs"` local store, which post-#1716 **is** the journal-backed store (not a memory no-op). Overwrite→restore→origA byte-verify and real-reopen durability both pass.
- `minVersion` is set on every append path + recovery scan + repack target; the whole-segment pin assumption holds because per-shard appends are version-monotone within a segment.
- Pin raise (before `MarkSnapshotReady`) / lower (recompute after delete commits) ordering is correct and monotone-conservative.

### Accepted notes (not blockers; follow-ups welcome)

1. **Startup pin-recompute window** (`runtime.go:511`): the pin is reconstructed at the *end* of `AddShare`, after `sharesSvc.AddShare` starts the block store. There is a sub-millisecond window where `pinVersion=0` before reconstruction. In practice reclaim is write-/timer-triggered and no client I/O occurs during `AddShare`, so a pinned segment cannot realistically be reclaimed in that window — but on a local-only share restarting with disk already over `MaxLocalBytes`, it is theoretically racy. Clean fix = set the pin inside `sharesSvc.AddShare` before background reclaim can run. Low severity; tracking as a follow-up.

2. **`RestoreToVersion` reimplements ceiling replay inline** rather than parameterizing `rebuildIdx` (design §3). The inline tombstone/truncate/newest-wins logic was verified faithful to `recover()`, but it is a duplicate that can drift if `recover()`'s semantics change. Maintenance debt, not a correctness bug.

3. **Pre-#1718 local-only journal snapshots** (`JournalVersion=0` + data, from the narrow #1716→#1718 window) are ambiguous with the legitimate empty-FS-snapshot case and cannot be safely re-materialized. Such snapshots' restores were already broken (returned zeros, previously `t.Skip`'d), so this is not a regression; no real deployment holds one. A guard would false-positive on the legitimate memory-local and empty-snapshot paths, so none was added.

Un-drafting for CI. Merge on green + Copilot addressed.

marmos91:
## Copilot review — addressed (commit 3439e00e)

1. **`store.go:702` `make([]byte, iv.length)` "won't compile":** it does compile — Go's `make` accepts any integer length type (CI is green, which proves it). The length is also already bounded: the loop skips `iv.length <= 0` and `PayloadLen` is validated by `scanValidRecords`. No 32-bit target exists (amd64/arm64 only). No change.

2. **+ 3. `minVersion` signed/unsigned hazard:** fixed — `segmentMeta.minVersion` is now `atomic.Uint64` end-to-end. `noteMinVersion` and `pinned()` compare in `uint64` with no `int64` round-trip, so the overflow-past-`MaxInt64` → negative → huge-`uint64` path Copilot flagged is gone. (`0 = unset` sentinel still holds; Versions start at 1.)

4. **Pre-#1718 `JournalVersion=0` guard:** added, with a correction. The bare `JournalVersion==0 && manifest non-empty` form false-positives on two legitimate paths — an empty-FS snapshot (correct to rewind-to-empty) and a **non-journal memory local store** (which legitimately reports `JournalVersion=0` and restores from its own CAS, not `RestoreToVersion`; this breaks `TestRestoreSnapshot_Integration/LocalOnlyRestore`). The guard now requires all three: `snap.JournalVersion==0 && snap.ManifestCount>0 && bs.JournalVersion()>0`, which fires only for a journal-backed store whose live log has real records but whose snapshot row carries no watermark — the actual pre-#1718 case — and fails closed instead of silently tombstoning.

Verified after the fixes: `go build ./...` + `-tags integration` clean, `go vet` clean, `pkg/block/journal` + `pkg/controlplane/runtime` full suites green (incl. the memory-store LocalOnlyRestore and the un-skipped journal-fs byteverify/crash-reopen).

PR #1729, opened 2026-07-16
chore(metadata): delete dead post-journal local bookkeeping (li:/ro:, local_chunk_index/rollup_offsets, BlockChunkCommit.Local)

marmos91:
## What

Pure deletion of the local block-store bookkeeping that the journal switchover superseded. Net **−2160 lines** across 47 files; no behavior change.

## Why it's dead

Post journal-switchover (#1716) + local-only restore (#1718), the journal's `.idx` sidecars are the authoritative local index and the local store is journal-backed. `BlockChunkCommit.Local` is genuinely always-zero — its only writer, `legacy_migration.go`, now always returns a zero `LocalChunkLocation` (the migration reads standalone chunks straight from the legacy remote; the journal local tier is `(payloadID,offset)`-keyed, not hash-keyed). With `.Local` always zero, `DefaultCommitBlock` never wrote a local-index entry, so nothing read it back.

## Deleted

- **badger** `li:` local chunk index (`keyLocalChunk`, encode/decode, `WalkLocalLocations`, `hashFromLocalChunkKey`, Put/Get/Delete) and `ro:` rollup offsets (`rollup.go`).
- **SQL** `local_chunk_index` + `rollup_offsets` tables. New drop migrations: **postgres `000039_drop_local_bookkeeping`**, **sqlite `000006_drop_local_bookkeeping`** (up drops both tables; down recreates both — DDL copied from pg `000036`/`000009`, sqlite `000003`/`000001`). Go impls (`postgres/rollup.go`, `sqlite/rollup.go`, and the Put/Get/Delete/Walk/scan methods in both `block_record_store.go`) removed.
- **memory** `localChunks` + `rollupOffsets` maps (store/transaction/reset), `rollup.go`.
- **Interfaces/wiring**: `metadata.LocalChunkIndex`, `metadata.RollupStore` (+ `ErrRollupOffsetRegression`), both embeddings in `metadata.Store`/`Transaction`, `blockCommitter.metadata.LocalChunkIndex` embedding, `BlockGCReclaimer.LocalIndex` + its two `DeleteLocalLocation` calls + the `LocalIndex: mds` wiring in `blockgc.go`.
- **Type**: `BlockChunkCommit.Local` field and the now-unused `block.LocalChunkLocation` type.
- **storetest**: deleted `local_index_ops.go` + `rollup_store_suite.go`; dropped the `LocalIndexOps` registration and the `.Local`/`GetLocalLocation` cases (plus the redundant `InTxRollback` and `ZeroLocalLocationNotWritten` subtests) from `commit_block_ops.go`.

## Schema-version bumps (silent-break guard)

- postgres `backupTables`: dropped `local_chunk_index` + `rollup_offsets`; `postgresSchemaVersion` **5 → 6**.
- sqlite `backupTables`: dropped `rollup_offsets`; `sqliteSchemaVersion` **1 → 2**.

## storetest suites removed

- `RunRollupStoreSuite` (+ the 4 per-backend `rollup_test.go` files) and the `LocalIndexOps` conformance subtest. Per-backend `synced_hash_store_test.go` that reused the deleted rollup helper now has its own local store ctor. `relaxed_durability_test.go` repointed off `SetRollupOffset` onto the durable `PutBlockRecord` write path.

## LocalChunkLocation decision

**Deleted** — after removing `.Local`, the index methods, and the migration plumbing, nothing referenced it.

## Verification

- `go build ./...` and `go build -tags integration ./...` — clean
- `go vet ./...` — clean
- `golangci-lint run` on all touched trees — 0 issues
- `go test ./pkg/block/... ./pkg/metadata/... ./pkg/controlplane/runtime/...` — all pass; sqlite conformance applied migration `000006` cleanly. Postgres migration `000039` is up+down reversible (down recreates both tables); it runs against a live DB under the integration harness.

## Notes

- Did **not** touch `FileChunk.LastSyncAttemptAt` / `LocalPath` / `BlockStoreKey` (separate follow-up).
- Draft for adversarial review before un-drafting.


marmos91:

Pure −2160-line deletion; verified against the deadness audit and re-run locally. **No blockers.**

- `go build ./...` + `-tags integration` clean; `go vet` + `golangci-lint` clean.
- Every metadata + block conformance suite green **without** the deleted surfaces (badger/memory/postgres/sqlite/storetest, engine, journal) — this is the proof the removed index/rollup/`.Local` paths were dead.
- Migrations reversible: `000039` (pg) / `000006` (sqlite) drop both tables; down-migrations recreate them from the original DDL.
- Silent-break vector handled: `postgresSchemaVersion` 5→6, `sqliteSchemaVersion` 1→2, and `local_chunk_index`/`rollup_offsets` removed from `backupTables` (remaining name hits are version-history comments only).
- `LocalChunkLocation` type fully removed (unused after the `.Local` collapse). `FileChunk.LastSyncAttemptAt` correctly left intact (live in the syncer claim-timeout) — its field-slimming siblings (`LocalPath`/`BlockStoreKey`) are a separate follow-up PR.

The one non-mechanical edit — repointing `relaxed_durability_test.go`'s `SetRollupOffset` durability assertion onto the `PutBlockRecord` durable path — is sound (the removed inline-fsync coverage was redundant with it) and the suite passes.

marmos91:
Addressed in 27aa31b5. You're right that the comment read as incomplete next to a migration that drops both tables. Clarified: `local_chunk_index` was never part of the **sqlite** snapshot payload (it's not in sqlite's `backupTables` — only pg backs it up), so dropping that table doesn't change the sqlite snapshot format; only removing the `rollup_offsets` section does, which is why `sqliteSchemaVersion` bumps for it alone. (pg is different — it backed up both, so `postgresSchemaVersion` 5→6 covers both sections.)

PR #1730, opened 2026-07-16
chore(block): slim FileChunk — drop dead LocalPath

marmos91:
## What

Drops the dead `block.FileChunk.LocalPath` field and its `cache_path` persistence column.

## Why dead

- No production constructor sets `LocalPath` — the three `FileChunk` constructors set only ID/Hash/DataSize/State.
- No behavioral reader consumes it. Its only touch was the pg/sqlite `cache_path` column round-trip, which post-journal-switchover only ever stored/loaded an empty string (the local store is journal-backed).
- The two partial indexes whose predicate is `cache_path IS NOT NULL` (`idx_file_blocks_pending`, `idx_file_blocks_remote`) are consequently dead — no query filters on `cache_path`, and the predicate is never true post-journal. Dropped with the column.

## Scope / migrations

- `pkg/block/types.go`: remove `LocalPath` field (+ stale lifecycle comment).
- pg/sqlite `objects.go`: remove `cache_path` from SELECT lists, INSERT column lists (placeholders renumbered), EXCLUDED, scan target, and args.
- Migrations: **pg `000040_drop_cache_path`**, **sqlite `000007_drop_cache_path`**. Up drops the two dead indexes then the column; down re-adds column + indexes (DDL copied from origin `000010` / `000001`). SQLite drops the indexes first because it refuses to drop a column referenced by an index predicate.
- storetest/tests: removed 34 `LocalPath:` composite-literal assignments (fixtures only — no reads/assertions existed).

## Snapshot schema version — BUMPED (required)

`file_blocks` **is** in `backupTables` for both backends (pg `snapshot_store.go`, sqlite `snapshot_store.go`), so dropping `cache_path` changes the dumped row shape. Bumped `postgresSchemaVersion` 6→7 and `sqliteSchemaVersion` 2→3, matching the established precedent (v4 dropped `link_counts`, v6 dropped `local_chunk_index`/`rollup_offsets`). No snapshot code references `cache_path` by name (it dumps via generic column introspection), so only the version constants + comments changed. Restore of pre-bump snapshots is rejected with `ErrSchemaVersionMismatch`, consistent with prior bumps (pre-1.0 backups are not forward-portable across schema versions).

## Deferred / kept

- **`BlockStoreKey` — NOT touched.** The original audit called it dead, but it has a **live reader**: `IsRemote()` (`types.go`) evaluates `b.BlockStoreKey != ""` as the legacy dual-read fallback for Pending rows, and `IsRemote()` is called in production (memory/badger objects.go). Removing it is a behavioral change (rewriting `IsRemote`'s legacy branch) — a separate follow-up gated on confirming the legacy dual-read window is closed. (The audit's `NeedsCASMigration` method no longer exists in the tree.)
- **`LastSyncAttemptAt` — kept.** Live in `pkg/block/engine/syncer.go` stale-claim reclaim.

## Verification

- `go build ./...` and `go build -tags integration ./...` — pass
- `go test ./pkg/block/... ./pkg/metadata/...` — 3492 passed / 27 packages (sqlite migration applies cleanly; badger/memory/sqlite conformance green)
- `gofmt -s`, `go vet`, `golangci-lint run` — clean


marmos91:

Scoped correctly to the safe subset after the deadness audit proved stale. **No blockers.**

- The audit's `BlockStoreKey`-is-dead verdict was **wrong** — `IsRemote()` (`types.go`) reads `BlockStoreKey != ""` and is called in production (`memory/objects.go`, `badger/objects.go` ×2). Correctly **deferred** here; removing it needs a separate deadness call on the legacy `IsRemote` dual-read window (behavioral change). `NeedsCASMigration` doesn't exist in-tree. Only genuinely-dead `LocalPath` dropped.
- **SQL placeholder alignment** (the silent-corruption risk) verified by hand: `cache_path` (position 4) consistently removed from the SELECT list, INSERT column list, `VALUES` ($10→$9 / ?10→?9), `Scan` targets, and the ON-CONFLICT `EXCLUDED` clause on both backends; remaining columns keep order. 3492 conformance tests pass.
- **Snapshot schema-version bump required + done:** `file_blocks` is in `backupTables` both backends → `postgresSchemaVersion` 6→7, `sqliteSchemaVersion` 2→3.
- **Migrations reversible:** sqlite drops the two partial indexes before the column (SQLite requires it), pg mirrors; downs re-add column + indexes from original DDL. The dropped `idx_file_blocks_{pending,remote}` partial indexes are dead — their `cache_path IS NOT NULL` predicate is never true post-journal (they index zero rows).
- `LastSyncAttemptAt` untouched (live in syncer).

Build clean (both tags), vet + golangci-lint clean.

PR #1731, opened 2026-07-16
chore(block): remove dead BlockStoreKey field + legacy IsRemote dual-read

marmos91:
## What

Removes the dead `FileChunk.BlockStoreKey` field + `block_store_key` column, and simplifies `IsRemote()` to `State == BlockStateRemote` (now identical to `IsFinalized()`).

## Why this is safe

`IsRemote()` was `State==Remote || (State==Pending && BlockStoreKey != "")` — a legacy dual-read. Findings:

- **Only 3 production callers** of `IsRemote()`, all the by-hash dedup-donor lookup (`memory/objects.go:493`, `badger/objects.go:390`, `badger/objects.go:1141`).
- **pg/sqlite `GetByHash` already filter `WHERE state=2` in SQL** and never call `IsRemote()`, so they already ignored the Pending clause.
- **Eviction / GC never use `IsRemote()`.**
- **No current-era code assigns a non-empty `BlockStoreKey`** — grep confirms every writer was test scaffolding; production only echoed persisted values back on DB scan.
- **Documented legacy rows are zero-Hash** → unreachable by `GetByHash` (keyed on real BLAKE3).

**Accepted risk (bounded, non-corrupting):** a hypothetical intermediate-era badger row (`state=Pending` + non-empty key + non-zero Hash) becomes a dedup MISS → ONE redundant re-upload (a self-consistent CAS write). Not corruption, not an eviction change. Given no prod users + delete-eagerly, we accept it. No promote-migration: SQL already ignores the clause, badger has no SQL migration path, and the blast radius is non-corrupting.

## Changes

- `pkg/block/types.go` — drop `BlockStoreKey` field, simplify `IsRemote()`, update lifecycle doc comments.
- `pkg/block/filechunk.go` — comment-only updates.
- pg/sqlite `objects.go` — drop `block_store_key` from SELECT lists, INSERT column list + VALUES (placeholders renumbered $9->$8 / ?9->?8), ON CONFLICT, scan target, and nullable guard.
- **Migrations:** pg `000041_drop_block_store_key`, sqlite `000008_drop_block_store_key` (up drops, down re-adds; no index references the column).
- **Snapshot schema-version bumps** (file_blocks is a backup table): pg `postgresSchemaVersion` 7->8, sqlite `sqliteSchemaVersion` 3->4.
- Tests — removed `BlockStoreKey` scaffolding fields; kept all `State` / `IsRemote` / `LastSyncAttemptAt` assertions.

`badger`/`memory` need no store change: JSON round-trip auto-drops the removed Go field, and old blobs carrying `"BlockStoreKey"` are ignored on unmarshal.

## Verification

- `go build ./...` and `go build -tags integration ./...` green.
- `go test ./pkg/block/... ./pkg/metadata/...` — 3492 passed, 27 packages (sqlite migration applies via conformance suite).
- `gofmt -s -w .`, `go vet ./...`, `golangci-lint run` — 0 issues.
- Placeholder alignment re-verified: both stores now 8 columns -> 8 placeholders -> 8 args in every INSERT.

### DIFF
--- a/pkg/block/engine/audit_state_test.go
+++ b/pkg/block/engine/audit_state_test.go
@@ -73,13 +73,12 @@ func seedAuditTestStore(t *testing.T) (store *metadatamemory.MemoryMetadataStore
 			off := uint64(bi) * 4096
 			blockID := payloadID + "/" + jstr(int(off))
 			fb := &block.FileChunk{
-				ID:            blockID,
-				Hash:          h,
-				State:         block.BlockStatePending,
-				BlockStoreKey: "cas/" + h.String()[0:2] + "/" + h.String()[2:4] + "/" + h.String(),
-				DataSize:      4096,
-				LastAccess:    now,
-				CreatedAt:     now,
+				ID:         blockID,
+				Hash:       h,
+				State:      block.BlockStatePending,
+				DataSize:   4096,
+				LastAccess: now,
+				CreatedAt:  now,
 			}
 			if err := store.Put(ctx, fb); err != nil {
 				t.Fatalf("Put block %s: %v", blockID, err)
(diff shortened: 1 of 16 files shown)


marmos91:

No blockers. Verified against the CONDITIONAL-SAFE deadness finding and re-run locally.

- **`IsRemote()` correctly collapsed** to `return b.State == BlockStateRemote` (now == `IsFinalized`). The legacy `&& BlockStoreKey != ""` clause is gone; pg/sqlite already ignored it (SQL `WHERE state=2`), eviction/GC never used it. Worst-case behavioral delta remains a bounded, non-corrupting dedup-miss → re-upload for a hypothetical intermediate-era badger row — accepted (no prod users, delete-eagerly).
- **INSERT placeholder alignment verified by hand** (the silent-corruption risk): both backends now 8 columns → 8 placeholders (`$1..$8` / `?1..?8`), `block_store_key` removed cleanly; 3492 conformance tests pass.
- **Snapshot schema-version bump done** (`file_blocks` is in `backupTables`): pg 7→8, sqlite 3→4.
- **Migrations reversible**: up drops (`IF EXISTS`), down re-adds the column from original DDL; no index referenced it.
- `LastSyncAttemptAt` + `State` untouched. `block_store_key` gone everywhere except down-migration DDL + version-history comments.

Build clean (both tags), vet + golangci-lint clean.

marmos91:
Addressed — reworded to describe the observable contract (hash collisions allowed; GetByHash may return either colliding row; a dedup miss just re-PUTs, never data loss) and dropped the stale `engine.uploadOne` reference.

PR #1732, opened 2026-07-16
feat(metadata): gate SQL manifest write on BlocksDirty — attr-only PutFile writes zero file_block_refs

marmos91:
## The write-amplification bug

`PutFile(File)` conflated attr mutation with manifest mutation. For a regular file, the postgres and sqlite backends ran `DELETE + INSERT×M` on `file_block_refs` on **every** `PutFile` — so a chmod / utimes / close / rename rewrote the entire chunk list. On the hot CLOSE path (`flushPendingWrite`) an M-chunk file paid an M-row manifest rewrite for a size/mtime-only update.

## The mechanism: `BlocksDirty`

Added a transient, request-scoped `BlocksDirty bool` to `metadata.File` (tagged `json:"-"`). It is **not persisted**: `json:"-"` keeps it out of the Badger `f:` blob, and the SQL backends write explicit column lists so a struct field never reaches a row. It defaults to `false`.

Both SQL backends now run `putFileChunkRefs` **only when `file.BlocksDirty`**. Attr-only writes leave it false and skip the manifest write entirely — end-state identical (the old DELETE+INSERT of the same rows was a no-op), zero rows written. Memory/Badger ignore the flag and keep holding `Blocks` inline (no regression).

## Manifest-changing sites that set the flag

- `pkg/metadata/file_modify.go` — truncate-down (`PruneChunkRefsToSize`), set only inside the prune branch (grow does not set it).
- `pkg/metadata/sparse.go` — `PunchHole` (rewrites `Blocks`).
- `pkg/metadata/block_record_store.go` — `ProjectManifestToBlocks`, the single funnel for carve/rollup commit (`DefaultCommitBlock`), the #953 reap+re-carve (`ReapSupersededManifest`), and `coordinator.ReprojectBlocks`.
- `pkg/controlplane/runtime/shares/coordinator.go` — `PersistFileChunks` (post-Flush Blocks+ObjectID).
- `internal/adapter/common/copy_payload.go` and `clone.go` — CopyPayload/Clone wholesale manifest replacement. **These two were not in the scoping plan's list** — they legitimately replace `dstFile.Blocks` before PutFile, so they set the flag too.

CREATE (`file_create.go`) constructs an empty `Blocks`, so the gate correctly writes nothing and there are no pre-existing rows — the first real manifest write is always `ProjectManifestToBlocks`/`PersistFileChunks` post-carve, both of which set the flag. No newly-carved file loses its refs.

## Also folded in (#8b)

`pkg/metadata/store/sqlite/file_block_refs.go` — converted the row-at-a-time INSERT loop into a batched multi-row INSERT (200 rows/batch to stay under SQLite's bound-variable limit). Postgres already batched via `pgx.Batch`.

## No schema bump

Write-path optimization only. Row shape and table framing are unchanged; a chmod that used to write a DELETE+INSERT no-op now writes nothing, end-state byte-identical for every manifest-changing op. Snapshot backup/restore is a raw table dump/load of `file_block_refs`, so it bypasses the gate — no restore regression.

## Correctness proof (the required new test)

`testChunkRef_ChmodDoesNotRewriteRefs` (in `storetest/truncate_block_refs.go`, fans out to all backends): loads a 4-chunk file, chmods it with mode-only `SetFileAttributes`, and asserts **zero** manifest writes, then truncates and asserts exactly one. Row-count alone cannot prove "zero writes" (a DELETE+INSERT of the same M rows leaves the same count), so both SQL stores expose a test-only `PutFileChunkRefsCallCount()` (an `atomic.Int64` incremented only when a PutFile runs past the gate). The test baselines the counter after seeding and asserts delta 0 for chmod, delta 1 after truncate.

## Badger unchanged

Adding the field is inert for Badger (its `PutFile` ignores the flag and keeps writing the full blob). The badger `f:` split is the separate high-risk Track B.

## Verification

`go build ./...` + `-tags integration` green; `go test ./pkg/metadata/... ./pkg/block/...` 3496 pass; `go test -race ./pkg/metadata/...` 2627 pass; gofmt/`go vet`/golangci-lint clean.


marmos91:

The write-amp win. No blockers; verified the data-loss-critical paths myself.

- **Gate correct:** `if file.Type == FileTypeRegular && file.BlocksDirty` guards `putFileChunkRefs` on both SQL backends. `BlocksDirty` is transient (`json:"-"`, not persisted, kept out of the badger blob).
- **Create→carve→persist is safe** (the scary regression): the first real manifest write after a carve routes through `block_record_store.go` `ProjectManifestToBlocks` (sets flag) and `coordinator.go` `PersistFileChunks` (sets flag), so a newly-carved file's refs always persist. CREATE with empty Blocks correctly skips (nothing to write). SQL snapshot restore is a raw table dump/load — no PutFile, bypasses the gate, no restore regression → no schema bump needed.
- **All 6 manifest-changing sites set the flag:** truncate-prune, PunchHole, ProjectManifestToBlocks (carve/#953-reap/reproject), PersistFileChunks, and CLONE + copy-payload (found beyond the original list — they replace `dstFile.Blocks` wholesale).
- **Zero-writes proof is real:** I ran it — `TestConformance/TruncateChunkRefOps/ChmodDoesNotRewriteRefs` PASSES. It asserts the `PutFileChunkRefsCallCount` delta is 0 across a chmod (not just row-count, which can't distinguish a no-op DELETE+INSERT) and 1 after a truncate.
- Full `pkg/metadata/...` + `pkg/block/...` conformance green across badger/memory/pg/sqlite (badger ignores the flag → unchanged). #8b (sqlite batched INSERT) folded in.

Badger `f:`-blob split is deliberately NOT here — that's the separate high-risk Track B (snapshot/GC hash-scan). Build clean (both tags), vet + scoped golangci-lint clean. (The whole-repo lint's phantom issue pointed at a sibling worktree — stale shared-cache artifact, not this code.)

Issue #1687, opened 2026-07-14
Metadata wall is badger DB.Sync vs write-pipeline RWMutex contention, not concurrent-sync serialization (#1573)


marmos91:
## Summary

The metadata workload wall (`dfsbench metadata`: 4 threads × create+write-4k+close, ~250 ops/s, dead last vs ZeroFS 354 / JuiceFS ~1990) is **not** caused by concurrent `badger DB.Sync` calls serializing against each other. Two group-commit attempts to coalesce those syncs both failed to move it. A fresh Linux mutex profile shows the real cause: `DB.Sync` (RLock) contends with badger's **own write pipeline** (`doWrites`/`ensureRoomForWrite`, Lock) for badger's value-log RWMutex — *sync-vs-write*, not *sync-vs-sync*.

## What was tried (both ineffective)

1. **PR #1684** — group-commit at the `flushPendingWrite` call site. Measured 256→250 ops/s (no change). Reverted.
2. **Branch `perf/1573-badger-dbsync-groupcommit`** (this investigation) — group-commit leader at `syncIfRelaxed`, the universal chokepoint for *all* durable syncs (WithTransaction + SetRollupOffset + durability ticker). The coordinator is correct and provably coalesces (unit test: 32 concurrent durable commits → 7 barrier passes, 4.6×). But same-VM A/B: treatment 249–285 vs baseline 211–254 ops/s (+12–19%, within rig noise), and the mutex profile is **unchanged**. Shelved, not merged.

## Evidence — treatment mutex profile (4-thread metadata load, Linux)

```
      flat  flat%   sum%        cum   cum%
    23.26s 63.82% 63.82%     23.26s 63.82%  sync.(*RWMutex).RUnlock (inline)
     6.62s 18.18% 81.99%      6.62s 18.18%  sync.(*Mutex).Unlock
     6.56s 17.99%   100%      6.56s 17.99%  sync.(*RWMutex).Unlock
         0     0%   100%     23.25s 63.81%  badger/v4.(*DB).Sync            <- via syncIfRelaxed (leader active)
         0     0%   100%      5.26s 14.45%  badger/v4.(*DB).ensureRoomForWrite  <- the WRITE side of the same lock
```

- **63.8%** of all mutex delay is `DB.Sync` → RWMutex.RUnlock, reached through `syncIfRelaxed` (43% WithTransaction, 20% SetRollupOffset). Essentially identical to baseline — the group-commit leader did **not** collapse it.
[...]

PR #1733, opened 2026-07-16
feat(metadata): flush→relaxed — move metadata fsync off the NFS/SMB write+commit ack path (#1687)

marmos91:
Moves the metadata `db.Sync` off the NFS/SMB WRITE+COMMIT ack path (#1687) — the felt-latency win. This is durability-critical: a bug here silently truncates an ACK'd file (#588 class), so the relax ships together with a load-bearing crash-safety net.

The relaxed-transaction machinery already existed (`withRelaxedTransaction` / `RelaxedTransactor`, badger+postgres `WithTransactionRelaxed`, `RelaxedDurability` defaulted on). This PR wires `flushPendingWrite` through it and adds the safety net — no new relaxed-txn plumbing.

## The 3 pieces (none is safe alone)

**Piece 1 — the relax.** `flushPendingWrite` / `FlushPendingWriteForFile` gain a `durable bool`. At commit: `durable` → strict `WithTransaction` (inline fsync); otherwise `withRelaxedTransaction` (bounded-lag background fsync). Shutdown passes `durable=true`. The dead `FlushPendingWrites` (zero callers) is deleted.

**Piece 2 — startup size reconciliation (the safety net).** `reconcileMetadataSizeFromJournal` runs in `AddShare` after the block store is built and BEFORE the share is registered (invisible to handlers). For each `id` in the local journal's `ListFiles`, it grows `metadata.Size` up to `journal.FileSize(id)` — **max-only, never shrinks** — under a strict transaction. A crash between a relaxed size commit and its background fsync therefore cannot truncate ACK'd data: the size is restored from the journal's durable high-water mark. Gated on `LocalBlockStoreID != ""`; the `*journal.Store` is reached via `engine.Store.Local()` (new accessor) whose `ListFiles`/`FileSize` are journal-native interface methods served from the in-memory index post-`recover()`.

**Piece 3 — durability classification (per sign-off).**

| Call site | durable |
|---|---|
| NFS v3 WRITE (`flushStableWrite`) | `stable >= FileSyncWrite` (FILE_SYNC only) |
| NFS v3 COMMIT | `false` |
| NFS v4 COMMIT | `false` |
| NFS v4 CLOSE | `false` |
| SMB WRITE | `false` ← the big win (per-write fsync removed) |
| **SMB CLOSE** | **`true` (STRICT — #1267 / MS-SMB2 3.3.5.10)** |
| **SMB FLUSH** | **`true` (STRICT — explicit client fsync)** |
| SMB copychunk / set-zero-data | `false` |
| metadata DEALLOCATE / ALLOCATE (sparse) | `true` (rare, precedes strict manifest PutFile) |
| Shutdown | `true` |

NFS v4 WRITE always reports UNSTABLE4 and never flushes inline — no change.

## Stayed STRICT (verified)
- SMB CLOSE + FLUSH (table above); the #1267 CLOSE durability comment stays true.
- `file_modify.go` SetAttr/truncate — a shrink is NOT covered by max()-only reconciliation, left strict, untouched.
- FILE_SYNC WRITE.

## The merge gate — crash-safety test
`pkg/controlplane/runtime/flush_relaxed_crash_test.go` models an ACK'd relaxed write that crashed before the metadata fsync: write N bytes journal-durable (`WriteToBlockStore` + `CommitBlockStore`, no rollup, no `FlushPendingWriteForFile`), so `metadata.Size` is still 0 while the journal holds N. After `simulateRestart` (real close+reopen + fresh `AddShare` → reconciliation), it asserts `Size == N` and the N bytes read back intact. Runs on badger + sqlite (memory/postgres skip — no durable reopen / needs integration tag).

This is a genuine gate: with the reconciliation call neutralized, both badger and sqlite FAIL with `ACK'd data truncated: metadata.Size = 0 after restart, want 262144`; with it, both PASS.

A badger store-level unit test (`TestRelaxedDurability_FlushPendingWriteRouting`) pins the routing via `InlineSyncCountForTest`: `durable=false` does not advance the inline-sync count, `durable=true` advances it by one.

## Notes
- **#1718 resolved:** the local-only snapshot-restore path that could have made `journal.ListFiles/FileSize` untrustworthy after a restore landed (#1727) and rebuilds the journal index correctly via re-materialize, so reconciliation-after-restore is fine.
- **sqlite is best-effort:** it has no strict/relaxed split (WAL + `synchronous=NORMAL` unconditionally, no `RelaxedTransactor`), so `durable=true` falls back to `WithTransaction` and fsyncs only at checkpoint. Pre-existing; documented in `sqlite/config.go`, not changed here.

## Verification (all green)
`go build ./...` + `-tags integration`; `go test ./pkg/metadata/... ./pkg/block/... ./internal/adapter/...` (58 ok / 0 fail); `go test -race ./pkg/controlplane/runtime/... ./pkg/metadata/...` (0 fail, no data race); `gofmt -s`; `go vet ./...`; `golangci-lint` (0 issues).

Draft — adversarial review before un-draft.

### DIFF
--- a/internal/adapter/nfs/v3/handlers/commit.go
+++ b/internal/adapter/nfs/v3/handlers/commit.go
@@ -211,7 +211,10 @@ func (h *Handler) Commit(
 	authCtx, authErr := h.GetCachedAuthContext(ctx)
 	if authErr == nil {
 		// Flush pending metadata for this specific file
-		flushed, metaErr := metaSvc.FlushPendingWriteForFile(authCtx, handle)
+		// Relaxed metadata commit (#1687): COMMIT already forced the block data
+		// durable above; the size/mtime fsync is deferred to the background syncer,
+		// with the journal size reconcile on restart as the crash-safety net.
+		flushed, metaErr := metaSvc.FlushPendingWriteForFile(authCtx, handle, false)
 		if metaErr != nil {
 			logger.WarnCtx(ctx.Context, "COMMIT: metadata flush failed", "handle", fmt.Sprintf("0x%x", req.Handle), "error", metaErr)
 			// Continue - content is flushed, metadata will be fixed eventually
--- a/internal/adapter/nfs/v3/handlers/write.go
+++ b/internal/adapter/nfs/v3/handlers/write.go
@@ -350,7 +350,10 @@ func (h *Handler) flushStableWrite(
 	if err := common.CommitBlockStore(ctx.Context, blockStore, payloadID); err != nil {
 		return err
 	}
-	if _, err := metaSvc.FlushPendingWriteForFile(authCtx, handle); err != nil {
+	// FILE_SYNC (stable>=2) promised durable metadata → strict inline fsync.
+	// DATA_SYNC/UNSTABLE defer it via the relaxed path (#1687); a later COMMIT or
+	// the journal size reconcile on restart makes the size durable.
+	if _, err := metaSvc.FlushPendingWriteForFile(authCtx, handle, stable >= FileSyncWrite); err != nil {
 		if stable >= FileSyncWrite {
 			// FILE_SYNC requires durable metadata; surface the failure.
 			return err
--- a/internal/adapter/nfs/v4/handlers/close.go
+++ b/internal/adapter/nfs/v4/handlers/close.go
@@ -104,7 +104,9 @@ func (h *Handler) handleClose(ctx *types.CompoundContext, reader io.Reader) *typ
 			"error", metaErr, "client", ctx.ClientAddr)
 	} else {
 		fileHandle := metadata.FileHandle(ctx.CurrentFH)
-		flushed, flushErr := metaSvc.FlushPendingWriteForFile(authCtx, fileHandle)
+		// Relaxed metadata commit (#1687): NFSv4 CLOSE is not an fsync boundary
+		// (clients use COMMIT for durability); defer the metadata fsync.
+		flushed, flushErr := metaSvc.FlushPendingWriteForFile(authCtx, fileHandle, false)
 		if flushErr != nil {
 			logger.Warn("NFSv4 CLOSE metadata flush failed",
 				"error", flushErr, "client", ctx.ClientAddr)
--- a/internal/adapter/nfs/v4/handlers/commit.go
+++ b/internal/adapter/nfs/v4/handlers/commit.go
@@ -138,7 +138,9 @@ func (h *Handler) handleCommit(ctx *types.CompoundContext, reader io.Reader) *ty
 	// Flush pending metadata writes (deferred commit optimization)
 	// This is critical - without this, file size and other metadata changes
 	// are not persisted to the store, causing data loss on server restart.
-	flushed, metaErr := metaSvc.FlushPendingWriteForFile(authCtx, fileHandle)
+	// Relaxed metadata commit (#1687): the size/mtime fsync is deferred; restart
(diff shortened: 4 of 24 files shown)


marmos91:
## Orchestrator review (pre-un-draft) — verified independently, highest rigor

Durability-critical (silent-truncation class). Verified every safety-relevant claim myself.

- **Crash-safety gate is genuinely load-bearing** (the key check): I neutralized `reconcileMetadataSizeFromJournal` (`if true { return nil }`) and re-ran — the test correctly **FAILS**: `ACK'd data truncated: metadata.Size = 0 after restart, want 262144` on both badger and sqlite. Restored → passes. So piece 1 (relax) without piece 2 (reconcile) = data loss, and the test proves piece 2 prevents it. Not a hollow green.
- **Reconciliation is max-only, never shrinks:** `if int64(f.Size) >= journalSize { continue }` + a strict in-txn re-check before the `PutFile`. A relaxed size-commit lost to a crash is recovered up to `journal.FileSize` on share start; a shrink is never applied (which is why `SetAttr`/truncate correctly stays strict — it's untouched).
- **Durability classification matches sign-off exactly:** relaxed = SMB per-WRITE, NFS UNSTABLE/v3+v4 COMMIT/v4 CLOSE, ioctls; strict = NFS FILE_SYNC (`stable >= FileSyncWrite`), **SMB CLOSE (true), SMB FLUSH (true)** — the #1267/MS-SMB2 comment stays accurate, no rewrite.
- Reconciliation hook placed in `AddShare` after Phase 2 / before Phase 3 (share not yet handler-visible); reaches the journal via the existing `local.LocalStore` `ListFiles`/`FileSize` methods (through a new `engine.Store.Local()` accessor) — cleaner than a `*journal.Store` type-assert. Gated on a local journal-backed store.
- Build clean (both tags); metadata+block+adapter suites + `-race` green; the badger `InlineSyncCountForTest` routing test confirms `durable=false` skips the inline fsync and `durable=true` keeps it.

The two badger comments that asserted "size stays synchronous" are correctly rewritten to the reconciliation-backed invariant. sparse.go PunchHole/Allocate → `durable=true` (conservative, rare, precede a strict manifest PutFile) — reasonable. #1718-after-restore is resolved (#1727 re-materialize rebuilds the journal index).

marmos91:
## Copilot review — addressed (commit pushed)

Two of these were real hardening of the safety-net path itself — good catches:

1. **`service.go` reconcile error handling** — fixed. Now only `metadata.IsNotFoundError(err)` is treated as an orphan (skip); any other error (I/O, corruption) **returns** rather than silently leaving `metadata.Size` stale. Added a nil-file guard too.
2. **`int64(f.Size)` overflow** — fixed. Both the pre-check and the in-txn re-check now compare in `uint64` with a `journalSize < 0` guard, so a `Size > MaxInt64` can never cast negative and shrink — the max-only invariant holds unconditionally.
4. **`service_test.go` COMMIT test** — fixed. Now passes `durable=false` (COMMIT is relaxed post-#1687), matching the comment and exercising the relaxed routing.

3. **`io.go:398` pop-before-flush loses pending state on error** — valid, but **pre-existing** (the pop-before-flush predates #1687; strict flush had the same behavior). The restart reconciliation already backstops the durability (a lost relaxed size-commit is recovered on next share start). Improving the in-session retry semantics (keep/restore state on flush error) is a separate robustness change orthogonal to this PR — tracking as a follow-up rather than widening #1687's scope.

Re-verified after the fixes: build clean, the crash-safety gate still passes (and still genuinely fails when reconciliation is neutralized), metadata + shares suites green.

Issue #1739, opened 2026-07-16
bench: harness must compare apples-to-apples across the whole matrix (mount opts, config, log level, caching)


marmos91:
## Mandate
The dfsbench matrix must run **every backend under identical, controllable conditions** — same mount options, same client config, same log level, same cache sizing, same durability tier, same fio/warp workload params, same warmup/measure window. No benchmark number is trustworthy until this holds. Where a backend's native access mode genuinely differs (FUSE vs NFS), normalize via the existing re-export path (`internal/dfsbench/backend/reexport.go`) or document the irreducible difference explicitly — never leave it a silent confound.

## Why (what triggered this)
The apparent metadata-create gap vs JuiceFS is **substantially a harness artifact**, not a DittoFS deficiency:
- Server sustains ~15,600 creates/s over loopback (64µs/op); the rig sees 523/s (1.9ms/op) → server is ~97% idle at rig load. The 1.9ms is NFS **round-trip**, not server work (see #1735).
- `Service.CreateFile` = 27µs — the store is not the wall.

So the gap is dominated by transport + harness asymmetries stacked on top. This issue tracks neutralizing the asymmetries.

## Confounds found so far
1. **Per-op logging.** DittoFS logs 2 synchronous INFO lines per create (lumberjack, mutex-serialized file write on the create critical path); JuiceFS logs nothing per-op by default. `create.go:95,330`. Measured ~26µs/op (~12%) single-thread. Default log level = INFO. → **PR #1738** demotes NFS per-op InfoCtx→DebugCtx; harness must also pin log level equal across the matrix.
2. **Mount-opts asymmetry.** DittoFS NFS mounted `actimeo=0,nolock` (attr cache OFF → revalidation round-trip per op) + single connection (no `nconnect`); JuiceFS FUSE = full page-cache. `dittofs.go:237`. → drop `actimeo=0`, add `nconnect`, match FUSE caching tier (branch `perf/bench-fair-nfs-mount-opts`).
3. **Transport.** JuiceFS native FUSE (no wire); DittoFS native NFS (full round-trip). Use `reexport.go` for a controlled same-transport A/B, or document as irreducible.
4. **Pending full audit:** durability tier, cache size, meta locality, fio params, background-work CPU steal during measure, cold-read eviction method, restart re-seed, S3 config parity.

## Plan
1. Land #1738 (log demotion) + pin log level equal across all backends.
[...]

Issue #1735, opened 2026-07-16
perf: close the create per-op gap vs JuiceFS — cross-op inode/dirent cache + binary codec + relaxed-tier write-back coordinator


marmos91:
Follow-up from #1687 / #1692. After flush→relaxed shipped, a store-level trace shows the metadata store is NOT the bottleneck:

| layer | per create |
|---|---|
| `Service.CreateFile` (relaxed badger) | **27µs** (~36k/s) |
| end-to-end over NFSv3 | **~1900µs** (523 ops/s) |

The store is ~70× faster than the end-to-end number, so **~98% of a create is spent above the metadata store** — in the NFS adapter (XDR decode/encode, RPC dispatch, per-op work/allocation, round-trip). This matches the known read-path finding (engine 47k IOPS → 8–13k end-to-end = per-op NFS RPC decode + metadata lookup + allocation, BENCHMARKS.md).

**This is where the JuiceFS/ZeroFS metadata gap actually lives** (DittoFS 523 vs JuiceFS ~1682 ops/s). Ruled out as the cause: durability (relaxed), badger (27µs), parent-dir contention (`DirTimesTracker`; separate-dir is *slower*), group-commit (refuted ×3).

Next: a running-server pprof under a create workload to localize the ~1.9ms in the adapter (CPU-bound decode/alloc vs latency/round-trip vs a synchronous per-op op). Fix likely reduces per-op allocations / redundant RPCs; some portion may be inherent NFS round-trip.

marmos91:
## Confirmed as the #1 perf target by the fair rig run (2026-07-17)

With the apples-to-apples harness (#1739: durability matched, mount opts identical, log level pinned), DittoFS metadata is **486 ops/s — still last** vs JuiceFS 1766 and ZeroFS 947. This is the honest correction to the earlier "mostly artifact" read: neutralizing async/actimeo/logging did NOT close the metadata gap, so the deficit is real and it is **this adapter/RPC wall**, not durability/store.

Evidence it's the userspace NFS adapter, not the store:
- `Service.CreateFile` = 27µs; server sustains ~15,600 creates/s over loopback (64µs/op) — the store is not the wall.
- Rig = 486 ops/s (~2ms/op). knfsd (kernel) re-exporting a fast local-meta FUSE mount (JuiceFS) does 1766 ops/s on the SAME wire — so a kernel NFS server + specialized meta beats our userspace adapter + badger per-create path.
[...]

PR #1741, opened 2026-07-17
bench: apples-to-apples harness — durability + mount opts + log level parity (#1739)

marmos91:
Makes the dfsbench matrix compare every backend under identical conditions (#1739). Validated by a fair rig run: with these fixes DittoFS goes from "dead last on writes" to **leading rand-write (6.9× JuiceFS)** and beating ZeroFS across the board — because competitors are no longer acked non-durably.

## Changes (all `internal/dfsbench/`)
- **Durability parity:** knfsd export `async` → `sync` (`reexport.go`) and Samba `strict sync = yes` (`smb.conf.tmpl`). Competitors were acked from knfsd/smbd RAM before their FUSE layer committed, while DittoFS commits durably — so writes compared durable-vs-volatile. Now both ack from stable storage (DittoFS's tier: local-durable + async-S3, matched by JuiceFS `--writeback` which is kept).
- **Mount-opt parity:** native DittoFS/zerofs mounts had `actimeo=0` (attr-cache off → per-op revalidation RPC) while the re-export client mounts had the kernel default. All client mounts now `actimeo=1,nconnect=4` (+`nolock` on v3), v3 and v4.1.
- **Log-level parity:** benched DittoFS server pinned to `DITTOFS_LOGGING_LEVEL=WARN` so per-op logging is uniformly off.

Fair-run result (medium, warm): DittoFS leads rand-write-4k (4611 IOPS) and mixed-rw (6115), ties seq-read; the remaining metadata deficit is the userspace NFS adapter (#1735), now isolated as the real target.

Part of #1739.

### DIFF
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -1,45 +1,100 @@
 # DittoFS Performance Benchmarks
 
 Performance of DittoFS (S3 backend) against the S3-backed network filesystems the
-`dfsbench` harness supports — **JuiceFS, rclone, s3fs, and ZeroFS** — plus
-**local-disk** as the raw-hardware ceiling. All systems run on one disposable
-Scaleway VM, driven by the same `fio` workloads over the same protocol.
+`dfsbench` harness supports, on the **apples-to-apples harness**
+([#1739](https://github.com/marmos91/dittofs/issues/1739)): every backend runs
+under **identical** conditions — durability matched (competitors re-exported
+`sync`, not `async`), identical NFS mount options (`actimeo=1,nconnect=4`), and a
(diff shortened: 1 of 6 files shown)


marmos91:
Addressed the Copilot comment: added explicit `tcp` to the re-export client mount opts (`vers=N,tcp,actimeo=1,nconnect=4`) — `nconnect` is TCP-specific and the native mounts set `tcp` explicitly, so this keeps the mounts truly identical. The red check was an unrelated flake (`TestParentMtimeContention` timing ratio, in pkg/metadata which this PR doesn't touch) — root-cause fix in #1745.

PR #1763, opened 2026-07-18
docs(bench): durability-first narrative rewrite + verified sync-to-S3 tier

marmos91:
Rewrites `docs/BENCHMARKS.md` as a narrative, docs-website-ready page and folds in the newly-verified sync-to-S3 (remote) durability tier.

## What changed
- **Durability-first structure**: premise (speed is meaningless without stating the guarantee) → test environment + fairness method → results grouped by tier (writeback / local-durable / synchronous-to-S3) → mixed throughput+IOPS table → plain-language summary → verified reproduce section.
- **New verified result**: the synchronous-to-S3 tier is confirmed end-to-end on real object storage (data is in the bucket the instant a write acks; the local tier's identical test leaves the bucket empty). A same-machine, 6-run study puts **DittoFS and JuiceFS at a dead heat** (~26 ops/s each), both ~10× faster than s3fs write-through. rclone's no-cache mode is excluded (its flush errors, leaving a 0-byte object — not the same guarantee); goofys can't run a create workload (no metadata engine).
- **Cleaned for publication**: dropped superseded prior-cycle tables and internal jargon; no issue/PR numbers, no credentials (only placeholder env-var names in the reproduce section); verified the `dfsbench` CLI commands and system labels against the real tool.

Docs-only.

### DIFF
--- a/docs/BENCHMARKS.md
+++ b/docs/BENCHMARKS.md
@@ -1,349 +1,206 @@
 # DittoFS Performance Benchmarks
 
-Performance of DittoFS (S3 backend) against the S3-backed network filesystems the
-`dfsbench` harness supports, on the **apples-to-apples harness**
-([#1739](https://github.com/marmos91/dittofs/issues/1739)): every backend runs
-under **identical** conditions — durability matched (competitors re-exported
-`sync`, not `async`), identical NFS mount options (`actimeo=1,nconnect=4`), and a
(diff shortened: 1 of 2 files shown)


marmos91:
Addressed both: reproduce-command systems pinned to `-nfs3` (bare labels expanded to all protocols), and the durability guide resynced (fixed the removed anchor link, updated the `remote` row from 'not yet benchmarked' to the measured JuiceFS parity).

Issue #1793, opened 2026-07-19
Remove inert local fs block-store config knobs (post-journal)


marmos91:
The journal write-back cache replaced the old append-log + log-blob local tier, but the per-share local `fs` store still *accepts* several config knobs that no longer do anything. Docs PR #1792 relabels them "legacy / inert"; this issue tracks actually deleting them so the config surface tells the truth.

### Inert knobs (accepted, no effect)
`FSStoreOptions` (`pkg/block/local/fs/fs.go`) already flags these — its own comment: *"no longer control anything … Drop them once the shares service stops reading them."*

- `max_log_bytes` — no longer gates writes; the journal caps on-disk usage via its own `MaxLocalBytes` budget + eviction (`ErrLocalStoreFull`). Only feeds `Stats` now. (Still resolved/plumbed in `pkg/controlplane/runtime/shares/service.go`.)
- `rollup_workers` — no rollup pool exists; carve runs on the journal's own age/size gate.
- `stabilization_ms` — superseded by the journal carve batching gate.
- `orphan_log_min_age_seconds` — no separate append-log to orphan-sweep.

### Work
1. Drop the fields from `FSStoreOptions` and the `RollupWorkers`/`StabilizationMS`/`SyncEveryWrite`/`OrphanLogMinAgeSeconds`/`MaxLogBytes` plumbing in `shares/service.go`.
2. Decide `max_log_bytes`: either wire it to the journal's `MaxLocalBytes` cap (if we want a per-share cache-size knob) or delete it too. Leaning delete — the deduced local-store size already covers this.
3. Emit a startup warning on any of these keys (same pattern as the ignored `use_append_log`) for one release, then reject.
4. Update `docs/guide/configuration.md` to drop the rows once removed.

Pre-1.0, no prod users → prefer straight deletion over a long deprecation window.

Refs #1692

marmos91:
One more for the same cleanup so nothing is forgotten:

- `use_append_log` (`fs` store config) — already **accepted-but-ignored**: appending is mandatory, and setting it just logs a startup warning. Delete the key handling in the same pass and drop it from the docs.

Full legacy/inert set to remove: `max_log_bytes`, `rollup_workers`, `stabilization_ms`, `orphan_log_min_age_seconds`, `use_append_log`.

marmos91:
Inert fs knobs removed in #1794 (the config surface now matches the journal design). The residual `max_log_bytes` decision is noted for follow-up.

[...]

PR #1794, opened 2026-07-19
chore(block): drop inert local fs-store rollup/append-log knobs

marmos91:
Follow-up to the docs refresh (#1792): actually delete the dead local `fs`-store knobs instead of documenting them as "legacy" forever.

The journal write-back cache replaced the old append-log + rollup local tier. These `FSStoreOptions` knobs were left wired up but control nothing — the journal carves and evicts on its own age/size/disk budget:

- `rollup_workers`
- `stabilization_ms`
- `sync_every_write`
- `orphan_log_min_age_seconds`
- `use_append_log` (was already accepted-but-ignored, warn-only)

### What changed
- Removed the four fields from `FSStoreOptions` (`pkg/block/local/fs/fs.go`).
- Removed the config-parsing blocks that read those keys (+ the `use_append_log` warning) in `CreateLocalStoreFromConfig` (`shares/service.go`). The keys are now silently ignored like any unknown config key.
- Updated the test fixtures / config-parsing tests that set them.

Net −70/+17 across 6 files. Build, vet, and the affected package tests (`fs`, `shares`, touched `engine` fixtures) are green.

### Not in this PR
`max_log_bytes` is intentionally kept: unlike the four above it still resolves through the config schema (`blockstore.local.max_log_bytes`), the deduced-defaults machinery, the `dfs config show` / `start` output, a `Stats` size hint, and the `localstore_append_log_limit_bytes` Prometheus metric. Removing it is a bigger, observable-surface change (metric + schema + CLI), tracked in #1793 — it should decide whether to wire it to the journal's `MaxLocalBytes` cap or delete it outright.

Refs #1793

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -302,22 +302,17 @@ server will start.
 
 These keys live inside the per-share `local` block store's `config` JSON
 (passed via `dfsctl store block local add --config '{...}'` or the REST API).
-They only take effect when the local store type is `fs`. (A legacy
-`use_append_log` key is accepted but ignored — appending is mandatory — and
-logs a startup warning.)
+They only take effect when the local store type is `fs`. Legacy keys from the
+pre-journal design — `use_append_log`, `rollup_workers`, `stabilization_ms`,
+`orphan_log_min_age_seconds` — are no longer parsed; if present they are
+silently ignored.
 
 | Key | Type | Default | Description |
 |-----|------|---------|-------------|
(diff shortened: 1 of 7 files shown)


marmos91:
Addressed the Copilot review. All six comments were the same class — stale `rollup`/`append-log` wording after the journal switchover.

Fixed (comments my knob removal made newly-misleading):
- `service.go`: the `max_log_bytes` seeding comment (it's a Stats size hint, not an append-log pressure budget) and the `chunk_size` comment (`carve chunker`).
- `write_bench_test.go`: the const comment now documents its `MaxLogBytes` role (maxDisk is passed 0 = uncapped).

Intentionally left: the two test-function names (`TestCreateLocalStoreFromConfig_AppendLogMandatory`, `_RollupSurvivesCallerCancel`). Those names/notes are pre-journal-switchover residue that predates this branch (this PR removes config knobs; it didn't introduce them), and their bodies already carry accurate notes explaining the switchover. Renaming test functions is churn better handled in a dedicated naming-cleanup sweep than piecemeal here.

PR #1796, opened 2026-07-19
feat(journal): opt-in per-read integrity + self-heal on durable tiers

marmos91:
aa
## The gap

`journal.Store.ReadAt` served a warm read piece with a raw `seg.fd.ReadAt(out, loc.Offset+subOff)` — **no per-read integrity check**. Integrity today = startup-recovery CRC + remote cold-fetch BLAKE3. On-disk corruption of a valid segment *between recovery and a warm read* (bit rot, or a bug mutating cached segment bytes) returned silently-wrong bytes — the worst failure mode. The Wall-A switchover dropped the old per-read verify deliberately, for read speed.

## The design (tier-gated CRC32 + typed error + engine heal)

Opt-in, layered so the journal stays remote-agnostic and the default path stays byte-for-byte the current fast raw read.

- **Journal `Store.VerifyReads`** (set at share construction from the resolved durability tier): `writeback → false` (raw fast read, **unchanged**); `local-durable` and `remote → true`.
- When on, `ReadAt`s warm-piece branch reads the **covering record + trailing CRC** (via the existing `readRecordAt`, reachable through the interval `recOff` already carried on the plan — no second resolve), verifies the per-record CRC32, then slices out the requested sub-range. This adds read amplification (whole record vs sub-range) — the accepted cost of the opt-in tiers.
- On mismatch the journal returns a typed **`*CorruptRangeError{FileID, Offset, Len}`** — **never** silent bytes, and **never** `cold` (cold zero-fills, which for a local-only read is the silent-zeros failure mode).
- The **engine** read path catches the typed error: remote present → invalidate + re-fetch the range through the **existing** BLAKE3-verified hydrate path (`ensureAndReadFromLocal`; the fresh `Hydrate` supersedes the corrupt interval by version) → return healed bytes; local-only (`!HasRemoteStore()`) → **fail closed** with `ErrIntegrityCheckFailed` (maps to NFS3ERR_IO). No parallel fetch; the existing cold/hydrate machinery is reused verbatim.

## The default (writeback) path is unchanged

When `VerifyReads` is off, the guard (`s.verifyReads.Load()`) is false and control goes straight to the original raw `seg.fd.ReadAt` — byte-identical, no extra pread/CRC/allocation (only an atomic-bool load). A unit test corrupts a byte and asserts the off path returns the raw corrupt byte with no error (proving it took no covering-record read).

## Tests (real journal store, run under `-race`)

- **remote-backed + verify on** — a warm read of a byte corrupted on disk after upload **self-heals**: re-fetches from remote and returns the correct bytes.
- **local-only + verify on** — the warm read **fails closed** (`ErrIntegrityCheckFailed`), never zeros/garbage.
- **verify off (writeback default)** — the read returns raw (unverified) bytes, documenting the fast path does not verify.
- journal-level: verify-on detection returns `*CorruptRangeError` with `cold=false`.

`go test ./pkg/block/... -race` → 880 passed. Lint/vet/integration build clean.

Part of the Wall-A/three-wall integrity follow-up (does not close it).

> Note: also drops two `FSStoreOptions` knobs (`rollup_workers`/`stabilization_ms`, removed by #1794) still set by the clone local-only test — the engine test package did not build on develop without this.

### DIFF
--- a/docs/guide/durability.md
+++ b/docs/guide/durability.md
@@ -72,6 +72,62 @@ absent these are honored unchanged:
   [Configuration → require_durable_commit](configuration.md#require_durable_commit)
   for the full semantics.
 
+## Read integrity: per-read verification & self-heal
+
+Warm reads (bytes served from the local journal without a remote round-trip) are
+verified **per durability tier** — there is no separate knob, the tier you pick
+above decides it:
+
+| Tier | Warm-read check | On corruption |
+|---|---|---|
+| **`writeback`** | none (raw fast read) | n/a — integrity comes from startup-recovery CRC + remote cold-fetch BLAKE3 |
+| **`local`** *(default)* / **`remote`** | per-record **CRC32** on every warm read | self-heal (remote) or fail closed with **EIO** (local-only) |
+
+On the durable tiers, every warm read re-reads the covering journal record and
+checks its stored CRC32 before returning the requested bytes:
+
+- **Remote-backed share** — on a CRC mismatch the range is **self-healed**: the
+  covering chunk is re-fetched from the remote (S3) block store, which is
+  BLAKE3-verified on the way in, re-hydrated into the local journal, and the read
+  returns the *correct* bytes. The remote is the source of truth; the local tier
+  is a cache.
+- **Local-only share** — there is no good copy to heal from, so the read **fails
+  closed** with `EIO` (`NFS3ERR_IO` / `STATUS_DATA_CHECKSUM_ERROR`). It never
+  returns silently-wrong or zero-filled bytes.
+
+This catches on-disk corruption of a valid segment that happens *after* startup
+recovery (bit rot, or a bug mutating cached segment bytes) — the case the
+startup-recovery CRC and the remote cold-fetch BLAKE3 don't cover on their own.
+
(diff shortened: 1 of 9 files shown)


marmos91:
Addressed the review:

- **FileID not covered by CRC (index.go)** — real gap, fixed. `verifiedRead` now checks the covering record actually belongs to the requested FileID (`string(rec.fileID) != string(id)` → `CorruptRangeError`) before trusting its payload, so a flipped FileID byte or a stale `recOff` mapping can no longer return another file's bytes silently. Added `TestVerifiedReadDetectsFileIDCorruption` (flips a FileID byte the CRCs don't cover, asserts it's caught, never cold).
- **Misleading "record CRC mismatch" message (record.go)** — generalized to "record integrity check failed" since verification also fails for structural reasons (bad magic/header CRC, truncated body), not only payload CRC.
- **Whole-record allocation on a small verified read (index.go)** — this is the documented read-amplification cost, on the durable tiers only. `readRecordAt` already rejects `PayloadLen > maxPayload` before allocating, and a record is one WriteAt (bounded by the protocol max write, ~1 MiB), not an unbounded slab — so no OOM path. Left as the accepted amplification the feature is opt-in-per-tier precisely to avoid on the fast default.

PR #1798, opened 2026-07-19
perf(journal): overlap within-file carve block commits (#1717)

### LINKED ISSUE #1717: Carver throughput optimization (FastCDC + BLAKE3 dominate the ingest pipeline)
**Gated on the journal switchover (#1716/#1692) merging first** — the switchover changes the carve invocation path, so optimize against the merged code, not the branch.

## Motivation (measured, new journal path, develop)
Decomposed the ingest→carve→commit pipeline (8 MiB seq, real badger, nil sealer, mem remote, Apple Silicon):

| stage | 1 writer | 8 writers |
|--|--:|--:|
| Ingest only (`WriteAt`→segment) | 3928 MB/s | 7508 MB/s |
| + Carve (FastCDC+BLAKE3, no commit) | 356 MB/s | 2246 MB/s |
| + Carve + commit (badger) | 209 MB/s | 1174 MB/s |

**The carver is ~90% of the pipeline cost** — raw ingest is nearly free (memory-bandwidth append); carve drops throughput ~11× single-thread. The metadata commit (`db.Sync`) is the *smaller* contributor for bulk writes. Carve scales ~6.3× across 8 cores (CPU-bound, shards cleanly).

## Caveats before chasing this
- **Measure-don't-assume / profile the treatment, not the mechanism** (see the perf-attempts ledger). Carve at 2.2 GB/s parallel is already an order of magnitude above S3 upload bandwidth, so it does NOT gate WAN uploads — only a very fast local sink. Confirm carve actually gates a real target before investing.
- **arm64-inflated:** the lukechampine/blake3 lib is AVX-512-only → generic fallback on Apple Silicon. On amd64 servers (AVX-512) BLAKE3 is much faster, so the Mac number understates production. **Re-profile on the amd64 bench VM first** — the win may be smaller than the Mac number implies.
- nil sealer here; real AEAD sealing adds further carve cost (so the carver's share is even larger in prod).

## Candidate levers (profile first, ranked hypothesis)
- **BLAKE3 SIMD on arm64** (NEON path) — only helps local dev; amd64 already fast. Low prod value.
- **FastCDC tuning** — larger Min chunk = fewer boundary evaluations (#1569 chunk_size knob already exists); the gearhash rolling-hash over every byte is the CDC cost.
- **Sealer/AEAD cost** — measure its share once wired.
- Parallelism is already good (per-shard); not a lever.

## Acceptance
A cpuprofile of carve on the amd64 VM showing the actual hot function split (FastCDC gearhash vs BLAKE3 vs pack), and a lever only if it moves a measured, carve-gated workload. Do NOT ship a micro-optimization that the profile doesn't justify.

marmos91:
## Why

Closes #1717 / #36 — the single-file sequential-write throughput lever.

A single large file's carve committed **one block at a time**: `carveRun`'s flush closure called `sink.CommitBlock` (PutBlock to S3 + metadata commit) synchronously, then flipped, before the next block was flushed. Consecutive blocks therefore uploaded serially at ~one S3 PUT RTT each (~9 MiB/s) regardless of link bandwidth. The per-file carve fan-out (#1785) does nothing for the common single-file seq-write case, because one large file is a single carve goroutine.

## What (Option A from the design)

Packing stays **sequential** (the single-buffered chunker reads the run in order). Only the `CommitBlock` (upload + commit) of successive packed blocks now overlaps, via a bounded per-run worker pool owned by the journal:

- New journal `Config.CarveUploadConcurrency` (default 8, falls back via `withDefaults`). Peak carve RAM is `window x (CarveBlockSize + one overhang chunk)` — kept modest.
- Each packed block gets its **own** buffer (from the arena pool) so its bytes stay live while its upload runs concurrently with the next block's packing.
- A small ordered-completion chain (`carve_dispatch.go`) applies the synced flips in submission/watermark order.

New file `carve_dispatch.go` holds the `carveDispatcher`; `carveRun` is rewritten to pack → `submit` → `wait`.

## How the three critical invariants are preserved

1. **Commit strictly precedes flip, per block.** A worker calls `flipUpTo` only after its own `CommitBlock` returned `nil`. A record flips to `synced` only once its block is durable remotely — unchanged from the sequential path, just moved into the worker.

2. **Flips apply in watermark order even if a later block's upload finishes first.** Each worker waits on its predecessor's completion channel (`prev`) before flipping, so `flipIdx` advances monotonically. A later block whose upload finishes first still blocks on the chain until every earlier block has flipped. Test `TestCarveFlipsInWatermarkOrder`: with the earliest block held and every later block committed, **no** record flips until the earliest block is released.

3. **Dedup visibility restricted to committed state.** `engineDeduper.IsChunkDurable` reads the synced-hash store, which is only written inside `CommitBlock`'s metadata transaction (after PutBlock). Packing stays sequential and consults that same committed oracle, so a block being committed concurrently has **not** yet marked its hashes durable — block K can never observe block K-1's uncommitted hash as durable. Worst case a duplicate chunk is re-packed, which the content-addressed commit collapses to a no-op. No optimistic in-flight dedup was added.

Mid-run failure: the first `CommitBlock` error records the error and stops the watermark at that block; every later worker sees `proceed=false` and skips its flip. Already-uploaded later blocks become GC-reclaimable orphans, matching the existing PutBlock-first semantics. In-flight uploads drain (no forced cancel of earlier, still-committing blocks).

## Tests (deterministic, `-race`)

`carve_dispatch_test.go`, a gated fake sink with per-block release:
- `TestCarveConcurrencyBounded` — up to `window` (and never more than) CommitBlocks run concurrently.
- `TestCarveFlipsInWatermarkOrder` — flips apply in watermark order; holding the earliest block blocks all flips even though later blocks committed (also covers the crash-safety "no record flips whose block hasn't committed").
- `TestCarveConcurrentCommitErrorStopsWatermark` — a mid-run commit error surfaces and stops the watermark; blocks before the failure flip, the failed block's records stay dirty.

All existing carve/journal tests pass unchanged.

## Verification

- `go test ./pkg/block/... -race` — 884 passed.
- New concurrency tests pass `-count=3 -race`.
- `gofmt -s`, `go vet ./...`, `golangci-lint run ./pkg/block/journal/...` (0 issues), `go build -tags integration ./...` all clean.

### DIFF
--- a/pkg/block/journal/carve.go
+++ b/pkg/block/journal/carve.go
@@ -401,9 +401,12 @@ func (s *Store) carveRun(ctx context.Context, sh *shard, id FileID, run []interv
 		}
 	}
 
-	if packErr != nil {
-		// Abandon the half-packed block (return its slot/buffer) and drain the
-		// blocks already in flight; do not advance the watermark to the run end.
+	if packErr != nil || disp.aborted() {
+		// A read/dedup error (packErr) or an in-flight commit failure (aborted)
+		// ends the run. Abandon the half-packed block (return its slot/buffer) and
+		// drain the blocks already in flight, but submit nothing more: advancing
+		// the watermark or committing the tail past a failure only adds orphan
+		// uploads. disp.wait returns the commit error in watermark order.
 		disp.discard(arenap, arena)
 		if err := disp.wait(); err != nil {
 			return err
--- a/pkg/block/journal/carve.go
+++ b/pkg/block/journal/carve.go
@@ -21,11 +21,12 @@ var carveScratchPool = sync.Pool{New: func() any {
 	return &b
 }}
 
-// carveArenaPool recycles the arena backing the pending chunk copies handed to the
-// sink. Both production sinks consume CarveChunk.Data synchronously inside
-// CommitBlock (localBlockSink reads only len; engineBlockSink seals/frames into its
-// own buffer before returning) — neither retains it — so the arena is safe to reuse
-// once flush's CommitBlock returns and pending is cleared.
+// carveArenaPool recycles the per-block arenas backing the pending chunk copies
+// handed to the sink. Both production sinks consume CarveChunk.Data synchronously
+// inside CommitBlock (localBlockSink reads only len; engineBlockSink seals/frames
+// into its own buffer before returning) — neither retains it — so a block's arena
+// is safe to return to the pool once its CommitBlock has returned. Each concurrent
+// block owns a distinct arena so overlapping commits never share backing bytes.
 var carveArenaPool = sync.Pool{New: func() any {
 	var b []byte
 	return &b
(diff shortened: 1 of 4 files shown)


marmos91:
Addressed both Copilot review comments in 7d99cc83:

- **carve.go tail-on-abort**: the post-pack path now takes the discard+drain branch on `disp.aborted()` too (not just a pack-side error), so a commit failure no longer falls through to `flush(runEnd)`. No tail block or run-end watermark is submitted past a known failure.
- **carve_dispatch.go early abort**: `d.abort` is now set the moment `CommitBlock` returns non-nil, before the worker reaches the head of the flip chain, so packing stops promptly instead of submitting more soon-to-be-orphan blocks. `firstErr` is still recorded in the chain (watermark order), so the earliest-watermark failure is the one returned — the crash-safety ordering is unchanged; these are cost/orphan reductions only.

All carve/journal tests pass `-count=3 -race`; full `./pkg/block/...` 884 pass `-race`; golangci-lint 0 issues; `go build -tags integration ./...` clean.

Issue #1801, opened 2026-07-19
Upgrade from v0.26.0 silently loses all local block data (journal switchover has no legacy-format reader)


marmos91:
Empirically confirmed on a Linux VM: upgrading a persistent deployment from **v0.26.0** to **develop** (post journal-switchover) causes **silent, total loss of file content on every share type** — local-only *and* S3-remote-backed.

## Behaviour
- Server starts cleanly, no error/warning/migration prompt.
- Control-plane DB + badger metadata open forward-compatibly (users, stores, shares, names, sizes, mtimes all survive).
- **Every file reads back as all-zeros of the correct length.**

## Root cause
v0.26.0 stores local block data as `blobs/*.blob` + `logs/<share>/*.log`. develop's journal recovery (`pkg/block/journal/store.go` `Open`) scans only for `journal/*.seg`, finds none, and **silently initialises an empty journal**, ignoring the on-disk legacy data. Remote-backed shares also serve zeros: the bytes are still in S3 (GC marks live hashes) but develop does not cold-fetch them by manifest on an empty local journal (the dropped `readLocalByHash` path).

The old bytes are NOT deleted — they remain on disk and are recoverable by a migration.

## Fix (staged)
1. **Guardrail (release-gating):** detect a pre-journal `blobs/`+`logs/` layout at local-store open and refuse to silently empty-init — fail loud so no data is served as zeros.
2. **Remote-backed auto-migrate:** seed cold intervals from the surviving metadata manifest so reads hydrate from S3.
3. **Local-only migrate:** read the legacy `blobs/`+`logs/` format and re-ingest into the journal.

Release must not ship without at least (1).

marmos91:
Follow-up (remote auto-migrate): PR #1803 (held, draft) makes drained remote-backed upgrades smooth (VM byte-identical), but before it can merge it needs a coverage check so un-carved data fails loud instead of reading zeros. Plan: (1) defer the legacy-dir archive until after the check; (2) file-level coverage check (manifest covers each file's size — needs metadata-store file enumeration + conformance); (3) VM re-verify both paths. The guardrail already keeps upgrades safe in the meantime.

marmos91:
Fixed by #1831 (async log-only migration for uncompacted local-only pre-journal shares) on top of the #1802 guardrail and #1803 remote-backed migration.

Issue #1465, opened 2026-06-30
e2e: remaining product-feature gaps after suite revival (#1456)


marmos91:
Tracking the e2e failures that are **genuine product gaps** (not test/harness drift), found by debugging the revived suite on a real Linux kernel-mount VM. Test/harness clusters are already fixed in #1462 (merged), #1463, #1464.

Each item below has a confirmed root cause; each is real product engineering, not a test edit.

## Cross-protocol byte-range lock enforcement
`TestCrossProtocolLocking` / `TestCrossProtocolLockingByteRange` (XPRO-01, XPRO-03).
After fixing identity/permission setup (NFS+SMB both as uid0), the locks still don't conflict: an exclusive NFS lock does **not** block an SMB exclusive/write lock (`Expected error with "lock would block" ... got nil`). The unified lock manager isn't enforcing cross-protocol conflicts. XPRO-02 (lease break) and XPRO-04 (data integrity) pass.

## NFSv4 byte-range locking returns EIO
`TestNFSv4Locking` / `TestNFSv4BlockingLock`: `LockFileRange` → `remote I/O error` (EIO) instead of acquiring the lock or returning a proper lock status. NLM/NFSv4 lock path.

## NFSv4.2 user xattr not honored
`TestNFSv42XattrRoundTrip`, `TestNFSv42XattrPersistAcrossRestart`, `TestCrossProtocolXattrParity`: `setfattr user.* ` → `Operation not supported`. SETXATTR/GETXATTR handlers exist (`internal/adapter/nfs/v4/handlers/{set,get}xattr.go`, `register_v42.go`), so this is a negotiation/advertisement gap (server not advertising xattr support, or the mount negotiating a lower minor version).

## NFSv4.2 CLONE returns wrong data
`TestNFSv42Clone`: `clone content must equal source content` / `clone tail must keep cloned content` — CLONE produces incorrect content.

## NFSv4.2 sparse (SEEK_HOLE / DEALLOCATE) wrong
`TestNFSv42Sparse`: `hole should not start before the punch` (hole offset 0 < 262144) — punch-hole/seek-hole semantics off.

## Cross-protocol errno conformance
[...]

PR #1803, opened 2026-07-19
feat(block): smooth remote-backed upgrade off the pre-journal local layout

marmos91:
## Stacked on #1802

This builds on the guardrail branch `fix/1801-legacy-upgrade-guardrail` (PR #1802), which is **not yet merged**. The base is `develop`, so until #1802 merges this PR shows both commits; rebase onto develop once #1802 lands.

## What

#1802 made a legacy (pre-journal `blobs/`+`logs/`) local dir **fail loud** rather than silently serve zeros. This PR lets a **remote-backed** share upgrade smoothly instead:

- On startup, a remote-backed share whose local dir carries the legacy layout **archives** `blobs/` -> `blobs.pre-journal-backup` and `logs/` -> `logs.pre-journal-backup` (non-destructive), so the journal opens clean.
- It then **seeds a cold interval per FileChunk manifest extent** so a subsequent read faults the bytes back in from the remote (BLAKE3-verified cold fetch). The metadata `File.Blocks` manifest survives the upgrade intact, so no bytes are lost.
- **Local-only shares keep the fatal guardrail** — their on-disk bytes are the sole copy and can't be recovered by an empty journal (that path is a separate follow-up).

## How it's wired

- `fs.FSStoreOptions.MigrateLegacyLayout` gates archive-vs-refuse at the exact site the guardrail checks (same `dir`, no path drift). Only remote-backed callers set it. `FSStore.MigratedFromLegacy()` reports whether an archive happened.
- `shares.ConfigureBlockStore` passes `remoteConfigured` into `CreateLocalStoreFromConfig`, and after `bs.Start` calls `SeedColdFromManifest` when the store migrated.
- `SeedColdFromManifest` is the **same cold-seed loop already used by snapshot restore** — extracted from `runtime.seedColdFromManifest` into the shares package and reused by both. No parallel fetch path, no hash-indexed local read.

## Idempotency / safety

After archiving, a second start finds no legacy dir and opens clean (reads still cold-fetch from the intact manifest). A partial/failed seed loses nothing — the remote bytes + manifest are untouched, worst case is a re-run.

Startup seeding is O(files); marked with a `ponytail:` note — lazy per-read seeding is the upgrade path if it ever bites a huge-file-count share.

## Tests

- **Unit (fs):** `TestNewWithOptionsMigratesLegacy` — archives the legacy dirs, opens clean, reports `MigratedFromLegacy`, and is idempotent across restarts.
- **Integration (shares):** `TestLegacyRemoteMigration_ReadsColdFromRemote` — full engine over a memory remote: write small + multi-MB files, carve to remote, wipe the local journal and plant a legacy layout, reopen with migration + cold-seed, and confirm every file reads back **byte-identical** (cold-fetched, not zeros) with the legacy dirs archived aside.
- VM byte-identical proof (v0.26.0 -> branch on the same dirs) recorded in the PR thread by the coordinator.

### DIFF
--- a/pkg/block/local/fs/fs.go
+++ b/pkg/block/local/fs/fs.go
@@ -31,6 +31,13 @@ type FSStoreOptions struct {
 	BackpressureMaxWait time.Duration
 	MaxLogBytes         int64
 	ChunkParams         chunker.Params
+	// MigrateLegacyLayout, when set, makes NewWithOptions archive a detected
+	// pre-journal blobs/+logs/ layout aside (instead of refusing to open) so the
+	// journal opens clean. Only a remote-backed share should set this: the
+	// authoritative bytes then live in the remote store and are re-materialized
+	// from the surviving metadata manifest via cold seeding. A local-only share
+	// must leave it false so the guardrail stays fatal.
+	MigrateLegacyLayout bool
 }
 
 // FSStore is the journal-backed local store. It embeds *journal.Store and
@@ -39,11 +46,20 @@ type FSStoreOptions struct {
 type FSStore struct {
 	*journal.Store
 
-	dir            string
-	maxDisk        int64
-	maxLogBytes    int64
-	fileChunkStore block.EngineFileChunkStore
-	durable        atomic.Bool
+	dir                string
+	maxDisk            int64
+	maxLogBytes        int64
+	fileChunkStore     block.EngineFileChunkStore
+	durable            atomic.Bool
+	migratedFromLegacy bool
+}
+
+// MigratedFromLegacy reports whether NewWithOptions archived a pre-journal
+// blobs/+logs/ layout aside while opening this store. The caller uses it to
+// trigger a one-time cold-seed of the journal from the surviving metadata
+// manifest so reads fault the bytes back in from the remote store.
+func (s *FSStore) MigratedFromLegacy() bool {
+	return s.migratedFromLegacy
 }
 
 var (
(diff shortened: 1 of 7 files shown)


marmos91:
## Coordinator audit + release recommendation

**Audit: correct for the common case.**
- The `seedColdFromManifest` → `shares.SeedColdFromManifest` extraction is a **pure behaviour-preserving move** (body byte-identical; snapshot restore now calls the shared func) — no restore regression, minimal new surface. Good reuse.
- Archive is non-destructive (`blobs/`→`blobs.pre-journal-backup`), gated opt-in so **local-only shares keep the fatal guardrail**. VM proof: v0.26.0→branch, 4 files, all sha256 **byte-identical** cold-fetched from S3. Rebased onto develop (guardrail #1802 already merged); build + fs/shares tests green, lint clean.

**One thing to confirm before this is unconditionally safe:** un-flushed local data at upgrade time. `SeedColdFromManifest` seeds only chunks that have a FileChunk manifest row. A ranged read for data that was written but never carved/uploaded (no row, not in S3) would seed nothing → **could read zeros silently** — the exact failure the guardrail prevents. The bytes are preserved in the `*.pre-journal-backup` dir either way (no destruction), but this edge case wants a test asserting *loud error, not silent zeros* for a not-fully-drained share.

**Recommendation:** the guardrail (#1802, merged) already makes the release **safe** (all upgrades fail loud, zero data loss). This PR makes drained remote-backed upgrades **smooth**. Since it's a same-day data-plane change touching the restore path with the un-drained edge unverified, my lean is **ship the tag with the guardrail now, land this as a fast-follow after a day of develop bake** — unless you want smooth remote upgrades in this tag and accept the drained-only caveat + release note. Your call; not merging autonomously.

marmos91:
**Held (draft) pending a coverage-check follow-up — not merging as-is.**

Verified byte-identical for drained remote shares, but there's a silent-zeros edge: for data written but never carved to S3 at upgrade time, the migration seeds cold reads only from the manifest, and the journal zero-fills uncovered ranges — so that data would read as zeros (it's preserved in the `.pre-journal-backup` archive, not destroyed). This is exactly what the guardrail (#1802) prevents, and it's the *realistic* case since v0.26.0 doesn't durably upload small datasets (#1465).

A payload-based check can't catch it: a fully-un-carved file has **zero** chunks, so it never appears in `EnumeratePayloads`. The safe version needs:
1. **Defer the archive** until after the check (open journal → seed → verify → *then* archive), so a failed check leaves the legacy dirs in place and a restart hits the guardrail (loud) instead of an empty journal (silent zeros).
2. A real **file-level coverage check** (manifest covers every file's recorded size) — which needs a file-enumeration method on `metadata.Store` across all backends + conformance.
3. VM re-verification of both the drained-happy-path and the un-drained-refuses-path.

The release (v0.27.0) is already safe via the guardrail, so this is a proper follow-up, not a rush. Tracked in #1801.

PR #1806, opened 2026-07-20
fix(smb): stop double-breaking parent dir leases on delete-on-close

### LINKED ISSUE #1701: Flaky: smb2.dirlease.unlink_*_initial_and_close double-break (lease_break_info.count==2) in smbtorture conformance

marmos91:
Fully fixes the SMB directory-lease defects behind #1701 (the flaky `smb2.dirlease.unlink_{same,different}_{initial,set}_and_close` conformance tests) and the create-heavy SMB3 stall. Every fix has a deterministic lock-layer test, and all four `unlink` dirlease tests now pass on the smbtorture suite (memory + badger-fs).

### 1. Break directory leases by parent lease key, not by client (`unlink_same_*`)
`OnDirChange` suppressed dir-lease breaks for the whole originating client (`lock.Owner.ClientID == originClientID`). Per MS-SMB2 §3.3.4.20 and the Samba object-store rule, a directory-content change must break **every** parent dir lease whose key differs from the change's parent lease key — including other-key leases held by the *same* client. On entry removal we now suppress by parent key only; the client exclusion stays for NFS delegations and add/rename.

### 2. Serialize RH dir-lease breaks (`unlink_different_*`)
Revoking a directory RH lease waits for the client's acknowledgment (the handle-caching bit). When one change breaks several dir leases (different parent keys → no lease spared), they must be delivered **one at a time**: send the first, wait for the ack, then the next. `OnDirChange` sent them all at once → the client saw `lease_break_info.count==2`. Now `OnDirChange` dispatches only the first and marks the rest `BreakDeferred`; each `AcknowledgeLeaseBreak` delivers the next.

### 3. Clear mirrored siblings on acknowledge (create-heavy stall)
A break mirrors its stage onto sibling records sharing a lease key without a wire notification, but `AcknowledgeLeaseBreak` cleared only the one record `findLeaseByKey` resolves. Mirrored siblings stayed `Breaking=true`, so `WaitForBreakCompletion` blocked until the force-complete timeout on every following op — the SMB3 fio stall. Now every record sharing the acked key is cleared.

### 4. Break parent dir leases when recycling a file
`RemoveFile`'s recycle (trash) branch returned before `notifyDirChange`, so on a `--enable-trash` share a delete left caching clients with a stale listing. `RemoveDirectory` already notifies; `RemoveFile` now does too. (Caught in review.)

## Tests
- `TestOnDirChange_BreaksOriginatorsOtherDirLease` / `_SuppressesParentKeyMatchedDirLease` — parent-key suppression (both verified red→green).
- `TestOnDirChange_SerializesMultipleDirLeaseBreaks` — one break immediate, the next on ack.
- `TestAcknowledgeLeaseBreak_ClearsMirroredSiblings` — the stall (verified red→green).
- `TestRemoveFile_RecycleNotifiesDirChange` — recycle notify (verified red→green).
- Full `pkg/metadata/lock` (race) + `pkg/metadata` + `internal/adapter/smb/handlers` green; **all four smbtorture `unlink_*` dirlease tests pass**.

Closes #1701.

### DIFF
--- a/internal/adapter/smb/handlers/close.go
+++ b/internal/adapter/smb/handlers/close.go
@@ -577,20 +577,16 @@ func (h *Handler) Close(ctx *SMBHandlerContext, req *CloseRequest) (*CloseRespon
 					h.purgeBlockStorePayload(ctx.Context, deleteTargetHandle, removedPayloadID, openFile.Path, "CLOSE")
 					h.restoreParentDirFrozenTimestamps(authCtx, deleteParentHandle)
 
-					// Break parent directory leases: deletion changes directory
-					// content. When docSetterKeysDiffer, clear the closer's
-					// parent key so no suppression applies — all dir leases break.
-					if docSetterKeysDiffer {
-						savedKey := openFile.ParentLeaseKey
-						savedHas := openFile.HasParentLeaseKey
-						openFile.HasParentLeaseKey = false
-						openFile.ParentLeaseKey = [16]byte{}
-						h.breakParentDirLeasesForContentChange(ctx, authCtx, openFile)
-						openFile.ParentLeaseKey = savedKey
-						openFile.HasParentLeaseKey = savedHas
-					} else {
-						h.breakParentDirLeasesForContentChange(ctx, authCtx, openFile)
-					}
+					// Removing the entry already broke the parent directory's
+					// leases: RemoveFile / RemoveDirectory notify the dir-change
+					// listener, which breaks every parent dir lease (except the
+					// originator / suppressed key) to None. The parent-key
+					// suppression was applied via the authCtx propagated above
+					// (closer's key when it matches the DOC-setter's key; none
+					// when they differ, so all dir leases break). Dispatching a
+					// second break here would re-break a lease the client may have
+					// already acked and re-cached in the meantime, sending a
+					// duplicate LEASE_BREAK for one logical change.
 
 					if h.NotifyRegistry != nil {
 						parentPath := GetParentPath(openFile.Path)
--- a/pkg/metadata/file_recycle_test.go
+++ b/pkg/metadata/file_recycle_test.go
@@ -6,6 +6,7 @@ import (
 	"testing"
 
 	"github.com/marmos91/dittofs/pkg/metadata"
+	"github.com/marmos91/dittofs/pkg/metadata/lock"
 	"github.com/marmos91/dittofs/pkg/metadata/store/memory"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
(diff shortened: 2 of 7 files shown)


marmos91:
Review pass caught a real regression: removing the redundant delete-path break exposed a latent `RemoveFile`/`RemoveDirectory` asymmetry — `RemoveFile`'s **recycle (trash) branch** returned before `notifyDirChange`, so on a `--enable-trash` share a file delete would no longer break the parent's dir leases (stale listing for a caching client). The old explicit break masked it.

Fixed at the root: `RemoveFile`'s recycle branch now notifies dir-change like `RemoveDirectory`'s already does. Added `TestRemoveFile_RecycleNotifiesDirChange` (confirmed it fails without the fix: 0 notifications). Both new tests are deterministic (verified red→green). gofmt/vet clean; full `pkg/metadata` + `pkg/metadata/lock` + smb-handlers suites green.

PR #1808, opened 2026-07-20
fix(journal): serialize per-file carve metadata commits (SSI conflict / EDEADLK)

marmos91:
## Problem

The within-file carve dispatcher (#1717, `CarveUploadConcurrency`) fires `CommitBlock` concurrently for a single file's carve run. Each commit re-projects `File.Blocks` via `ProjectManifestToBlocks` → `PutFile` on the **same File row**. Under badger's SSI, two overlapping read-modify-writes of one row abort as a transaction conflict; under real load the retry budget exhausts and the raw `ErrConflict` surfaces to the carver — the SMB/NFS client sees `EDEADLK` (`Resource deadlock avoided`) on create-heavy and seq-write workloads. This blocked the SMB3 write/metadata benchmark cells (all DittoFS write cells stalled).

## Root cause

Confirmed by a new reproduction test: 8 concurrent same-file commits produce 22 SSI conflicts. The File row is the *only* shared key — per-payloadID serialization drops conflicts to exactly zero, proving no other hot row.

## Fix

A fixed **striped mutex** (`carveCommitLocks`, 256 stripes) serializes a payloadID's metadata commit. Key properties:

- The `PutBlock` upload stays **outside** the lock, so successive blocks' uploads still overlap — the entire point of the #1717 dispatcher is preserved.
- Distinct files take distinct stripes and commit concurrently; a stripe collision only serializes two unrelated files' short commit txns (harmless).
- Fixed array bounds memory — a long-lived share carving many files never accumulates a per-file mutex the way a keyed map would.

This mirrors the existing house pattern (`PendingWritesTracker.GetFlushLock`, which serializes concurrent same-file NFS COMMITs for the same reason).

## Verification

- New regression test `TestLocalBlockSink_ConcurrentSameFileCommit_NoSSIConflict`: fails before (22 conflicts), passes after (0), green under `-race`.
- `go test -race ./pkg/block/engine/... ./pkg/block/journal/...` — 358 pass.
- `go test ./pkg/metadata/...` — 2648 pass.
- `go build ./...`, `go vet`, `gofmt` clean.

Unblocks re-running the SMB3 write/metadata benchmark.

### DIFF
--- a/pkg/block/engine/blocksink.go
+++ b/pkg/block/engine/blocksink.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"fmt"
 	"math"
+	"sync"
 
 	"lukechampine.com/blake3"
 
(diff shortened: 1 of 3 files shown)


marmos91:
Addressed Copilot's note: added a compile-time guard so a non-power-of-two `numCarveCommitStripes` fails the build (the mask in `forKey` assumes it). 568b1fe3.

Issue #1810, opened 2026-07-20
SMB adapter hot-reload briefly drops the TCP listener on share/store config change


marmos91:
## Symptom

Adding a metadata/block store or creating a share on a running dfs (via `dfsctl`/API) hot-reloads the SMB adapter, which **closes and reopens the port-12445 listener**. Any SMB client that connects during that ~few-second window gets a refused/closed connection — mount.cifs reports `Server abruptly closed the connection` / `mount error(112): Host is down`. Existing SMB connections are also dropped by the swap.

## Repro (on a fresh VM)

1. `dfs start` (control-plane secret + admin pass), `dfsctl login`, `dfsctl adapter enable smb --port 12445`, wait for port.
2. `dfsctl store metadata add` + `store block local add` + `store block remote add s3` + `dfsctl share create --name /bench ...`.
3. Immediately `mount -t cifs -o port=12445,guest,vers=3.0 //127.0.0.1/bench /mnt` → **fails 'Host is down'**.
4. Wait ~3s and retry the same mount → **succeeds** (NEGOTIATE/SESSION_SETUP/TREE_CONNECT all STATUS_SUCCESS).

The dfs log shows `SMB shutdown signal received error=context canceled` (old adapter torn down) followed shortly by a fresh listener + successful handshake — confirming a restart, not a crash (dfs PID is unchanged; SSI=0, PANIC=0).

## Impact

- Any config change (share/store add) momentarily kills all SMB mounts and drops live SMB sessions.
- Surfaced by the dfsbench SMB3 matrix: every dittofs SMB cell failed at mount because the harness mounts right after `share create`. Worked around in the harness with a bounded mount-retry (separate PR), but the underlying listener-drop is a real server-side defect.

## Fix direction

Keep the SMB TCP listener alive across a hot-reload — swap the handler/config without closing the accept socket — so config changes don't interrupt in-flight or new connections (NFS does not exhibit this).

marmos91:
Fixed in #1833: Port=0 now resolves to the adapter default before the unchanged-listener comparison, so a hot-reload no longer drops the TCP listener. Default ports are single-sourced in pkg/controlplane/models.

PR #1811, opened 2026-07-20
test(dfsbench): retry SMB mount across the adapter hot-reload window

marmos91:
## Problem

Every dittofs SMB3 benchmark cell failed at mount (`Server abruptly closed the connection` / `Host is down`, exit 32), so the SMB3 matrix produced zero dittofs results.

## Root cause

Not SSI, not a protocol bug — a **timing race**. The harness enables the SMB adapter and `waitPort`s it, then adds the metadata/block stores and creates the share. That share-create hot-reloads the running adapter, which briefly closes and reopens the port-12445 listener. The harness mounts *immediately* after share-create, landing in that ~few-second down-window. Verified manually: once the adapter settles, the exact same guest/vers=3.0 mount succeeds (full NEGOTIATE/SESSION_SETUP/TREE_CONNECT `STATUS_SUCCESS`).

The server-side listener-drop is tracked separately in #1810; this PR makes the harness robust to it.

## Fix

Retry the mount up to 8× (2s apart, ctx-aware) so it rides out the reload instead of failing the cell. Harmless for NFS (succeeds first try).

## Verification

Manual repro on the bench VM: local-only share mounts first try; a share created right after `adapter enable` fails once then succeeds on retry. Build/vet/gofmt clean.

### DIFF
--- a/internal/dfsbench/backend/dittofs.go
+++ b/internal/dfsbench/backend/dittofs.go
@@ -424,10 +424,14 @@ func dittofsMount(ctx context.Context, proto Protocol) (string, error) {
 	// gone). waitPort above only proves the listener was up right after `adapter
 	// enable`, not after the later share-create reload. Retry so the mount rides
 	// out the reload instead of failing the whole cell over a ~few-second gap.
+	const mountAttempts = 8
 	var err error
-	for attempt := 0; attempt < 8; attempt++ {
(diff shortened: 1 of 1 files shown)


marmos91:
Addressed Copilot: the loop no longer sleeps after the final attempt, so a genuine mount failure surfaces immediately instead of after an extra 2s.

Issue #1843, opened 2026-07-21
Field-deployment observations (tiered capacity / upgrade safety / disk-full) — seeking your perspective


raffaelegiustini-create:
## Field-deployment observations — would value your perspective

Hi @marmos91 — first, thank you for the incredibly fast turnaround on every issue we've filed; the root-causes and fixes have been excellent. This isn't a bug report — it's a set of **operational patterns** we've hit running dittofs on a real deployment (Hetzner Helsinki, 8 GiB box, ~60 GiB `pin` share, Cubbit DS3 remote), and we'd love your **consideration / how you see these** as the product hardens for demanding production use. Three themes:

### 1. Tiered capacity — "share larger than the local disk" (the core value prop for us)
This is the scenario our key-account cares most about: small local cache + effectively unlimited capacity on Cubbit. Today it doesn't hold yet — auto-eviction never reclaims the existing footprint. On v0.28.1 the negative-underflow is fixed and the evict **mechanism** works, but `disk_used` isn't seeded from the on-disk store at startup, so the tier grows unbounded to a full disk (details + suggested fix in #1527). This feels like the single biggest gate for the cloud-tiering promise.

### 2. Upgrade safety / reversibility
Upgrading a data-bearing box across a metadata-format change was risky in practice: the migration ran on the startup path and looked like a hang on a large store (#1622-class), and once the metadata was migrated, **rolling back to the prior binary left the share unreadable** (old build can't load the new metadata format) — we recovered only via the durable snapshots + a disable→enable dance. For production, operators will expect upgrades to be **safe, observable (progress), and reversible** (or at least clearly one-way with a guard). Curious how you're thinking about the upgrade/rollback contract.

### 3. Graceful behavior near disk-full
On a near-full disk the box wedged (startup blocking on S3, writes with no headroom), and there's currently no self-service reclaim path once eviction is unavailable (GC only reclaims true orphans; deletes sit in the recycle bin; snapshots hold blocks). A fail-fast/degraded-read mode + an operator "reclaim now" lever would help a lot.

### What's clearly solid
To be balanced — data integrity (post #1267/#1274/#1435), reads/streaming, GC correctness (#1433), and write throughput (now ~4–6× after #1567/#1717, ~2× off rclone on our path) are all in good shape and improving fast.

[...]

Issue #1527, opened 2026-07-01
Local-store disk_used_bytes underflows to negative → LRU eviction never fires, --local-store-size not enforced


raffaelegiustini-create:
## Local-store `disk_used_bytes` underflows to **negative** → LRU/TTL eviction never fires, `--local-store-size` not enforced

On `develop` (8e376d36 / aed3ae2), a share's local fs tier grows **unbounded to the physical disk** regardless of retention (`lru`, `ttl`, or `pin`) and regardless of `--local-store-size`. Root cause is visible in the metrics: the local-store used-bytes counter has gone **negative**.

### The smoking gun (`/metrics`, :9090)
```
dittofs_localstore_disk_limit_bytes{share="/drive"}  4.294967296e+10   # 40 GiB  ✅ cap set correctly from --local-store-size
dittofs_localstore_disk_used_bytes{share="/drive"}  -4.28867584e+08    # -0.43 GB  ❌ NEGATIVE
dittofs_localstore_evictions_total                   0
dittofs_localstore_evicted_bytes_total               0
dittofs_localstore_backpressure_total                0
```
Meanwhile the on-disk block dir actually holds **~61 GB**:
```
$ du -sh /var/lib/dittofs/blocks
61G
```

So `disk_limit` is correct (40 GiB), but `disk_used` has **underflowed below zero**. `ensureSpace` compares used vs limit; since used (−0.43 GB) is far under the 40 GiB limit, it concludes there's plenty of room → **`lruEvictOne` is never called** (`evictions_total = 0`), no backpressure, and no `ErrDiskFull` — the tier just keeps growing until the physical disk fills.

### Repro
1. `dfsctl share edit /drive --retention lru --local-store-size 40GiB`
2. `systemctl restart dfs` (confirmed retention=lru, disk_limit=40 GiB after restart)
3. Local tier already ~61 GB (well over the 40 GiB cap). Wrote a fresh 1 GB file over NFS.
4. Result: `du(blocks)` grew 65.5 GB → 66.2 GB, `evictions_total` stayed **0**, disk free kept dropping. No eviction, no ErrDiskFull.
5. Same under `pin`: at 61 GB local with a 40 GiB cap, reads/writes do **not** hit `ErrDiskFull` (contradicting the "pinned bounded tier fails ErrDiskFull once the working set exceeds it" warning in `service.go`).

### Why it matters
[...]

PR #1846, opened 2026-07-22
fix(journal): default local-store cap + background dead-ratio GC (#1844)

### LINKED ISSUE #1844: Uncapped local store grows unbounded → disk-full wedge (default cap + dead-ratio GC)
## Summary

An **uncapped** local store (writeback tier with no `--local-store-size`) grows its on-disk
footprint **without bound** under sustained random overwrites, until the physical disk fills and
the node wedges (unrecoverable — SSH userspace starved, hard reboot hangs). This is the
synthetic-load reproduction of the field observation in #1843 (theme 1) and the same root family
as #1527 (LRU never fires): when the cap is unset, `ensureSpace` is a no-op, so dead append-log
records are never reclaimed.

Discovered during the DittoFS benchmark matrix: the `dittofs-s3-writeback` (uncapped) cell running
`rand-write-4k` over a 1 GiB file wedged the bench VM hard.

## Root cause (code)

The journal write path gates capacity through `ensureSpace` (`pkg/block/journal/reclaim.go`):

```go
func (s *Store) ensureSpace(ctx context.Context, needed int64) error {
	if s.cfg.MaxLocalBytes <= 0 {
		return nil          // <-- no cap set => no eviction, no backpressure
	}
	for s.diskBytes.Load()+needed > s.cfg.MaxLocalBytes { ... evict/backpressure ... }
}
```

Every `WriteAt` **appends** a new record; an overwrite makes the previously-covering record *dead*
but does not free it. Reclaiming dead bytes only happens through the eviction/GC path, which
`ensureSpace` drives — and that path is entirely skipped when `MaxLocalBytes <= 0`. So an uncapped
store accumulates dead records forever.

## Measured repro (unit-level, deterministic, no VM)

A local test (`pkg/block/journal`, `Config{}` = uncapped, 128 MiB address space, random-4k
overwrites, no drain) isolates *which* resource is unbounded:

```
writes=0        diskBytes=0.00GiB  unsynced=0.00GiB  segs=16  liveIvs=1      heapAlloc=0MiB
writes=40000    diskBytes=0.15GiB  unsynced=0.09GiB  segs=16  liveIvs=23219  heapAlloc=5MiB
writes=120000   diskBytes=0.46GiB  unsynced=0.12GiB  segs=17  liveIvs=31974  heapAlloc=2MiB
writes=200000   diskBytes=0.77GiB  unsynced=0.12GiB  segs=19  liveIvs=32709  heapAlloc=3MiB
writes=400000   diskBytes=1.54GiB  unsynced=0.12GiB  segs=22  liveIvs=32768  heapAlloc=3MiB
```

- **`diskBytes` grows linearly and unbounded** (1.54 GiB written for a 128 MiB live footprint =
  ~12x amplification, still climbing) — the dead records never get reclaimed.
- **`liveIvs` plateaus at 32768** (= address space / 4 KiB) — the in-memory index is bounded.
- **`heapAlloc` stays flat (~3 MiB)** — RAM is *not* the leak.
- **`unsynced` plateaus** — the dirty-byte counter is bounded.

So the unbounded resource is **on-disk segment bytes**. On the bench VM this filled the volume and
wedged the box; on the field 8 GiB box (#1843) it manifests the same way.

## Proposed fix (keeps the fast append path)

The append-only write path is the performance win and must stay; the bug is *no bound + no
reclaim* when uncapped. Three coordinated parts, none touching the common-case hot path:

1. **Always have a ceiling** — default `MaxLocalBytes` from the store volume's free space when
   unset (a fraction of the filesystem at the mount), so `ensureSpace` always has a bound. This is
   the #1527/#1843 fix. Below the cap the writer runs at full append speed.
2. **Dead-ratio-triggered background GC/compaction** — drive compaction off per-segment `deadBytes`
   (segment ≥ X% dead → rewrite its live remainder, free the old), *independent of whether a cap is
   set*, so `diskBytes` tracks `live × small constant`. Standard log-structured GC; append speed is
   untouched because it's background.
3. **Backpressure at the ceiling** — when the writer genuinely outruns GC, `ensureSpace` throttles
   near the cap (the backpressure path exists; it just needs a cap to engage). Degrades to "slower
   writes" instead of "wedged box."

Net: normal/bursty writes keep full append speed (GC keeps the cap far away); only a sustained
overwrite-outrunning-GC workload is throttled — which beats an unrecoverable wedge.

The measured repro converts directly into a **regression test**: same workload, assert `diskBytes`
stays bounded under the (defaulted) cap.

## Relationship to existing issues

- #1843 theme 1 ("tier grows unbounded to a full disk; `disk_used` not seeded") — this is the
  mechanism + a deterministic repro.
- #1527 (closed — negative underflow fixed) — same failure family (eviction never reclaims the
  existing footprint); the remaining gap is "no cap set ⇒ no reclaim at all".

marmos91:
Fixes #1844.

## Problem
An uncapped local store (`MaxLocalBytes` unset) left the write-path capacity gate (`ensureSpace`) a permanent no-op, so dead append-log records left behind by overwrites were never reclaimed. On-disk usage grew linearly with total writes (not live data) until the disk filled and the node wedged. Confirmed by measurement: a random-4k overwrite workload over a 128 MiB live footprint grew `diskBytes` past 1.5 GiB while the in-memory index and Go heap stayed flat.

## Fix
- **Default cap from free space.** `Open` derives a soft `MaxLocalBytes` (80% of the store volume's free space) when unset, so `ensureSpace` always has a ceiling. A statfs probe failure degrades to the old unbounded posture rather than failing `Open`.
- **Background dead-ratio GC.** A loop started by `Open` periodically runs the existing `GC` pass, repacking high-dead-ratio segments so on-disk bytes track live bytes regardless of whether a cap is configured. Cancelled and joined in `Close` before segment files are closed.
- The existing write-path backpressure now engages once a cap is set (safety valve when writes outrun GC).

## Tests
- `TestOpenSetsDefaultLocalCapWhenUnset` — guards the root cause: an unset cap must now be defaulted.
- `TestGCReclaimsDeadOverwrites` — overwrites a 256 KiB footprint ~32 MiB worth, asserts a GC pass reclaims the dead space and on-disk bytes stay bounded to a few segments (fails hard on the pre-fix unbounded behaviour).

`go test -race ./pkg/block/journal/ ./pkg/block/engine/` green; gofmt/vet clean.

Relates to field report #1843 (theme 1) and #1527.

### DIFF
--- a/pkg/block/journal/bounded_growth_test.go
+++ b/pkg/block/journal/bounded_growth_test.go
@@ -1,0 +1,68 @@
+package journal
+
+import (
+	"context"
+	"testing"
+)
+
+// TestOpenSetsDefaultLocalCapWhenUnset guards the root cause of the disk-full
+// wedge: an unset MaxLocalBytes used to leave the write-path capacity gate a
+// permanent no-op, so dead records were never reclaimed and disk usage grew
+// without bound. Open must now derive a soft cap from the volume's free space.
+func TestOpenSetsDefaultLocalCapWhenUnset(t *testing.T) {
+	s, err := Open(t.TempDir(), Config{}, newFakeRemote(), SystemClock())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = s.Close() }()
+
+	if s.cfg.MaxLocalBytes <= 0 {
+		t.Fatalf("expected Open to default MaxLocalBytes from free space, got %d", s.cfg.MaxLocalBytes)
+	}
+}
+
+// TestGCReclaimsDeadOverwrites is the wedge regression test. It overwrites a
+// tiny live footprint thousands of times so sealed segments fill with dead
+// (superseded) records — the exact shape that grew on-disk bytes without bound
+// before the fix — then asserts a GC pass reclaims that dead space, leaving
+// disk usage tracking live bytes rather than total writes.
+func TestGCReclaimsDeadOverwrites(t *testing.T) {
+	// SegmentSize at the 1 MiB floor so overwrites seal many segments; single
+	// shard is deterministic; a generous explicit cap keeps eviction/backpressure
+	// out of the way so this isolates the dead-ratio repack path.
+	const segSize = 1 << 20
+	cfg := Config{SegmentSize: segSize, ShardCount: 1, MaxLocalBytes: 1 << 30}
+	s, err := Open(t.TempDir(), cfg, newFakeRemote(), SystemClock())
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = s.Close() }()
+	ctx := context.Background()
+
+	const blk = 4096
+	const slots = 64 // 256 KiB live footprint, rewritten in place
+	data := make([]byte, blk)
+	for i := range 8000 { // ~32 MiB appended for 256 KiB of live data
+		off := int64(i%slots) * blk
+		if err := s.WriteAt(ctx, "f", off, data); err != nil {
+			t.Fatalf("WriteAt #%d: %v", i, err)
+		}
+	}
+
+	before := s.diskBytes.Load()
+	if _, err := s.GC(ctx, GCOptions{}); err != nil {
+		t.Fatalf("GC: %v", err)
+	}
+	after := s.diskBytes.Load()
+
+	if after >= before {
+		t.Fatalf("GC reclaimed nothing: before=%d after=%d", before, after)
+	}
+	// ~32 MiB was appended for 256 KiB of live data. After repack the surviving
+	// bytes must track live data (a few segments), not total writes. A handful of
+	// segments is a generous bound that still fails hard if dead records are left
+	// unreclaimed (the pre-fix unbounded-growth behaviour).
+	if bound := int64(4 * segSize); after > bound {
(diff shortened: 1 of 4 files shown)


marmos91:
Addressed the review comments (pushed):

- **Windows build** — `diskFreeBytes` moved behind build tags (`statfs_unix.go` uses `unix.Statfs`, `statfs_windows.go` uses `GetDiskFreeSpaceEx`); `GOOS=windows` now compiles.
- **Comment accuracy** — `Config.MaxLocalBytes` no longer references the nonexistent `defaultMaxLocalBytes`; it describes the free-space default (`defaultMaxLocalBytesFreeFraction`).
- **Quiet shutdown** — the background GC loop now also suppresses `errClosed` (not just `context.Canceled`) so a pass racing `Close` doesn't log a spurious warning.
- **Robust test** — `TestOpenSetsDefaultLocalCapWhenUnset` skips when the free-space probe is unavailable, matching `Open`'s documented degrade-to-unset behaviour.
- Re: "`for i := range 8000` doesn't compile" — that's valid Go 1.22+ (repo builds on 1.25 and it passed), but switched to a plain counted loop to keep it unambiguous.

Both platforms build, `go test -race ./pkg/block/journal/ ./pkg/block/engine/` green, gofmt/vet clean.

Issue #1828, opened 2026-07-21
Unify the metadata-store backends (SQL family + KV family) behind a shared base — JuiceFS as prior art


marmos91:
## Summary

DittoFS ships **four fully separate metadata backends** (`pkg/metadata/store/{sqlite,postgres,badger,memory}/`), each re-implementing the entire `metadata.Transaction` contract (`pkg/metadata/store.go:258`). The two SQL backends are **near-verbatim copies** of each other (~1,400 lines apiece, ~95% shared modulo `?n`↔`$n` and `mapDBError`↔`mapPgError`). The two KV-shaped backends (badger, memory) share a large conceptual surface but no code.

This issue proposes collapsing the four impls into **two engine families behind one shared base**, exactly as JuiceFS does with `sql.go` / `redis.go` / `tkv.go` behind `baseMeta`. Estimated net deletion: **~1,300–1,600 lines** with no behavioural change, gated by the existing `storetest` conformance suite at every step.

> **State note (current develop):** the shared-helper split is already underway. `pkg/metadata/store/internal/{quota,sqlcodec,sqlstat,txretry}/` are **all populated** — `quota` (#1193/#1787), `sqlcodec`, `sqlstat` (#1826, statfs builder), and `txretry` (#1827, the SQL transaction backoff). So the retry/backoff skeleton and statfs builder are **already shared** between sqlite and postgres — "P0" below is effectively done. This issue tracks the remaining, larger convergence (the CRUD bodies → one SQL impl, and the KV family).

---

### A.1 sqlite vs postgres `transaction.go` — per-method classification

Files: `pkg/metadata/store/sqlite/transaction.go` and `pkg/metadata/store/postgres/transaction.go` (~1,400 L each). A normalized diff (rewriting `$n`→`?n`, `mapPgError`→`mapDBError`, `pgx.ErrNoRows`→`sql.ErrNoRows`, type/receiver renames) leaves only **~40 genuinely-differing source lines**, all concentrated in 4 methods. Everything else is identical.

| Method | Class | Notes |
|---|---|---|
| `GetFile` | **(a) identical** | same `blockRefsAggExpr` fold, same `sqlcodec.FileRowToFileWithNlinkAndBlocks` |
| `DeleteFile` | (a) identical | same pre-read + delta tracking |
| `GetChild` / `SetChild` / `DeleteChild` | (a) identical | same `ON CONFLICT ... DO UPDATE` upsert |
[...]

PR #1849, opened 2026-07-25
perf(metadata): narrow the WRITE hot path off GetFile+PutFile

### LINKED ISSUE #1819: perf: sqlite metadata backend ~19x slower than badger on random writes (single-writer + per-op amplification)
Fresh SMB3 benchmark (medium, rand-write-4k IOPS): badger ~7,300-8,500, postgres ~1,090-1,765, **sqlite ~340-390** (~19-22x below badger). Block store identical across all three — only the metadata backend differs; each 4k write drives a metadata write.

## Root cause (not a durability confound)
Both sqlite (WAL + synchronous=NORMAL, no per-commit fsync) and badger (relaxed, SyncWrites=false) are effectively relaxed, so the gap is NOT sqlite-fsync-vs-badger-relaxed.

1. **Single-writer serialization (dominant).** pkg/metadata/store/sqlite/store.go:114 bounds the pool to one connection (SetMaxOpenConns(1)) — SQLite is architecturally single-writer. Concurrent rig writers fully serialize (SQLITE_BUSY backoff in transaction.go:116-173). Badger is a concurrent LSM; disjoint-key writers never block. Explains badger >> postgres (concurrent MVCC, IPC cost) >> sqlite (zero write concurrency).
2. **Pure-Go SQLite VM** (glebarez/go-sqlite; cgo avoided due to shared GORM driver registration, store.go:13-16) — high fixed per-statement cost.
3. **Per-op write amplification.** A data write does SELECT + a full 20-column UPDATE inodes (transaction.go:257-280) + a file_block_refs DELETE+INSERT manifest rewrite (transaction.go:449-453). Attr-only writes already skip the manifest via the BlocksDirty gate (#1715), but data writes pay it.

## Levers
- **Real fix (beats the single-writer floor):** a write coalescer / group-commit batching pending inode updates into one transaction on the single connection — amortizes BeginTx/Commit + WAL frame over many ops.
- **Trim per-op work (needs a targeted path, not a one-liner — PutFile is generic):** an attr-only PutFile fast path that narrows the UPDATE to size/mtime/ctime and skips the manifest rewrite when the chunk list is unchanged on in-place overwrite.
- **Not pursuing:** cgo driver (intentionally avoided); pragmas already sensibly tuned.

## Note
Single-writer is a hard architectural ceiling — sqlite will never match badger on write-bound concurrency. But 340 IOPS is well below that ceiling, so the coalescer + per-op trimming are worthwhile. badger remains the recommended backend for write-heavy workloads.

marmos91:
## Problem

The metadata data-write path was an order of magnitude slower on the SQL backends than on badger (field SMB3 rand-write-4k: badger ~7,300, postgres ~1,090–1,765, sqlite ~340). The block store is identical across all three, so this is purely the metadata backend.

## It is not the disk and not durability

Measured in-process, same machine, same disk, durability equalized (badger `SyncWrites=false`, sqlite WAL+`synchronous=NORMAL`), driving the identical write op against a populated inode table:

| backend | IOPS |
|---|---|
| badger | 20,049 |
| sqlite (`GetFile`+`PutFile`) | 5,526 |

Badger runs on the *same fs* and is ~4x faster, so the gap is what SQL does per op. The CPU profile is the tell: sqlite spends ~53% in the SQL bytecode VM (compile + execute) and ~31% in the per-commit WAL write; badger does one key-value put.

## Root cause

A data WRITE only needs to grow `size` and stamp `mtime`/`ctime`. Both commit paths did a full-row read-modify-write instead: `GetFile` (with its aggregate block-refs read) + `PutFile` (a 20-column rewrite). A single narrow `UPDATE` of the columns a write actually changes runs ~10x faster than that pair (53k vs 5.5k IOPS), verified to affect exactly one row.

## Fix

An optional `DataWriteApplier` a store transaction can implement to do the update as a narrow statement — grow `size` to `max(old,new)` so an out-of-order write never shrinks the file, stamp the times, clear setuid/setgid for a non-root writer, keep the `usedBytes`/quota deltas exact. Both write paths use it when present and fall back to the full read-modify-write otherwise, so backends adopt it incrementally.

Applied to **both** commit paths — the immediate one and, importantly, the pending-write flush behind `deferredCommit`, which is the **default** and therefore the path most deployments actually take.

- **sqlite**: `SELECT` old + narrow `UPDATE`, one transaction. **5,526 → 20,341 IOPS (3.7x)** — now level with badger.
- **postgres**: single statement — `RETURNING` references the pre-update CTE, so the old-size read and the guarded in-place update collapse into one round-trip. See the durability caveat below.
- **badger**: unchanged — it is in-process with no SQL-compile or round-trip tax and already at the target rate, so the fallback is its optimal path.

## Postgres: the SQL shape is not its wall

Measured on a co-located postgres 16 (8 vCPU), same op, scattered over 5k inodes:

| `synchronous_commit` | `GetFile`+`PutFile` | `ApplyDataWrite` | gain |
|---|---|---|---|
| `on` (default) | 3,999 | 4,132 | +3% |
| `off` | 9,160 | 13,940 | **+52%** |

At default durability postgres is bound by the per-commit WAL fsync, not by the statement shape — removing the fsync alone is worth 2.3x. Once that barrier is gone the narrow write is worth another ~52%, so the change is the right shape and is kept, but **it does not by itself move postgres at default durability**. Closing the postgres half of #1819 needs a commit-level lever, which is a separate change and not one to take by quietly weakening durability.

## Two real bugs this surfaced

- The postgres `ApplyDataWrite` never planned at all (`~` on an untyped parameter is ambiguous), so it failed every call. CI stayed green because the postgres jobs run with the default `deferredCommit`, which never reached the immediate path — the exact blind spot the new gated tests close.
- The hot-row benchmark targeted the share root, a *directory*. Re-pointed at a regular file it reads 5,526 rather than an inflated ~7,000, which is also much closer to the field baseline.

## Soundness

- Full sqlite conformance suite green; postgres conformance green (two `TestPostgresPutFile_*` failures on the bench box reproduce with all of this PR's files removed — pre-existing, unrelated).
- Dedicated soundness tests green on **both** sqlite and postgres: grows to `max(old,new)`, never shrinks on out-of-order writes, clears/keeps setuid correctly, exact usedBytes/quota delta, correct `ErrNotFound`/`ErrIsDirectory`.
- `-race` clean under concurrent writers across metadata, sqlite, badger, and memory.
- The postgres soundness test and write-path benchmark are gated on `DITTOFS_TEST_POSTGRES_DSN`, matching the existing postgres suites.

Closes #1819 for sqlite; the postgres commit-fsync wall is called out above and needs its own change. The narrow-write primitive carries directly into the SQL-family unification (#1828).

### DIFF
--- a/pkg/metadata/data_write.go
+++ b/pkg/metadata/data_write.go
@@ -1,0 +1,20 @@
+package metadata
+
+import (
+	"context"
+	"time"
+)
+
+// DataWriteApplier is the hot-path metadata update for a data WRITE. A store
+// transaction implements it to collapse the read-modify-write a WRITE performs
+// — grow size to max(old, new), stamp mtime/ctime, optionally clear setuid/setgid
+// — into a single narrow statement, instead of GetFile (full-row read) plus
+// PutFile (full-row rewrite). It returns the resulting size so the caller can
+// synthesize post-op attrs without reading the row back.
+//
+// Optional: transactions that do not implement it fall back to GetFile+PutFile.
+// Only regular files are affected; a missing or non-regular target returns
+// ErrNotFound.
+type DataWriteApplier interface {
+	ApplyDataWrite(ctx context.Context, handle FileHandle, newSize uint64, now time.Time, clearSUID bool) (finalSize uint64, err error)
+}
--- a/pkg/metadata/data_write.go
+++ b/pkg/metadata/data_write.go
@@ -13,8 +13,9 @@ import (
 // synthesize post-op attrs without reading the row back.
 //
 // Optional: transactions that do not implement it fall back to GetFile+PutFile.
-// Only regular files are affected; a missing or non-regular target returns
-// ErrNotFound.
+// Only regular files are affected: a missing target returns ErrNotFound, and a
+// target that exists but is not a regular file returns ErrIsDirectory and is
+// left untouched.
 type DataWriteApplier interface {
 	ApplyDataWrite(ctx context.Context, handle FileHandle, newSize uint64, now time.Time, clearSUID bool) (finalSize uint64, err error)
 }
--- a/pkg/metadata/deferred_flush_applier_test.go
+++ b/pkg/metadata/deferred_flush_applier_test.go
@@ -1,0 +1,95 @@
+package metadata_test
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+	"time"
+
+	"github.com/marmos91/dittofs/pkg/metadata"
+	"github.com/marmos91/dittofs/pkg/metadata/store/sqlite"
+	"github.com/stretchr/testify/require"
+)
+
+// newApplierFixture builds a Service over a sqlite store. sqlite implements
+// DataWriteApplier, so the deferred pending-write flush takes the narrow
+// single-statement path rather than the GetFile+PutFile fallback. The memory
+// store used by the other service tests does not implement it, so this is the
+// only place the fast path is exercised end to end through the Service.
+func newApplierFixture(t *testing.T) (*metadata.Service, metadata.Store, metadata.FileHandle, string) {
+	t.Helper()
+	ctx := context.Background()
+	store, err := sqlite.NewSQLiteMetadataStore(ctx,
+		&sqlite.SQLiteMetadataStoreConfig{Path: filepath.Join(t.TempDir(), "m.db"), AutoMigrate: true},
+		metadata.FilesystemCapabilities{
+			MaxReadSize: 1048576, PreferredReadSize: 1048576,
+			MaxWriteSize: 1048576, PreferredWriteSize: 1048576,
+			MaxFileSize: 1 << 62, MaxFilenameLen: 255,
+			MaxPathLen: 4096, MaxHardLinkCount: 32767,
+			SupportsHardLinks: true, SupportsSymlinks: true,
+			CaseSensitive: true, CasePreserving: true, TimestampResolution: 1,
+		})
+	require.NoError(t, err)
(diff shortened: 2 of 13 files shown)


marmos91:
Ran the postgres half on a real postgres 16 (8 vCPU, co-located). Two things came out of it, and I have rewritten the description to match.

The postgres fast path was broken. `~` applied to an untyped parameter is ambiguous, so the statement never planned and every call failed. CI stayed green because the postgres jobs run with the default `deferredCommit`, which never reached the immediate path. That blind spot is now closed by a gated soundness test that actually drives the path on postgres.

Chasing that blind spot turned up the bigger miss: the pending write flush behind `deferredCommit` is the default path, and it was still paying the same `GetFile` plus full row `PutFile`. It now goes through the applier too, so the win lands in the configuration people actually run.

On numbers, postgres is honest but unexciting at default durability:

| `synchronous_commit` | `GetFile`+`PutFile` | `ApplyDataWrite` | gain |
|---|---|---|---|
| `on` (default) | 3,999 | 4,132 | +3% |
| `off` | 9,160 | 13,940 | +52% |

Postgres is bound by the per commit WAL fsync, not by the statement shape. The narrow write is clearly the right shape once that barrier is gone, but it does not move postgres on its own. That half of #1819 needs a commit level lever and should not be bought by weakening durability.

sqlite is unchanged in substance: 5,526 to 20,341 IOPS, level with badger. The baseline moved slightly because the hot row benchmark was pointed at the share root, a directory. On a regular file it reads 5,526, which also sits much closer to the field baseline.

Bench VM has been returned to stopped.

PR #1851, opened 2026-07-27
fix(journal): persist cold intervals so evicted ranges are not read as holes (#1850)

### LINKED ISSUE #1850: P0: pre-journal→journal migration leaves share serving all-zeros on read (data intact in *.pre-journal-backup)
## P0: after the pre-journal→journal migration, reads return all-zeros (no remote fetch) — data intact in backup, but the share is unreadable

On this deployment (v0.29.0, `pin` share, Cubbit DS3), **file reads through the mount return all-zeros**, silently. The user discovered it opening files (corrupt). Source data is intact, so this is a **serving/migration bug**, not data loss — but the share is effectively unreadable right now.

### Evidence
```
# read 1 MiB of a known ROM through the NFS mount:
read 1MB: 0.11s   rx-from-network: 0 KB   non-zero bytes: 0 / 1048576   # all zeros, NO Cubbit fetch
head -c 16 <file> -> 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00 00     # should be the ROM header
```
So dfs returns zeros **locally without contacting Cubbit** (0 network) for blocks it should be serving/fetching.

### Block store state
```
dfsctl store block stats -> Blocks Total 26774, Blocks Local 0, Blocks Remote 26774, Local Disk Used 25 MB
du blocks/shares/drive -> 64 GiB, of which:
  40G  blobs.pre-journal-backup   (0000000000000000.blob = 1 GiB, real data: 103M/104M non-zero in first 100MB)
  15G  logs.pre-journal-backup
  18M  journal                     (active store — near-empty)
```
So the migration moved the real block data into `*.pre-journal-backup` (verified: **real, non-zero content**), and the active `journal` store is ~18 MB. Blocks read as `Remote` (0 Local), yet reads don't fetch from Cubbit — they return zeros. drain-uploads = 0 (fully synced), so the data is also on Cubbit.

### Impact / timeline
Streaming/reads worked fine **before** the journal migration (earlier in this deployment). After the pre-journal→journal migration, reads return zeros. It went unnoticed because byte *counts* were correct (correct length, zero content) — only opening the files revealed it.

### What I need
The data is safe (intact `*.pre-journal-backup` on disk + synced to Cubbit), so no rush to recover blindly — I'd rather not guess. **How do you want to recover a share whose journal migration left it serving zeros?** Options I can see: re-run the migration from the `*.pre-journal-backup` blobs, rebuild the journal from Cubbit, or restore the pre-journal layout + a matching binary. I have the box live and can run whatever diagnostic/repair you want. **Not deleting the `*.pre-journal-backup` — it's the golden copy.**

Also: this means the migration's "success" wasn't verified end-to-end (reads never checked content). A post-migration content check (sample read + hash vs the pre-journal blob) before declaring success — and before any cleanup of the backups — would have caught it.

marmos91:
Fixes #1850.

## What breaks

A cold interval marks a range that is durable on the remote but holds no local bytes. It owns no record, and the interval index is rebuilt purely from segment records at recovery — so every cold interval was lost on restart. The range then came back as a POSIX hole: `journal.ReadAt` zero-filled it and reported `cold=false`, so the engine never hydrated, and the read returned zeros of the right length **with no remote request at all**.

Two ways in, both silent:

- **eviction** (`reclaim.go evictSegment`) flipped a dropped segment's intervals cold *in memory only*;
- the **pre-journal upgrade** (`service.go` → `SeedColdFromManifest`) and **snapshot restore** seed cold intervals from the surviving manifest against an otherwise empty journal.

The field report is the second: a migrated share served ~40 GiB of intact, fully-synced data as zeros after the next restart, because nothing re-seeds once the legacy dirs are archived (`hasLegacyLocalLayout` is false, so `MigratedFromLegacy()` never fires again).

The first is wider than the reported case: any remote-backed share that evicted a segment and restarted read those ranges as zeros.

## Fix

Cold intervals persist in a `cold.log` side-log, fsynced **before** the eviction that needs it unlinks its segment (markers-first: the opposite order can lose them, and a marker for bytes that are still local only costs a needless fetch).

They cannot live in the segment stream itself — a segment holding only cold markers has no live warm interval, so eviction and GC would reclaim it and take the markers with it.

Recovery replays entries by `Version` like any record, so a later warm write shadows a stale entry and the tombstone/truncate passes clip them exactly as they clip warm intervals. Recovery also compacts the log once the live set has drifted below it. A torn tail costs only the torn entry.

`SeedCold` now takes a whole file's extents, so seeding a manifest costs one durable write per file rather than one per extent (~27k fsyncs on the reporter's box otherwise).

## Tests

Each fails without the fix and passes with it:

- `TestColdIntervalSurvivesReopen` — evict, reopen, read: must still report cold.
- `TestSeedColdSurvivesReopen` — the seeded flavor.
- `TestColdLogSupersededByLaterWrite` — a stale entry must not shadow a newer local write.
- `TestColdLogTornTailKeepsIntactEntries` — crash mid-append costs only the torn entry.
- `TestLegacyRemoteMigration_ReadsColdAfterRestart` — the field sequence end to end (migrate → restart → read, with no read before the restart, which would warm the journal and mask it).

`go test ./...` clean with `-race` on `journal`, `engine`, `local/fs`. One unrelated pre-existing failure on this machine, `pkg/discovery/wsd TestResponder_StartStopNoDeadlock`, reproduces identically on clean `develop`.

## Not in this PR

- Recovery guidance for the affected deployment (`dfsctl share warm` rebuilds the local tier from the manifest and is immune to this bug — it never consulted cold intervals).
- The migration still does not verify content before declaring success, nor reap its own `*.pre-journal-backup` archive (the 55 GiB in #1843). Both follow-ups.

### DIFF
--- a/pkg/block/engine/flush.go
+++ b/pkg/block/engine/flush.go
@@ -101,28 +101,24 @@ func (bs *Store) DrainRollups(ctx context.Context) error {
 	return err
 }
 
-// ResetLocalState clears the local store's per-payload append-log state so
-// post-restore reads resolve purely through the restored CAS manifest.
-// The snapshot-restore orchestration calls this BEFORE the metadata Reset()
-// + Restore() (not after): clearing the overlay first leaves no dirty
-// intervals for a background rollup worker to flush into the freshly-restored
-// metadata, so a file modified in place after the snapshot is not served from
-// a stale append-log record overlaid on the restored CAS bytes.
-// SeedCold marks [offset, offset+length) of payloadID as remote-durable-but-not-
-// local so a subsequent read faults it in from the remote store rather than
-// zero-filling. Snapshot restore calls it (per restored FileChunk extent, remote
-// shares only) to re-arm cold reads after ResetLocalState wiped the local tier.
-// No-op when the local store has no remote-hydration support (e.g. the in-memory
-// test store), which the caller only hits on non-remote paths anyway.
-func (bs *Store) SeedCold(ctx context.Context, payloadID string, offset, length int64) error {
+// SeedCold marks payloadID's extents — each an {offset, length} pair — as
+// remote-durable-but-not-local so a subsequent read faults them in from the
+// remote store rather than zero-filling. Callers pass a whole file's extents at
+// once: the local tier persists the markers, and one durable write per file beats
+// one per extent when a caller is seeding an entire manifest. Snapshot restore
+// and the pre-journal upgrade both use it (remote-backed shares only) to arm cold
+// reads over a local tier that holds none of the bytes. No-op when the local store
+// has no remote-hydration support (e.g. the in-memory test store), which the
+// caller only hits on non-remote paths anyway.
+func (bs *Store) SeedCold(ctx context.Context, payloadID string, extents [][2]int64) error {
 	type coldSeeder interface {
-		SeedCold(ctx context.Context, id journal.FileID, offset, length int64) error
+		SeedCold(ctx context.Context, id journal.FileID, extents [][2]int64) error
 	}
 	cs, ok := bs.local.(coldSeeder)
 	if !ok {
 		return nil
 	}
-	return cs.SeedCold(ctx, journal.FileID(payloadID), offset, length)
+	return cs.SeedCold(ctx, journal.FileID(payloadID), extents)
 }
 
 // RestoreToVersion rewinds the local journal to a snapshot's version watermark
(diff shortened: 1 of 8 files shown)


marmos91:
Good catches from Copilot — all three were real, fixed in 5326e5d3.

- **Directory not fsynced on create.** Right, and it defeated the point: syncing `cold.log` says nothing about its *directory entry*, so a crash right after the first eviction could come back to no cold log at all. Now flushes the directory when the log is created, using the same `fsyncDir` the segment path uses.
- **Rename not durable.** Same class, fixed after the compaction rename — and after the `os.Remove` when compaction empties the log, which had the same hole.
- **`FileIDLen` unbounded.** `uint16(len(e.id))` would silently truncate an over-long FileID and make every following entry unreplayable. Now rejected up front against `maxFileIDLen`, matching `writeDataRecord` and friends.

PR #1855, opened 2026-07-27
fix(dfsctl): render nfs-config show instead of crashing on a nil table renderer (#1854)

### LINKED ISSUE #1854: dfsctl share nfs-config show panics with a nil pointer dereference
Reported by a user evaluating DittoFS on Proxmox VE (v0.29.0 release binary):

```
root@dittofs:~# dfsctl share nfs-config show /export
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x65ceb5]

github.com/marmos91/dittofs/internal/cli/output.PrintTable(...)
        internal/cli/output/table.go:20
github.com/marmos91/dittofs/cmd/dfsctl/cmdutil.PrintResource(...)
        cmd/dfsctl/cmdutil/util.go:279
github.com/marmos91/dittofs/cmd/dfsctl/commands/share.runNFSConfigShow(...)
        cmd/dfsctl/commands/share/nfs_config.go:108
```

`runNFSConfigShow` calls `cmdutil.PrintResource(os.Stdout, cfg, nil)` unconditionally. In the default (table) format `PrintResource` hands that nil `TableRenderer` to `output.PrintTable`, which calls `data.Headers()` on a nil interface. So the command crashes for every invocation that does not pass `-o json`/`-o yaml`.

Two defects:

1. **No renderer for the resource.** The other four `PrintResource(..., nil)` call sites (`netgroup show`, `adapter settings` x2, `share show`) all guard with `if format != output.FormatTable` first; this one does not, and has no table renderer at all.
2. **`PrintTable` cannot survive a nil renderer.** A missing renderer is a caller bug, but it should be an error, not a SIGSEGV that takes down the CLI.

Fixing both: a field/value renderer for the NFS config, and a nil guard at the `PrintTable` seam so no future `show` command can crash this way.

Same reporter also raised documentation/help/log text referencing flags and subcommands that do not exist or sit at a different command path, and NFS interop trouble with Proxmox VE. Filing those separately.

marmos91:
Fixes #1854. Reported by a user evaluating DittoFS on Proxmox VE, on the v0.29.0 release binary.

`dfsctl share nfs-config show <name>` crashed with a SIGSEGV on every invocation that did not pass `-o json`/`-o yaml`: it called `PrintResource(os.Stdout, cfg, nil)` unconditionally, and in table format that nil `TableRenderer` reached `Headers()` through a nil interface inside `PrintTable`.

Two fixes, because there are two defects:

- **The command had no renderer.** Added `NFSConfigDetail`, a field/value table matching the `NetgroupDetail` idiom: netgroup (spelled out as all-clients-allowed when unset), squash, the anonymous UID/GID only when the export overrides them, the auth flavors, and READDIRPLUS un-negated from the stored `DisableReaddirplus`.
- **`PrintTable` could not survive a nil renderer.** It now returns an error pointing at `-o json`. A missing renderer is a caller bug, but it should not take down the CLI. Four other call sites (`netgroup show`, `adapter settings` x2, `share show`) pass nil too and are only safe because they check the output format first — nothing structural was stopping the next one from crashing the same way.

## Tests

- `TestPrintTableNilRenderer` — verified to reproduce the reported SIGSEGV without the guard.
- `TestNFSConfigDetail_Rows_Defaults` / `_Overrides` — cover the nil-pointer anonymous IDs and the negated READDIRPLUS bool.

No command or flag definitions changed, so `docs/guide/cli.md` does not need regenerating.

Same reporter also raised docs/help/log text referencing flags and subcommands that do not exist, and NFS interop trouble with Proxmox VE (`showmount`-based probing plus our non-standard default port). Both tracked separately.

### DIFF
--- a/cmd/dfsctl/commands/share/nfs_config.go
+++ b/cmd/dfsctl/commands/share/nfs_config.go
@@ -3,6 +3,7 @@ package share
 import (
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
 
 	"github.com/marmos91/dittofs/cmd/dfsctl/cmdutil"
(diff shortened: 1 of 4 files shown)


marmos91:
Declining this one. A typed nil (`(*TableData)(nil)`) stored in the interface would indeed slip past `data == nil`, but it cannot arise here: every renderer in this codebase is a value type — `NetgroupDetail`, `ShareDetail`, the new `NFSConfigDetail` — so there is no pointer to be nil in the first place, and a caller would have to go out of their way to construct one.

Pulling `reflect` into the print path to defend against that trades a real, observed crash (a `nil` literal, which is what every one of the five call sites actually passes) for a hypothetical one, and leaves the general case uncovered anyway — a renderer whose `Headers()` panics for any other reason is equally fatal. If we ever want that guarantee properly, the answer is a `recover` around the render, not a nil probe. Not worth it today.

PR #1865, opened 2026-07-27
fix(stores): stamp the on-disk format and refuse state from a newer release

### LINKED ISSUE #1864: Upgrade safety: no on-disk format version, so a downgrade serves zeros instead of refusing
Splitting theme 2 of #1843 out of the parent thread so it can be tracked and closed on its own.

## What the field report described

Upgrading a data-bearing box across a metadata-format change was risky in practice: the migration ran on the startup path and looked like a hang on a large store, and once the metadata had migrated, rolling back to the prior binary left the share unreadable. Recovery was manual, via durable snapshots and a disable→enable dance on a live box.

## What the code actually does

**No on-disk format-version marker gates live store open anywhere.** The `*SchemaVersion` constants in each backend's `snapshot_store.go` gate backup payload streams only (`Snapshot`/`RestoreSnapshot`), never `Open`.

**The one boot guard wired to a hard exit is dead code.** `block.ErrLegacyLayoutDetected` has zero producers — `.cas-migrated` appears in no Go code since the migration tool was removed. The guard that does fire, `fs.ErrLegacyLocalFormat`, is a different sentinel, so `handleLoadSharesError` falls through to `logger.Warn("Failed to load some shares")` and the daemon boots anyway with the share silently absent.

**Formats have changed without a version bump.** Recent examples:

| Layer | Change | Landed | Old binary on new state |
|---|---|---|---|
| badger | inline `fBlocks` → sibling `fm:<uuid>` manifest | v0.29.0 | manifest not found; matters for remote-backed reads |
| journal | `cold.log` cold-interval side log | v0.29.1 | log never opened; evicted ranges become holes |
| badger | nlink → `l:<uuid>` key | v0.29.0 | nlink silently 0 |
| sqlite | `schema_migrations` ahead of the binary | ongoing | opens, then per-query SQL errors |

`fileVersion` stayed `0x01` across the `fm:` split, so an older binary decodes the record cleanly rather than rejecting it.

## Measured on a fresh VM

A local-only badger share written on v0.28.1, upgraded to v0.29.1, written to again, then rolled back to v0.28.1 read **every file correctly**. A local-only share resolves chunks from the journal's own per-payload index, so the metadata manifest is not on that read path. The exposure is specific to remote-backed shares, where the manifest and the cold-interval log are what make a fetch possible — which is the reporter's configuration.

## Scope

- Stamp the on-disk format version per store; refuse to open state from a newer release, hard-exit 78 with a message naming both versions and the recovery options.
- Route the existing `ErrLegacyLocalFormat` through the same hard exit instead of warn-and-continue.
- Warn before a one-way migration starts; add a progress heartbeat to the migration loops (they run before the API binds, so logs are the only channel).
- Document the upgrade/rollback contract.

Note the guard is forward-looking: already-released binaries cannot be made to refuse. It makes rollbacks safe from the next release onward.

Parent: #1843

marmos91:
Closes #1864. Theme 2 of #1843 — upgrade safety and reversibility.

## What prompted this

A field deployment reported that upgrading a data-bearing box was risky: the migration ran on the startup path and looked like a hang on a large store, and after rolling back to the prior binary the share was unreadable. They recovered via snapshots and a disable→enable dance on a live box.

## What I verified in the code

Two defects, both confirmed by grep rather than inference:

1. **The exit-78 boot guard was dead code.** `block.ErrLegacyLayoutDetected` had zero producers — `.cas-migrated` appears nowhere in Go source since the `.blk` migration tool was removed.
2. **The guard that does fire warn-and-continues.** `fs.ErrLegacyLocalFormat` is a different sentinel, so `handleLoadSharesError` fell through to `logger.Warn("Failed to load some shares")` and the daemon booted anyway — healthy-looking, with the share silently absent.

And the gap behind the report: **no on-disk format-version marker gates live store open anywhere.** The `*SchemaVersion` constants in each backend gate backup payload streams only, never `Open`. Formats have moved without a version bump — the badger `fm:` manifest split and the journal `cold.log` both landed with no marker, and `fileVersion` stayed `0x01` across the former.

## What I could NOT reproduce

On a fresh Scaleway VM I ran two rollbacks end to end:

- local-only badger share, v0.28.1 → v0.29.1 → v0.28.1, with files written under each release
- remote-backed share with a small local cache, 300 MiB written and drained to S3, local tier evicted to 104K, v0.29.1 → v0.29.0

**Both read every file back with matching SHA-256.** A local-only share resolves chunks from the journal's own per-payload index, so the metadata manifest is not on that read path; and the evicted remote-backed share still resolved through the manifest. I could not produce the silent-zeros rollback, so this PR does not claim to fix the reporter's exact failure — their jump crossed the pre-journal migration, which is a larger step than either tested here.

What this PR does fix is the class: the two verified defects above, plus the missing contract that let both format changes ship unguarded.

## The change

- **Format stamp per store.** Local journal keeps a `format` file beside `cold.log`; badger uses an `fmt:store` key; sqlite and postgres need no new state — they compare the highest applied `schema_migrations` version against the newest migration the binary embeds. That check now runs **unconditionally**, not only when `AutoMigrate` is set, since it defaults off and left the schema uninspected.
- **Refuse and stop.** State newer than the build wraps `block.ErrFutureFormat`; boot exits 78 with a directive naming both versions and the two recovery paths. `ErrLegacyLocalFormat` now takes the same hard exit.
- **Pre-flight warning** before a one-way migration — what changes, that the previous release cannot read the result, snapshot first. No ack flag: an unattended `apt upgrade && systemctl restart` must not strand a box.
- **Progress heartbeat** (5s) in the three migration loops whose cost scales with stored data. Deliberately no job/REST/`--watch` surface: these run before the API binds, so logs are the only channel that exists when it matters.

## Scope note

The guard is forward-looking. Already-released binaries cannot be made to refuse, so this makes rollbacks safe from the next release onward; for the current generation the levers are the pre-flight warning and the documented snapshot step.

## Verification

`go build ./...`, `go vet ./...`, and `go test ./pkg/... ./cmd/...` all clean. New tests per backend: state stamped one version above the build refuses with `errors.Is(err, block.ErrFutureFormat)`; at or below opens; unstamped opens and gets stamped. sqlite and postgres assert the refusal with `AutoMigrate` both on and off.

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -915,7 +915,7 @@ build would serve the stored files as zeros rather than fail, so the daemon
 stops instead.
 
 Upgrades are one-way. To recover, either:
-  - reinstall the release that wrote this share and start again, or
+  - reinstall the release that wrote this state and start again, or
   - restore the snapshot taken before the upgrade.
 
 No data has been modified.`, err)
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -20,6 +20,7 @@ import (
 	"github.com/marmos91/dittofs/pkg/adapter/smb"
 	"github.com/marmos91/dittofs/pkg/auth/kerberos"
 	"github.com/marmos91/dittofs/pkg/block"
+	"github.com/marmos91/dittofs/pkg/block/local/fs"
 	"github.com/marmos91/dittofs/pkg/config"
 	"github.com/marmos91/dittofs/pkg/controlplane/api"
 	"github.com/marmos91/dittofs/pkg/controlplane/models"
@@ -41,10 +42,10 @@ var isTerminal = func(fd uintptr) bool {
 }
 
 // EX_CONFIG is the exit code per sysexits(3) — "configuration error".
-// Used by the legacy-layout boot guard when a share directory still
-// contains pre-v0.16 `.blk` files without a `.cas-migrated-v1` sentinel.
-// The operator migrates with an older release's migrate-to-cas command
-// before retrying — this release no longer ships it.
+// Used by the format-mismatch boot guard when a share's on-disk state was
+// written by a release this build cannot read, in either direction: newer
+// state after a downgrade, or a pre-journal layout awaiting migration.
+// Serving either would return zeros for stored files, so boot stops instead.
 const EX_CONFIG = 78
 
 // exitFn is the production exit path for the legacy-layout boot guard.
@@ -834,8 +835,8 @@ func handleLoadSharesError(err error, stderr *os.File) bool {
 	if err == nil {
 		return false
 	}
-	if errors.Is(err, block.ErrLegacyLayoutDetected) {
-		_, _ = fmt.Fprintln(stderr, formatLegacyLayoutDirective(err))
+	if errors.Is(err, block.ErrFutureFormat) || errors.Is(err, fs.ErrLegacyLocalFormat) {
+		_, _ = fmt.Fprintln(stderr, formatFormatMismatchDirective(err))
 		exitFn(EX_CONFIG)
 		// Unreachable in production (exitFn == os.Exit terminates).
 		// Defensive: in-process tests stub exitFn to NOT terminate.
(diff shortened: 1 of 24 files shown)


marmos91:
On the `blockstore:` prefix (Copilot): keeping `store:` deliberately, and not adding `ErrFutureFormat` to `TestSentinelMessages_HaveBlockstorePrefix`.

The convention holds for sentinels that are about blocks. This one is not — the metadata stores return it too, so the prefix would be actively wrong on the path that fires most often. From the VM run:

```
Refusing to start: failed to load metadata stores: failed to create metadata store "m1": store: on-disk format is newer than this build: metadata store is at format version 2, this build reads up to 1.
```

`blockstore:` there would tell an operator to go looking at the block store while the badger metadata store is what refused. The deviation is documented in both `errors.go` and the test comment so the exclusion does not read as an oversight.

PR #1867, opened 2026-07-27
fix(metadata): reseed quota cache on memory Reset and RestoreSnapshot

viquitorreis:
Reset() never touched the quota cache, it zeroed usedBytes and cleared every other map but left per identity usage stale. Now clears it under quotaMu, alongside usedBytes.
RestoreSnapshot recomputed usedBytes from the restored files but left the quota cache untouched, so GetQuotaUsage kept reporting pre-restore (now-stale) per identity values after a successful restore. Rebuilt the per uid/gid maps from the restored files and reseed via quota.Cache.Seed, mirrorring the usedBytes recompute right above it.

sqlite/postgres/badger already reseed correctly via initUsedBytesCounter on restore, this was memory only.

Adds conformance coverage in storetest (QuotaUsageAfterRestore, plus Reset/Restore assertions in ResetThenRestoreConformance) so the divergence is caught for all four backend going forward.

Found while scoping the basestore extraction in #1828.

# Description

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Test coverage improvement

## Changes Made

Please provide a clear list of changes:

- `memory`'s `Reset()` never cleared the per-identity quota cache (only `usedBytes` and the other maps) now clears it under `quotaMu`.
- `memory`'s `RestoreSnapshot` recomputed `usedBytes` from the restored files but left the quota cache stale now rebuilds per uid/gid maps from the restored files and reseeds via `quota.Cache.Seed`, mirroring the `usedBytes` recompute right above it.
- sqlite/postgres/badger already reseed correctly via `initUsedBytesCounter` on restore this was memory-only.
- Added conformance coverage in `storetest`: `QuotaUsageAfterRestore`, plus Reset/Restore assertions in `ResetThenRestoreConformance`, so the divergence is caught for all four backends going forward.

## Testing

- [x] Unit tests pass (`go test ./...`)

## Testing Instructions

1. `go build ./...`
2. `go test -race ./pkg/metadata/store/... -run 'TestConformance|TestSnapshotConformance|TestResetThenRestoreConformance' -v`

## Checklist

## Additional Notes

This PR is still open/draft the `internal/basestore` extraction (the main scope agreed in #1828) is in progress and will land as follow-up commits here.

### DIFF
--- a/pkg/metadata/store/internal/basestore/base.go
+++ b/pkg/metadata/store/internal/basestore/base.go
@@ -36,27 +36,27 @@ func NewBaseStore() *Base {
 
 // GetUsedBytes returns the current total logical bytes used by regular files.
 // This is an O(1) atomic read, safe for concurrent access without locks.
-func (s *Base) GetUsedBytes() int64 {
-	return s.usedBytes.Load()
+func (b *Base) GetUsedBytes() int64 {
+	return b.usedBytes.Load()
 }
 
 // GetQuotaUsage returns per-identity usage for the given scope and id.
 // O(1) map read under quotaMu. A missing key returns a zero UsageStat.
-func (s *Base) GetQuotaUsage(scope metadata.QuotaScope, id uint32) (metadata.UsageStat, error) {
-	s.quotaMu.Lock()
-	defer s.quotaMu.Unlock()
-	return s.quota.Get(scope, id), nil
+func (b *Base) GetQuotaUsage(scope metadata.QuotaScope, id uint32) (metadata.UsageStat, error) {
+	b.quotaMu.Lock()
+	defer b.quotaMu.Unlock()
+	return b.quota.Get(scope, id), nil
 }
 
 // ApplyQuotaDelta folds a per-identity usage delta into the in-memory usage
 // cache. Called post-commit (matching usedBytes).
-func (s *Base) ApplyQuotaDelta(delta map[quota.Key]metadata.UsageStat) {
+func (b *Base) ApplyQuotaDelta(delta map[quota.Key]metadata.UsageStat) {
 	if len(delta) == 0 {
 		return
 	}
-	s.quotaMu.Lock()
-	defer s.quotaMu.Unlock()
-	s.quota.Apply(delta)
+	b.quotaMu.Lock()
+	defer b.quotaMu.Unlock()
+	b.quota.Apply(delta)
 }
 
 // AddUsedBytes atomically adds delta (positive or negative) to the used-bytes
--- a/pkg/metadata/store/internal/basestore/base.go
+++ b/pkg/metadata/store/internal/basestore/base.go
@@ -1,0 +1,95 @@
+package basestore
+
+import (
+	"sync"
+	"sync/atomic"
+
+	"github.com/marmos91/dittofs/pkg/metadata"
+	"github.com/marmos91/dittofs/pkg/metadata/store/internal/quota"
+)
+
+// Base holds the usage accounting state shared byte-for-byte across all
+// metadata store backends. The total logical bytes used by regular files, and
+// per-identity (uid/gid) usage for quota enforcement. Each backend
+// embeds Base and is responsible for seeding it at the startup. The seeding
+// mechanism differs per backend (SQL SUM + GROUP BY, KV full scan, or none)
+// and stays outside this type
+type Base struct {
+	// usedBytes tracks the total logical bytes used by regular files. Updated
+	// atomically on every size-changing operation.
+	usedBytes atomic.Int64
+
+	// quota tracks per-identity usage (bytes + file count) for regular files,
+	// keyed by owner uid / gid. Updated from each comitted transaction's deltas
+	// Guarded by quotaMu.
+	quotaMu sync.Mutex
+	quota   *quota.Cache
+}
+
+// NewBaseStore returns a zero usage Base ready to be embedded. Backends still
+// seed it from their own durable state at startup.
+func NewBaseStore() *Base {
+	return &Base{
+		quota: quota.NewCache(),
+	}
+}
+
+// GetUsedBytes returns the current total logical bytes used by regular files.
+// This is an O(1) atomic read, safe for concurrent access without locks.
+func (s *Base) GetUsedBytes() int64 {
+	return s.usedBytes.Load()
+}
+
+// GetQuotaUsage returns per-identity usage for the given scope and id.
+// O(1) map read under quotaMu. A missing key returns a zero UsageStat.
+func (s *Base) GetQuotaUsage(scope metadata.QuotaScope, id uint32) (metadata.UsageStat, error) {
+	s.quotaMu.Lock()
+	defer s.quotaMu.Unlock()
+	return s.quota.Get(scope, id), nil
+}
+
+// ApplyQuotaDelta folds a per-identity usage delta into the in-memory usage
+// cache. Called post-commit (matching usedBytes).
+func (s *Base) ApplyQuotaDelta(delta map[quota.Key]metadata.UsageStat) {
+	if len(delta) == 0 {
+		return
+	}
+	s.quotaMu.Lock()
+	defer s.quotaMu.Unlock()
+	s.quota.Apply(delta)
+}
+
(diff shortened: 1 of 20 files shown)


viquitorreis:
@marmos91 

This grew a bit past the original memory only bug: found the same quota cache not cleared gap in badger's Reset (same class of bug you flagged for memory, just undiscovered there too until the new storetest coverage ran against it).

Ready for a look whenever you have time.

viquitorreis:
@marmos91 The CI failure (TestFiletimeToTime_WindowsEpoch) is unrelated to this PR, confirmed it's pre -existing on develop with none changes made on this branch.
Root cause: FiletimeToTime in internal/adapter/smb/types/filetime.go returns time.Unix(sec, rem*100) without .UTC() and time.Unix returns local time in Go, so on any machine/runner with a negative UTC offset, Jan 1 1601 00:00 UTC rolls back to Dec 31 1600 locally. Deterministic, not flaky fails on any UTC-negative TZ (reproduced standalone with TZ=America/Sao_Paulo).

Not touching it since it's protocol adapter code, out of scope for this PR. Flagging so it doesn't block the merge on something unrelated happy to open a separate one-line fix (.UTC() on the return) if useful, just let me know.

marmos91:
@viquitorreis Nice find on the badger Reset, that one was hiding in plain sight.

I pulled the branch and checked it end to end: builds clean, vet/golangci-lint/gofmt all clean,
`go test -race ./pkg/metadata/...` green. I also reverted your two fixes one at a time to make sure
the new assertions actually bite, and they do, QuotaUsageAfterRestore and ResetThenRestoreConformance
both fail without them, on memory and badger. That's the part I care most about, so 👍.

Two things before this can go in:

1. **This targets `main`.** It needs to go to `develop`. `main` only ever moves by fast-forward from
   `develop` at release time,  merging a feature branch into it directly breaks that invariant. Please retarget and rebase onto `develop`; it's a fair bit ahead now, including v0.29.1.

2. **Scope.** The title and first commit are the quota fix; commits 2 and 3 are the basestore
   extraction from #1828. The refactor itself is clean and mechanical, but bundling them means the
   bugfix can't be reverted on its own, and it makes this harder to review than it needs to be. Can
   we split the quota fix into its own PR against develop so it lands now, and keep the basestore
   work here?

pkg/metadata/store/internal/basestore/base.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/marmos91/dittofs/pkg/metadata/store/internal/quota"
+)
+
+// Base holds the usage accounting state shared byte-for-byte across all
+// metadata store backends. The total logical bytes used by regular files, and
+// per-identity (uid/gid) usage for quota enforcement. Each backend
+// embeds Base and is responsible for seeding it at the startup. The seeding
+// mechanism differs per backend (SQL SUM + GROUP BY, KV full scan, or none)
+// and stays outside this type
+type Base struct {
```
marmos91, line 17: Design question, not a blocker, embedding by pointer promotes all seven methods onto the four
exported store types. We only need `GetUsedBytes` and `GetQuotaUsage` (those satisfy metadata.Store).
The other five come along for free, and four of them are reachable from outside the repo:
`ApplyQuotaDelta` is safe because it takes an internal `quota.Key`, but `SetUsedBytes(int64)`,
`AddUsedBytes(int64)`, `ResetUsage()` and `Seed(...)` only take public types, so
`memStore.SetUsedBytes(999)` compiles for anyone importing pkg/metadata/store/memory.

Making the field unexported (`base *basestore.Base`) plus two one-line forwarders per backend is
8 tiny funcs instead of 20 leaked mutators. Your call.

pkg/metadata/store/internal/basestore/base.go
```diff
@@ -0,0 +1,95 @@
...
+// O(1) map read under quotaMu. A missing key returns a zero UsageStat.
+func (s *Base) GetQuotaUsage(scope metadata.QuotaScope, id uint32) (metadata.UsageStat, error) {
+	s.quotaMu.Lock()
+	defer s.quotaMu.Unlock()
+	return s.quota.Get(scope, id), nil
+}
+
+// ApplyQuotaDelta folds a per-identity usage delta into the in-memory usage
+// cache. Called post-commit (matching usedBytes).
+func (s *Base) ApplyQuotaDelta(delta map[quota.Key]metadata.UsageStat) {
```
marmos91: Receiver name flips halfway down: `(s *Base)` for the first three, `(b *Base)` for the rest.
Pick one.

marmos91:
Your TZ analysis is right and I reproduced it: nice catch:

```
    TZ=America/Sao_Paulo go test -run TestFiletimeToTime_WindowsEpoch ./internal/adapter/smb/types/
    filetime_test.go:75: FiletimeToTime(1) year = 1600, want 1601   FAIL

    TZ=UTC   ... ok
```

So yes, please open the one-line `.UTC()` PR, that's a real latent bug and I'd take it.

But it isn't what's failing here. GH runners are UTC, so that test passed in your run:
`TestFiletimeToTime_WindowsEpoch` shows PASS in the log. The red job is postgres, and it's your
own new assertion:

```
    postgres_conformance_test.go:118: post-Reset GetQuotaUsage(user, 1000) = {Bytes:14680064 Files:2}, err=<nil>, want zero
    --- FAIL: TestResetThenRestoreConformance (0.14s)
    FAIL  github.com/marmos91/dittofs/pkg/metadata/store/postgres  20.321s
```

[https://github.com/marmos91/dittofs/actions/runs/30364329064/job/90309862686](https://github.com/marmos91/dittofs/actions/runs/30364329064/job/90309862686)

Root cause is the same bug you're fixing, in the one backend that got missed. `postgres/reset.go`
truncates every table and returns without touching the counters, memory, sqlite and badger all
call `ResetUsage()` now, postgres doesn't. Should be a one-liner after the COMMIT:

```go
    if _, err := pgRaw.Exec(ctx, "COMMIT").ReadAll(); err != nil {
        return fmt.Errorf("reset: commit: %w", err)
    }
    s.ResetUsage()
    return nil
```

Worth saying: this is your conformance test working exactly as intended. It already caught badger,
now it's caught postgres, that's the whole argument for putting the assertion in storetest instead
of per-backend. Please fix it here rather than deferring, it's the same defect and the same PR.

Two things worth flagging for you separately: the `.UTC()` bug is real but invisible in CI, so it'll only ever bite contributors in negative-offset timezones, worth an issue even if the fix is trivial. And the badger stale `shareCache` I found earlier is a third instance of this same "Reset clears durable state, forgets the in-memory mirror" pattern: three of four backends had a variant of it.

viquitorreis:
@marmos91 Heads up on ordering: I pushed this rebase before splitting #1883 out, so right now this branch's history still has the raw quota-fix commits alongside the basestore extraction duplicated with #1883, not yet cleaned up. Once #1883 merges I'll rebase this onto the updated develop, which should let git drop the now-upstream duplicate commits automatically, leaving just the basestore work. Flagging so the current commit list doesn't read as unresolved feedback it's just sequencing, will be clean shortly after #1883 lands.

marmos91:
Closing this as superseded by #1883.

#1883 was split out of this branch and carries the whole behavioural change. Comparing both against current `develop`, the fixes are identical:

| gap | here | #1883 |
|---|---|---|
| `memory.Reset` clears the quota cache | yes | yes |
| `memory.RestoreSnapshot` reseeds the quota cache | yes | yes |
| `badger.Reset` zeroes `usedBytes` + quota cache | yes | yes |
| `postgres.Reset` zeroes `usedBytes` + quota cache | yes | yes |
| `sqlite.Reset` | already correct on `develop` — refactored here, untouched there | n/a |
| conformance coverage for the restore + reset paths | yes | yes |

The only thing this branch has that #1883 does not is the `basestore.Base` extraction — hoisting `usedBytes` and the `quota.Cache` out of the four backends into a shared embedded type. That is a refactor, not a fix, and it belongs to the metadata unification tracked in #1828, where the SQL and KV families get collapsed properly rather than one type at a time. Nothing is lost by closing here: the bug fix lands via #1883, and the extraction is picked up by that work.

#1883 has since been rebased onto current `develop` and re-verified. Thanks for splitting it out — that was the right call.

Issue #1767, opened 2026-07-19
Cold-read benchmark barrier broken for DittoFS: drain-uploads stalls + evict is a no-op below 1GiB


marmos91:
Surfaced by the QoS benchmark cold-read pass. Two distinct failures make it impossible to force a genuinely cold (from-S3) read on DittoFS in the harness:

## 1. `drain-uploads` stalls (local + writeback tiers)
The cold barrier runs `dfsctl system drain-uploads` to push local blocks to S3 before evicting. It fails:
```
drain uploads stalled: no upload progress within 5m0s: context canceled
```
Zero upload progress in 5 minutes — the async uploader appears stuck, not merely slow. Seen on `dittofs-s3` and `dittofs-s3-writeback` (badger + sqlite). May be aggravated by a large local cache (more pinned/unsynced bytes to flush).

## 2. Evict is a no-op below ~1GiB (remote tier)
The cold barrier requires local disk to drop ≥80% after evict; instead it does not move:
```
cold barrier failed: local disk only fell 477MiB->477MiB (want >=80% drop); the cold pass would measure locally-served reads, not S3
```
The active log blob is never sealed/evicted while the working set is under ~1GiB, so a "cold" read would actually be served warm. This looks like a recurrence of the previously-seen evict-below-1GiB gap.

## Impact
- Cold-read-from-S3 numbers for DittoFS cannot be produced in the benchmark; warm-tier numbers are unaffected.
- Benchmark workaround: run warm-only (`--evict-cache=false`); cold-read documented as a gap.

## Expected
`drain-uploads` should make progress (or fail fast with a real error) and evict should be able to seal + drop the active blob regardless of working-set size, so a cold pass measures true from-S3 latency.

marmos91:
## Triage: root-caused both paths; ownership split

### Path 1 — `drain-uploads` stalls
The watchdog is working as designed: `runWithIdleWatchdog` (`internal/controlplane/api/handlers/system.go`) samples `UploadProgress()` every 1s and trips `drain uploads stalled` after `DefaultDrainStallTimeout` (5m) of **zero concluded upload attempts**. A merely-slow drain bumps `completedSyncs`/`failedSyncs` per ~4 MiB CAS chunk, so a flat counter for 5m means one upload is genuinely **wedged, not slow**.

[...]

PR #1875, opened 2026-07-28
perf(journal): pack carve blocks across dirty runs

### LINKED ISSUE #1872: Carve emits one remote object per contiguous dirty run, serially — a randomly-written share drains at ~130 KB/s
Surfaced while investigating #1767 (DittoFS cold cells missing from the QoS matrix). Four of five DittoFS tiers failed their cold barrier because `dfsctl system drain-uploads` did not return within 15 minutes. The mechanism is measured, not inferred.

## What carve does

`carveFile` (pkg/block/journal/carve.go:233) splits the dirty set into maximal contiguous runs and calls `carveRun` per run. Two consequences:

1. **One remote object per run.** `carveRun` flushes at `CarveBlockSize` *and* unconditionally at the run end (carve.go:412), so a 4 KiB run commits its own block even though `CarveBlockSize` is 4 MiB.
2. **No overlap across runs.** `carveRun` ends with `disp.wait()` (carve.go:414), blocking until that run commits before the next run begins. The dispatcher overlaps commits *within* a run only.

So the cost of a drain scales with the *number of scattered writes*, not the bytes, and every one of them is a serial round-trip.

## Measured

Added to `pkg/block/journal/carve_scatter_test.go` (fake sink, no network):

```
scattered: 200 runs, 800 KiB total -> 200 blocks committed (1 if packed by size)
scattered: 50 runs -> 50 commits, peak concurrent commits = 1
```

800 KiB of data costs 200 remote round-trips. Peak concurrency is 1.

## Impact

At an S3 RTT of ~30ms this is ~33 writes/s, i.e. **~130 KB/s of drain throughput** on a randomly-written file. In the benchmark, one `rand-write-4k` cell (~2365 IOPS x 30s = ~71k scattered writes) leaves a backlog needing ~35 minutes to drain.

Not benchmark-only. A writeback-tier share taking random writes accumulates a durability backlog that drains at that rate, which affects how far behind the remote the share sits at any moment, and makes `drain-uploads` before a snapshot or maintenance appear to hang. The strict `ack-on-S3` tier is unaffected (it uploads at COMMIT, so nothing accumulates).

Observed on the QoS matrix run: `dittofs-s3`, `dittofs-s3-writeback`, `dittofs-sqlite-s3`, `dittofs-postgres-s3` all timed out at 15m with the server making progress the whole time (the 5m stall watchdog never fired). `dittofs-s3-remote` drained instantly and produced all 6 cold cells.

## Direction

Hoist the dispatcher and arena from `carveRun` up into `carveFile` so blocks pack across runs and commits overlap. Only the FastCDC chunker has to reset at a hole; the packing container does not. The complication is that the watermark/flip chain is currently per-run and would have to become per-file.

Filed from the investigation in #1767; that issue also has a second failure (evict no-op below 1GiB) which is now fixed and verified separately.

marmos91:
Closes #1872.

## The defect

`carveFile` split a files dirty set into maximal contiguous runs and carved each independently. Two consequences, both measured:

1. **One remote object per run.** `carveRun` flushed at `CarveBlockSize` *and* unconditionally at the run end, so a 4 KiB run committed its own block even with a 4 MiB `CarveBlockSize`.
2. **No overlap across runs.** `carveRun` ended with `disp.wait()`, blocking until its commits landed before the next run began. The dispatcher overlapped commits only *within* a run.

So the cost of a drain scaled with the *number of scattered writes*, not the bytes, and each was a serial round-trip. On a bench VM the scattered tail of a real drain moved ~2MiB per 30s (~68 KiB/s) with network TX matching — about 21 serial PUTs/s.

## The change

Hoist the block builder and the dispatcher from the run to the file. A block now spans runs and commits overlap across them. Only the FastCDC chunker restarts at a hole — chunk boundaries must not depend on bytes the run does not contain — while the block is just a container, and every `CarveChunk` already carries its own file offset, so a block holding chunks from several runs needs no format change.

The dispatcher now spans the whole file, which is what keeps the flip correct: watermarks ascend monotonically across runs (runs are processed in file order), so the existing ordered flip chain walks the full snapshot with one index and a block straddling two runs still commits and flips exactly once, in watermark order.

`ReapSupersededManifest` moves after the drain. It was correct at run end only because that run own rows were committed by then; with a block able to straddle runs that is no longer true until the whole file has drained. Runs are disjoint, so reaping each span afterwards never deletes a sibling run fresh rows.

## Verified

New tests, which **fail on develop** with exactly the right diagnosis:

```
--- FAIL: TestCarveScatteredRunsPackBySize
    committed 200 blocks for 800 KiB across 200 runs, want <= 1 —
    blocks are being cut per run rather than per CarveBlockSize
--- FAIL: TestCarveScatteredRunsCommitsOverlap
    peak concurrent commits = 1 across 64 blocks, want >= 2 —
    commits are serialized at run boundaries
```

With the change: 200 scattered 4 KiB writes commit **1** block instead of 200, and commits overlap.

Full `pkg/block/journal` suite green (it covers flip ordering, crash-mid-commit re-carve and dedup, which is what this refactor risks), and `go test -race ./pkg/block/...` green.

## What is NOT yet verified

This has **not** been run on hardware. The drain-rate improvement is inferred from the object-count reduction, not measured end to end, and I would rather say so than imply a number. The rig run that confirms it is the same one that closes out #1767 — I will report what it actually shows.

Reviewers: the flip path sets a records on-disk synced bit, and flipping it before the bytes are durable is silent data loss on recovery (`flipUpTo` own comment spells this out). The watermark-ordering argument above is the load-bearing part of this change and is worth attacking directly.

### DIFF
--- a/pkg/block/journal/carve.go
+++ b/pkg/block/journal/carve.go
@@ -213,7 +213,13 @@ func (s *Store) carveCandidates(sh *shard, opts CarveOptions, now, maxAge int64)
 }
 
 // carveFile snapshots one file's live dirty intervals, splits them into maximal
-// contiguous runs (a hole resets FastCDC), and carves each run.
+// contiguous runs (a hole resets FastCDC) and packs every run through one shared
+// block builder, so a single block can carry chunks from several runs.
+//
+// Ending a block at each run boundary would cost one remote object — and one
+// round-trip — per dirty range, which for a randomly written file is one per
+// write. Only the chunker has to restart at a hole; the block is just a
+// container, and every chunk carries its own file offset.
 func (s *Store) carveFile(ctx context.Context, sh *shard, id FileID, res *CarveResult) error {
 	sh.mu.Lock()
 	fi := sh.index[id]
@@ -230,37 +236,13 @@ func (s *Store) carveFile(ctx context.Context, sh *shard, id FileID, res *CarveR
 		return nil
 	}
 
-	for start := 0; start < len(snap); {
-		end := start + 1
-		for end < len(snap) && snap[end].fileOff == snap[end-1].end() {
-			end++
-		}
-		if err := s.carveRun(ctx, sh, id, snap[start:end], res); err != nil {
-			return err
-		}
-		start = end
-	}
-	s.maybeResetDirtyClock(sh, id)
-	return nil
-}
-
-// carveRun streams one contiguous dirty run through FastCDC, dedups each chunk,
-// packs novel chunks into blocks (flushed at CarveBlockSize and at the run's end),
-// and flips the run's records to synced as the durable frontier advances.
-func (s *Store) carveRun(ctx context.Context, sh *shard, id FileID, run []interval, res *CarveResult) error {
-	c := chunker.NewChunkerWithParams(s.cfg.ChunkParams)
-	rr := &runReader{s: s, sh: sh, ivs: run}
-
-	fileOff := run[0].fileOff
-	flipIdx := 0
-	// Offsets of every chunk this run tiles (novel or deduped), so the run-end
-	// reap keeps this run's own rows and deletes only superseded ones (#953).
-	newOffsets := make(map[int64]struct{})
-
 	// disp overlaps successive blocks' CommitBlock (upload + commit) while packing
 	// stays sequential. It owns the bounded worker pool, the per-block buffers and
 	// the ordered flip chain; flush hands it a completed block or a bare watermark.
-	disp := newCarveDispatcher(ctx, s, sh, id, run, res, &flipIdx)
+	// It spans the whole file, so a block straddling two runs still commits and
+	// flips exactly once, in watermark order.
+	flipIdx := 0
+	disp := newCarveDispatcher(ctx, s, sh, id, snap, res, &flipIdx)
 
 	// Each packed block gets its OWN buffer (cap one block plus one overhang chunk)
 	// so its bytes stay live while its CommitBlock runs concurrently with the next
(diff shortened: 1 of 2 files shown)


marmos91:
Do not merge — this regresses on hardware. Converting to draft.

Ran the same probe that motivated the issue (`dittofs-s3-writeback`, `large`, `rand-write-4k` + `seq-read`) against a build of this branch rebased onto develop (so it includes #1874), sampling the store every 30s during the drain:

```
drain round 0: unsynced=644MiB pending_uploads=0     (30 samples, ~15 min, flat)
FAIL dittofs-s3-writeback-nfs3: evict: dfsctl system drain-uploads: exit status 1
Error: ... context deadline exceeded (Client.Timeout exceeded while awaiting headers)
```

On develop the same drain *progresses* (49 → 34 → 45 → 22 → 2 → 2 MiB per 30s). With this change it does not move at all.

The failure mode is the informative part. The client timed out at 15m and the server never returned a 504 — with #1874 in the build, the stall watchdog only stays quiet while blocks are committing. So blocks were committing the whole time while `UnsyncedBytes` never fell, which means **records are never being flipped synced**: the carve re-carves the same ranges indefinitely, uploading without ever converging.

So the watermark-ordering argument in the PR description is wrong somewhere. Passing the whole `snap` to the dispatcher with one shared `flipIdx` does not advance the flip the way per-run slices did. That needs to be understood properly before this is re-attempted — this is the path that sets the on-disk synced bit, and #1850 is what happens when it is wrong.

The unit tests do not catch it: they assert block *count* and commit overlap, and both pass. Whatever regression test comes with the fix has to assert that a scattered carve leaves `UnsyncedBytes() == 0` under a sink that mimics the real commit path, not just that fewer blocks were written.

Keeping the branch for the diagnosis. #1872 stays open.

marmos91:
Closing in favour of #2070, which stages this work so the flip-chain change comes last.

The diagnosis in the comment above stands and is carried over. What changes is the sequencing: the serial-runs defect and the one-object-per-run defect are separable, and only the second one touches `flipUpTo`. Running the runs concurrently — each keeping its own dispatcher and its own `flipIdx` — makes the drain converge without going near the path that sets the on-disk synced bit. Whether the block hoist is needed on top of that is a question for the rig, not for a rewrite.

Not deleting the branch: `perf/1872-carve-pack-across-runs` stays as the reproduction for the stall, and step 3 of #2070 has to explain it before re-attempting. #1872 stays open.

PR #1883, opened 2026-07-28
fix(metadata): clear and reseed the quota cache on Reset and RestoreSnapshot

viquitorreis:
# Description

`Reset()` and `RestoreSnapshot()` replace a store's durable contents, but three
backends left the in-memory usage mirrors behind — the store-wide `usedBytes`
counter and the per-identity (uid/gid) quota cache. Both are caches over the
durable rows, so a stale one silently reports and enforces quota against data
that no longer exists.

Same gap in three places:

- **memory** `Reset()` cleared every map and zeroed `usedBytes`, but never
  touched the quota cache.
- **memory** `RestoreSnapshot()` recomputed `usedBytes` from the restored
  files but left the quota cache at its pre-restore contents.
- **badger** `Reset()` (`db.DropAll`) wiped every key and left both counters
  untouched.
- **postgres** `Reset()` truncated every table and left both counters
  untouched.

`sqlite` already did this correctly, which is why the divergence went
unnoticed. badger and postgres reseed correctly on *restore* (via
`initUsedBytesCounter`); only their `Reset` path was affected.

Relates to #1828

## Type of change
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] Test coverage improvement

## Changes Made
- `memory.Reset` clears the quota cache under `quotaMu`, alongside `usedBytes`.
- `memory.RestoreSnapshot` rebuilds the per-owner buckets from the restored
  files and folds them into a freshly-cleared cache, reusing the existing
  `quota.Delta` accumulator rather than hand-rolling the user/group split.
- `badger.Reset` and `postgres.Reset` zero `usedBytes` and clear the quota
  cache after the drop/truncate commits.
- New conformance coverage in `storetest`, so all four backends are held to
  the invariant:
  - `SnapshotConformance/QuotaUsageAfterRestore` — `GetQuotaUsage` after a
    restore must reflect the restored files.
  - `ResetThenRestoreConformance` — post-`Reset` usage must be zero, and
    post-`Restore` must be reseeded.
  - the 14 MiB fixture total is now a named constant shared by both suites,
    so a fixture change can't leave a stale assertion passing.

## Testing
Verified the new assertions are real guards: reverting the three source fixes
and re-running the suites reproduces the bug —

    memory:  QuotaUsageAfterRestore  = {Bytes:0 Files:0}, want {Bytes:14680064 Files:2}
    memory:  post-Reset GetQuotaUsage = {Bytes:14680064 Files:2}, want zero
    badger:  post-Reset GetQuotaUsage = {Bytes:14680064 Files:2}, want zero

- [x] `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- [x] `go test ./...` and `go test -race ./pkg/metadata/...`
- [ ] postgres integration suite (build tag `integration`) — not runnable
      locally, left to CI

## Testing Instructions
1. `go build ./...`
2. `go test -race ./pkg/metadata/store/... -run 'Conformance|Snapshot|Reset'`
3. For postgres (build tag `integration`, needs a running Postgres):

```
docker run -d --name dittofs-postgres -e POSTGRES_USER=postgres -e POSTGRES_PASSWORD=postgres -e POSTGRES_DB=dittofs_test -p 5432:5432 postgres:16
DITTOFS_TEST_POSTGRES_DSN="host=localhost port=5432 dbname=dittofs_test user=postgres password=postgres sslmode=disable" go test -race -tags=integration ./pkg/metadata/store/postgres/... -v
```

## Checklist

## Additional Notes
Split out of #1867 so the bugfix can land on `develop` independently of the
`basestore` extraction; #1867 is superseded and closed, and the extraction it
carried is tracked by #1828.

Not covered here: #1941 is the same class of staleness in a different cache —
badger's `Reset` also leaves the decoded-share cache (`shareCache`) populated,
and nothing invalidates it on restore, so a restored read-only share can keep
serving the destination's stale read-write `ShareOptions`. That is a record
cache, not a usage counter, and is untouched by this PR.

### DIFF
--- a/pkg/metadata/store/memory/snapshot_store.go
+++ b/pkg/metadata/store/memory/snapshot_store.go
@@ -408,15 +408,41 @@ func (s *MemoryMetadataStore) RestoreSnapshot(ctx context.Context, r io.Reader) 
 
 	// Re-initialize transient state.
 	s.sessions = make(map[string]*ShareSession)
+	gUID := make(map[uint32]*metadata.UsageStat)
+	uUID := make(map[uint32]*metadata.UsageStat)
 
-	// Recompute usedBytes from files (only regular files count).
+	// Recompute usedBytes from files (only regular files count)
+	// and reconstruct the quota cache from the restored files
+	// since UNIX/NFS stores based on user and groupd, we need to aggregate the usage stats for each user and group.
 	var totalBytes int64
 	for _, fd := range s.files {
 		if fd.Attr != nil && fd.Attr.Type == metadata.FileTypeRegular {
 			totalBytes += int64(fd.Attr.Size)
+
+			u := uUID[fd.Attr.UID]
+			if u == nil {
+				u = &metadata.UsageStat{}
+				uUID[fd.Attr.UID] = u
+			}
+
+			u.Bytes += int64(fd.Attr.Size)
+			u.Files++
+
+			g := gUID[fd.Attr.GID]
+			if g == nil {
+				g = &metadata.UsageStat{}
+				gUID[fd.Attr.GID] = g
+			}
+
+			g.Bytes += int64(fd.Attr.Size)
+			g.Files++
 		}
 	}
 	s.usedBytes.Store(totalBytes)
+
+	s.quotaMu.Lock()
+	s.quota.Seed(uUID, gUID)
+	s.quotaMu.Unlock()
 
 	return nil
 }
--- a/pkg/metadata/store/memory/snapshot_store.go
+++ b/pkg/metadata/store/memory/snapshot_store.go
@@ -408,30 +408,30 @@ func (s *MemoryMetadataStore) RestoreSnapshot(ctx context.Context, r io.Reader) 
 
 	// Re-initialize transient state.
 	s.sessions = make(map[string]*ShareSession)
-	gUID := make(map[uint32]*metadata.UsageStat)
-	uUID := make(map[uint32]*metadata.UsageStat)
+	groupUsage := make(map[uint32]*metadata.UsageStat)
+	userUsage := make(map[uint32]*metadata.UsageStat)
 
 	// Recompute usedBytes from files (only regular files count)
 	// and reconstruct the quota cache from the restored files
-	// since UNIX/NFS stores based on user and groupd, we need to aggregate the usage stats for each user and group.
+	// since UNIX/NFS stores based on user and group, we need to aggregate the usage stats for each user and group.
 	var totalBytes int64
 	for _, fd := range s.files {
 		if fd.Attr != nil && fd.Attr.Type == metadata.FileTypeRegular {
 			totalBytes += int64(fd.Attr.Size)
 
-			u := uUID[fd.Attr.UID]
+			u := userUsage[fd.Attr.UID]
 			if u == nil {
 				u = &metadata.UsageStat{}
-				uUID[fd.Attr.UID] = u
+				userUsage[fd.Attr.UID] = u
 			}
 
 			u.Bytes += int64(fd.Attr.Size)
 			u.Files++
 
-			g := gUID[fd.Attr.GID]
+			g := groupUsage[fd.Attr.GID]
 			if g == nil {
 				g = &metadata.UsageStat{}
-				gUID[fd.Attr.GID] = g
+				groupUsage[fd.Attr.GID] = g
 			}
 
 			g.Bytes += int64(fd.Attr.Size)
@@ -441,7 +441,7 @@ func (s *MemoryMetadataStore) RestoreSnapshot(ctx context.Context, r io.Reader) 
 	s.usedBytes.Store(totalBytes)
 
 	s.quotaMu.Lock()
-	s.quota.Seed(uUID, gUID)
+	s.quota.Seed(userUsage, groupUsage)
 	s.quotaMu.Unlock()
 
 	return nil
--- a/pkg/metadata/storetest/reset_conformance.go
+++ b/pkg/metadata/storetest/reset_conformance.go
@@ -53,6 +53,13 @@ func ResetThenRestoreConformance(t *testing.T, factory SnapshotableStoreFactory)
 		t.Fatalf("Reset: %v", err)
 	}
 
+	// Quota assertion: Reset must clear the per-identity cache, not just the
+	// share/file maps a stale cache here silently keeps enforcing limits
+	// against data that no longer exists.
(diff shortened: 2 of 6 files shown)


marmos91:
Fix looks right. Reverted the 4 prod files locally and confirmed the new assertions actually fail without them, with sqlite staying green.

One thing next door, not from this PR: badger's `Reset` drops the DB but leaves the in-memory record caches populated, and nothing invalidates them on restore either. Snapshot a read-only share, restore it over a destination whose same-named share is read-write and warm:

    post-restore GetShareOptions.ReadOnly = false, want true

`runtime/snapshot.go` calls `Reset` then `RestoreSnapshot` with no share re-registration in between, so it's reachable. Same invariant you're fixing, different failure mode — I'd give it its own issue rather than fold it in, so this stays cleanly revertable.

Postgres is the one piece I couldn't verify locally (needs the integration tag + a live DB) — worth un-drafting so that job runs before merge.

Two small nits inline.

pkg/metadata/store/memory/snapshot_store.go
```diff
@@ -408,15 +408,41 @@ func (s *MemoryMetadataStore) RestoreSnapshot(ctx context.Context, r io.Reader)
 
 	// Re-initialize transient state.
 	s.sessions = make(map[string]*ShareSession)
+	gUID := make(map[uint32]*metadata.UsageStat)
+	uUID := make(map[uint32]*metadata.UsageStat)
```
marmos91, line 412: These read as UUID / GUID at a glance. sqlite and postgres already call the same two maps userUsage / groupUsage — worth matching so the three seed sites grep as one thing.

pkg/metadata/storetest/reset_conformance.go
```diff
@@ -69,6 +76,13 @@ func ResetThenRestoreConformance(t *testing.T, factory SnapshotableStoreFactory)
 		t.Fatalf("Restore post-Reset: %v", err)
 	}
 
+	// Quota assertion: Restore must reseed the cache from the restored files,
+	// not leave it at the post-Reset zero state.
+	const wantBytes = int64((8 << 20) + (6 << 20)) // 14 MiB total
```
marmos91, line 81: Same constant is repeated in snapshot_conformance.go:883, and it is really a property of populateTestData (8 MiB + 6 MiB). Worth exporting one const from storetest next to the fixture, so changing a file size cannot leave a stale assertion passing for the wrong reason.

marmos91 (commented):
Two of the nits from my earlier comment, inline so they are easier to act on.

viquitorreis:
@marmos91 Addressed all 3: fixed the run-on comments (snapshot_store.go and reset_conformance.go), renamed gUID/uUID to groupUsage/userUsage to match sqlite/postgres, and extracted the 14 MiB fixture size into a single exported const (PopulateTestDataUsedBytes) next to populateTestData, used by all three assertion sites.

Opened #1941 for the badger shareCache staleness you found, kept separate as suggested.

Ready for another look whenever you have time.

marmos91 (approved):
LGTM

marmos91:
Picked this back up and rebased it onto current `develop` (it was ~140 commits behind). The rebase was clean — no conflicts, and nothing in the intervening drift touched these paths.

Two changes on top of the original four commits:

- `memory.RestoreSnapshot` now accumulates into the existing `quota.Delta` and folds it with `cache.Reset()` + `cache.Apply(...)`, instead of hand-rolling the uid/gid split into two `map[uint32]*UsageStat`. Same result — `Delta.Add` already buckets a file into both the user and group scopes, and every contribution here is positive so `Apply`'s clamp/delete-on-zero branches never fire (a zero-byte regular file keeps `Files:1`, exactly as `Seed` would).
- The two new conformance subtests were a line-for-line clone of each other, so `QuotaUsageAfterRestore` is folded into the existing restore test, renamed `UsageCountersAfterRestore`. Same invariants, one fewer full snapshot/restore cycle per backend. The fixture-size constant is unexported — nothing outside `storetest` referenced it.

Verified the assertions genuinely bite rather than passing vacuously: reverting the three source fixes and re-running reproduces the bug on both affected backends —

```
memory:  UsageCountersAfterRestore  GetQuotaUsage(user, 1000) = {Bytes:0 Files:0}, want {Bytes:14680064 Files:2}
memory:  post-Reset GetQuotaUsage(user, 1000) = {Bytes:14680064 Files:2}, want zero
badger:  post-Reset GetQuotaUsage(user, 1000) = {Bytes:14680064 Files:2}, want zero
```

Green: `go build ./...`, `go vet ./...`, `gofmt -l .`, full `go test ./...`, and `go test -race ./pkg/metadata/store/... ./pkg/metadata/storetest/...`. The postgres integration suite needs a live DB and is left to CI.

Also checked the lock discipline while I was in here: every `quotaMu` critical section in all four backends is a leaf that only calls `quota.Cache` methods, so `mu` → `quotaMu` is the only ordering that exists and there is no inversion to deadlock on.

Heads up: maintainer edits are off on this PR, so I could not push the rebase to the branch directly. The rebased and verified branch is on the upstream repo as `fix/quota-reset-rebased`. Either enable "Allow edits by maintainers" and I will push it here, or rebase your branch onto it.

marmos91:
Correction to my previous comment: this had already been merged (`ec39e9e`) by the time I posted it, so the note about maintainer edits and a rebased branch is moot. The rebase I did was against the pre-merge state; it was clean, and the merged content is the same fix.

Two cleanups from that pass did not make it in, since they were written on top of the rebase rather than on the merged commit. They are on `refactor/quota-restore-cleanup` (net -50 lines, no behaviour change):

- `memory.RestoreSnapshot` uses the existing `quota.Delta` accumulator instead of hand-rolling the uid/gid split into two maps. `Delta.Add` already buckets each file into both scopes, and every contribution here is positive, so `Apply`'s clamp and delete-on-zero branches never fire — a zero-byte regular file still keeps `Files:1`, exactly as `Seed` did.
- `QuotaUsageAfterRestore` was a line-for-line clone of `UsedBytesAfterRestore`; folded into one `UsageCountersAfterRestore`, saving a full snapshot/restore cycle per backend.

Green on build, vet, gofmt, and the metadata store suites with `-race`. No PR opened — say the word if you want one.

viquitorreis:
@marmos91 No worries on the maintainer-edits note, makes sense given the timing. I'll take refactor/quota-restore-cleanup and open the PR myself want to actually trace through the Delta.Add bucketing before I ship it. Appreciate the deadlock-ordering check too, good habit to pick up.

viquitorreis:
@marmos91 Went to cherry-pick bc7a1d89 onto current develop to open the PR, and it came back empty after conflict resolution the Delta/Apply change is already in develop, landed via #1947. No PR needed here, closing the loop.

Issue #1868, opened 2026-07-28
perf: SeedColdFromManifest does one IsSynced round-trip per chunk, blocking startup on large stores


marmos91:
## What

`SeedColdFromManifest` (`pkg/controlplane/runtime/shares/service.go`) issues one `metaStore.IsSynced(ctx, row.Hash)` **per chunk**, inside the per-payload loop. For a 55 GiB store at 1 MiB chunks that is roughly 56k metadata round-trips on a path that blocks startup.

This is the real driver behind the "migration looks like a hang on a large store" report in #1843. #1865 added a 5s progress heartbeat to this loop, which makes the wait legible — it does not make it shorter.

## Why it is shaped this way

The comment at the call site explains the correctness constraint, and it is a real one: whether a chunk reached the remote is answered by the synced-hash store, not by `FileChunk.State`. The carve path records synced markers and leaves the row state at `Pending` for the life of the payload, so reading `State` here would report every chunk unsynced. That trap is worth preserving — see #1850, where confusing these two is exactly what made cold ranges unrecoverable.

So the fix is not "read State instead". It is to stop paying a round-trip per chunk for an answer the store could give in bulk.

## Options

1. **Batch the lookup.** `SyncedHashStore` (`pkg/metadata/synced_hash_store.go`) exposes only single-hash `IsSynced`. Adding a batched variant — one query per payload with the payload's hash set, or a prefix scan for KV backends — collapses O(chunks) round-trips to O(payloads). Costs a new method on the interface and its conformance suite, which cuts against keeping the interface surface small, so it needs to earn its place with a measurement.
2. **Drop the per-chunk answer entirely.** `report.unsynced` is only used to decide whether the legacy archive is safe to reap. A per-payload answer ("any chunk unsynced") may be sufficient for that decision, in which case the loop can short-circuit after the first unsynced chunk in a payload.

Option 2 is cheaper and possibly enough; worth checking what consumes `report.unsynced` before adding interface surface for option 1.

## Verification

Measure first — a store large enough for the round-trips to dominate, timing the seed before and after. A fix that does not show up against the enumerate-and-seed baseline is not worth the interface change.

marmos91:
Heads-up for whoever takes this: #1881 touches the same loop in `SeedColdFromManifest` — it adds an error-level log and a counter for manifest rows whose ID cannot be placed, so the range they describe is no longer skipped in silence.

[...]

Issue #1888, opened 2026-07-29
Unplaceable-row guard is unreachable for the rows cold seeding skips, so they still read as silent zeros


marmos91:
## Summary

#1881 added a guard so that a manifest row which cannot be placed refuses the read with `ErrManifestInconsistent` instead of being treated as a sparse hole and zero-filled. The guard lives in the cold-read covering walk (`findRowCoveringOffset` / `resolveCovering`), which is only reached when the journal reports the range as **cold**.

`SeedColdFromManifest` skips exactly those rows (`pkg/controlplane/runtime/shares/service.go:3141`) — it logs, increments `report.unplaceable`, and continues without adding an extent. So the range gets no cold interval.

A range with no interval at all is a plain hole: `readAtInternal` (`pkg/block/engine/read_internal.go:34-46`) calls `local.ReadAt`, gets `cold=false`, and returns the zero-filled buffer at `if !cold { return n, nil }` — without ever consulting the manifest, and therefore without ever reaching the guard.

The two changes compose into the opposite of the intent: the rows the guard was written for are precisely the rows that seeding removes from the guard's reach.

## The claim this falsifies

The comment justifying the non-fatal skip, at `service.go:3135-3140`, states:

```
// A later read that finds nothing covering the offset it wants, on a payload
// holding a row like this, refuses with ErrManifestInconsistent rather than
// serving zeros.
```

That is true only on a read that has already been classified cold. After a migration or restore in which the row was skipped, and once the rest of the file has been hydrated warm, a read of that range is `cold=false` and returns silent zeros — on **every** metadata backend, indefinitely.

I wrote both the skip and the guard in the same PR, and the comment asserting the safety property was the reason the skip was allowed to be non-fatal. The property does not hold.

## Verification

The journal's behaviour is empirically confirmed (throwaway probe, since deleted): a journal file holding a single interval at `[4096,8192)` returns `DataExtents = [[4096 8192]]`, while `ReadAt(0, 8192)` returns `n=8192, cold=false, err=nil` with the leading page zeroed.

Mechanism in `pkg/block/journal/index.go`: `:254-262` maps gaps to `hole`; `:301-305` treats an unknown file as all-hole with `cold=false`; `:332-337` clears on `hole` without setting the cold flag — only a genuine cold interval sets it.

What is **not** verified is whether any store in the field actually holds an unplaceable row in this state. That requires a manifest dump.

## Related asymmetry, same root

[...]

Issue #1885, opened 2026-07-29
dfs never finishes startup: synchronous per-file badger scan (shares/service.go) blocks all listeners for 10+ min after any restart


raffaelegiustini-create:
## Summary

After **any restart**, `dfs` never finishes startup: it spends **10+ minutes (and counting) in a synchronous per-file scan of the badger metadata store** before it binds any listener, so NFS (12049), SMB (1445) and the control plane (8090) never come up. The drive is fully unavailable during this window, and on this dataset the window has not ended after 10 minutes.

The running process had 27h uptime and served fine — **only because it was never restarted**. The moment it is restarted (in my case, to upgrade v0.29.0 → v0.29.1 as suggested in #1879), it does not recover. Reproduced on **both v0.29.0 (`5c30c801`) and v0.29.1 (`03c0565`)**.

## Root cause — goroutine 1, `[runnable]`, on the synchronous startup path

`SIGQUIT` goroutine dump. Goroutine 1 (the main startup goroutine) is **`[runnable]`** — actively running, not blocked on a lock — here:

```
cmd/dfs/main.go:32
cmd/dfs/commands/root.go:35
cmd/dfs/commands/start.go:253
pkg/controlplane/runtime/init.go:264
pkg/controlplane/runtime/runtime.go:482
pkg/controlplane/runtime/shares/service.go:702
pkg/controlplane/runtime/shares/service.go:781
pkg/metadata/store/badger/files.go:163
pkg/metadata/store/badger/files.go:187      ← badger iterator over files
pkg/metadata/store/badger/files.go:188
pkg/metadata/store/badger/encoding.go:348
pkg/metadata/store/badger/encoding.go:357   ← encoding/json decode, per file
```

So `shares` service startup (`service.go:702 → :781`) iterates **every file** in the badger store, JSON-decoding each, **synchronously before listeners are bound**. On this store it does not complete in a usable time.

## Evidence it is a slow scan, not a lock deadlock or memory problem

- Goroutine 1 is `[runnable]` (making progress), not `semacquire`/`chan`/futex-blocked.
- `dfs` RSS is **0.2 GB** — not GC/allocation pressure. (The other busy goroutines are idle runtime GC workers.)
- `/proc/<pid>/io` `read_bytes` is **flat across samples** — working from already-mapped badger data, pure CPU.
- Sustained **~1.4 cores** (140–147% CPU) for the whole window, log completely silent after `Carve dispatcher started`.
[...]

PR #1889, opened 2026-07-29
fix(block): re-seed cold intervals a share lost under an earlier build

### LINKED ISSUE #1887: Pre-journal cold re-seed is gated on a one-shot flag, so an already-migrated share never re-seeds
## Summary

Three fixes for the pre-journal cohort — #1851 (durable `cold.log`), #1859 (verify before reaping the archive) and #1865 (on-disk format stamp) — are all gated on `MigratedFromLegacy()`, which is true **only for a share that performs the archive rename during that particular open**.

A share migrated by an earlier build returns `false` on every subsequent start, forever. It never re-enters any of those paths. So the cohort that the three fixes were written for — shares carrying pre-journal data — is exactly the cohort none of them reach, unless the share happened to migrate on the very start that first ran the patched binary.

## Evidence

`pkg/block/local/fs/fs.go:74`

```
// MigratedFromLegacy reports whether NewWithOptions archived a pre-journal
// blobs/+logs/ layout aside while opening this store.
func (s *FSStore) MigratedFromLegacy() bool {
	return s.migratedFromLegacy
}
```

`pkg/controlplane/runtime/shares/service.go:1192`

```
if m, ok := localStore.(legacyArchiveMigrator); ok && m.MigratedFromLegacy() {
	if metaStore, ok := fileChunkStore.(metadata.Store); ok {
		report, serr := SeedColdFromManifest(ctx, bs, metaStore)
		...
		if verr := finishLegacyArchiveMigration(ctx, bs, m, report, config.Name); verr != nil {
```

`SeedColdFromManifest` and `finishLegacyArchiveMigration` are both inside that gate. The PR body for #1851 named this trap and did not close it: "nothing re-seeds once legacy dirs are archived (`hasLegacyLocalLayout` false, so `MigratedFromLegacy()` never fires again)".

## Consequence

A share migrated under v0.28.x or v0.29.0 seeded its cold intervals **in memory only** — that was #1850. Those intervals were lost on the next restart, and the affected ranges became ordinary POSIX holes that `ReadAt` zero-fills with `cold=false`: no hydration attempted, no error, stable checksum on re-read.

Upgrading such a share to v0.29.1 or later does not repair it. The binary now has a durable `cold.log`, but nothing writes the missing intervals into it, because the seed is gated on a migration that already happened under the old binary. The share opens with an empty `cold.log` and the ranges stay indistinguishable from sparse holes.

The bytes are not lost — they are on the remote, and the manifest is intact. The only thing missing is the local record that says "this range lives remotely, go fetch it".

## Proposed fix

Re-seed when the evidence says the local tier does not know about ranges the manifest covers, rather than when a rename happened on this start. The cheapest correct trigger is a share that has a manifest but an absent or empty `cold.log` and no local intervals for payloads the manifest describes.

Points worth settling in review:

- `SeedColdFromManifest` is O(files) with one `IsSynced` round-trip per chunk (#1868), so an unconditional re-seed on every open is not acceptable on a large store. It needs a completion marker, in the shape of the one #1886 added for the payload index.
- A re-seed must be idempotent and safe to interrupt: partial progress loses nothing, since the remote bytes and the manifest are untouched.
- `finishLegacyArchiveMigration` reaping the archive is gated behind the same flag, so an already-migrated share never verifies its archive either. Deciding whether to reap on a re-seed path is a separate call from deciding to seed — seeding is safe, reaping is not.

## Relationship to #1879

This is **not** a claimed root cause for #1879. Lost cold intervals zero whole ranges; #1879 reports exactly the first 4096 bytes with correct content resuming at offset 4096, which this does not explain.

What it does explain is why that reporter's exposure is not removed by upgrading: their share migrated under an earlier build, so v0.29.2 gives them a durable `cold.log` with nothing in it.

marmos91:
Fixes #1887.

## The defect

`SeedColdFromManifest` was gated on `MigratedFromLegacy()`, which is true only for the one open that actually performed the pre-journal archive rename. A share migrated by v0.28.x or v0.29.0 returns false on every subsequent start, forever. Those builds seeded the cold intervals in memory only (#1850), so the intervals were lost on the next restart and the ranges became ordinary POSIX holes: `ReadAt` zero-fills with `cold=false`, no hydrate, no error, stable checksum on re-read. Upgrading did not repair it — the newer binary has a durable `cold.log` and nothing writes the missing intervals into it.

## The fix

Two halves, both load-bearing.

**1. A durable completion marker, modelled on `cfg:payload-index-backfilled` from #1886.** `journal.Store` gains `ColdSeeded()` / `MarkColdSeeded()`, backed by an empty `<journal-dir>/cold-seeded` file fsynced through the directory. The seed trigger is no longer "this open migrated something" (a one-shot the affected stores have already spent) but "this local tier has never been seeded", which stays true until a seed finishes. The marker is written **last** — after the seed's `cold.log` entries are durable and after the archive verify/reap — so an interrupted open repeats rather than skips. The marker lives next to the `cold.log` it certifies, so a wiped journal re-seeds.

The decision itself moved out of `createBlockStoreForShare` into `seedColdIfNeeded` (`legacy_verify.go`), which is what the tests drive.

**2. `SeedCold` now seeds only the ranges the file index does not already cover.** This is required, not cleanup: a cold interval carries a fresh version, so seeding over a range the journal holds locally would *shadow* those bytes — including bytes not yet on the remote, which a cold read cannot fetch back. Without it, the once-per-store re-seed would be a data-loss regression on every healthy remote-backed share. With it, the seed is idempotent, safe to repeat, and safe on a store that is only partly missing its markers.

Cost: one manifest scan per local tier, once, on the first start after upgrade — not per restart.

## Scope

- **Reaping is not extended to this path.** `finishLegacyArchiveMigration` stays behind `MigratedFromLegacy()`: seeding is safe on a share that already migrated, deleting the archive is not. A re-seeded share therefore also skips the sample read-back verification, which only exists to decide whether the archive may be deleted. If we want the re-seed to verify (there is a case for it — it would catch a share whose remote cannot serve what the manifest claims), that is a separate change with its own failure policy, since there is no archive to fall back to.
- **`RestoreSnapshot` is unaffected and must stay that way.** It calls `SeedColdFromManifest` directly, not through `seedColdIfNeeded`, so the marker never makes a restore skip re-seeding — a restore legitimately needs to re-seed a share that was seeded before. `ResetLocalState` deletes each file from the journal index, so after a reset every manifest range is a hole and the new gap-only seed covers all of it, exactly as the old unconditional one did. Deliberately not routed through the marker path.
- **One-time `IsSynced` cost not addressed.** `SeedColdFromManifest` does one `IsSynced` round-trip per chunk to build the report, and the report is only consumed by the archive reap. On the re-seed path that work is dead. Cutting it means a signature change and belongs with #1868; it is a one-time cost per store here, not a per-restart one.

## Tests

Both new `shares` tests fail on `develop` with this change reverted:

- `TestColdReseed_AfterMigrationUnderAnEarlierBuild` — plants a pre-journal share, then strips `blobs/`+`logs/` by hand so nothing on disk says a migration happened (`MigratedFromLegacy()` false, manifest and remote intact, journal empty). Without the fix: `read 4096 bytes, byte-identical=false; served wrong bytes`. With it, every file reads back byte-identical.
- `TestColdReseed_SurvivesRestartAndIsRecordedOnce` — no read before the restart (a read would hydrate the bytes as warm records and mask a seed that was never persisted); asserts the marker exists and the re-seeded intervals survive a reopen.

Journal-level:

- `TestSeedColdLeavesLocalBytesAlone` — a seed spanning a locally-written range must not turn it cold or replace it; the ranges either side must still be seeded.
- `TestSeedColdIsIdempotent` — a repeated seed appends nothing.
- `TestColdSeededMarkerSurvivesReopen`.

`go build`, `go vet`, `gofmt -l`, `go test ./...`, `go test -race ./pkg/controlplane/... ./pkg/block/...` and `golangci-lint run` (0 issues) all pass.

### DIFF
--- a/pkg/block/journal/cold.go
+++ b/pkg/block/journal/cold.go
@@ -46,6 +46,10 @@ const (
 	coldHeaderSize = 31
 	coldLogName    = "cold.log"
 
+	// coldSeededName is the marker recording that this journal's cold log has
+	// been seeded from the caller's manifest. See ColdSeeded.
+	coldSeededName = "cold-seeded"
+
 	// coldCompactFloor keeps recovery from rewriting a log that is merely small:
 	// compaction only pays once the dead entries outweigh a few pages of I/O.
 	coldCompactFloor = 1024
@@ -251,6 +255,37 @@ func liveColdEntries(indexByShard []map[FileID]*fileIndex) []coldEntry {
 		}
 	}
 	return out
+}
+
+// ColdSeeded reports whether a caller has already seeded this journal's cold
+// log from its manifest and said so with MarkColdSeeded.
+//
+// A journal that opened empty over data living only on a remote store — an
+// upgrade that archived the previous local layout aside, a restore that wiped
+// the local tier — holds no interval for those ranges, so a read zero-fills
+// instead of fetching. The repair is a manifest scan, which is O(files) and
+// worth doing once rather than on every open. A store whose marker cannot be
+// read is reported unseeded: repeating the scan costs time, skipping it serves
+// zeros.
+func (s *Store) ColdSeeded() bool {
+	_, err := os.Stat(filepath.Join(s.dir, coldSeededName))
+	return err == nil
+}
+
+// MarkColdSeeded records that the cold log has been seeded from a manifest.
+// Call it only once the seed's entries are durable: an open interrupted before
+// this point leaves no marker and seeds again, whereas a marker written early
+// would let an incomplete seed be mistaken for a finished one.
+func (s *Store) MarkColdSeeded() error {
+	fd, err := os.OpenFile(filepath.Join(s.dir, coldSeededName), os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0o644)
+	if err != nil {
+		return fmt.Errorf("journal: create cold seed marker: %w", err)
+	}
+	if err := fd.Close(); err != nil {
+		return fmt.Errorf("journal: close cold seed marker: %w", err)
+	}
+	// The marker is an empty file, so only the directory entry has to reach disk.
+	return fsyncDir(s.dir)
 }
 
 // closeCold releases the append handle. Idempotent.
--- a/pkg/block/journal/cold_test.go
+++ b/pkg/block/journal/cold_test.go
@@ -196,3 +196,118 @@ func TestColdLogTornTailKeepsIntactEntries(t *testing.T) {
 		t.Error("the entry before the torn tail was dropped; only the torn one should be lost")
 	}
 }
+
+// TestSeedColdLeavesLocalBytesAlone is what makes a re-seed safe to run against a
+// store that is only partly missing its markers. A cold interval carries a fresh
+// version, so seeding a range the journal already holds would shadow those bytes
+// — and a chunk not yet on the remote cannot be fetched back, so the read would
+// fail or return zeros for data that exists only locally.
+func TestSeedColdLeavesLocalBytesAlone(t *testing.T) {
+	ctx := context.Background()
+	dir := t.TempDir()
+	cfg := Config{ShardCount: 1, SegmentSize: minSegmentSize}
+
+	s, err := Open(dir, cfg, newFakeRemote(), newFakeClock())
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
+	defer func() { _ = s.Close() }()
+
+	// A locally-written range in the middle of what the manifest describes.
+	want := bytes.Repeat([]byte{0x3C}, 4096)
+	if err := s.WriteAt(ctx, "f", 4096, want); err != nil {
+		t.Fatalf("WriteAt: %v", err)
+	}
+	if err := s.SeedCold(ctx, "f", [][2]int64{{0, 12288}}); err != nil {
+		t.Fatalf("SeedCold: %v", err)
+	}
+
+	got := make([]byte, 4096)
+	_, cold, err := s.ReadAt(ctx, "f", 4096, got)
+	if err != nil {
+		t.Fatalf("ReadAt over the written range: %v", err)
+	}
+	if cold {
+		t.Error("the seed shadowed a range the journal holds locally; the read now needs a remote copy that may not exist")
+	}
+	if !bytes.Equal(got, want) {
+		t.Error("the seed replaced locally-written bytes")
+	}
+	// The ranges on either side had nothing local, so they must be cold.
+	for _, off := range []int64{0, 8192} {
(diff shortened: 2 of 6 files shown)


marmos91:
Two things from review. Neither is a defect in the change; both are consequences that should be decided deliberately rather than discovered in the field.

## Every healthy remote-backed share now pays the scan once

Moving the gate from "this open performed the archive rename" to "this local tier has never been seeded" is the right fix, but `ColdSeeded()` is false for every share in existence today, since no marker file has ever been written. So on first open after upgrading, every remote-backed share runs the full manifest scan with one `IsSynced` round-trip per chunk — including the overwhelming majority that were never damaged and need nothing.

That is the same shape as #1885: a one-time O(N) startup cost that blocks listeners before anything binds. I told a reporter to upgrade last week and took their box offline for ten minutes doing exactly this, so it deserves an explicit decision rather than a footnote.

The reason I am not asking for it to be gated more tightly is that I do not think it can be. A share migrated under v0.29.0 and restarted is byte-for-byte indistinguishable from a healthy one: no archive directory, empty `cold.log`, intact manifest. Any cheap predicate that skips the healthy majority also skips the damaged cohort, which is the entire population this change exists for.

So the options are: everyone pays once, or the broken shares stay broken. Paying once is probably correct — but it belongs in the release notes as a stated upgrade cost, not as a surprise. The #1865 heartbeat does at least make the wait legible rather than looking like a hang.

Worth measuring before merge: the actual duration on a store of ~10k files, so the release note can quote a number instead of a shrug.

## A re-seeded share still has silent holes where rows are unplaceable

`SeedColdFromManifest` skips unplaceable rows, so a re-seed leaves those ranges with no cold interval — which makes them plain holes that read as zeros with `cold=false`, never reaching the guard added in #1881. That is #1888.

The interaction matters here specifically: this PR is the first thing that will run cold seeding against stores that have never had it, so it is also the first thing that will hit that gap at scale. It does not block this PR — the change is strictly an improvement over never seeding at all — but #1888 should land before anyone is told their share has been repaired, or the claim will be wrong for exactly the rows that motivated it.

## Confirmed good

The gap-only `SeedCold` change is required, not cleanup, and the reasoning is right: a cold interval carries a fresh version, so seeding over a locally-held range would shadow bytes that may not be on the remote yet. `TestSeedColdLeavesLocalBytesAlone` pins the property that matters — the read stays warm and returns the written bytes. Without that half, a once-per-store re-seed would have been a data-loss regression on every healthy share.

Checking `RestoreSnapshot` and confirming it calls `SeedColdFromManifest` directly, so the marker cannot make a restore skip re-seeding, was the right instinct and the right answer.

Issue #1894, opened 2026-08-05
Cold-fetch loop skips chunks after a hole instead of the actual next chunk


marmos91:
**Where:** `pkg/block/engine/fetch.go:372`
**Dimension:** `gaps` · **Area:** block-engine-read-fetch

## What

EnsureAvailableAndRead's per-window fetch loop: when resolveCovering(cur) finds no chunk at cur (fb==nil), it jumps cur straight to the NEXT 8MiB block boundary (`cur = (cur/BlockSize + 1) * BlockSize`) instead of to the next actual chunk start. Comment justifies this with 'real files are fully written (no holes)' — but sparse holes ARE a supported, documented case elsewhere in this same package (read_internal.go:27-28 'A genuinely sparse hole ... stays zero-filled', range.go:13-15 'Sparse holes ... are NOT skipped'). BlockSize=8MiB (pkg/block/types.go:17), FastCDC avg chunk ~4MiB (read_internal.go:103-104), so a block routinely holds 2+ chunks. If a hole precedes a real chunk that starts and ends before the next block boundary, that whole chunk is skipped from the fetch list — never dispatched to dispatchRemoteFetch/hydrateChunk.

## Why it matters

Caller re-reads via local.ReadAt after this returns and trusts the window is now warm. The skipped chunk's bytes were never hydrated, so the post-fetch read either re-reports cold (silently swallowed — see next finding) or, worse, gets zero-filled as if it were a genuine hole, corrupting a read of a sparse file that has real data after a hole within the same 8MiB block.

## Verification

[...]

Issue #1879, opened 2026-07-28
At-rest data corruption: leading 4 KiB block silently zeroed on a subset of files


raffaelegiustini-create:
# At-rest data corruption: leading 4 KiB block silently zeroed on a subset of files

## Summary

Several files in my drive are **corrupted at rest**: their **first 4096 bytes (exactly 4 KiB) are all zeros**, while the rest of the file is intact. For a PDF this wipes the `%PDF-…` header (and Acrobat refuses to open it: *"unsupported type or damaged file"*); for a JPEG it wipes the `FFD8FF…` SOI marker. The tail is fine (e.g. PDFs still end with a valid `startxref … %%EOF`).

The corruption is **silent** (no error at write or read time) and **deterministic** — repeated reads return the exact same bytes (stable md5), so it is genuinely stored that way, not a transient read glitch.

## Environment

- **dfs `v0.29.0`** (commit `5c30c801`, built 2026-07-25) — *current* build.
- The affected files were **written ~2026-06-21 by an earlier build** (the box was on an older version at upload time). **Re-uploading the same bytes on v0.29.0 does NOT reproduce the corruption** — the round-trip is byte-perfect. So this looks like a *write-path bug present in an earlier version* that left permanently-damaged objects at rest.
- Backend: single remote S3 store (Cubbit DS3). `Blocks Local = 0`, all blocks remote. Block store `fs`, `max_log_bytes = 8 GiB`.
- Access verified via both the NFS loopback mount on the server **and** an independent HTTP client (filebrowser `/api/raw`) — identical corrupt bytes from both, confirming the damage is below the access layer.

## Evidence (one example)

`accordoquadrotim2025.pdf`, size **5861209 bytes** (matches the known-good original exactly):

```
first 8 bytes : 00 00 00 00 00 00 00 00      # should be 25 50 44 46 ("%PDF")
leading zeros : 4096 bytes (exactly 4 KiB), first non-zero byte at offset 4096
"%PDF" search : not found anywhere in the file
tail (last 32): … 3391\n>>\nstartxref\n5860680\n%%EOF\n   # valid PDF trailer, intact
```

So exactly the **first 4 KiB block** is zeroed; everything after offset 4096 is the correct original content.

## Scope

In a single month folder (`Archivio/2026-06`, ~60 files), **5 files** were affected, mixed types:

- `accordoquadrotim2025.pdf` (5.86 MB)
- `P00011.pdf` (498 KB)
[...]

PR #1903, opened 2026-08-05
fix(block): fetch chunks after a sparse hole instead of skipping to the block boundary

marmos91:
Relates to #1894.

## Root cause

`EnsureAvailableAndRead` walks a read window resolving the chunk covering each offset. When `resolveCovering` returned nil — a sparse hole — the loop advanced `cur` to the next 8 MiB `BlockSize` boundary, justified in a comment by "real files are fully written (no holes)".

That justification is contradicted by this same package: `read_internal.go` documents that a genuinely sparse hole stays zero-filled, `range.go` that sparse holes are not skipped, and NFSv4.2 sparse is a shipped feature. FastCDC chunks average ~4 MiB against an 8 MiB `BlockSize`, so a block routinely holds two or more. A hole followed by real data inside the same block meant every chunk up to the boundary was dropped from the fetch list — never dispatched, never hydrated. The caller then re-reads via `local.ReadAt` trusting the window is warm, and the range comes back as zeros. Same silent-zeros class as #1879 / #1888.

## Fix

Step to the next actual chunk start instead of the block boundary, via `resolveNextChunkStart` — a sibling of `resolveCovering` in `read_internal.go` that mirrors its optional-interface shape.

The badger backend already had `GetFileChunkAtOrAfterOffset`, purpose-built for exactly this ("skip a hole straight to the next chunk in one indexed keys-only scan"), conformance-tested in `storetest`, but never wired to any caller. It is now the indexed path; other backends fall back to a manifest walk, matching what `resolveCovering` already does for them.

Two details worth flagging:

- The successor is queried at `off+1`, not `off`. A zero-length row parked exactly at `off` would otherwise answer with `off` itself and stall the loop.
- On the indexed path, a successor row whose ID cannot be placed is reported as `ErrManifestInconsistent` rather than skipped. Silently stepping over it would strand real data behind a hole and reintroduce the same silent-zero read. Both paths report `ErrManifestInconsistent` for an unplaceable successor row. An earlier revision guarded only the indexed path, on the reasoning that reaching the fallback proved the manifest held no unplaceable rows. **That reasoning was wrong**, for two independent reasons: the two lookups assert for their interfaces separately (`chunkAtOffsetResolver` vs `chunkAtOrAfterOffsetResolver`), so a backend indexing coverage but not succession takes the indexed branch for the first — never enumerating rows, so the unplaceable check never runs — and the walk for the second; and they are separate `ListFileChunks` snapshots, so a concurrent write or GC between them can introduce a row the first walk never saw. Either way the row would have been skipped and its bytes reclassified as hole — the very silent-zero read this PR closes.

## Test

`TestEnsureAvailableAndRead_ChunkAfterHoleIsFetched` — hole over [0, 1 MiB), real cold/evicted chunk at [1 MiB, 3 MiB), both inside the first block; read spans [0, 4 MiB). It asserts the chunk is hydrated and the real bytes come back, with the destination poisoned to 0xAA so an unhydrated range fails rather than hiding in zeros. It runs over both store shapes, since they resolve the successor by different means.

Verified failing before the fix (`git stash` of the two source files):

```
--- FAIL: TestEnsureAvailableAndRead_ChunkAfterHoleIsFetched
    --- FAIL: .../ListFileChunksFallback
        chunk after hole not hydrated: got 00000000000000000000000000000000…, want 2571fcbfb038cc6e174270550895d96e…
    --- FAIL: .../OffsetIndexed
        chunk after hole not hydrated: got 00000000000000000000000000000000…, want 2571fcbfb038cc6e174270550895d96e…
```

All zeros on both — the reported symptom exactly.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l .` clean
- `go test ./pkg/block/...` all ok
- `go test -race ./pkg/block/engine/...` ok (78.5s)

### DIFF
--- a/pkg/block/engine/cold_read_hole_test.go
+++ b/pkg/block/engine/cold_read_hole_test.go
@@ -1,0 +1,123 @@
+package engine
+
+import (
+	"bytes"
+	"context"
+	"math/rand"
+	"testing"
+
+	"github.com/marmos91/dittofs/pkg/block"
+	memorylocal "github.com/marmos91/dittofs/pkg/block/local/memory"
+	remotememory "github.com/marmos91/dittofs/pkg/block/remote/memory"
+	metadatamemory "github.com/marmos91/dittofs/pkg/metadata/store/memory"
+)
+
+// indexedChunkStore gives stubFileChunkStore the two offset-indexed lookups the
+// badger backend implements, so resolveCovering and resolveNextChunkStart take
+// their indexed branches instead of the ListFileChunks fallback. Both answers
+// are derived from the same rows the fallback would walk, so the two store
+// shapes must agree.
+type indexedChunkStore struct{ *stubFileChunkStore }
+
+func (s *indexedChunkStore) GetFileChunkAtOffset(ctx context.Context, payloadID string, off uint64) (*block.FileChunk, error) {
+	rows, err := s.ListFileChunks(ctx, payloadID)
+	if err != nil {
+		return nil, err
+	}
+	rw, err := findRowCoveringOffset(rows, off)
+	if err != nil || rw == nil {
+		return nil, err
+	}
+	return rw.fb, nil
+}
+
+func (s *indexedChunkStore) GetFileChunkAtOrAfterOffset(ctx context.Context, payloadID string, off uint64) (*block.FileChunk, error) {
+	rows, err := s.ListFileChunks(ctx, payloadID)
+	if err != nil {
+		return nil, err
+	}
+	var (
+		best    *block.FileChunk
+		bestOff uint64
+	)
+	for _, fb := range rows {
+		abs, ok := block.ParseChunkOffset(fb.ID)
+		if !ok || abs < off {
+			continue
+		}
+		if best == nil || abs < bestOff {
+			best, bestOff = fb, abs
+		}
+	}
+	return best, nil
+}
+
+// TestEnsureAvailableAndRead_ChunkAfterHoleIsFetched pins the sparse-hole step
+// in the per-window fetch loop. The payload is a hole over [0, 1 MiB) with a
+// real remote-resident chunk at [1 MiB, 3 MiB) — both inside the first 8 MiB
+// block, which is the routine shape because FastCDC chunks average well under
+// BlockSize.
+//
+// Advancing to the next block boundary on the uncovered first byte discards
+// every chunk up to that boundary, so the chunk after the hole is never
+// dispatched and never hydrated; the caller's post-fetch re-read then serves it
+// as zeros. The loop must instead resume at the next chunk start, so the read
+// spanning [0, 4 MiB) leaves the chunk warm in the local tier.
+//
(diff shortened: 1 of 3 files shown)


marmos91:
Both Copilot findings addressed in fa178c18. The first was a real bug, not a comment problem.

## Finding 1 — real defect, fixed in code

My comment claimed the fallback needed no unplaceable-row guard because `findRowCoveringOffset` would already have failed `resolveCovering` for that payload. Copilot is right that this is not guaranteed, and the reasoning fails for two independent reasons:

1. **The two lookups assert for their interfaces independently.** `resolveCovering` checks for `chunkAtOffsetResolver`; `resolveNextChunkStart` checks for `chunkAtOrAfterOffsetResolver`. Nothing couples them. A backend that indexes coverage but not succession takes the indexed branch for the first (which never enumerates rows, so the unplaceable check never runs at all) and the walk for the second. `gatedChunkStore` in syncer_shutdown_test.go is already exactly this shape — it implements `GetFileChunkAtOffset` and not the successor lookup.
2. **They are separate snapshots.** Each calls `ListFileChunks` on its own, so a concurrent write or GC between the two can introduce a row the first walk never saw.

The consequence is the silent-zeros class this PR exists to close: the scan skips the unplaceable row, the loop treats the range up to the next placeable start as hole, and those bytes come back zero-filled.

Fixed by reporting rather than skipping, on both branches. Note this is deliberately stricter than `findRowCoveringOffset`, which returns a covering row even when unplaceable rows exist elsewhere. That scoping is sound for coverage — the answer for that offset is unaffected. It is not sound for succession: an unplaceable row sits at an unknown offset, so it may itself be the true successor, and returning a later start would reclassify the bytes it holds as hole. So the successor lookup errors whenever such a row is present, even if a placeable candidate was found.

## Finding 2 — guarded

`off+1` wrapped at `math.MaxUint64`. Confirmed reachable in a unit test: without the guard the probe restarts at 0 and returns `(0, true)` — a successor *before* `off`, breaking the strictly-after contract exactly as described. Early-returns "no successor" now.

## Tests

`TestResolveNextChunkStart_UnplaceableRowIsReported` (both branches) and `TestResolveNextChunkStart_MaxOffsetDoesNotWrap`. Verified failing against the pre-guard code:

```
--- FAIL: TestResolveNextChunkStart_UnplaceableRowIsReported/ListFileChunksFallback
    resolveNextChunkStart = <nil>; want ErrManifestInconsistent
--- FAIL: TestResolveNextChunkStart_UnplaceableRowIsReported/OffsetIndexed
    resolveNextChunkStart = <nil>; want ErrManifestInconsistent
--- FAIL: TestResolveNextChunkStart_MaxOffsetDoesNotWrap
    resolveNextChunkStart(MaxUint64) = (0, true); want no successor
```

`go build ./...`, `go vet ./...`, `gofmt -l .` clean; `go test ./pkg/block/...` ok; `go test -race ./pkg/block/engine/...` ok (85s).

Issue #1899, opened 2026-08-05
WRITE_THROUGH/WRITE_UNBUFFERED request flag decoded but never honored


marmos91:
**Where:** `internal/adapter/smb/handlers/write.go:97`
**Dimension:** `gaps` · **Area:** smb-read-write
**Independently re-confirmed** by two separate audit runs with different lenses.

## What

DecodeWriteRequest parses req.Flags (write.go:97) into WriteRequest.Flags, whose doc comment (write.go:47-52) names bit 0x1 write-through and 0x2 unbuffered. Write() never reads req.Flags again anywhere in the function body — it always runs the same PrepareWrite -> WriteToBlockStore -> CommitWrite -> deferred-flush path (write.go:412-451). WriteToBlockStore (internal/adapter/common/write_payload.go) is a bare passthrough to engine.WriteAt with no sync flag, and the metadata flush is explicitly deferred (write.go:442-448, 'Relaxed (#1687): ... we can defer the metadata db.Sync off the per-WRITE ack path'). handlePipeWrite (write.go:528) also never looks at req.Flags.

## Why it matters

Per MS-SMB2 3.3.5.13, SMB2_WRITEFLAG_WRITE_THROUGH tells the server the write must be durable at the underlying object store before the response is sent, and WRITE_THROUGH/WRITE_UNBUFFERED on a named-pipe write in the 3.x dialect family MUST be rejected with STATUS_INVALID_PARAMETER. Here a client that sets write-through (databases, Hyper-V-over-SMB, FILE_FLAG_WRITE_THROUGH opens) gets ack'd exactly like a normal cached write, with data/metadata durability deferred to CLOSE/FLUSH — silently dropping the durability guarantee the client explicitly requested and depends on for crash correctness. The named-pipe rejection case is also completely unimplemented.

## Verification

[...]

Issue #1892, opened 2026-08-05
READ unconditionally does a synchronous metadata write (atime bump) on every read call


marmos91:
**Where:** `internal/adapter/smb/handlers/read.go:440`
**Dimension:** `perf` · **Area:** adapter-smb-rw
**Independently re-confirmed** by two separate audit runs with different lenses.

## Verification

Confirmed: read.go:440-443 calls metaSvc.SetFileAttributes(Atime) on every successful READ unless IsAtimeFrozen, with no throttle. Service.SetFileAttributes (file_modify.go:217+) does GetFile + permission checks + a persisting store write; the dirTimes coalescing path applies only to directories (dirTimeSet gate at 236), so file reads take the full read-modify-write. One metadata write per READ RPC on the SMB data plane — classic atime amplification, and the codebase already debounces the mtime-on-write analogue.

## Suggested fix

Debounce/batch atime updates the same way mtime writes are already debounced (e.g. only bump atime once per N seconds per open, mirroring smbDelayedWriteWindow), or defer via the existing FlushPendingWrite-style async path instead of a synchronous per-READ SetFileAttributes call.

---
Found by a data-plane audit of `origin/develop` @ 7eeec0da (metadata ↔ block store hot write/read path). Confirmed from source, proven reachable from a non-test caller, and checked against the governing spec before filing. Full report with the 322 MED/LOW findings is kept local, not filed.

marmos91:
Tracked under #1900 (data-plane audit umbrella).

marmos91:
Fixed in #1905 (squash-merged to develop as e6e6a416).

Per-READ atime writes are now coalesced through the existing per-`OpenFile` timestamp state, flushed on CLOSE. `DirTimesTracker` was deliberately not extended — its pending map has no evictor (bounded for directories, unbounded for files).

Two real defects were caught in review and fixed: concurrent READs reaching the lock out of sample order could move atime **backwards**, and the CLOSE flush could stomp a newer atime persisted by another handle's SET_INFO. Both now hold an only-ever-advance discipline. Note the CLOSE guard narrows rather than closes the window — it is a read-then-write with no cross-handle lock and no CAS in the store, stated as such in the code.

**Not covered, same defect class:** `write.go:482`, `query_directory.go:586`, and `ioctl_copychunk.go:565` all still do an unconditional per-op `SetFileAttributes(Atime)`.

PR #1904, opened 2026-08-05
fix(smb): honor WRITE_THROUGH by flushing durably before acking the write

marmos91:
## What

SMB2 WRITE decoded the request `Flags` field but never read it again, so a client setting `SMB2_WRITEFLAG_WRITE_THROUGH` (databases, Hyper-V over SMB, `FILE_FLAG_WRITE_THROUGH` opens) was acknowledged exactly like a cached write. Both halves of durability stayed deferred off the ack path — the block-store commit and the metadata fsync — silently dropping the crash-durability guarantee the client explicitly asked for.

Three changes:

1. **Write-through takes a synchronous durability path.** On `SMB2_WRITEFLAG_WRITE_THROUGH`, WRITE now runs the same strict pair the FLUSH handler runs — `blockStore.Flush(payloadID)` then `FlushPendingWriteForFile(..., durable=true)` — before the response is produced, and returns a mapped error instead of acking a guarantee that does not hold. No new primitives; both calls are the ones `flush.go` already uses.

2. **Named-pipe writes reject the bits.** `handlePipeWrite` returns `STATUS_INVALID_PARAMETER` for `WRITE_THROUGH`/`WRITE_UNBUFFERED`.

3. **The default path is untouched.** The non-write-through branch is byte-identical to before; the only added cost is one bit test that short-circuits before any dereference.

## Spec basis

- **MS-SMB2 2.2.21** — `SMB2_WRITEFLAG_WRITE_THROUGH` (0x1) "is not valid for the SMB 2.0.2 dialect"; `SMB2_WRITEFLAG_WRITE_UNBUFFERED` (0x2) "is not valid for the SMB 2.0.2, 2.1, and 3.0 dialects". Undefined bits are ignored.
- **MS-SMB2 3.3.5.13** — "If Connection.Dialect is not 2.0.2, and SMB2_WRITEFLAG_WRITE_THROUGH is set ... the server SHOULD indicate to the underlying object store that the write is to be written to underlying storage before completion is returned." Also: "If the server implements the SMB 3.x dialect family, if the write is executed on a named pipe and the Flags field is set to SMB2_WRITEFLAG_WRITE_UNBUFFERED or SMB2_WRITEFLAG_WRITE_THROUGH, the server MUST fail the request with STATUS_INVALID_PARAMETER."

## The two dialect gates (deliberately different)

They key off different things because the spec words them differently:

- **Honor gate** (write-through durability) is on **`Connection.Dialect`** — skipped only when the negotiated dialect is 2.0.2, where the bit is undefined. A connection with no recorded dialect honors the bit: over-flushing is safe, silently dropping a requested durability point is not.
- **Pipe reject gate** is on **what the server implements**, not what the connection negotiated, so it keys off the configured `MaxDialect` (`>= Dialect0300`). This is not a blanket reject: the repo default `MaxDialect` is `0x0210`, so a default-configured server keeps ignoring the bits exactly as the 2.x dialects require.

`WRITE_UNBUFFERED` on a regular file is treated as the caching hint it is, not a durability request — on its own it does not pull the strict commit onto the ack path.

## Tests

The repo has been burned before by durability tests that pass against a fake sink while never exercising real durability, so these assert on **the durability calls actually made**, not on the returned status. The metadata store is wrapped so the durable (`WithTransaction`) and relaxed (`WithTransactionRelaxed`) commit entrypoints are tallied separately, with both wrappers delegating to a real memory store. Supplying `WithTransactionRelaxed` is what makes the two countable — the plain memory store does not implement `metadata.RelaxedTransactor`, so without the wrapper the relaxed path silently collapses onto `WithTransaction` and the test would measure nothing.

- write-through → durable commit > 0, relaxed == 0
- default → durable == 0, relaxed > 0 (the deferred path, unchanged)
- unbuffered alone → durable == 0
- write-through on 2.0.2 → durable == 0 (bit ignored)
- pipe writes → table over six server/flag combinations, asserting reject on 3.0/3.1.1 and no flag-caused failure on 2.0.2/2.1

## Known boundaries

- The **block-store half** of the durability pair (`engine.Store.Flush`) has no injection seam reachable from a handler test: the engine resolves its local tier at share start, and closing the store fails the earlier `WriteAt` rather than the flush. It is the identical call FLUSH and CLOSE make, covered by the block-store conformance suites. Noted in the test file header.
- A share that opts into the **writeback tier** downgrades `durable=true` to `false` in `FlushPendingWriteForFile`. Write-through therefore inherits exactly the same metadata semantics as an explicit SMB FLUSH on such a share — pre-existing and deliberate, not introduced here. The block-store commit still runs.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .` all clean. `go test ./internal/adapter/smb/... ./pkg/block/...` and `go test -race ./internal/adapter/smb/...` all pass.

Relates to #1899

## Error path

A failed durability step does not return early. `CommitWrite` has already applied the size/mtime mutation to the pending-write state, so other sessions can observe the change through GETINFO whether or not the fsync landed. Returning early would skip the post-write bookkeeping — frozen-timestamp restore and the CHANGE_NOTIFY change notification — leaving watchers unaware of a modification they can already see, and dropping a SET_INFO(-1) timestamp freeze. The bookkeeping therefore still runs and the failure status is carried to the response instead.

### DIFF
--- a/internal/adapter/smb/handlers/close_flush_error_test.go
+++ b/internal/adapter/smb/handlers/close_flush_error_test.go
@@ -28,12 +28,13 @@ import (
 	metamemory "github.com/marmos91/dittofs/pkg/metadata/store/memory"
 )
 
-// setupFlushErrorShare wires a handler + runtime + memory metadata + memory
-// block store with a single read-write share, writes a small payload to a file
-// named "data" under the share root, and registers an OpenFile for it. Returns
-// the handler, a primed SMBHandlerContext, the file's metadata handle, and the
-// FileID of the registered OpenFile.
-func setupFlushErrorShare(t *testing.T) (*Handler, *SMBHandlerContext, metadata.FileHandle, [16]byte) {
+// setupWriteTestShare wires a handler + runtime + the caller's metadata store +
+// a memory block store with a single read-write share, writes a small payload to
+// a file named "data" under the share root, and registers an OpenFile for it. A
+// nil metaStore uses a plain memory metadata store. Returns the handler, a
+// primed SMBHandlerContext, the file's metadata handle, and the FileID of the
+// registered OpenFile.
+func setupWriteTestShare(t *testing.T, metaStore metadata.Store) (*Handler, *SMBHandlerContext, metadata.FileHandle, [16]byte) {
 	t.Helper()
 	ctx := context.Background()
 
@@ -49,7 +50,9 @@ func setupFlushErrorShare(t *testing.T) (*Handler, *SMBHandlerContext, metadata.
 	if _, err := cps.CreateMetadataStore(ctx, &models.MetadataStoreConfig{Name: "flushmeta", Type: "memory"}); err != nil {
 		t.Fatalf("CreateMetadataStore: %v", err)
 	}
-	metaStore := metamemory.NewMemoryMetadataStoreWithDefaults()
+	if metaStore == nil {
+		metaStore = metamemory.NewMemoryMetadataStoreWithDefaults()
+	}
 	if err := rt.RegisterMetadataStore("flushmeta", metaStore); err != nil {
 		t.Fatalf("RegisterMetadataStore: %v", err)
 	}
@@ -173,7 +176,7 @@ func setupFlushErrorShare(t *testing.T) (*Handler, *SMBHandlerContext, metadata.
 // engine.ErrStoreClosed (the same mapping seam fs.ErrPressureTimeout and the
 // other block-store content errors travel through).
 func TestClose_FlushFailure_SurfacedAsNonSuccess(t *testing.T) {
-	h, smbCtx, fileHandle, fileID := setupFlushErrorShare(t)
+	h, smbCtx, fileHandle, fileID := setupWriteTestShare(t, nil)
 
 	// Force the durable flush to fail: close the share's block store so the
 	// CLOSE-time engine.Flush short-circuits with engine.ErrStoreClosed.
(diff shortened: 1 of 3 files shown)


marmos91:
Assessed the review finding on the transaction tally. The factual claim does not hold for this fixture, but the robustness concern behind it does, and the tests are now hardened against it.

**Measured, not reasoned.** Instrumented the counter to print the raw tally for a single WRITE:

```
flags=0 (default)      status=0x00000000 durable=0 relaxed=1
flags=1 (WRITE_THROUGH) status=0x00000000 durable=1 relaxed=0
```

Exactly one counted transaction per WRITE in both cases. The per-WRITE atime update and the parent-directory atime update contribute **zero** to the tally — `SetFileAttributes` does not reach the `WithTransaction` / `WithTransactionRelaxed` entrypoints the wrapper counts on this store. So `relaxed` was not being inflated and the `relaxed == 0` assertion was not passing for the wrong reason. Had it been, that assertion would already have been failing rather than passing.

**The assertion has teeth.** Mutation test — reverting the fix by flipping the write-through flush back to `durable=false`:

```
--- FAIL: TestWrite_WriteThrough_CommitsMetadataDurably
    write-through WRITE performed no durable metadata commit (durable=0 relaxed=1):
    the requested durability point was dropped
```

**What changed anyway.** The suggested remedy — freezing atime and clearing `ParentHandle` — would tune the fixture to exactly the behavior #1892 is about to change, so I went the other way and made the measurement independent of it. Assertions are now stated as a *delta* between two runs over the same fixture that differ only in the flag, via a new `writeCommitDelta` helper. Any metadata write the WRITE path performs for its own reasons lands in both tallies and cancels, so the numbers stay pinned to the durability decision whether or not atime traffic starts or stops routing through those entrypoints:

- write-through: `durable +1`, `relaxed -1` (the durable commit *replaces* the deferred flush rather than running alongside it — a strictly stronger claim than the old `relaxed == 0`)
- unbuffered alone, and write-through on 2.0.2: `durable 0`, `relaxed 0`

`TestWrite_Default_StaysDeferred` deliberately keeps an **absolute** `durable == 0`. "Nothing on the default ack path fsyncs" is the invariant itself, so anything that starts taking the durable entrypoint during a plain WRITE should fail there regardless of cause. It is also what gives the delta its meaning: the baseline being subtracted is a measured zero, not an assumed one.

Re-verified after the change: `go build ./...`, `go vet ./...`, `gofmt -l .` clean; `go test ./internal/adapter/smb/... ./pkg/block/...` and `go test -race ./internal/adapter/smb/...` all pass; mutation test still fails with `added 0 durable metadata commit(s) over a plain WRITE, want 1`.

PR #1905, opened 2026-08-05
perf(smb): coalesce per-READ atime updates instead of writing metadata on every read

marmos91:
Relates to #1892.

## Problem

Every successful SMB READ called `metaSvc.SetFileAttributes(Atime)` synchronously unless the handle had a frozen LastAccessTime, with no throttle. `Service.SetFileAttributes` does a `GetFile` plus permission checks plus a persisting store write, so a sequential read cost one metadata read-modify-write per READ RPC on the data plane. The existing `DirTimesTracker` coalescing in the metadata service covers directories only.

## Fix

Coalesce the bumps per open handle, in the SMB adapter, next to the existing per-handle timestamp state:

- the first access on a handle still writes through, so open/read/close clients see unchanged store traffic;
- further accesses inside a 2s window are held in `OpenFile.SmbPendingAtime`;
- QUERY_INFO overlays the held value, so the newest access time stays observable via GETINFO/stat (a newer stored value from another handle still wins);
- SET_INFO touching LastAccessTime drops the held value, so a client's explicit set is never flushed over;
- CLOSE persists the held value, so a coalesced bump is not lost with the handle;
- `IsAtimeFrozen` still suppresses both the bump and the CLOSE flush;
- out-of-order samples from concurrent READ pipelines on one handle cannot lower the access time.

## Why not reuse an existing mechanism

- `DirTimesTracker` (pkg/metadata) is the closest analogue, but its pending map has no evictor. That is fine for a bounded set of directories and unbounded for files. It also lives below the point where the distinction matters: an explicit client `touch -a` reaches `Service.SetFileAttributes` with the same `SetAttrs{Atime}` as an implicit read bump, so the metadata layer cannot tell them apart.
- `smbDelayedWriteWindow` / sticky-write-time is the same shape of per-handle state and lives in the same file, which is where the new helpers were added, but it serves a different semantic (show the pre-write Mtime until the window expires) so folding them together would add conditionals rather than remove them.
- The FlushPendingWrite deferred path is keyed to write coordination and would not avoid the read-modify-write.

## Tests

New `internal/adapter/smb/handlers/timestamps_atime_test.go` drives the READ and CLOSE atime steps against a real metadata service and counts the `SetFileAttributes` calls that actually reach the store:

- 8 reads inside the window produce exactly 1 store write, and the store holds the first read's time (the other seven were elided, not overwritten);
- the newest access is still visible through the QUERY_INFO overlay, and a newer stored value wins;
- CLOSE persists the coalesced time, and a second CLOSE writes nothing;
- crossing the window writes again, so atime keeps advancing durably on long-lived handles;
- a frozen LastAccessTime suppresses the bump and the CLOSE flush;
- an out-of-order sample does not lower the access time.

Both behaviours were mutation-checked: with `noteSmbAccess` forced to always write (the pre-fix behaviour) the write-count test reports 8 writes instead of 1, and without the monotonic guard the out-of-order test reports the earlier time.

```
go build ./... && go vet ./... && gofmt -l .   # clean
go test ./internal/adapter/smb/... ./pkg/metadata/...   # all ok
go test -race ./internal/adapter/smb/...                # all ok
```

### DIFF
--- a/internal/adapter/smb/handlers/close.go
+++ b/internal/adapter/smb/handlers/close.go
@@ -263,6 +263,26 @@ func (h *Handler) Close(ctx *SMBHandlerContext, req *CloseRequest) (*CloseRespon
 				logger.Debug("CLOSE: metadata flushed", "path", openFile.Path)
 			}
 
+			// READ coalesces its LastAccessTime bumps, so the newest access time
+			// may still only live on the handle. Persist it before the
+			// frozen-timestamp restore below, which stays authoritative.
+			//
+			// Only ever advance it: another handle on the same file (or an NFS
+			// client) may have set a newer LastAccessTime since this handle's
+			// last READ, and SetFileAttributes assigns Atime directly, so an
+			// unconditional flush would move it backwards. The compare is a plain
+			// read-then-write — openFile.mu is per-handle and the store offers no
+			// compare-and-set for a single attribute — so it narrows that window
+			// rather than closing it. Losing the race costs at most
+			// smbAtimeUpdateWindow of access-time precision.
+			if pending := takeSmbPendingAtime(openFile); !pending.IsZero() && !openFile.IsAtimeFrozen() {
+				if cur, curErr := metaSvc.GetFile(authCtx.Context, openFile.MetadataHandle); curErr == nil && cur.Atime.Before(pending) {
+					if _, atimeErr := metaSvc.SetFileAttributes(authCtx, openFile.MetadataHandle, &metadata.SetAttrs{Atime: &pending}); atimeErr != nil {
+						logger.Warn("CLOSE: LastAccessTime flush failed", "path", openFile.Path, "error", atimeErr)
+					}
+				}
+			}
+
 			// Per MS-FSA 2.1.5.14.2: After flushing pending writes (which may overwrite
 			// frozen timestamps), restore any timestamps that were frozen via SET_INFO -1.
 			// The deferred commit flush sets Mtime/Ctime to the WRITE time, but if the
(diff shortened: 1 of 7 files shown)


marmos91:
Addressed the Copilot finding on the CLOSE flush — it is real.

Nothing serializes two opens on the same file: `openFile.mu` is per-handle, and `Service.SetFileAttributes` assigns `file.Atime = *attrs.Atime` directly (pkg/metadata/file_modify.go:475). So handle B's explicit SET_INFO (or an NFS client's SETATTR, or B's own read-through) can land between handle A's last READ sample and A's CLOSE, and the unconditional flush would stomp it backwards.

Fixed with the same monotonic discipline already applied to the coalesce path: CLOSE now reads the current attributes and flushes only when the pending time is strictly newer. This also skips the write entirely when the store is already ahead, so it costs nothing in the common case.

Explicit about what this does and does not buy: the compare is a plain read-then-write, not a compare-and-set. There is no per-file lock spanning both handles and the metadata store exposes no CAS for a single attribute, so the window is narrowed, not closed. The comment in close.go says exactly that. Losing the race costs at most `smbAtimeUpdateWindow` (2s) of access-time precision on a field Windows itself updates lazily — not worth a new cross-handle lock or a new store API.

Regression test `TestSmbAtime_CloseDoesNotLowerAnotherHandlesAtime`: handle coalesces a read, another handle sets a newer LastAccessTime, handle closes, stored value must stay the newer one. Verified it fails without the guard:

```
--- FAIL: TestSmbAtime_CloseDoesNotLowerAnotherHandlesAtime
    stored atime after close = 16:50:45.191, want the other handle's newer value 17:50:45.141
```

Not changed: the READ write-through path still assigns `time.Now()` unconditionally, so a client that set a *future* LastAccessTime via SET_INFO will see a subsequent READ lower it. That is pre-existing behaviour from before this PR and is what MS-FSA 2.1.5.4 specifies (a read sets LastAccessTime to the current time), so it is left alone.

```
go build ./... && go vet ./... && gofmt -l .   # clean
go test ./internal/adapter/smb/... ./pkg/metadata/...   # 26 packages ok, 0 failures
go test -race ./internal/adapter/smb/...                # 12 packages ok, 0 failures
```

Issue #1914, opened 2026-08-05
evictSegment leaks a segment busy claim when appendCold fails, permanently blocking reclamation


marmos91:
Found by the data-plane audit of `origin/develop` @ 7eeec0da (umbrella #1900), graded MED there. Re-triaged as data-integrity/security priority: unrecoverable space leak without a restart

### [MED] evictSegment leaks a segment's busy claim forever when appendCold fails, permanently blocking reclamation

- **Where:** `pkg/block/journal/reclaim.go:207` · `bugs` · area: block-journal-replay-gc
- **Verified:** Confirmed: evictSegment's appendCold branch returns `0, err` with no busy reset, and evict() (reclaim.go:94-97) just propagates. Sibling paths do reset — grep for busy.Store(false) hits only reclaim.go:406 (reclaimEmptied error branch) and 495 (gcShard after repack); the claim CAS sites are 182/402/491. A stuck busy permanently excludes the segment from evictable() (190) and pickVictim, so a single appendCold I/O/ENOSPC failure strands the segment (and its bytes) for the process lifetime — precisely under the disk-pressure condition eviction serves. MED: needs an appendCold failure, and impact is a space leak, not data loss.
- **Fix:** Reset the claim on failure, mirroring gcShard/reclaimEmptied: in evictSegment's appendCold error branch (and/or in evict()'s caller), add `seg.busy.Store(false)` before returning the error so a future eviction/GC pass can retry the segment.

---

### [MED] evictSegment scans the whole shard index once per evicted segment

- **Where:** `pkg/block/journal/reclaim.go:207` · `perf` · area: block-journal-replay-gc
- **Verified:** CONFIRMED: evictSegment (207-240) walks all of sh.index twice (collect coldEntry, then flip cold) filtering loc.SegmentID==seg.id, each pass under sh.mu; evict() (64-104) loops evictSegment until targetBytes is met. No reverse segment->intervals index exists, so N evicted segments cost O(N×I). REACHABLE from two non-test entrypoints: Evict (shares DrainLocalSynced admin drain, which evicts many segments) and ensureSpace, the write-path capacity gate.
- **Fix:** When targetBytes requires evicting multiple segments, snapshot per-segment interval buckets once (single pass over sh.index) up front and consume from that snapshot across the evict loop, instead of re-scanning the full index per segment.

marmos91:
Fixed in #1922 (squash-merged to develop).

[...]

PR #1922, opened 2026-08-05
fix(journal): release the segment claim when eviction's cold-log append fails

marmos91:
Relates to #1914.

## The leak

`evictSegment` takes an exclusive per-segment `busy` claim (CAS in `claimColdestEvictable`) and never released it when `appendCold` failed — it returned `0, err` straight through `evict()`. A stuck claim permanently excludes the segment from `evictable()`, `pickVictim`, and `reclaimEmptied`, so one cold-log I/O or ENOSPC failure strands that segment and its bytes for the life of the process, recoverable only by a restart — exactly under the disk pressure eviction exists to relieve.

Fixed with a `defer` over named returns that drops the claim on any failure, so it cannot rot when a future error path is added. Success still retains the claim, as before: `retireSegment` has already dropped the segment from `sh.sealed`, so the claim is unreachable.

## Making the retry actually safe

Releasing the claim only helps if a failed append leaves nothing half-done. It did not. `appendCold` packs the whole batch into one buffer and issues a single `Write`; Go loops the underlying syscall, so an ENOSPC part-way through returns an error *after* some bytes are already on disk — a torn entry at the tail. Nothing truncated it back and the `O_APPEND` fd stayed open, so the retry (or any later eviction sharing the fd) appended a valid batch *behind* the tear. `loadCold` stops at the first tear, so those valid, fsynced markers are dropped on every subsequent load — the segment gets unlinked and the range reads back as a zero-filled hole instead of fetching from remote.

`appendCold` now records the tail before the batch and truncates back to it when the write or fsync fails, restoring the log’s "only the tail can tear" invariant so a retry appends onto intact bytes.

## Audit of the other claim sites

Both other `busy` CAS sites were traced and are correct as-is:

- `reclaimEmptied` — releases explicitly on `retireSegment` failure; retains on success (segment already out of `sh.sealed`).
- `gcShard` — releases *unconditionally* after `repackSegment`. That is deliberate and must not be converted to the release-on-error shape: `repackSegment` has non-error paths that do not retire the victim (the defensive `remaining != 0` branch and the `testStopBeforeUnlink` seam), and those need the claim back.

## Test

`TestEvictReleasesClaimAfterColdLogFailure` forces the append to fail, asserts the segment survives, then restores the cold log and asserts the *same* segment is now reclaimable. Verified to fail before the fix on the claim assertion specifically (not merely on error propagation):

```
evict_test.go:186: segment 0 stayed claimed after the failed eviction: stat err=<nil>
--- FAIL: TestEvictReleasesClaimAfterColdLogFailure
```

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .` clean. `go test ./pkg/block/...` and `go test -race ./pkg/block/journal/...` all pass.

## Not addressed

The second (perf) finding on #1914 — `evictSegment` scanning the whole shard index once per evicted segment — is deliberately left out; it needs a restructured evict loop with a snapshotted per-segment bucket, which would enlarge this diff and its risk for no correctness gain.

### DIFF
--- a/pkg/block/journal/cold.go
+++ b/pkg/block/journal/cold.go
@@ -155,11 +155,23 @@ func (s *Store) appendCold(entries []coldEntry) error {
 	if len(buf) == 0 {
 		return nil
 	}
+	// A write that fails part-way (ENOSPC is the likely one: eviction runs because
+	// the volume this log shares with the segments is filling up) leaves a torn
+	// entry at the tail. Replay stops at the first tear, so any entry a later
+	// append puts behind it is lost for good — silent zeros for a range the caller
+	// went on to evict. Roll the tail back to where the batch started so the log
+	// keeps its "only the tail can tear" shape and the caller's retry appends onto
+	// intact bytes.
+	st, err := s.coldFD.Stat()
+	if err != nil {
+		return fmt.Errorf("journal: stat cold log: %w", err)
+	}
+	tail := st.Size()
 	if _, err := s.coldFD.Write(buf); err != nil {
-		return fmt.Errorf("journal: append cold log: %w", err)
+		return errors.Join(fmt.Errorf("journal: append cold log: %w", err), s.coldFD.Truncate(tail))
 	}
 	if err := s.coldFD.Sync(); err != nil {
-		return fmt.Errorf("journal: fsync cold log: %w", err)
+		return errors.Join(fmt.Errorf("journal: fsync cold log: %w", err), s.coldFD.Truncate(tail))
 	}
 	return nil
 }
--- a/pkg/block/journal/evict_test.go
+++ b/pkg/block/journal/evict_test.go
@@ -145,6 +145,48 @@ func TestEvictSkipsClaimedSegment(t *testing.T) {
 	}
 }
 
+// TestEvictReleasesClaimAfterColdLogFailure: a failed cold-log append must leave
+// the segment reclaimable. Eviction's claim is exclusive, so keeping it past the
+// failure would bar that segment from every later eviction and GC pass until the
+// process restarts, stranding its bytes under the disk pressure eviction relieves.
+func TestEvictReleasesClaimAfterColdLogFailure(t *testing.T) {
+	s, _ := evictStore(t, Config{})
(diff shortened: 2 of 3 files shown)


marmos91:
Checked this one against the source rather than from memory — the premise does not hold here, so leaving the code as-is. Evidence:

**1. The concrete type at the call site is `*os.File`, not an interface.**

`store.go:208` declares `coldFD *os.File`. Within the whole package it is assigned in exactly one place, `cold.go:141`, straight from `os.OpenFile` (which returns `*os.File`); the only other writes are `s.coldFD = nil` at `cold.go:211` and `cold.go:311`. There is no `io.Writer` indirection, no `bufio` wrapper, and no test double anywhere in the path, so the standard library guarantee does apply.

**2. `(*os.File).Write` cannot return a short write with a nil error.**

From `$(go env GOROOT)/src/os/file.go` in the toolchain that builds this repo (go1.26.5; `go.mod` requires >= 1.25.0):

```go
func (f *File) Write(b []byte) (n int, err error) {
	if err := f.checkValid("write"); err != nil {
		return 0, err
	}
	n, e := f.write(b)
	if n < 0 {
		n = 0
	}
	if n != len(b) {
		err = io.ErrShortWrite
	}
	epipecheck(f, e)
	if e != nil {
		err = f.wrapErr("write", e)
	}
	return n, err
}
```

`n != len(b)` sets `err = io.ErrShortWrite` unconditionally. The `e != nil` branch below can only *replace* that with the more specific underlying error (so an ENOSPC short write reports ENOSPC rather than `ErrShortWrite`) — it can never clear it. Both cases are non-nil, so both already take the rollback branch this PR adds. A partial write reaching `Sync()` with `err == nil` is not reachable.

**3. Not adding an `n != len(buf)` guard.**

It would be provably dead code for a concrete `*os.File`, and worse, it would imply to the next reader that short-write-with-nil-error is a live hazard on this path when the type system and the stdlib contract rule it out 40 lines from the declaration. The invariant that actually needed enforcing on this durability path — that a *failed* write leaves no torn bytes behind for a later append to hide behind — is what the rollback in this PR does.

Issue #1912, opened 2026-08-05
Copy paths bypass CRC verification: carve upload and GC repack can launder corrupt bytes into the remote store and new segments


marmos91:
Found by the data-plane audit of `origin/develop` @ 7eeec0da (umbrella #1900), graded MED there. Re-triaged as data-integrity/security priority: these do not corrupt data, they promote already-corrupt data to trusted, defeating checksums elsewhere in the pipeline

### [MED] Carve reads dirty local bytes for remote upload with an unverified raw pread — corrupted bytes get hashed and committed to the remote store as genuine content

- **Where:** `pkg/block/journal/carve.go:571` · `gaps` · area: block-journal-replay-recovery
- **Verified:** CONFIRMED. runReader.Read (carve.go:558-576) calls s.readPayload(rr.sh, iv.loc, rr.off, ...); readPayload (segment.go:~542) is a bare seg.fd.ReadAt with no CRC and no verifyReads gate — its own doc even notes ReadAt has the resolve-and-guard/verify path it skips. Carve hashes exactly those bytes (BLAKE3) and commits/uploads them, so latent local bit rot becomes permanent content-addressed remote data instead of a CorruptRangeError. Reachable: carveDispatcher→local.Carve (carve_dispatch.go:118) in prod, plus Flush/ManualSync. Fix: route runReader through the CRC-verifying record read (or verify record CRC before hashing) so carve fails closed.
- **Fix:** Gate runReader.Read (or readPayload itself, for this call site) on the store's verifyReads path — read via readRecordAt/verifiedRead and return an error instead of raw bytes when the record's CRC does not match, so carveRun aborts that run instead of uploading unverified/corrupted content.

---

### [MED] GC repack copies victim payload with raw ReadAt, bypassing the store's own CRC verification — silently launders bit-rot into the new segment

- **Where:** `pkg/block/journal/reclaim.go:675` · `gaps` · area: block-journal-replay-recovery
[...]

PR #1925, opened 2026-08-05
fix(block): verify record CRCs before copying payloads to remote or a new segment

marmos91:
Relates to #1912

## Problem

Three paths copied a journal record's payload forward with a raw, unverified pread and then re-hashed or re-CRC'd the copy. None of them *cause* corruption — they **promote already-corrupt bytes to trusted**, which is worse: the promoted copy becomes authoritative and passes every downstream integrity check.

1. **Carve upload** (`carve.go` `runReader.Read`) — BLAKE3-hashes the bytes and commits them to the remote store. Latent local bit rot becomes permanent content-addressed remote data that matches its own hash.
2. **GC repack** (`reclaim.go` `repackSegment`) — copies victim payload into a new segment, where `writeDataRecord` stamps a **fresh valid CRC** over the rotted bytes.
3. **Snapshot restore** (`store.go` `RestoreToVersion`) — re-materializes bytes as fresh records the same way. Not in the original report; found while grepping the other callers.

The store already had the right machinery and the read path already used it (`verifiedRead`, gated on `SetVerifyReads`, which production enables for non-writeback shares). So the read path failed closed while these three laundered.

`victimMarkers` had a related hole: `scanValidRecords` stops at the first bad record, so every tombstone and truncate marker behind it was silently dropped — a delete or size-down would come back to life once the victim was reclaimed.

## Fix

One shared primitive in `record.go`, reused by all four sites rather than hand-rolled per call site:

- `readVerifiedRecord` — header CRC + payload CRC + framed-FileID check. The FileID check matters because neither CRC covers those bytes, so a flipped FileID (or a stale `recOff`) otherwise passes and hands back another file's payload.
- `record.payloadRange` — bounds-checked slice of the verified payload by absolute segment offset.

`verifiedRead` was refactored onto the same primitive (identical semantics, less duplication), and `segment.go`'s unverified `readPayload` is deleted rather than left as a footgun.

## Failure policy, per site

The two sites need different answers:

- **Carve**: fail the run. The bytes stay dirty and local, nothing is uploaded, and the record's synced bit is not flipped. There is no remote copy to heal from for never-uploaded data, so loud failure is the only correct outcome — and the state stays recoverable.
- **GC repack**: keep the victim intact and abort. Silently dropping a victim's payload is unacceptable, and so is copying it forward under a fresh CRC. The victim is then quarantined via an in-memory `segmentMeta.corrupt` flag that `pickVictim` skips, so one damaged segment surfaces once instead of wedging reclamation for the whole shard (which would turn a data-integrity alarm into a disk-full outage).

## Cost

`BenchmarkCarve` (8 MiB, 6 runs, medians): **~9.9 ms/op → ~11.9 ms/op**, roughly **20% slower** carve throughput. The cost is a CRC32C pass plus one copy out of the verified record buffer instead of a pread straight into the chunker's buffer — you cannot trust any byte of a record until you have read all of it. Repack pays the same CRC pass, plus read amplification when only a fragment of a record is still live. Stated trade, not an accident: promoting corrupt bytes to authoritative is not worth 20%.

## Tests

`verify_copy_test.go` injects **real corruption** — byte flips written into the segment file through a separate fd — and asserts on the production entry points (`Carve`, `GC`), not internals:

- `TestCarveRefusesCorruptedRecord` — payload flip; asserts `*CorruptRangeError`, **zero bytes reaching the sink**, unsynced count unchanged, synced bit not flipped.
- `TestRepackRefusesTruncatedRecordStream` — payload flip; asserts GC refuses, victim survives, no orphan target segment, victim quarantined, and a second pass is a clean no-op.
- `TestRepackRefusesRecordFramingAnotherFile` — FileID flip. Distinct path: the record stream stays fully scannable (CRCs don't cover the FileID), so this exercises the move-loop check rather than the marker scan.

All three were confirmed to **fail against pre-fix source** (`SegmentsRepacked:1` and `err=<nil>` where the fix now errors), so they are not tautological.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .` clean; `go test ./pkg/block/...` and `go test -race ./pkg/block/journal/...` green.

### DIFF
--- a/pkg/block/journal/carve.go
+++ b/pkg/block/journal/carve.go
@@ -249,7 +249,7 @@ func (s *Store) carveFile(ctx context.Context, sh *shard, id FileID, res *CarveR
 // and flips the run's records to synced as the durable frontier advances.
 func (s *Store) carveRun(ctx context.Context, sh *shard, id FileID, run []interval, res *CarveResult) error {
 	c := chunker.NewChunkerWithParams(s.cfg.ChunkParams)
-	rr := &runReader{s: s, sh: sh, ivs: run}
+	rr := &runReader{s: s, sh: sh, id: id, ivs: run}
 
 	fileOff := run[0].fileOff
 	flipIdx := 0
@@ -544,15 +544,27 @@ func (s *Store) maybeResetDirtyClock(sh *shard, id FileID) {
 	fi.firstDirtyNanos = 0
 }
 
-// runReader streams a contiguous run's live bytes in file-offset order by
-// preading each interval's payload. It reuses the store's positioned-read path so
-// reads race nothing (segment fds are stable once created).
+// runReader streams a contiguous run's live bytes in file-offset order, serving
+// each interval out of its owning record's CRC-verified payload. Reads race
+// nothing (segment fds are stable once created).
+//
+// The verification is load-bearing rather than defensive: carve hashes whatever
+// it reads and commits it to the remote store as content-addressed data, so a
+// raw pread here would promote latent local bit rot to the authoritative copy of
+// the file. A failed check aborts the run instead, leaving the bytes dirty and
+// local — recoverable, and loud.
 type runReader struct {
 	s   *Store
 	sh  *shard
+	id  FileID
 	ivs []interval
 	i   int   // current interval
 	off int64 // bytes already read from the current interval
+	// rec is the verified record backing the current interval, kept so a record
+	// spanning several Read calls or several intervals is read and verified once
+	// rather than per call. recSeg pairs with rec.segOff to identify it.
+	rec    *record
+	recSeg uint64
 }
 
 func (rr *runReader) Read(p []byte) (int, error) {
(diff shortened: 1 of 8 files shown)


marmos91:
Checked this one rather than taking it on faith, because Go optimizes exactly this shape — and it does here. **No allocation occurs; not changing the code.**

The expression at `record.go:217` is a *direct comparison*:

```go
if string(rec.fileID) != string(id) {
```

The compiler elides `[]byte`→`string` conversions that appear directly in a comparison, lowering to a length check plus `runtime.memequal` with no copy. (`id` is `FileID`, whose underlying type is `string`, so `string(id)` is a no-op conversion.) The optimization only applies while the converted value stays inside the comparison — assigning it, passing it to a function, or using it as a map key would allocate. Neither happens here.

Two independent measurements:

**1. Generated assembly** (`go build -gcflags=-S`), the function body:

```
0x00b0 00176 (record.go:217)	CALL	runtime.memequal(SB)
```

`runtime.slicebytetostring` does not appear anywhere in `readVerifiedRecord`.

**2. Allocation count** (`testing.AllocsPerRun`, 1000 iterations, realistic 32-byte FileID):

```
allocs/op for `string(rec.fileID) != string(id)` = 0
```

So the suggested rewrite would trade identical machine code for a less readable expression. `bytes.Equal(rec.fileID, []byte(fid))` would actually be *worse* — that `[]byte(fid)` conversion is not in a comparison and would allocate.

Worth noting the allocation that *does* exist on this path is the record payload buffer in `readRecordAt`, which is inherent: you cannot verify a CRC without having read the whole payload. That is the ~20% carve cost documented in the PR description, and it is the intended trade.

Issue #1909, opened 2026-08-05
Crash-ordering silent zeros: acknowledged writes return correct size with all-zero content after device loss


marmos91:
## Summary

Under a real device-loss crash, an acknowledged write can come back as **correct file size with all-zero content** — a silent wrong-data read, not a truncation and not an error. Measured on hardware, not inferred from code.

The loss of *unsynced* data on a power cut is expected. The bug is the **asymmetry**: metadata size reaches durable storage while the block data does not, so the range degrades into an ordinary sparse hole and `ReadAt` zero-fills it. The client gets `EOK` and plausible-looking bytes. A correct outcome would be either a shorter file (size lost with the data) or a read error — never silent zeros.

This is the same observable failure as #1879 / #1888 / #1894, reached by a different mechanism: crash-ordering rather than an unplaceable manifest row.

## Evidence (SCW POP2-8C-32G, Ubuntu 6.8.0-106, ext4 on `dm-flakey` over a loop device)

Crash model: `dmsetup suspend --noflush --nolockfs`, then reload the table with `drop_writes`, so only bytes that genuinely reached the device survive. The writer is `SIGSTOP`ed **before** the device switch, so every acknowledged record in the log was answered while the device was healthy. Records are 4096 B, self-identifying (`REC%08d` + SHA-256), with the ACK log `fsync`ed to a separate disk outside the flaky device.

| Cell | Build | WRITE_THROUGH | Tier | ACKed | expected/actual bytes | bad |
|---|---|---|---|---|---|---|
| A2 | `develop` (pre-#1904) | on (ignored) | local | 1995 | 8171520 / 8171520 | **1995** |
| C2 | with #1904 | off | local | 2020 | 8273920 / 8273920 | **2020** |

Failure signature, identical in both:

```
first_bad=(0, 'ZEROS', 4096)   last_bad=(1994, 'ZEROS', 4096)
```

**File size came back exactly right. Every byte of content came back zero.** Record 0 was acknowledged ~6 seconds before the crash and still read as zeros.

Controls that make this meaningful:
- With #1904 **and** the flag set (cell B2): 0 bad out of 524 — so the rig can distinguish, and the box was not fsyncing regardless.
- `kill -9` does **not** reproduce this (both cells survived it) — page cache outlives process death. Only device-level loss exposes it.

## Why this matters beyond "don't crash"

- A reader cannot detect it. No error, no short read, checksum-free path returns zeros as if written.
[...]

PR #1929, opened 2026-08-05
fix(block,metadata): hold a committed file size to the durable extent

marmos91:
Relates to #1909.

Under real device loss an acknowledged write came back with the correct file size and all-zero content. This is the crash-ordering sibling of #1879 / #1888 / #1894: same observable failure, different mechanism, so their fixes do not cover it.

## Diagnosed ordering

On the deferred write ack (SMB WRITE, NFS UNSTABLE WRITE):

1. `common.WriteToBlockStore` -> `engine.WriteAt` -> `journal.appendRecord`, which never fsyncs (its own doc comment says so). The bytes sit in the page cache.
2. `metadata.FlushPendingWriteForFile(..., durable=false)` -> `flushPendingWrite` -> `withRelaxedTransaction`, committing the grown `File.Size`.

That size then reaches stable storage on the metadata store's own schedule, entirely decoupled from the data:

- badger relaxed: `runDurabilitySync` fsyncs every `durabilitySyncInterval` (100ms), unconditionally.
- badger strict / sqlite / postgres: `withRelaxedTransaction` falls back to the fully durable `WithTransaction`, so the size is durable inline, at ack time.

The journal only fsyncs at an explicit commit (`Store.Commit` -> `groupCommit` -> `segSync`), a segment seal, or close. A write stream not followed by COMMIT/CLOSE therefore leaves durable size ahead of durable data permanently, not for a 100ms window. That matches the reported signature exactly, including record 0 (acked ~6s before the crash) reading as zeros: its size was fsynced within 100ms, its bytes never were.

**`reconcileMetadataSizeFromJournal` is not responsible.** It grows size only up to `localStore.FileSize`, computed from the index `recover()` rebuilds from records that physically survived and passed both header and payload CRC (`readRecord` / `errTornRecord`; the tail scan stops at the first torn record). That is the physically-present extent — a subset of what was durable. It also never shrinks, so it can neither cause the bug nor fix it.

## Fix: a committed size may never describe non-durable bytes

Clamping at commit time, which is simultaneously the ordering fix.

- **journal**: each shard tracks `lastVersion` (stamped under `sh.mu` once a record's writes return) and `syncedVersion` (raised after a completed fsync in `groupCommit`, after a segment seal, seeded to the replayed maximum at recovery). `DurableExtent` reports how far a file's bytes are on stable storage.
- **engine / metadata / shares**: `engine.Store.DurableExtent` exposes it upward, and `metadata.Service` gains a resolver installed by the shares service, which owns both tiers. A local tier that cannot answer returns "unknown", never "nothing is durable".
- **`flushPendingWrite`** holds the committed size to that extent. Held-back bytes stay in the pending-write state, so readers still see the acknowledged size through the pending merge, and the first flush after the data is committed publishes them.

Every existing durability sync point already commits the block store before flushing metadata (`flushStableWrite` for NFS FILE_SYNC, NFS COMMIT, SMB CLOSE, SMB FLUSH), so the clamp is a no-op there and none of them change behaviour. The write path stays asynchronous — no new fsync anywhere on it.

This pairs with the restart-time half that already existed: a persisted size never runs ahead of durable data, and share start grows it back up to that data.

### Two sharp edges, both covered by tests

**Out-of-order writes.** Intervals are sorted by offset but durability is keyed on Version (append order), so a durable interval can sit *above* a non-durable one. Taking the max end over qualifying intervals would jump the gap and report a size covering the range that vanishes — the original bug, relocated. The scan therefore stops at the first non-durable interval. Gaps nobody wrote are a different thing entirely: genuine sparse holes that read as zeros correctly and must not hold the extent back, or a client writing only at a high offset would never get a size.

**fsyncgate.** Once the kernel reports a write-back failure it drops those pages, and the next fsync can return success without them ever reaching the device. A shard freezes its watermark permanently on the first fsync failure, mirroring the stickiness `groupCommit` already applies to `errSeq`/`syncErr` for the same reason.

### Why not the alternatives

- **Ordering by fsyncing the journal before the ack** is #1904's write-through; it destroys the deferred write path.
- **Clamping down at reconcile time** would truncate genuine sparse files (truncate-grow legitimately leaves size above any written extent) and remote-only extents. Clamping at commit time has neither problem: truncate goes through SetAttr, not this path, and reconcile still only ever grows.
- **Distinguishability** — making the range recoverable as not-a-hole, the way `cold.log` did for evicted ranges — needs a durable marker written at ack time. `cold.log` can afford its fsync (`cold.go` fsyncs before returning) because eviction is a background event; a per-write ack marker cannot, which is write-through again. Written without an fsync it dies in the same crash as the data.

This is not silent-zeros traded for silent-truncation. An NFS UNSTABLE WRITE is explicitly *not* a durability promise (RFC 1813 — the client must COMMIT, and the reply carries `Committed: UNSTABLE`), and neither is an SMB write without write-through. Losing an uncommitted write is the contract; the shorter file is what the protocol expects, and the write verifier (`serverBootTime`) changing across a restart is exactly how the client learns to re-send. A read error would instead poison the range against that recovery path.

### Per QoS tier (`resolveDurabilityTier`)

- **local** (default): the tier the bug was measured on. Size now trails the journal's fsynced extent.
- **writeback**: downgrades per-op FILE_SYNC metadata flushes to relaxed, widening the gap. Same clamp, same invariant.
- **remote** (`require_durable_commit`): CLOSE/COMMIT block until data is durable remotely, and those paths commit the journal first, so the extent covers everything and nothing is held back.

## Tests

`pkg/block/journal/durable_extent_test.go`:
- **device loss, not process death** — records committed, then more acked but never committed, then the segment is physically truncated back to its length at the last fsync and reopened. What survives is exactly what `DurableExtent` predicted. A plain reopen proves nothing here: the page cache serves un-fsynced bytes straight back.
- out-of-order durable-above-non-durable, genuine sparse holes, and the fsync-failure freeze.

`pkg/controlplane/runtime/size_durability_test.go` — the merge gate, across memory/badger/sqlite: performs the adapter's write ack with no block commit and asserts the persisted size never exceeds the durable extent, that the acked size stays visible through the pending state, and that committing the data releases the full size (a delay, not a truncation). Without the clamp it fails on all three backends with `persisted size 262144 exceeds the durable extent 0`.

Each assertion was confirmed to fail with its guard removed.

## Also fixed

`PendingWritesTracker.RecordWrite` never filled `PayloadID` on its update branch, so the field was empty whenever `PrepareWrite`'s `SetCachedFile` had created the state first — the common case. Nothing read it before; the clamp does.

## Not covered

`CommitWrite`'s non-deferred path publishes size without routing through `flushPendingWrite`, so it is unclamped. It is unreachable in production — `SetDeferredCommit(false)` has test-only callers — so it is left alone rather than duplicating the lookup into dead code.

### DIFF
--- a/pkg/block/engine/engine.go
+++ b/pkg/block/engine/engine.go
@@ -448,6 +448,23 @@ func (bs *Store) LocalForTest() local.LocalStore { return bs.local }
 // metadata file sizes against the journal's durable high-water mark (#1687).
 func (bs *Store) Local() local.LocalStore { return bs.local }
 
+// DurableExtent reports how far a payload's bytes are on stable storage: bytes
+// below the returned offset survive an unclean shutdown, bytes above it were
+// only buffered and are gone after one. ok is false when the local tier cannot
+// answer (no local store, or one with no durability model), which callers must
+// read as "unknown", never as "nothing is durable". Callers publish file sizes
+// against it so a persisted size never describes bytes a crash would take away,
+// which would leave the range reading as a hole full of zeros.
+func (bs *Store) DurableExtent(ctx context.Context, payloadID metadata.PayloadID) (int64, bool) {
+	reporter, ok := bs.local.(interface {
+		DurableExtent(ctx context.Context, payloadID string) (int64, bool)
+	})
+	if !ok {
+		return 0, false
+	}
+	return reporter.DurableExtent(ctx, string(payloadID))
+}
+
 // LocalStore returns nil: the journal-backed local tier is a per-file byte
 // cache, not a content-addressed block.Store, so there is no hash-namespace to
 // sweep. The journal self-manages local segment reclaim (dead-byte GC +
--- a/pkg/block/journal/durable_extent_test.go
+++ b/pkg/block/journal/durable_extent_test.go
@@ -1,0 +1,217 @@
+package journal
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"os"
+	"testing"
+)
+
+// TestDurableExtent_PredictsWhatSurvivesDeviceLoss is the proof behind the size
+// invariant: whatever DurableExtent reports before an unclean shutdown is
+// exactly what the store still has after one. A caller that publishes a file
+// size at or below that number can never end up describing bytes the device
+// never took — bytes whose absence would read back as a hole full of zeros
+// instead of an error.
+//
+// The crash model is device loss, not process death: reopening from the same
+// tmpfiles would otherwise be served the un-fsynced tail straight out of the
+// page cache and prove nothing. So the segment is physically truncated back to
+// the length it had at the last fsync, which is precisely what a device that
+// dropped every unflushed write would have kept.
+func TestDurableExtent_PredictsWhatSurvivesDeviceLoss(t *testing.T) {
+	const rec = 4096
+	dir := t.TempDir()
+	cfg := Config{ShardCount: 1}
+	s, err := Open(dir, cfg, newFakeRemote(), SystemClock())
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
+	ctx := context.Background()
+	id := FileID("acked.bin")
+
+	write := func(recIdx int, fill byte) {
+		t.Helper()
+		if err := s.WriteAt(ctx, id, int64(recIdx)*rec, bytes.Repeat([]byte{fill}, rec)); err != nil {
+			t.Fatalf("WriteAt rec %d: %v", recIdx, err)
+		}
+	}
+
+	// Two records the client was told are on stable storage.
+	write(0, 'a')
+	write(1, 'b')
+	if err := s.Commit(ctx, id); err != nil {
+		t.Fatalf("Commit: %v", err)
+	}
+	durable, ok := s.DurableExtent(ctx, id)
+	if !ok || durable != 2*rec {
+		t.Fatalf("DurableExtent after Commit = (%d, %v), want (%d, true)", durable, ok, 2*rec)
+	}
+	durableTail := s.shards[0].active.tail.Load()
+	segID := s.shards[0].active.id
+
+	// Two more the client acknowledged but nothing ever committed. They are in
+	// the page cache: readable now, gone after device loss.
+	write(2, 'c')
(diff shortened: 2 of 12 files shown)


marmos91:
Copilot finding addressed in 07171629 — with a correction to the suggested shape.

**Facts first.** A negative extent is unreachable today. `journal.DurableExtent` returns either `(0, false)` or a `size` that starts at 0 and is only ever assigned `iv.end()` when it grows; `iv.end()` is `fileOff + length`, and `appendRecord` rejects `offset < 0` outright (`segment.go:268`). Every hop above it (`durableExtentFor`, `shares.durableExtent`, `engine.DurableExtent`, `FSStore.DurableExtent`) either returns literal `(0, false)` or passes the value through untouched. There is no `-1` sentinel anywhere in the chain — "unknown" travels in the separate `ok` bool, never in band.

**But the guard is not dead code, and the suggested form was not enough.** Two things:

1. The cast does not sit behind the journal. It sits behind `Service.SetDurableExtentResolver`, which is exported — any caller can install a resolver returning anything, so the input is not fixed by construction at the point of the widening.

2. `end >= 0 &&` would have prevented the wrap but still **failed open**: a negative would skip the clamp and publish the size unclamped, which is the same outcome as the wrap. For this invariant the unusable answer has to withhold everything, not everything-but-nothing. So the extent is floored (`uint64(max(end, 0))`) rather than guarded — a broken resolver now holds the size back at 0.

Confirmed by test rather than asserted: `TestDeferredSizeCommit_NonsenseExtentWithholdsEverything` installs a resolver returning `(-1, true)` and requires the persisted size to stay 0. Replacing the floor with a plain cast fails it with `persisted size = 65536 on a negative extent, want 0`.

Worth noting the contrast with the `os.File` short-write guard declined on #1922: there the stdlib contract made it unreachable and the failure would have been loud. Here the value crosses an injectable exported boundary and the failure is silent — the clamp stops clamping, no error, no log, and #1909 is quietly back.

`go build`/`go vet`/`gofmt` clean; `./pkg/block/... ./pkg/metadata/... ./pkg/controlplane/...` and `-race ./pkg/block/...` all pass. Rebased onto current develop (#1927).

Issue #1911, opened 2026-08-05
Truncate/PunchHole leave manifest rows past the new boundary; cold-read hydrate resurrects removed bytes (no clamp in Hydrate seam)


marmos91:
## Summary

`hydrateChunk` writes a chunk's **full** `fb.DataSize` at the chunk's absolute offset with no clamp against the file's current size or hole map. Two callers leave manifest rows whose byte range extends past a boundary the file has since moved, so a cold read can resurrect bytes that were logically truncated away or punched out.

Both were found by the data-plane audit (#1900) as MED, and the umbrella recorded them as wanting **the same clamp in the `Hydrate` seam** as #1894 — which merged as #1903 without them. Filing so the remaining half is not lost.

## Mechanism (Truncate, `pkg/block/engine/readwrite.go:151`)

- The drop predicate is only `b.Offset >= newSize` (`readwrite.go:151-160`), so a chunk **straddling** `newSize` is kept verbatim, retaining its original `Hash`/`Size`.
- journal `Store.Truncate` (`store.go:708-730`) only *clips* the local interval to `newSize`; it does not re-mark the range dirty, so nothing re-carves the straddling manifest row. Unlike the PunchHole case there are no dirty bytes, so the tail chunk's segment is already fully synced and immediately evictable (`reclaim.go:190`).
- After eviction, a cold read at any offset inside that chunk resolves the un-narrowed row, and `hydrateChunk` (`fetch.go:83`) writes the full `fb.DataSize` at the chunk's absolute offset **with no size clamp**.
- The hydrate record carries a higher version than the truncate marker — `store.go:683-685` explicitly does not fence later versions — so bytes past `newSize` come back into the local index.
- Re-extending the file then serves those stale bytes where a zero hole is required.

`PunchHole` (`readwrite.go:262`) has the same shape at the hole boundary.

## Impact

Bounded but real. It needs eviction + a cold read + a re-extend to become user-visible, and metadata `Size` still clamps ordinary reads — which is why the audit graded it MED rather than HIGH. But the outcome is stale data served where zeros are required, which is a correctness failure of the same family as #1879 / #1888 / #1894 / #1909, and it silently grows local storage past the file's own size in the meantime.

## Suggested fix

[...]

PR #1933, opened 2026-08-05
fix(block): hold a cold-read hydrate to the bytes its manifest row still claims

marmos91:
Relates to #1911.

## Confirmed, and partly refuted

**Truncate half: confirmed.** A truncate landing inside a chunk keeps that chunk s manifest row verbatim (`readwrite.go`, drop predicate is only `b.Offset >= newSize`). journal `Store.Truncate` merely clips the local interval and never re-marks it dirty, so nothing re-carves the row and the tail segment — already fully synced — is immediately evictable. After eviction a cold read resolves the un-narrowed row and `hydrateChunk` writes the full `DataSize` at the chunk s absolute offset, restoring the removed tail above the version of the truncate marker. Re-extending then serves those stale bytes where a zero hole is due.

The new test fails on develop with:

```
re-extended hole: byte 0 is 0x3a, want a zero hole — removed bytes were resurrected
```

**PunchHole half: refuted as reported.** The punch overwrites the range with zeros through the local write path, and the carve of those zeros supersedes the punched rows, so the manifest still tiles the file exactly and the hole resolves only to the zero rows. The punch test passes both before and after this change; it is kept as a regression guard and now also asserts the tiling structurally (`assertManifestTiles`), so it is a structural refutation rather than one lucky green run.

## The fix

Two halves, both load-bearing:

1. **The shared seam** (`fetch.go` `hydrateChunk`) writes at most the bytes the row claims. Every production hydrate — the demand cold read (`inlineFetchOrWait`) and `fetchResolvedBlock`, used by both `WarmAll` and prefetch — routes through this one function, so the guard covers all of them rather than one named path.
2. **The boundary** (`readwrite.go` `Truncate`) narrows the straddling row s `DataSize` to the surviving prefix, which is what gives the clamp something to clamp to. Without it `DataSize == len(data)` and the clamp is a no-op.

Removing either half makes the truncate test fail again — the clamp alone leaves `manifest row at 3182115 still claims up to 4194304, past the new size 3688209`, the narrowing alone still resurrects `0x3a`.

The remote chunk is untouched. `readChunkVerified` reads by locator length and verifies the full bytes against the hash, so only the row s claim changes; RefCount is unchanged since the file still references the chunk, just less of it.

## Sparse files

Unaffected by construction: an unshrunk row claims its chunk in full, so the clamp is inert on every ordinary path. `TestColdReadSparseFile_HydratesAtHighOffset` pins it — a file written only at 2 MiB still hydrates there after eviction, and the leading gap, the gap in front of the data, and the range past it all still read as zeros.

## Verification

The failure needs eviction plus a cold read plus a re-extend, so each test drives the whole sequence: write, carve, upload, shrink, force-evict (`evictAll` fails the test if nothing was actually evicted, so it can never silently no-op), cold read, then re-extend and assert the hole.

```
go build ./... && go vet ./... && gofmt -l .   # clean
go test ./pkg/block/...                        # all ok (engine 34.5s, journal 31.7s)
go test -race ./pkg/block/engine/...           # ok 88.4s
```

Also folded in a small reuse win the simplifier found: `assertManifestTiles` had its own copy of the manifest projection, and now shares the new `manifestRefs` helper (−40/+8, error messages byte-identical).

## Coordination

Does not overlap PR #1929 — that adds a per-shard durable watermark and `DurableExtent` in the journal/metadata layers, which is about which bytes survive a crash. This is about which bytes a file logically still has, and touches only `pkg/block/engine`.

### DIFF
--- a/pkg/block/engine/cold_read_overwrite_test.go
+++ b/pkg/block/engine/cold_read_overwrite_test.go
@@ -5,7 +5,6 @@ import (
 	"math/rand"
 	"testing"
 
-	"github.com/marmos91/dittofs/pkg/block"
 	"github.com/marmos91/dittofs/pkg/block/engine"
 	remotememory "github.com/marmos91/dittofs/pkg/block/remote/memory"
 	"github.com/marmos91/dittofs/pkg/metadata"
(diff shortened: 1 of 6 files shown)


marmos91:
Both Copilot findings were real. Addressed in 13527f76.

**Finding 1a — `DataSize == 0` skipping the clamp: valid, and the more serious of the two.** It was not reachable through the path this PR adds — `Truncate` narrows only in the `b.Offset < newSize` branch (the `b.Offset >= newSize` case drops the block and continues), so `newSize - b.Offset` is always at least 1, and truncating exactly to a chunk boundary fails the `>` test and is not narrowed at all. But that is not the only producer: `blocksink.go` writes `DataSize: uint32(len(c.Data))`, `CopyPayload` copies `b.Size` through, and `read_internal.go` already defends against "a zero-length row parked exactly at off", so the codebase treats zero-length rows as a live possibility.

What decides it is which paths can carry such a row into the hydrate. A zero-`DataSize` row can never be selected by `findRowCoveringOffset` (it needs `target < abs+DataSize`), so the demand cold read is safe — but `WarmAll` walks the enumerated rows directly and skips only nil rows and unparseable IDs, with no `DataSize` filter, handing the row straight to `fetchResolvedBlock` → `hydrateChunk`. So the sentinel did disable the guard on a live path, silently. The default is now inverted: the row is authoritative unconditionally, and a row claiming nothing writes nothing.

**Finding 1b — `int(fb.DataSize)` on 32-bit: valid in principle, unreachable in practice, fixed anyway.** `DataSize` is `uint32` and no 32-bit target is built, and FastCDC chunks are MiB-scale so `> MaxInt32` cannot occur — but the wrap would have silently skipped the clamp, and a `uint64` comparison on both sides costs nothing and removes the reasoning burden entirely.

**Finding 2 — contradicting definitions: valid.** `pkg/block/types.go` is now the authoritative one and describes both cases (full chunk length, except where a shrink narrowed the row to its surviving prefix, with the remote chunk still holding and still hash-verified over all of its bytes). The comment in `fetch.go` no longer redefines the field; it refers to it.

New test `TestHydrateChunk_WritesOnlyWhatTheRowClaims` covers all three claims (full / narrowed / nothing) and fails against the previous shape:

```
--- FAIL: TestHydrateChunk_WritesOnlyWhatTheRowClaims/row_claims_nothing
    byte 0 is 0xab, want a zero hole — hydrate wrote past the row's claim of 0
```

Re-verified on top of the current develop (rebased over #1926/#1928/#1929): `go build ./... && go vet ./... && gofmt -l .` clean, `go test ./pkg/block/...` all ok, `go test -race ./pkg/block/engine/...` ok 83.1s.

Issue #1900, opened 2026-08-05
Audit: data-plane (metadata ↔ block hot path) — 8 HIGH findings


marmos91:
Data-plane audit of `origin/develop` @ **7eeec0da**, scoped to the hot write/read path between the metadata stores and the block stores.

**565 raw findings judged → 330 confirmed** after adversarial verification: **8 HIGH · 102 MED · 220 LOW**.

Only the 8 HIGH are tracked here. The 322 MED/LOW findings are kept in a local report rather than filed — they are mostly naming, duplication, and stale-comment cleanup, and would drown the issue tracker.

Every finding below passed three gates before filing: confirmed from source, proven reachable from a non-test caller somewhere in the tree, and checked against the governing spec (RFC 1813 / 7530 / 8881 / 7862, MS-SMB2) so that fixing it will not break conformance. Five of the eight were reported independently by two separate audit runs using different lenses.

### Data integrity / correctness

- [x] #1894 — Cold-fetch loop skips chunks after a hole instead of the next actual chunk (`pkg/block/engine/fetch.go:372`). Silent zeros; same class as #1888, different mechanism.
- [x] #1897 — sqlite `GetLinkCount` swallows all query errors as `count=0`, defeating the caller's unlink safety net (`pkg/metadata/store/sqlite/files.go:183`). Data loss on the delete path.
- [x] #1895 — `Identity.gidSet` lazy init mutated without a lock while the `Identity` is shared per-session (`pkg/metadata/auth_identity.go:214`). Concurrent map write is a fatal runtime throw.

### Security

- [x] #1898 — Netgroup export access check enforced only for NFSv3 MOUNT; SMB and NFSv4 bypass it entirely (`pkg/controlplane/runtime/netgroups.go:169`).

### Protocol conformance

- [x] #1899 — `WRITE_THROUGH` / `WRITE_UNBUFFERED` request flag decoded but never honored (`internal/adapter/smb/handlers/write.go:97`). Durability acknowledged but not provided.

### Hot-path performance

- [x] #1893 — Every SMB WRITE takes a process-wide mutex plus a durable-store lookup, serializing all writers (`internal/adapter/smb/handlers/write.go:376`).
- [x] #1892 — Every SMB READ does a synchronous metadata write for the atime bump, with no debounce (`internal/adapter/smb/handlers/read.go:440`).
[...]

PR #1935, opened 2026-08-06
fix(lock): close the wait-for-graph check-then-act window and bound two lease-key sweeps

marmos91:
Fixes the MEDIUM findings the data-plane audit raised against `pkg/metadata/lock/`.

## Correctness

**Wait-for-graph check-then-act.** `WouldCauseCycle` (RLock) and `AddWaiter` (separate Lock) were a check-then-act pair, and the only production caller — `parkLockOnConflict` — did real work between them (`TryReserveAsync`, `generateAsyncId`, `context.WithTimeout`, `PendingLockRegistry.Register`). Two SMB LOCK requests racing to park could each observe an acyclic graph and then together commit the edges that close a cycle. They are now one `TryAddWaiter(waiter, owners) bool` holding a single write lock across the check and the insert; on refusal nothing is recorded. The park path inserts the edges before reserving the async slot so concurrent callers see the wait immediately, and a deferred `RemoveWaiter` unwinds them on every path that does not end up parking. On the parked path `resumePendingLock` owns them and already defers `RemoveWaiter` at entry, so timeout, cancellation, success and error all prune.

The undetected cycle was never a permanent wedge — every parked lock is bounded by `asyncBlockingLockTimeout` — so this converts a late `LOCK_NOT_GRANTED` into an immediate, correct one.

`TestWaitForGraph_TryAddWaiter_ConcurrentRingCannotCloseCycle` fires a full ring of wait edges at once; every edge is individually cycle-free against the empty graph, so the old pair could admit all of them.

## Performance

**Two full-manager sweeps replaced by the existing reverse index.** `hasLeaseKeyOnOtherFile` (called under `lm.mu` on every lease-granting SMB CREATE) and `SetLeaseEpoch` both ranged all of `lm.unifiedLocks` to find records for a single lease key. Both now probe `leaseKeyIndex[leaseKey]`, which is derived state reconciled from the authoritative slice on every mutation and already used by `findLeaseByKey` for exactly this lookup. It ref-counts every bucket holding the key, so it is a complete candidate set — `SetLeaseEpoch` still reaches every matching record, not just the first. Work now scales with the files actually bound to that key instead of with every lock on the share.

`TestSetLeaseEpoch_IndexMatchesFullScan` and `TestHasLeaseKeyOnOtherFile_IndexMatchesFullScan` assert the index-driven answer equals the full-sweep answer across a mixed fixture, and `indexes_test.go` asserts index/`unifiedLocks` consistency across the whole lease lifecycle.

## Clarity

- Grace-period `Operation` was a struct of three independent bools (`IsNew`/`IsTest`/`IsReclaim`) that admitted contradictory combinations; it is now an enum. The sole production caller sets it explicitly, so the `OpNew`-as-zero-value change is inert.
- `NewDeadlockError`, `NewGracePeriodError` and `NewLockLimitExceededError` accepted diagnostic values and dropped them on the floor; they now appear in the message. `RegisterClient` builds its error via the existing `errors.NewConnectionLimitError` instead of an inline literal.
- `ValidFileLeaseStates`/`ValidDirectoryLeaseStates` unexported (no external users), unused deprecated re-export vars deleted from `lock_exports.go`, and eight duplicated `// Package lock ...` headers detached from the `package` clause so `types.go` is the only package doc.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .` clean. `pkg/metadata/lock` 451 tests pass and pass under `-race`; `internal/adapter/smb/...` + `pkg/adapter/nfs/...` 803 tests pass. No behaviour change outside the lock scope.

Relates to #1900

### DIFF
--- a/internal/adapter/smb/handlers/lock_async.go
+++ b/internal/adapter/smb/handlers/lock_async.go
@@ -43,19 +43,27 @@ func (h *Handler) parkLockOnConflict(
 	lockSeqEnabled bool,
 	lockSeqIndex uint32,
 	lockSeqNumber uint8,
-) (uint64, bool) {
-	// Deadlock detection: probe the conflicting holder via TestLock and
-	// check the WFG. If granting our wait would close a cycle, refuse to
-	// park — the inline-retry fallback then exits with LOCK_NOT_GRANTED.
+) (asyncId uint64, parked bool) {
+	// Deadlock detection: probe the conflicting holders via TestLock and record
+	// the wait edges in the WFG. TryAddWaiter refuses when they would close a
+	// cycle, and the inline-retry fallback then exits with LOCK_NOT_GRANTED.
+	// The edges go in before the async slot is reserved so concurrent callers
+	// see the wait immediately; the deferred unwind prunes them unless we park,
+	// where resumePendingLock prunes them on every exit path.
 	owners := h.conflictHolders(openFile.MetadataHandle, fileLock)
 	if h.LockWaitGraph != nil && len(owners) > 0 {
-		if h.LockWaitGraph.WouldCauseCycle(fileLock.OpenID, owners) {
+		if !h.LockWaitGraph.TryAddWaiter(fileLock.OpenID, owners) {
 			logger.Debug("LOCK: deadlock detected, refusing to park",
 				"waiter", fileLock.OpenID,
 				"owners", owners,
 				"messageID", ctx.MessageID)
 			return 0, false
 		}
+		defer func() {
+			if !parked {
+				h.LockWaitGraph.RemoveWaiter(fileLock.OpenID)
+			}
+		}()
 	}
 
 	if !ctx.TryReserveAsync() {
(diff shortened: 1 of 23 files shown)


marmos91:
Checked the acyclicity-assertion finding — it does not hold, and I verified rather than reasoned about it.

The claim is that wouldCauseCycle(owner, []string{owner}) only catches an explicit self-edge, because canReach has no from == to base case, so a committed multi-node ring would slip through. canReach does not need that base case: it checks waitSet[to] directly at every hop, so on a ring A->B->C->A the call canReach(A, A) recurses A -> B -> C and matches at C, whose edge set contains A. The visited map only stops a node being expanded twice; it does not stop that node matching as a target.

Empirically, on a 3-node ring built with no self-edges:

wouldCauseCycle(A, [A]) = true
wouldCauseCycle(B, [B]) = true
wouldCauseCycle(C, [C]) = true

with an acyclic X->Y->Z chain as a control returning false for all three, so this is detection and not a blanket true.

That said, the finding was worth raising: the assertion genuinely reads as if it only catches self-edges, which is exactly why it could later be 'corrected' into something weaker. I have pinned the behaviour with TestWouldCauseCycleDetectsRingWithoutSelfEdge, which asserts both the ring and the acyclic control, plus a comment explaining why the direct-edge check is what makes it work. Pushed as ebe1c1e0; full lock suite still green.

PR #1938, opened 2026-08-06
fix(smb): guard the OpenFile fields WRITE mutates and coalesce its atime writes

marmos91:
Three MEDIUM findings from the data-plane audit, all reported against the SMB2 WRITE handler. All three were confirmed real; one sub-claim in the first was refuted (see below).

## 1. PayloadID was mutated without the handle lock (bugs)

`OpenFile`'s struct doc listed PayloadID among the "immutable fields ... safe to access without the mutex". That was false: a file created empty has no payload until its first WRITE publishes the one the metadata store allocated, and SET_REPARSE_POINT and COPYCHUNK replace it. So the doc was wrong, not the code.

Unguarded readers raced that write — CLOSE's block-store commit gate, `flushFileCache` on session/tree teardown, and the durable-handle snapshot. Added `GetPayloadID` / `SetPayloadID` accessors that take `mu`, converted every reader and writer, and corrected the struct doc to name PayloadID, Path, FileName and ParentHandle as mutable.

Refuted sub-claim: the audit also cited `set_reparse_point.go:256` as an unguarded mutation. It already mutates inside `openFile.mu.Lock()`; no change was needed there.

## 2. FileName was read without the lock while rename mutated it (bugs)

SET_INFO rename rewrites the name, path and parent on a live handle. WRITE read the name to classify an ADS write and to address the change notification, and read the parent to bump the directory's access time — none of it under the lock.

WRITE now snapshots name, path and parent once under the read lock and uses that snapshot throughout, so a rename landing mid-WRITE cannot leave the ADS classification and the change notification describing different names. Both rename publication sites in set_info.go now publish the triple under the write lock.

`updateBaseObjectTimestampsForADSWrite` took the parent off the OpenFile itself, which reintroduced the live read one call deep; it now receives the caller's snapshot, so the base-object lookup cannot target a different directory than the one its name was derived from.

## 3. Two extra unbatched metadata round trips per WRITE (perf)

Every WRITE issued a `SetFileAttributes` for the file's LastAccessTime and another for the parent directory's, on the ack path. The parent bump serialized every WRITE to any file in a directory on that one metadata row — the same read-modify-write contention shape already fixed elsewhere for mkdir nlink.

Both now coalesce per handle, reusing the mechanism the READ path already uses rather than adding a second one. The file's newest access time is held on the handle for QUERY_INFO to overlay and CLOSE to persist, so nothing observable is lost. The parent's is dropped rather than deferred, since a directory has no handle here to carry it; a parent's access time therefore trails the newest write by at most the coalescing window.

## Verification

`go build ./...`, `go vet ./...` and `gofmt -l .` are clean. `go test ./internal/adapter/smb/...` passes 2056 tests, and `go test -race ./internal/adapter/smb/...` is green across all 12 packages.

The PayloadID race was reproduced under `-race` before the fix — write in the WRITE handler against the read in `flushFileCache`. `write_race_test.go` keeps that pairing as a regression guard: it drives the real WRITE handler on one side, so the reader half is production code rather than a stand-in.

The rename race was likewise reproduced pre-fix with both sides unguarded, and showed up as three distinct races (FileName, ParentHandle, Path). It is not kept as a test: both sides of that pairing now take the handle lock, which leaves the detector nothing to observe, so the guard there is the lock itself.

Relates to #1692

### DIFF
--- a/internal/adapter/smb/handlers/close.go
+++ b/internal/adapter/smb/handlers/close.go
@@ -208,8 +208,15 @@ func (h *Handler) Close(ctx *SMBHandlerContext, req *CloseRequest) (*CloseRespon
 	// (silent write truncation, #1267). The mapped status is recorded and
 	// applied to the response below; handle teardown still runs unconditionally
 	// so the failed flush does not leak the open-file/lease state.
+	// Snapshot the path once for the rest of CLOSE. SET_INFO rename rewrites it
+	// on a live handle, so reading it per use would let a rename landing
+	// mid-close produce log lines, a delete target and a parent path that each
+	// name a different file.
+	closePath := openFile.GetPath()
+
 	var flushFailStatus types.Status
-	if !openFile.IsDirectory && openFile.PayloadID != "" {
+	payloadID := openFile.GetPayloadID()
+	if !openFile.IsDirectory && payloadID != "" {
 		blockStore, bsErr := common.ResolveForWrite(ctx.Context, h.Registry, openFile.MetadataHandle)
 		if bsErr != nil {
 			// A ResolveForWrite failure is NOT a block-store content/durability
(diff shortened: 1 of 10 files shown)


marmos91:
All four findings addressed, and the sweep behind them turned up something bigger.

**1. buildPersistedDurableHandle (durable_context.go) — real, and the serious one.** The handle lock is now held across the WHOLE snapshot rather than per field, because Path, FileName and ParentHandle are rewritten together by rename and this row is replayed on reconnect: reading them at different instants could persist one rename's name against another's parent, which no later operation would correct. Note the PayloadID and FileName reads inside that block went back to direct field access deliberately — calling the accessors while already holding RLock risks deadlock, since sync.RWMutex.RLock is not safely reentrant when a writer is queued.

I checked the lock ordering this introduces: the only production caller holds h.durablePurgeMu, so the order is durablePurgeMu then openFile.mu. The other two durablePurgeMu acquisitions (purgeDisconnectedConflicts, SeedFromDurableHandles) do not hold openFile.mu, so there is no inversion.

**2. flushFileCache (handler.go) — real.** Snapshots the path once; its three log lines could otherwise name different paths for one operation.

**3. CLOSE (close.go) — real.** Snapshotted once at the top of Close() and used throughout the function, rather than patching only the two lines flagged. Partial replacement would have left some lines reading a snapshot and others reading live, which is harder to reason about than either. This also covers the delete-target Path and the GetParentPath call, not just logging.

**4. noteSmbParentAccess coverage — added.** TestNoteSmbParentAccessCoalescesWithinWindow covers first-write, suppression inside the window, the boundary, and asserts parent bumps do not leak into the file's own pending-atime slot (they are dropped rather than held, unlike noteSmbAccess). Plus a nil-handle case.

## The sweep

I then swept the package rather than stopping at the four. There are **262 references** to openFile.Path / .FileName / .ParentHandle across 15 non-test files. This PR and the rename path itself account for the ones in write.go, close.go, durable_context.go, handler.go and set_info.go; the rest — read.go, flush.go, query_info.go, ioctl_*.go, stub_handlers.go — are still live unguarded reads.

I deliberately did NOT expand this PR to cover them. Locking 262 sites by hand would be a large mechanical diff, hard to review, and nothing would stop the next handler reading the field directly. Filed as #1940 with the recommendation to unexport the fields so the compiler enforces accessor use, plus the two traps found here: snapshot the group under one RLock rather than calling accessors in sequence, and never call an accessor while already holding the lock.

Worth flagging that query_info.go should be audited before the log-only sites — it composes a response the client consumes, so any torn read there is observable rather than cosmetic.

Verification: build/vet/gofmt clean, full SMB suite green across all 12 packages, and go test -race ./internal/adapter/smb/handlers/ clean. Rebased onto current develop; diff is 10 files, all mine.

PR #1942, opened 2026-08-12
fix(nfs4): consume the owner seqid when a sequenced operation fails

chessman:
# Description

## Type of change

## Changes Made

* Seqid should advance even in case of some errors. RFC:
```
The client MUST advance the sequence number for the CLOSE, LOCK, LOCKU, OPEN,
OPEN_CONFIRM, and OPEN_DOWNGRADE operations. This is true even in the event
that the previous operation that used the sequence number received an error.
The only exception to this rule is if the previous operation received one of
the following errors: NFS4ERR_STALE_CLIENTID, NFS4ERR_STALE_STATEID,
NFS4ERR_BAD_STATEID, NFS4ERR_BAD_SEQID, NFS4ERR_BADXDR, NFS4ERR_RESOURCE,
NFS4ERR_NOFILEHANDLE, or NFS4ERR_MOVED.
```

* CLOSE can fail when byte-range locks are actually held - not merely when lock stateids exist, since a lock stateid stays valid after LOCKU for as long as the file remains open.

In theory, CLOSE can automatically close the locks instead of returning NFS4ERR_LOCKS_HELD, this is what the Linux NFS server does.

## Testing

## Testing Instructions

It's not easy - usually RELEASE_LOCKOWNER is sent before CLOSE. I can reproduce this error only by running a long test suite. Unfortunately, I cannot provide a short reproduction besides unit tests.

## Checklist

### DIFF
--- a/internal/adapter/nfs/v4/state/manager.go
+++ b/internal/adapter/nfs/v4/state/manager.go
@@ -760,19 +760,7 @@ func (sm *StateManager) onLeaseExpired(clientID uint64) {
 				delete(sm.lockStateByOther, lockState.Stateid.Other)
 
 				// Remove actual locks from unified lock manager
-				if lm := sm.lockManagerFor(lockState.FileHandle); lm != nil {
-					ownerID := lockState.LockOwner.LockManagerOwnerID()
-					handleKey := string(lockState.FileHandle)
-					locks := lm.ListUnifiedLocks(handleKey)
-					for _, l := range locks {
-						if l.Owner.OwnerID == ownerID {
-							_ = lm.RemoveUnifiedLock(
-								handleKey,
-								l.Owner, l.Offset, l.Length,
-							)
-						}
-					}
-				}
+				sm.removeOwnerLocksLocked(lockState)
 
 				// Remove lock-owner from lockOwners map
 				if lockState.LockOwner != nil {
@@ -1077,7 +1065,7 @@ func (sm *StateManager) OpenFile(
 	fileHandle []byte,
 	shareAccess, shareDeny uint32,
 	claimType uint32,
-) (*OpenFileResult, error) {
+) (result *OpenFileResult, err error) {
 	// Grace period checks (before acquiring sm.mu)
 	switch claimType {
 	case types.CLAIM_NULL:
@@ -1158,9 +1146,9 @@ func (sm *StateManager) OpenFile(
 		// New owner: create it
 		clientRecord := sm.clientsByID[clientID]
 		owner = &OpenOwner{
-			ClientID:     clientID,
-			OwnerData:    make([]byte, len(ownerData)),
-			LastSeqID:    0, // Will be set to seqid after success
+			ClientID:  clientID,
+			OwnerData: make([]byte, len(ownerData)),
+
 			Confirmed:    false,
 			OpenStates:   make([]*OpenState, 0),
 			ClientRecord: clientRecord,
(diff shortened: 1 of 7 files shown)


internal/adapter/nfs/v4/state/openowner.go
```diff
@@ -26,6 +28,114 @@ const (
...
+}
+
+// ValidateSeqID checks whether a seqid is valid for this owner.
+//
+// Per RFC 7530 Section 9.1.7:
+//   - Expected = LastSeqID + 1 (with wrap: 0xFFFFFFFF -> 1, not 0)
+//   - seqid == expected -> SeqIDOK
+//   - seqid == LastSeqID -> SeqIDReplay
+//   - else -> SeqIDBad
+func (os *ownerSeq) ValidateSeqID(seqid uint32) SeqIDValidation {
```
marmos91: `os` as the receiver name shadows the `os` package (same on `consumeSeqidOnError` below). Nothing imports it in this file today, so it compiles, but it's a trap for whoever adds the first `os.` call here.

marmos91 (commented):
Thanks for this, clear write-up, and the two halves clearly belong together: the `LockStates > 0` predicate is what made the seqid bug reachable in the first place.

I read it against `develop` (the `state` package is identical there, so it applies) and the logic holds up. Nice touch that the defers sit under `sm.mu`, and that DENIED was already covered on the value path.

Two asks:

- Could you retarget to `develop`? `main` is release-only here. Should be a clean retarget.
- All the new tests prove a failure *does* consume the seqid. Could you add one for the other direction — that BAD_STATEID / BAD_SEQID / a replay leave it untouched? That's the side where a regression is silent.

One thing I'd rather track separately: CLOSE now drops the lock-owner, but a new lock-owner is only accepted at `lockSeqid == 1` (`manager.go:1890`), where nfsd adopts whatever the client sends. Probably unreachable from Linux — happy to open an issue for it if you agree.

One nit inline.

internal/adapter/nfs/v4/state/openowner.go
chessman: thank you, fixed

chessman:
Thank you for the review. I added more tests and rebased to develop.

Regarding this one:

> One thing I'd rather track separately: CLOSE now drops the lock-owner, but a new lock-owner is only accepted at lockSeqid == 1 (manager.go:1890), where nfsd adopts whatever the client sends. Probably unreachable from Linux — happy to open an issue for it if you agree.

Yes, the spec allows that:
> When a sequence value associated with a state-owner is unavailable
      for checking because the state-owner is unknown to the server, it
      takes no part in the comparison.

There is a weird statement, "The server SHOULD assign a seqid value of one for the first request issued for any given state-owner" but it looks like they just wanted to say that the number should be greater than zero.

internal/adapter/nfs/v4/state/manager.go
```diff
@@ -1577,6 +1590,67 @@ func (sm *StateManager) CloseFile(stateid *types.Stateid4, seqid uint32) (*OpenS
...
+		}
+		ownerID := lockState.LockOwner.LockManagerOwnerID()
+		for _, l := range lm.ListUnifiedLocks(string(lockState.FileHandle)) {
+			if l.Owner.OwnerID == ownerID {
+				return true
+			}
+		}
+	}
+	return false
+}
```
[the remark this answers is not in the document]
chessman: I think it's overkill.

marmos91:
LGTM

PR #1943, opened 2026-08-17
fix(runtime): surface adapter restart failures, hold entries until stopped, drop dead client tracking

marmos91:
Three findings from the data-plane audit, all in the Runtime composition layer.

## Adapter restart failures were swallowed

`adapters.Service` restarted an adapter and discarded the error, so a restart that
failed left the entry marked running with nothing behind it. The error now
propagates, and the registry entry is held until the adapter has actually stopped
rather than being dropped at the start of teardown. New tests in
`adapters/service_test.go` cover both the failed-restart and the
stopped-before-removal orderings.

## Client registry tracked shares nothing ever populated

`clients.Registry` carried an `AddShare` / `RemoveShare` / `ListByShare` trio and a
`Shares` slice on every record. No caller outside the package ever invoked them, so
`Shares` was empty on every record the API returned. The same applies to
`ClientRecord.User` and to most of the protocol detail structs: `NfsDetails`
`AuthFlavor` / `UID` / `GID` and `SmbDetails` `Dialect` / `Domain` / `Signed` /
`Encrypted` were declared, serialized, and never assigned.

`GET /clients`, `GET /clients?share=`, and `dfsctl client list` were therefore
reporting permanently blank columns as though they were real state. The dead
methods and fields are removed along with the tests that covered them, which drops
the `mu`-guarded mutable field set to nothing — an entry's fields are now written
at `Register` and never mutated again, so the lock comment matches what the lock
actually protects.

**This changes the shape of the client JSON payload and the `dfsctl client list`
output.** Every field removed was always empty or zero, so no information is lost,
but a consumer parsing those keys will no longer find them.

`docs/guide/cli.md` is regenerated from the Cobra change, not hand-edited.

## Verification

`go build ./...`, `go vet ./...`, and the runtime / API-handler / dfsctl package
trees all pass. `go run ./cmd/gendocs` produces no diff against the committed
`cli.md`.

### DIFF
--- a/cmd/dfsctl/commands/client/client.go
+++ b/cmd/dfsctl/commands/client/client.go
@@ -11,7 +11,7 @@ var Cmd = &cobra.Command{
 	Short: "Manage connected clients",
 	Long: `Manage connected NFS and SMB clients on the DittoFS server.
 
-Use these commands to inspect which clients are currently connected, filter by protocol or share, and forcefully disconnect misbehaving sessions. All operations require admin privileges.
+Use these commands to inspect which clients are currently connected, filter by protocol, and forcefully disconnect misbehaving sessions. All operations require admin privileges.
 
 Examples:
   # List all connected clients across NFS and SMB
@@ -19,9 +19,6 @@ Examples:
 
   # Show only NFS clients
   dfsctl client list --protocol nfs
-
-  # Show clients connected to a specific share
-  dfsctl client list --share myshare
 
   # Disconnect a specific client by its ID
   dfsctl client disconnect nfs-42 --force`,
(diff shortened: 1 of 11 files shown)


marmos91:
Addressed the timeout-path comment. The entry is now marked cancelled when a stop times out, and UpdateAdapter excludes cancelled entries from the preserve-listener shortcut, so a reload can no longer return success for an adapter whose context is already cancelled. typeClaimedLocked reports the cancelled state distinctly rather than as already running. TestUpdateAdapter_NoPreservedListenerAfterStopTimeout covers it and fails against the previous behaviour.

On the waitable completion signal: errCh is created with make(chan error, 1) and the serve goroutine sends exactly once, so a send that lands after the timeout parks in the buffer rather than blocking, and a later stop attempt receives it. The completion cannot be lost, so I have not restructured around a done channel.

Issue #1944, opened 2026-08-17
Support master-key rotation: retired-key map so old wrapped blocks stay decryptable


marmos91:
## Summary

`KeyProvider` implementations in `pkg/block/encryption/keyprovider/` hold exactly one master key. There is no set of retired keys, so a master key can never be changed without losing access to everything already written under the old one.

## Current behaviour

Both providers embed a single `aesGCMKEK` holding one `masterKey` / `masterKeyID`:

- `local.go` — key read from `key_file`
- `kmip.go` — key fetched from the HSM at startup by `key_uid`

Every encrypted frame records the identifier of the master key that wrapped its block key. `Unwrap` compares that identifier against the single held id and returns `ErrWrongMasterKey` on any mismatch — the recorded id is used only to reject, never to look anything up.

The consequence: pointing a share at a different master key (writing a new `key_file`, or changing `key_uid`) makes every block wrapped under the previous key **permanently unreadable**. There is no recovery path once the old key material is gone, and no re-wrap tooling.

The docs used to describe rotation as a supported procedure and claimed parity with SSE-KMS / KES / Vault Transit, all of which keep retired key versions decryptable. That documentation has been corrected to state plainly that rotation is unsupported and destroys data. This issue tracks the missing capability itself.

## Why it matters

Master-key rotation is a routine compliance requirement (periodic rotation, rotation after suspected exposure, rotation on personnel change). Operators reasonably expect an envelope-encryption design that stamps a key id into every frame to support it — the frame format was built for exactly this. Today the only safe answer is "never change the key", which is not one an operator can live with indefinitely.

## Proposed change

Give `Unwrap` a lookup keyed by `masterKeyID`:

- `Wrap` always uses the current master key and records its id (unchanged).
- `Unwrap` resolves the frame's recorded id against a set of `{current} ∪ {retired...}` and fails with `ErrWrongMasterKey` only when the id matches none of them.

This mirrors how Vault Transit and SSE-KMS keep old key versions live for decrypt-only. It changes no frame bytes — the id is already on the wire.

Complementary and separable: a bulk re-wrap pass (read with the old key, write with the new) so an old key can eventually be decommissioned rather than retained forever.

## Open design question

**The config surface is the undecided part, and is why this was split out rather than implemented inline.** Needs a decision on:

- How retired keys are supplied per provider — a list of `key_file` paths and a list of `key_uid`s? A directory? An ordered list where the first entry is current?
[...]

PR #1946, opened 2026-08-17
fix(block): share decorator passthrough, drop dead append contract, correct rotation docs

marmos91:
Closes the MED-severity findings scoped to the block-store decorator layer from the 2026-08-05 data-plane audit.

## 1. Shared decorator passthrough

`EncryptedRemote` (`pkg/block/encryption`) and `Decorator` (`pkg/block/compression`) carried ~150 lines of plumbing that was byte-identical apart from the receiver name — the block-keyed forwards, the CAS/block inner assertions, `Has`, `Delete`, `DeleteLegacyChunk`, `Close`, `HealthCheck`, `Healthcheck`, `Durable`, and `GetRange`.

That plumbing now lives once in `remote.Passthrough`, an embeddable struct in the new `pkg/block/remote/passthrough.go`, alongside two package-level helpers:

- `remote.CASInner(s)` — a function rather than a method, so embedding does not put an exported handle on the untransformed CAS surface into either decorator's public method set.
- `remote.SliceRange(full, offset, length)` — the shared bounds/clamp contract, so the two `GetRange` error contracts can no longer drift apart. The conformance suite asserts those sentinels for both decorators.

Each decorator now keeps only what its transform actually changes: `Put`, `Get`, `GetRange`, `Head`, `Walk`, `SealChunk`, `ReadChunk`, `sealLayer`. Encryption still overrides `Close` so it closes its key provider as well as the inner store; compression inherits the forward, having no store-level resource of its own.

The exported method sets of both decorators are unchanged — verified by reflecting over both types on this branch and on `develop` and diffing the results.

## 2. Dead `block.BlockStoreAppend` contract

`block.BlockStoreAppend` had zero implementations and zero consumers, and its godoc asserted that `pkg/block/local/fs.FSStore` implemented it. It does not: `FSStore` is payload-keyed (`WriteAt` / `ReadAt` / `Hydrate` / `Commit`) and satisfies neither `BlockStoreAppend` nor `block.Store`. `blockstoretest` also documented a `BlockStoreAppendConformance` entrypoint that was never written.

The interface is deleted and the claims that repeated it are corrected across `pkg/block/blockstore.go`, `pkg/block/doc.go`, `pkg/block/blockstoretest/`, `docs/internals/contributing.md`, `docs/internals/implementing-stores.md`, and the store-contract rule in `CLAUDE.md`.

Also removed: `pkg/block/remote/doc.go`, a second `// Package remote` comment that contradicted the accurate one in `remote.go` on both the backend list (it named a filesystem remote that does not exist) and the key format.

Two more comments that were false on every claim they made are cut back to what is true — the `blake3Sum` provenance note in `blockstoretest/conformance.go` cited a file, a symbol, and a caller that all do not exist.

## 3. Master-key rotation documented truthfully

Every `KeyProvider` holds exactly one master key. `Unwrap` returns `ErrWrongMasterKey` whenever a frame's recorded key id is not that one, and there is no keyring of retired keys. An operator following the rotation procedure previously documented in the KMIP provider — write a new key, restart — would permanently orphan every block wrapped under the old key. `docs/internals/encryption-design.md` compounded it by stating outright that "existing blocks remain decryptable", and the package docs claimed parity with key-management services that keep retired key versions decryptable.

This PR makes the documentation truthful rather than implementing rotation: the rotation procedure and the parity claim are removed, and the package docs, the design doc, and the KMIP provider now state plainly that rotation is unsupported and that changing a share's master key destroys access to everything already written under the old one.

Implementing the retired-key map is deliberately out of scope — the config surface (how retired keys are supplied, identified, and retained) is an open design question. Tracked separately.

Relates to #1944

## Verification

- `go build ./...`, `go vet ./...`, `gofmt -l .` — all clean
- `go test ./...` — full suite green
- `go test -race` on `pkg/block`, `pkg/block/compression`, `pkg/block/encryption/...`, `pkg/block/remote/...` — green
- `blockstoretest.BlockStoreConformance` still runs against both decorators (all three AEADs, both compression algorithms) and both remote backends; `RemoteBlockStoreConformance` still runs against the memory and s3 backends
- Exported method sets of both decorators diffed against `develop`: identical

No Cobra command or flag changed, so `docs/guide/cli.md` is untouched.

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -92,7 +92,9 @@ composition layer over six sub-services: `adapters/`, `stores/`, `shares/`, `mou
 
 7. **Store contracts** live in `pkg/metadata/storetest/` — any new metadata backend must pass
    that conformance suite. Block stores have a unified suite in `pkg/block/blockstoretest/`
-   (`BlockStoreConformance` + `BlockStoreAppendConformance`).
+   (`BlockStoreConformance` for the hash-keyed `block.Store` surface,
+   `RemoteBlockStoreConformance` for the block-keyed one). The local tier is
+   payload-keyed and is covered by its own package tests, not by that suite.
 
 ## Code comments
 
--- a/docs/internals/contributing.md
+++ b/docs/internals/contributing.md
@@ -287,17 +287,17 @@ DittoFS uses a Service-oriented architecture where **stores are simple CRUD inte
 
 **Block Store (Local):**
 
-1. Implement `pkg/block/local.LocalStore` interface. It embeds the
-   content-addressed `block.BlockStoreAppend` surface (`Put`, `Get`,
-   `GetRange`, `Has`, `Delete`, `Head`, `Walk`, plus `AppendWrite` /
-   `DeleteAppendLog`) and adds per-payload admin, lifecycle, rollup-drain,
-   and retention methods — see `pkg/block/local/local.go` for the full list.
-2. Storage is content-addressed (keyed by `block.ContentHash`), not by
-   block ID/offset
+1. Implement the `pkg/block/local.LocalStore` interface. It is the
+   journal-native surface keyed by `(payloadID, offset)` — `WriteAt`,
+   `ReadAt`, `Hydrate`, `Commit`, `Truncate`, `Delete` — plus carve,
+   eviction, lifecycle and retention methods. See `pkg/block/local/local.go`
+   for the full list.
+2. Storage is payload-keyed, NOT content-addressed: the local tier is a
+   per-file byte cache, so it does not implement `block.Store`.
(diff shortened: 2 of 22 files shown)


marmos91:
Both comments addressed in ca4a8cae.

The wrapped store is now unexported. Passthrough holds it as inner behind a NewPassthrough constructor, so embedding no longer promotes a handle to the untransformed store into the decorators' public surface. Each decorator keeps its own unexported reference for the ChunkReader and ChunkSealer capability access its transform participates in. go doc on EncryptedRemote now reports only unexported fields, so the bypass is gone.

The doc.go references to a BlockStore contract now name block.Store, on both lines.

PR #1948, opened 2026-08-17
fix(block): journal reclaim/recovery correctness and local-store cleanups

marmos91:
Batch of MED findings from the 2026-08-05 data-plane audit, covering the journal's reclaim/recovery paths and the local-store tier. Relates to #1900.

## Correctness

**`FSStore.FileSize` and `DataExtents` skipped the legacy fault-in seam.** `WriteAt`, `ReadAt`, `Delete` and `Truncate` each drain a payload's archived pre-journal log via `materializeLegacy` before touching it; these two delegated straight through. Both answer without ever issuing a read, so an un-drained payload reported a zero/short size and an empty extent list — and NFSv4.2 SEEK/READ_PLUS take the extents as authoritative, so a missing extent reads as a genuine sparse hole and a sparse-aware copy silently drops the bytes behind it. `FileSize` has no error in its signature, so a drain failure now reports "not found" and the caller falls back to its own size source rather than trusting a truncated answer. `DurableExtent` is deliberately left alone: it only ever under-reports durability, which is the safe direction for that contract.

**Tombstone and truncate-marker durability now goes through the shard commit leader.** Both paths issued a private `fd.Sync()` instead of `sh.groupCommit()`, so a burst of concurrent deletes hashing to one shard paid a full disk barrier each — exactly what the group-commit leader exists to eliminate. The record's bytes are written under `sh.mu` before the call, and `groupCommit`'s contract already covers rotation (`sealInPlace` fsyncs the segment it seals), so durability is unchanged.

**`evictSegment` now flips cold only the files it logged.** The cold-marker scan already collects every file backed by the segment; the second pass re-walked the whole shard index. It now flips exactly the set it wrote cold-log entries for. Safe because the segment is sealed and busy-claimed, so nothing can append to it and no repack can repoint into it during the unlocked `appendCold` window. This also tightens the invariant: the flipped set is now a subset of the durably-logged set, where the full re-walk could in principle mark an interval cold that had no cold-log entry behind it.

## Reclaim and read-path cost

- `pickVictim` rebuilt the entire shard live-byte map on every call, making a GC pass O(victims x intervals) under the same lock the read path takes. It is now summed once per pass by `shardLiveBytes`, and `repackSegment` keeps the map in step as it retires a victim and records its target. The pass holds `carveMu` throughout, and stale entries can only under-report dead bytes, which is conservative in the safe direction.
- `findMove` linear-scanned `moves` once per repointed interval, giving O(moves^2) under `sh.mu` for a heavily-overwritten victim. Now bucketed by record Version, which identifies a single written record.
- `verifiedRead` allocated a whole-record buffer per piece per read. `readRecordAt`/`readVerifiedRecord` take an optional scratch buffer and report what they used, and warm reads recycle it through a pool capped at the chunker's average cut. Recovery passes nil, since it keeps every payload it scans and the records must not share one buffer.

## Local-store cleanups

- `journal.Store.FileCount()` replaces `len(ListFiles(...))` in the two Stats paths. `ListFiles` built and returned a slice of every FileID under every shard lock just to take its length — on the metrics scrape path, which documented itself as O(1)-ish.
- `SetRetentionPolicy` was a no-op in every `LocalStore` implementation while remaining a required interface method that live callers invoked with real config. Removed from the interface, both implementations, the engine, and the shares service. Only the pin/non-pin distinction ever reached the block store; the rest is metadata the API reports back, and `UpdateShare` now says so. The two internals docs that described the no-op are updated.
- `MemoryStore` grew its buffer by allocating exactly the new size and copying, making a sequential file O(n^2) to fill. Extends through `append` so the runtime's geometric growth amortizes the copy.

## Notes for review

This branch was recovered from a raw git index after the previous session was killed between `reset --soft` and its squash commit. It was re-verified from scratch and audited for recovery artifacts; two were found and fixed (an import-grouping artifact in `journal/index.go`, and the `ListFiles` doc comment displaced onto the new `FileCount`). A follow-up pass found no others.

## Verification

`go build ./...`, `go vet ./...`, `gofmt -l .` clean. `go test ./...` passes; `go test -race ./pkg/block/...` passes.

### DIFF
--- a/docs/internals/architecture.md
+++ b/docs/internals/architecture.md
@@ -329,8 +329,9 @@ Under disk pressure the journal evicts whole **fully-synced** segments
 approx-LRU; only ranges already carved to the remote qualify, so eviction
 never destroys the only copy of dirty bytes. Eviction is health-gated: while
 the remote is unhealthy, cold-marking a range would strand unrecoverable
-bytes, so eviction is paused. There is no pin/ttl/lru retention knob
-(`SetRetentionPolicy` is a no-op on the journal-native store).
+bytes, so eviction is paused. The journal-native store has no pin/ttl/lru
+retention knob; only the pin/non-pin distinction reaches it, as the switch that
+enables or disables eviction.
 
 The journal's own garbage collection (repack) only relocates live cache bytes
 between local segments to reclaim dead space — it **never** touches the remote
--- a/docs/internals/implementing-stores.md
+++ b/docs/internals/implementing-stores.md
@@ -457,8 +457,7 @@ type LocalStore interface {
     UnsyncedBytes() int64
     Evict(ctx context.Context, targetBytes int64) (journal.EvictResult, error)
     SetEvictionEnabled(enabled bool)
-    // ... plus lifecycle (Start, Close), Stats, Healthcheck, and a
-    // no-op SetRetentionPolicy retained for admin-path compatibility.
+    // ... plus lifecycle (Start, Close), Stats, and Healthcheck.
 }
 ```
 
--- a/pkg/block/blockstore.go
+++ b/pkg/block/blockstore.go
@@ -164,7 +164,7 @@ type Store interface {
 //
 // Note: the narrowed LocalStore interface that lands keeps
 // additional lifecycle / admin methods (Truncate, EvictMemory
-// SetRetentionPolicy, SetEvictionEnabled, Stats, ListFiles
+// SetEvictionEnabled, Stats, ListFiles
 // GetStoredFileSize, Healthcheck, SyncFileChunks, SyncFileChunksForFile
 // Flush, Start, Close) as a strict admin-superset of BlockStoreAppend.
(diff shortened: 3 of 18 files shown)


marmos91:
On the writeLocked allocation: leaving it as is, deliberately.

The append of a zero slice is the mechanism that gets the amortized growth. Sizing the buffer to exactly end on each extending write is what made a sequential fill O(n^2), which is the finding this change closes. What remains is one temporary zero slice per extending write when capacity already suffices — a constant-factor allocation, not a complexity problem, in an in-memory store used for tests and ephemeral shares.

The in-place variant works (reslice within cap, clear the newly exposed region so the zero-fill gap semantics survive a prior Truncate), but it trades the current version's obvious correctness for a subtle one, and it is not worth another full CI cycle on a green PR for a store that never touches disk. Worth revisiting if this store ever becomes a data-path backend.

PR #1951, opened 2026-08-17
feat(encryption): support master-key rotation via a retired-key set

### LINKED ISSUE #1944: Support master-key rotation: retired-key map so old wrapped blocks stay decryptable

marmos91:
Closes #1944.

Implements the design decided in [this comment](https://github.com/marmos91/dittofs/issues/1944#issuecomment-5316205324).

## The change

`Wrap` is unchanged — always the current master key, recording its id. `Unwrap` now resolves the id recorded in the frame against `{current} ∪ {retired...}` and returns `ErrWrongMasterKey` only when it matches none. No frame bytes change; the id was already on the wire and was previously used only to reject.

```yaml
encryption:
  key:
    kind: local
    file: /etc/dittofs/keys/2026-08.key    # current — everything new is wrapped here
    retired_files:                         # decrypt-only
      - /etc/dittofs/keys/2026-02.key
```

KMIP mirrors it with `key_uid` + `retired_key_uids`. Both retired lists are additive and default to empty, so every existing config behaves exactly as it did.

## One correction to the issue's premise

The issue says the id is "derived from the key material", and that this makes operator-assigned ids expensive. That is not what the code does — `local` generates a UUID once in `GenerateKeyFile` and stores it *in* the key file as `master_key_id`; `kmip` uses the operator-supplied `key_uid` verbatim. Both are already assigned-at-generation and travel with the key, which is the Tink/KMS model. So a retired key identifies itself and nothing in config restates an id. This makes the change smaller than the issue anticipated and removes that open question entirely.

## Startup behaviour

- **Unreadable retired key** — logged and skipped, not fatal. Blocks under it become unreadable, which is the state the daemon was already in before it was configured; refusing to start would take every other share offline with it. A missing *current* key stays fatal.
- **Duplicate id** — rejected. Two keys claiming one id makes which one `Unwrap` picks arbitrary. Covers both "same file listed twice" and "current key also listed as retired".
- Retired keys share the current key's passphrase (one `DITTOFS_ENCRYPTION_PASSPHRASE` env var); documented, not changed here.

## Tests

`rotation_test.go` covers the property the whole thing exists for — a block wrapped under the old key still unwraps after rotation — plus: an id belonging to no configured key is still `ErrWrongMasterKey` (the lookup did not become accept-anything), unreadable-retired degrades, duplicate ids are rejected, and `Close` zeroes retired material too.

No KMIP-specific rotation test: those round-trips need a live HSM and are skipped in CI, and the fake server serves one canned key. The shared path (`loadRetiredKeys`, `keyForID`) is exercised by the local tests; the KMIP side only adds "fetch N uids".

## Still missing, and the docs now say so

No bulk re-wrap and no way to enumerate which blocks reference a given key. Until both exist there is no supported way to move blocks off a retired key, so a retired key must be treated as keep-forever — dropping it from the config or deleting the material is still unrecoverable data loss. `docs/guide/encryption.md` leads with that hazard rather than burying it.

## Verification

`go test -race ./pkg/block/encryption/... ./pkg/controlplane/runtime/... -count=1` green. `golangci-lint` clean on the changed packages.

### DIFF
--- a/docs/guide/encryption.md
+++ b/docs/guide/encryption.md
@@ -65,14 +65,26 @@ encryption:
     kind: local               # local | kmip
     # kind=local
     file: /etc/dittofs/keys/share.key
+    retired_files:            # optional, decrypt-only (see rotation below)
+      - /etc/dittofs/keys/share-2026-02.key
     # kind=kmip
     endpoint: kms.example.com:5696
     server_ca: /etc/dittofs/kmip/ca.pem
     client_cert: /etc/dittofs/kmip/client.pem
     client_key:  /etc/dittofs/kmip/client.key
     key_uid: 12345-abcde-...
+    retired_key_uids:         # optional, decrypt-only
+      - 09876-zyxwv-...
     timeout_ms: 5000
 ```
+
+`file` / `key_uid` name the **current** master key: everything written from
+now on is wrapped under it. `retired_files` / `retired_key_uids` name keys
+that are used for decryption only. Both retired lists default to empty, so
+a config written before rotation existed keeps behaving exactly as it did.
+
+All retired keys share the current key's passphrase
+(`DITTOFS_ENCRYPTION_PASSPHRASE`); there is no per-file passphrase.
 
 ### AEAD cipher choices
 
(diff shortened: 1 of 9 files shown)


marmos91:
Both Copilot comments addressed in fa7185d9.

**Key material on failed construction** — correct, and it applied more widely than the comment noted. `loadRetiredKeys` now zeros the just-loaded key and everything already accumulated before returning either `ErrInvalidConfig`, and each constructor zeros the *current* master key when retired loading fails — that one leaks on the same paths and had no `Close` to catch it either. `aesGCMKEK.Close` now reuses the same `zeroKey` / `zeroKeys` helpers instead of open-coding the loops.

`TestLoadRetiredKeys_ZeroesOnRejection` drives this directly with a stub loader that retains the material it hands out. Confirmed it fails without the fix:

```
--- FAIL: TestLoadRetiredKeys_ZeroesOnRejection (0.00s)
    rotation_test.go:415: key 0 not zeroed after rejection: eeeeeeee...
```

**KMIP coverage** — fair, and I was wrong to wave it off in the PR body on the grounds that the fake serves one canned key. Making it serve responses *in order*, one per connection, is about ten lines, and since `newKMIPProvider` fetches the current key first and then each retired uid in configured order, new-then-old is deterministic. That is enough to drive a real rotation with genuinely distinct key material per uid.

Added `TestKMIP_RetiredUIDStillUnwraps` (block wrapped under the retired uid still unwraps; asserts the fetched current/retired material landed in the right slots), `TestKMIP_RetiredUIDValidation` (both duplicate-id rejections), and `TestKMIP_UnreadableRetiredUIDDegrades`. All run against the fake, so they execute in CI rather than skipping like the live-HSM tests. Also dropped that last test's timeout to 300ms — it was sitting on the default five seconds waiting for a reply that never comes.

`go test -race ./pkg/block/encryption/...` green, `golangci-lint` clean.

PR #1953, opened 2026-08-18
fix(block): resolve overlapping manifest rows to the greatest start

### LINKED ISSUE #1934: resolveCovering fallback picks first covering row, badger's indexed path picks greatest-start — latent disagreement if overlapping rows ever survive
## Summary

`resolveCovering`'s `ListFileChunks` fallback returns the **first** covering row it finds, which is order-dependent. Badger's indexed path instead picks the greatest-start row. So the two implementations can disagree about which chunk covers an offset whenever more than one row covers it.

Today this does not bite: the carve re-tiles the manifest, so overlapping rows do not survive and exactly one row covers any offset. The inconsistency is latent, not active.

## Why it is worth recording anyway

This is the mechanism that the PunchHole half of #1911 would have needed to be a real bug. That half was **refuted** — the punch overwrites the range with zeros through the local write path and the carve supersedes the punched rows, verified structurally with `assertManifestTiles` rather than just by a passing read. The refutation holds *because* the manifest tiles exactly.

So the safety of that path currently rests on an invariant (rows never overlap) that is maintained by the carve, not enforced at the lookup. If any future change lets overlapping rows survive — a partial carve, a new supersede path, a migration that rewrites rows — the two resolvers would start disagreeing, and the failure mode is the silent-zeros / stale-bytes family (#1879, #1888, #1894, #1909, #1911), which returns wrong data with no error.

Same family as the existing `#953` overlap note in `cold_read_overwrite_test.go`.

## Suggested direction

Make the two paths agree explicitly rather than by coincidence. Either:
- have the fallback pick greatest-start too, matching the indexed path; or
- assert the tiling invariant at the lookup and report `ErrManifestInconsistent` when it is violated — consistent with how #1903 chose to report unplaceable successor rows rather than skip them.

The second is stricter and fits the direction the block engine has been moving all session: prefer a loud inconsistency error over a silently-chosen wrong row.

Low priority — this is defence against a future change, not a live defect. Worth doing when someone is next in that code.

Found while fixing #1911. Related: #1903, #1911, and the `#953` note in `cold_read_overwrite_test.go`.

marmos91:
Closes #1934.

`resolveCovering`'s `ListFileChunks` fallback returned the **first** row covering an offset — order-dependent — while badger's indexed `GetFileChunkAtOffset` takes the greatest-start row. The fallback now picks the greatest start too, so both paths answer a read alike.

## The issue's premise is wrong: overlap is reachable today

#1934 recorded this as latent, on the grounds that the carve re-tiles the manifest so overlapping rows never survive. They do survive:

- `Truncate` narrows a straddling row to the new size (`narrowChunkRow`), leaving it ending exactly at the truncation point.
- A later write re-carves from an *earlier* chunk boundary, emitting a new row that starts inside the narrowed row's range.
- Nothing reaps the narrowed row's tail, so both rows now claim the same bytes.

Reproduced on **unmodified develop** with a randomized mutation soak (write / truncate / punch-hole, carve after each, drain local so reads go cold). 3 of 24 seeds hit manifest overlap, e.g. `row at 2894422 extends to 9863070, next row starts at 8208892`.

## Why not report it as an inconsistency

That was the issue's suggested direction and this PR's first attempt. It is wrong. With the error in place, the soak turns a cold read of a double-covered offset into a hard failure:

```
READ BROKEN at double-covered offset 8142960: cold read hydrate failed:
blockstore: file manifest inconsistent: offset 8142960 covered by both
".../3948656" and ".../8142960"
```

Those reads return **correct data** today. Since overlap is reachable through ordinary truncate+write, refusing would convert working reads into EIO on every non-badger backend. Loud beats silent only when the silent answer is wrong; here it isn't.

Greatest-start is also the *right* row: it is the newer one, holding what the last write put there. Badger already returns it. Under the tiling invariant nothing changes at all.

## Scope note

The soak also surfaced a **separate, pre-existing bug**, present on develop and unaffected by either variant of this change: after `PunchHole`, a cold read of the punched range does not read back as zeros (`PunchHole`'s contract is that it must). 5 of 24 seeds. Filed separately — not fixed here.

Test: overlapping rows resolve to the greatest start; an offset only the older row covers still resolves to it.

### DIFF
--- a/pkg/block/engine/manifest_inconsistent_test.go
+++ b/pkg/block/engine/manifest_inconsistent_test.go
@@ -81,9 +81,9 @@ func TestFindRowCoveringOffset_WellFormedRowResolves(t *testing.T) {
 	}
 }
 
-// Overlap is an invariant the carve maintains, not one the lookup may assume:
-// the row this walk reaches first need not be the one the indexed badger path
-// (largest start) returns, and both read back as ordinary data.
+// Non-overlap is an invariant the carve maintains, not one the lookup may
+// assume: the row this walk reaches first need not be the one the indexed
+// badger path (largest start) returns, and both read back as ordinary data.
 func TestFindRowCoveringOffset_OverlappingRowsAreReported(t *testing.T) {
 	rows := []*block.FileChunk{
 		{ID: "payload-1/0", DataSize: 8192},
--- a/pkg/block/engine/manifest_inconsistent_test.go
+++ b/pkg/block/engine/manifest_inconsistent_test.go
@@ -81,23 +81,25 @@ func TestFindRowCoveringOffset_WellFormedRowResolves(t *testing.T) {
 	}
 }
 
-// Non-overlap is an invariant the carve maintains, not one the lookup may
-// assume: the row this walk reaches first need not be the one the indexed
-// badger path (largest start) returns, and both read back as ordinary data.
-func TestFindRowCoveringOffset_OverlappingRowsAreReported(t *testing.T) {
+// Overlapping rows resolve to the greatest start, matching the indexed badger
+// lookup. A truncate narrows a straddling row to the new size and a later write
+// re-carves from an earlier boundary, so the narrowed row keeps claiming bytes
+// the new row also covers; the newer row holds what the last write put there.
+func TestFindRowCoveringOffset_OverlapResolvesToGreatestStart(t *testing.T) {
 	rows := []*block.FileChunk{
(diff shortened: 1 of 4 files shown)


marmos91:
Reworked since the last review: the fix is now greatest-start rather than reporting `ErrManifestInconsistent`, because the soak showed overlap is reachable and erroring breaks live reads. Details in the updated description. Requesting a fresh review of the final state.

PR #1954, opened 2026-08-18
fix(smb): claim durable handles by ID on reconnect

### LINKED ISSUE #1931: SMB durable-handle reconnect: Get and Consume are not atomic, so a concurrent sibling can be claimed instead of the validated handle
## Summary

On the SMB durable-handle reconnect path, `Get` and `Consume` are two separate store calls. The handler validates the handle `Get` returns, then claims it via `Consume`. Nothing holds a lock across the two, so a sibling handle inserted or removed in that window can make `Consume` act on a different handle than the one that was validated.

## Why it is only now reachable

Before #1926, the FileID index was a single-value key, so a file effectively had one findable durable handle and the two calls could not disagree. #1926 makes multiple handles per file persist correctly, which is the right fix — and it makes this pre-existing looseness in the reconnect design observable.

#1926 already closed the *deterministic* half of this: `Get` had no `ORDER BY` on the SQL backends and used map iteration order on memory, so it could validate one handle and destroy another with no concurrency at all. That is fixed — both now select the lowest-keyed live entry, and a conformance assertion that failed 5/20 runs on memory pre-fix passes 20/20.

What remains is the genuinely concurrent case: the selection is now consistent, but not atomic.

## Failure scenario

1. Client reconnects; handler calls `Get(fileID)` → handle A, validates it (owner, lease, app-instance-id, whatever the path checks).
2. Concurrently, handle A expires and is reaped, or a new handle B is inserted with a lower key.
3. Handler calls `Consume(fileID)` → returns B, or returns nothing.
4. The handler proceeds believing it claimed the handle it validated.

Consequences range from a failed reconnect (benign) to claiming a handle whose validation was never performed (not benign).

## Suggested direction

Options, roughly in increasing order of invasiveness:

- **Consume by handle ID rather than by FileID** on the reconnect path, so the claim names exactly what was validated. Likely the smallest correct change — the handler already has the handle from `Get`.
- Perform validation inside the same store transaction as the consume.
- Add a compare-and-consume primitive (`ConsumeIfMatches(fileID, handleID)`).

The first is worth trying first: if the caller can name the handle it validated, the window closes without new store API surface.

## Scope

Found by the code-reviewer gate on #1926 and deliberately left out of that PR to avoid widening a data-correctness fix into a reconnect-protocol change. Related: #1915, #1926, and the durable-handle work in #1908.

marmos91:
Closes #1931.

## Problem

On the SMB durable-handle reconnect path, `Get` and `Consume` were two separate store calls, each keyed independently by a *secondary* index — FileID for V1 (DHnC), CreateGuid for V2 (DH2C). The handler validated the handle `Get` returned, then claimed one via `Consume`. Nothing held a lock across the pair, so a sibling handle inserted or reaped in that window could be claimed instead of the one that was validated.

#1926 made multiple durable handles per file persist correctly and fixed the *deterministic* half (both sides now select the lowest-keyed live entry). What remained was the genuinely concurrent case: the selection was consistent, but not atomic.

## Fix

Take the first option the issue suggests — consume by handle ID rather than by FileID/CreateGuid. The handler already holds the record `Get` returned, so naming it closes the window with no new store API surface.

`ConsumeDurableHandleByFileID` + `ConsumeDurableHandleByCreateGuid` collapse into one `ConsumeDurableHandle(ctx, id string)`:

- **memory** — map lookup + delete under the existing mutex.
- **badger** — `getDurableHandleTx` + `deleteDurableHandleTx` inside one `db.Update` txn.
- **sqlite / postgres** — `DELETE FROM durable_handles WHERE id = $1 RETURNING …`, keyed on the primary key instead of a subquery over a secondary index.

`validateAndRestore` loses its `consume` closure parameter and calls `durableStore.ConsumeDurableHandle(ctx, handle.ID)` directly — both reconnect paths now share one claim, so the V1/V2 asymmetry in the closure disappears. The V1-via-DH2C fallback (zero CreateGuid) still *looks up* by FileID; only the claim changes.

Net −164 lines across interface, four backends, and the conformance suite.

## Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean.
- `go test -count=1 ./pkg/metadata/... ./internal/adapter/smb/...` green.
- Conformance `DurableHandles/Consume` passes on memory, badger, sqlite, **and postgres** (run against a local 18.6 instance with `-tags=integration`). It asserts a sibling on the same FileID survives, stays reachable by FileID afterward, and that re-consuming an already-claimed ID reports nil rather than falling through to the sibling — the exact substitution this fixes.
- The two `TestPostgresPutFile_*` integration failures seen locally reproduce identically on unmodified `develop` (unrelated env config, "host is required").

### DIFF
--- a/internal/adapter/smb/handlers/durable_context.go
+++ b/internal/adapter/smb/handlers/durable_context.go
@@ -509,10 +509,9 @@ func processV1Reconnect(
 	// Non-destructive lookup: validation-failure paths (share/path/username/
 	// access mismatch) must leave the persisted handle reclaimable until
 	// expiry — a transient mismatched retry should not destroy a valid
-	// durable open. The TOCTOU window is closed on the *success* path
-	// (claimDurableHandleByFileID below) which atomically re-reads via
-	// Consume; a second concurrent reconnect that also passes validation
-	// will find the handle already gone and return OBJECT_NAME_NOT_FOUND.
+	// durable open. The success path claims this exact handle by ID
+	// (validateAndRestore); a second concurrent reconnect that also passes
+	// validation finds it gone and returns OBJECT_NAME_NOT_FOUND.
 	handle, err := durableStore.GetDurableHandleByFileID(ctx, fileID)
 	if err != nil {
 		logger.Warn("processV1Reconnect: store error", "error", err)
@@ -547,10 +546,7 @@ func processV1Reconnect(
 	checkPath := persistedHasLease
 
 	openFile, status, restoreErr := validateAndRestore(ctx, durableStore, metaSvc, handle, sessionID, username,
-		sessionKeyHash, shareName, filename, checkPath,
-		func(ctx context.Context) (*lock.PersistedDurableHandle, error) {
-			return durableStore.ConsumeDurableHandleByFileID(ctx, fileID)
-		})
+		sessionKeyHash, shareName, filename, checkPath)
 	return openFile, handle.LeaseState, handle.LeaseEpoch, handle.OriginalFileID, status, restoreErr
 }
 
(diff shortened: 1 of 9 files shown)


marmos91:
@copilot review

PR #1957, opened 2026-08-18
fix(smb): publish the OpenFile name triple as one atomic value

### LINKED ISSUE #1940: SMB: OpenFile.Path/FileName/ParentHandle read without the handle lock across most of the handler package (262 sites)
`OpenFile.Path`, `FileName` and `ParentHandle` are mutable — SET_INFO rename rewrites all three on a live handle — but they are read without the handle lock from most of the SMB handler package.

This came out of the sweep behind #1938. That PR fixed the WRITE seam and, after review, the three sites that a reviewer and Copilot between them identified. The sweep showed the problem is much wider than the sites anyone named.

## Scale

262 references to `openFile.Path` / `.FileName` / `.ParentHandle` across 15 non-test files:

```
set_info.go        88     close.go           44     write.go           21
query_info.go      21     read.go            19     handler.go         17
ioctl_sparse.go    13     stub_handlers.go    8     set_reparse_point.go 8
flush.go            8     durable_context.go  5     ioctl_fsctl.go      4
query_directory.go  3     lock.go             2     ioctl_copychunk.go  1
```

Not all are unguarded — set_info.go's are the rename path itself and write.go/close.go/durable_context.go/handler.go are now handled by #1938 — but the majority in read.go, flush.go, query_info.go, ioctl_*.go and stub_handlers.go are live reads with no lock.

## Why this is not cosmetic

Most of these are log-line arguments, which invites dismissing them as harmless. A Go string is a pointer plus a length and is not read atomically. A read racing a write can observe the pointer from one value and the length from another, which is not a wrong-looking log line but a potential out-of-bounds read. `go test -race` treats it as a genuine race because it is one.

Severity is not uniform, and flattening it would be a mistake:

- **Persisted / observable** — the value is written somewhere durable or returned on the wire. #1938 found one of these: `buildPersistedDurableHandle` could persist a durable-handle row mixing one rename's name with another's parent, and durable handles are replayed on reconnect, so it is durable wrong state rather than a transient glitch. Any remaining instance in `query_info.go` (which answers the client) is in this class.
- **Logging-only** — real race, bounded blast radius.

An audit of `query_info.go` in particular is worth doing before the log sites, since it composes a response the client consumes.

## Why per-site locking is the wrong fix

#1938 added `GetPath()`, `GetFileName()` and `GetPayloadID()` accessors and snapshots the group under one `RLock` where several fields are used together. Sprinkling that across 262 sites would be a large, mechanical, hard-to-review diff, and nothing stops the next handler from reading the field directly.

The durable fix is to make it impossible: rename the fields to unexported (`path`, `fileName`, `parentHandle`) so every access must go through an accessor, and let the compiler find the call sites. That is a single mechanical refactor whose correctness the type checker verifies, rather than a judgement call repeated 262 times.

Two things to preserve when doing it:

1. Where a caller needs more than one of the three, it must snapshot them together under one `RLock` — not call two accessors in sequence. Calling `GetPath()` then `GetFileName()` reintroduces exactly the torn pairing that #1938 fixed in the durable-handle path.
2. Do not call an accessor while already holding the handle lock. `sync.RWMutex.RLock` is not safely reentrant when a writer is queued, so a nested call can deadlock. #1938 hit this: `buildPersistedDurableHandle` had to switch back to direct field reads once it took `RLock` for the whole snapshot.

## Lock ordering, for whoever picks this up

`h.durablePurgeMu` is taken before `openFile.mu` on the disconnect-persist path. The other two `durablePurgeMu` acquisitions (`purgeDisconnectedConflicts`, `SeedFromDurableHandles`) do not hold `openFile.mu`, so there is currently no inversion. Preserve that ordering.

Relates to #1938, #1900.

marmos91:
Closes #1940.

`OpenFile.Path`, `.FileName` and `.ParentHandle` were mutable exported fields. SET_INFO rename rewrites all three on a live handle, while the rest of the handler package read them with no lock from ~280 sites.

## Approach

The issue proposed unexporting the fields so the compiler forces every access through an accessor. That does not work here: all 156 files are `package handlers`, so unexported fields stay directly reachable and the compiler enforces nothing. Unexporting would have bought a one-time sweep and no lasting guarantee.

Instead the three fields become one immutable `OpenName` published through an `atomic.Pointer`:

```go
type OpenName struct {
	Path         string
	FileName     string
	ParentHandle metadata.FileHandle
}

func (f *OpenFile) Name() *OpenName   // never nil, safe under the handle lock
func (f *OpenFile) SetName(n OpenName)
```

A single load yields all three fields from the same rename. That removes the whole class rather than the named instances, and it disposes of both hazards the issue calls out:

- **Torn pairing** — `GetPath()` then `GetFileName()` could straddle a rename. There is now one load, so a caller that needs two or three fields takes one snapshot and cannot mix generations.
- **Reentrancy deadlock** — `RLock` is not safely reentrant with a writer queued, so an accessor could not be called while already holding `mu`. An atomic load has no such constraint; `buildPersistedDurableHandle` no longer needs its direct-field-read workaround.

Call sites were rewritten from compiler error positions rather than by text matching, so no site was missed and none was rewritten by accident.

## Torn pairings actually found and fixed

These were live bugs, not just unguarded reads:

- **CLOSE delete-on-close** (`close.go`) — the removal notification built its parent path from the CLOSE-entry snapshot while the delete target came from a later one, with a metadata flush in between. A rename landing in that window made the notification name the new file under the old parent.
- **SET_INFO stream rename** (`set_info.go`) — six separate reads: the destination directory from one, the move's source name from another, the notification path from a third. A concurrent rename would have moved the wrong child out of the wrong directory.
- **`checkShareModeConflict` / `isFileOrBaseDeletePending`** (`handler.go`) — a path derived from one read, compared against a second read of the same field.
- **`buildFileStreamInformation`** (`query_info.go`) — `baseName` from the function snapshot, but the ADS enumeration re-read the parent, so streams could be listed under a different parent than the base name came from.
- **MFsymlink conversion** (`close.go`, `set_reparse_point.go`) — the purge and notification re-read the path *after* the file had been removed under a different snapshot's name.

`buildPersistedDurableHandle` still holds `mu` across the row build (it reads other mutable fields), and both rename sites still hold `mu` across their read-modify-write so two renames cannot interleave. The `durablePurgeMu` → `openFile.mu` ordering is unchanged.

## Verification

- `TestOpenName_ConcurrentRenameAndRead_NoRace` hammers `SetName` against four concurrent readers and asserts the triple always agrees with itself. Confirmed it fails — data race plus torn-name assertion — when the atomic is swapped back for a plain pointer, so it pins the invariant rather than merely passing.
- `go build ./...`, `go vet ./...`, `golangci-lint`, `gofmt` clean.
- `go test -race ./internal/adapter/smb/...` green across all 12 packages.

### DIFF
--- a/internal/adapter/smb/handlers/auth_priming_test.go
+++ b/internal/adapter/smb/handlers/auth_priming_test.go
@@ -57,7 +57,7 @@ func TestClose_PrimesAuthContextFromOpenFile(t *testing.T) {
 		TreeID:    treeID,
 		SessionID: sess.SessionID,
 		// No MetadataHandle / PayloadID — Close skips the BlockStore /
-		// metadata-flush paths entirely, but Step 2b primer still runs.,
+		// metadata-flush paths entirely, but Step 2b primer still runs.
 	}).WithName(OpenName{Path: "/share/a.txt"})
 	h.StoreOpenFile(openFile)
 
@@ -181,7 +181,7 @@ func TestRead_DeniesOnGrantedAccessStripped(t *testing.T) {
 	openFile := (&OpenFile{
 		FileID:        [16]byte{0xDE, 0xAD},
 		DesiredAccess: uint32(types.FileReadData), // requested
-		GrantedAccess: 0,                          // DACL stripped it,
+		GrantedAccess: 0,                          // DACL stripped it
 	}).WithName(OpenName{Path: "/share/secret.txt"})
 	h.StoreOpenFile(openFile)
 
@@ -207,7 +207,7 @@ func TestRead_AllowsOnGrantedAccess(t *testing.T) {
 	openFile := (&OpenFile{
 		FileID:        [16]byte{0xBE, 0xEF},
 		DesiredAccess: 0,                          // pre-DACL not relevant
-		GrantedAccess: uint32(types.FileReadData), // DACL granted,
+		GrantedAccess: uint32(types.FileReadData), // DACL granted
 	}).WithName(OpenName{Path: "/share/public.txt"})
 	h.StoreOpenFile(openFile)
 
@@ -234,7 +234,7 @@ func TestWrite_DeniesOnGrantedAccessStripped(t *testing.T) {
 	openFile := (&OpenFile{
 		FileID:        [16]byte{0xCA, 0xFE},
 		DesiredAccess: uint32(types.FileWriteData), // requested
-		GrantedAccess: 0,                           // DACL stripped it,
+		GrantedAccess: 0,                           // DACL stripped it
 	}).WithName(OpenName{Path: "/share/ro.txt"})
 	h.StoreOpenFile(openFile)
 
--- a/internal/adapter/smb/handlers/auth_priming_test.go
+++ b/internal/adapter/smb/handlers/auth_priming_test.go
@@ -52,14 +52,13 @@ func TestClose_PrimesAuthContextFromOpenFile(t *testing.T) {
 		ShareName: "/share",
 	})
 
-	openFile := &OpenFile{
+	openFile := (&OpenFile{
 		FileID:    [16]byte{0x01, 0x02},
 		TreeID:    treeID,
 		SessionID: sess.SessionID,
-		Path:      "/share/a.txt",
 		// No MetadataHandle / PayloadID — Close skips the BlockStore /
-		// metadata-flush paths entirely, but Step 2b primer still runs.
-	}
+		// metadata-flush paths entirely, but Step 2b primer still runs.,
+	}).WithName(OpenName{Path: "/share/a.txt"})
 	h.StoreOpenFile(openFile)
 
 	// Build a ctx with zero session/tree state — mirrors the dispatcher,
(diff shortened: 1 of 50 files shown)


marmos91:
Addressed in 12863eca.

**Shared `emptyOpenName` pointer** — real footgun, fixed differently than suggested. Rather than lazily publishing a per-`OpenFile` empty, `Name()` now returns `OpenName` by value. That makes `openFile.Name().Path = x` a compile error rather than a silent corruption of shared state, so the mutation cannot be reintroduced at all, and it drops the package-level var instead of adding per-handle allocation.

**Stale `openFile.Path` reference in read.go** — fixed.

**Trailing-comma comments** — correct, and they were mechanical damage from the field move: the element separator was appended after each part's trailing comment instead of after its value. Swept the whole branch and fixed all six, not just the three flagged. The sweep also caught one your sample didn't: in `rename_scan_close_race_test.go` the comment "child of the directory being renamed" described `ParentHandle`, which moved into `WithName`, leaving the comment attached to `MetadataHandle` above it. Reworded and reattached.

PR #1960, opened 2026-08-18
fix(smb-conformance): excuse the self-contradictory channel-sequence CSN draw

### LINKED ISSUE #1923: smb2.replay.channel-sequence fails intermittently on badger-fs (byte-identical binary passes and fails)
## Summary

`smb2.replay.channel-sequence` fails intermittently against the badger-fs profile. It is **not** in `KNOWN_FAILURES.md` (it appears only in prose), so it flips the conformance verdict when it fails.

## Evidence

Two smbtorture runs on the same branch, 36 minutes apart, with a **byte-identical** `dfs` binary (`2bf7a95d…b28044`, built `-trimpath` — the only delta between the two commits was a `_test.go` file, which is not linked into the server):

| | run 31011120983 | run 31014070873 |
|---|---|---|
| `smb2.replay.channel-sequence` | **pass** | **fail** |
| replay tests executed | 56 | 56 |
| exit-124 timeouts | 7 (same suites) | 7 (same suites) |
| graded | success | failure |

Failure assertion:

```
source4/torture/smb2/replay.c:4673: failed to test CSN with replay flag
```

The suite ran to completion in both runs — this is not a truncation artifact. Both runs also hit the *same* 7 timeouts on the *same* suites, dying at the same test in each, so the timeouts are not what differed.

## Why this is filed separately

#1910 originally attributed the flipping verdict to grading that folded timeout-prone suites into pass/fail. That diagnosis was wrong — the grading was correct, and #1919 fixed a different (real) hole where timed-out and never-reached tests vanished silently from the report. What actually flipped the verdict is this test, and it is either a genuinely intermittent product bug in SMB2 replay / channel-sequence handling or a racy test.

Channel sequence is how a client's requests are ordered across a reconnect after a channel failure, so an intermittent failure here is worth understanding rather than blacklisting — it touches the same durable-handle / replay machinery as #1701, #1832 and #1908.

## Suggested triage

1. Reproduce by running the `smb2.replay` filter in a loop against a fixed binary to establish the failure rate.
2. Read `source4/torture/smb2/replay.c` around line 4673 to see exactly what CSN invariant is being asserted, and whether the test itself has a race (it would not be the first — see #1832, where a torture expectation was mutually exclusive with conformant behavior).
3. Only add a `KNOWN_FAILURES.md` row if it is established as a test-side problem; per the project rule, fix flakes rather than documenting them.

Do not blacklist it to make CI quiet — with #1919 merged, the report now distinguishes new failures from known ones and from inconclusive suites, so this failing is visible and actionable rather than noise.

marmos91:
## Root cause: an upstream torture-test defect, not a DittoFS bug

`smb2.replay.channel-sequence` flipped the badger-fs verdict between two runs of a
byte-identical binary (#1923). The failing run's log names the exact input:

```
Testing setinfo (replay: true) with CSN 0x7fff, expecting: NT_STATUS_FILE_NOT_AVAILABLE
WARNING!: replay.c:4627: status was NT_STATUS_OK, expected NT_STATUS_FILE_NOT_AVAILABLE
```

`test_channel_sequence_table` (samba `source4/torture/smb2/replay.c`) runs a 16-row
ChannelSequence table against one Open. Fourteen rows are fixed; two are drawn at
random — `rand() % 0x8000 + 0x7fff` giving `[0x7fff, 0xfffe]`, and `rand() % 0x8000`
giving `[0, 0x7fff]` — and both expect `STATUS_FILE_NOT_AVAILABLE` on the grounds
that the value is stale or too far ahead. Both ranges include `0x7fff`, which is
neither: MS-SMB2 §3.3.5.2.10 updates the Open when the unsigned 16-bit difference
"is less than or equal to 0x7FFF".

**The table is unsatisfiable when a draw lands there, whichever way a server resolves
a `+0x7fff` difference.** Accept it and the drawn row fails, because it demanded a
rejection. Reject it and rows 5, 6 and 10 — fixed rows that pass today — fail
instead, because each depends on `+0x7fff` being a forward step. No deterministic
server passes both arms. I verified this by running both policies over the full
table rather than reasoning about it.

Samba is in the first camp: `smbd_smb2_request_dispatch_update_counts` flips the
comparison only when `abs(cmp) > INT16_MAX`, which is false at `0x7fff`, so it
accepts and fails the drawn row too. Its `pre_request_count` gate cannot save it
either — `smbd_smb2_request_reply_update_counts` drains the gauges on every reply,
and no earlier row takes the branch that raises them.

DittoFS's `VerifyChannelSequence` matches Samba's arithmetic bit-for-bit at this
boundary. **No product code changes here.**

## Empirically confirmed on upstream Samba

The diagnosis does not rest on reading Samba's source. smbtorture 4.22.6 was run against
**smbd 4.22.6** (`quay.io/samba.org/samba-server:v0.8` — the same Samba build as the pinned
torture binary), with libc `rand()` forced to 0 via `LD_PRELOAD` so `csn_rand_high` draws
exactly `0x7fff` every time instead of 1-in-32768.

**Control, unmodified `rand()`** — Samba runs the full 16-row table and passes, including:

```
Testing write (replay: false) with CSN 0x7fff, expecting: NT_STATUS_OK
```

so Samba does treat `+0x7fff` from tracked `0` as a legal forward step.

**Forced boundary draw, same server, same binary:**

```
Testing write (replay: false) with CSN 0x7fff, expecting: NT_STATUS_FILE_NOT_AVAILABLE
WARNING!: replay.c:4627: status was NT_STATUS_OK, expected NT_STATUS_FILE_NOT_AVAILABLE
```

Same file, same line, same expected/actual pair as the DittoFS failure in #1923.
**Samba fails its own test on that draw**, while the same server passes the row demanding
the opposite answer.

Reproduction needs `durable handles = yes`, `kernel oplocks = no`, `kernel share modes = no`,
`posix locking = no`, and `smbd -F --no-process-group`.

## Frequency cross-check

If the real failure rate were higher than 1-in-32768 per draw, the diagnosis would be
incomplete. Across the last 60 SMB-conformance runs there were 192 `smbtorture / <profile>`
jobs: 145 completed successfully, 1 failed, and that failure was
`smb2.dirlease.unlink_same_set_and_close` — unrelated. So `channel-sequence` failed
**0 times in ~1750 draws**, against a model predicting 0.05 expected failures.

The single failure in #1923 is explained by selection bias: that run was examined *because*
it failed, not sampled at random.

## The change

`parse-results.sh` reclassifies a `channel-sequence` failure to `skip:` **only** when
that exact draw was printed during the test, in the same pre-pass that already
excuses connection flakes. The excuse is one-shot per draw. The test stays off
`KNOWN_FAILURES.md`, per the issue's explicit instruction not to blacklist it — any
other channel-sequence failure, including a genuine regression in the mechanism, is
still graded normally and still fails CI.

## Verification

Replayed both real CI logs from #1923 through the parser:

| | before | after |
|---|---|---|
| failing run (attempt 1 of `31014070873`) | exit 1, 1 new failure | exit 0, all known |
| passing run (`31011120983`) | exit 0, 498/38/40 | exit 0, 498/38/40 — identical |

Note the referenced `31014070873` looks green in the UI because attempt 2 was a
rerun; attempt 1 is the failure.

Plus 4 new `parse-results_test.sh` cases pinning the gate: the excused draw, a
channel-sequence failure *without* the draw (still red), one draw excusing only one
block, and no leak into the next test. `go test ./internal/adapter/smb/...` green;
shellcheck clean.

## Follow-up

Worth reporting upstream to Samba — the fix is to start the `csn_rand_high` range at
`0x8000` and end `csn_rand_low` at `0x7ffe`. Not filed yet.

Closes #1923

### DIFF
--- a/internal/adapter/smb/handlers/channel_sequence_test.go
+++ b/internal/adapter/smb/handlers/channel_sequence_test.go
@@ -114,3 +114,34 @@ func TestVerifyChannelSequence_SeedNonZero(t *testing.T) {
 		t.Fatal("write below seeded csn should be rejected")
 	}
 }
+
+// TestVerifyChannelSequence_ForwardBoundary pins the largest difference that
+// still counts as a forward step. MS-SMB2 §3.3.5.2.10 updates the Open when the
+// unsigned 16-bit difference "is less than or equal to 0x7FFF", so 0x7fff ahead
+// of the tracked sequence is accepted and 0x8000 ahead is not. Samba draws two
+// of its channel-sequence table rows at random from ranges that include 0x7fff
+// while expecting a rejection, which is why that draw can never pass.
+func TestVerifyChannelSequence_ForwardBoundary(t *testing.T) {
+	f := &OpenFile{}
+	if !f.VerifyChannelSequence(0, true) {
+		t.Fatal("seed write at csn=0 should be allowed")
+	}
+	if !f.VerifyChannelSequence(0x7fff, true) {
+		t.Fatal("csn=0x7fff is a forward step and must be allowed")
+	}
+	if f.channelSeq != 0x7fff {
+		t.Fatalf("expected tracked csn to advance to 0x7fff, got 0x%04x", f.channelSeq)
+	}
+	// Resending on the now-current sequence is still allowed.
+	if !f.VerifyChannelSequence(0x7fff, true) {
+		t.Fatal("write on the current csn should be allowed")
+	}
+	// One past the boundary reads as a wrapped (stale) counter, not a step.
+	g := &OpenFile{}
+	if !g.VerifyChannelSequence(0, true) {
+		t.Fatal("seed write at csn=0 should be allowed")
+	}
+	if g.VerifyChannelSequence(0x8000, true) {
+		t.Fatal("csn=0x8000 exceeds the forward window and must be rejected")
+	}
+}
--- a/internal/adapter/smb/handlers/channel_sequence_test.go
+++ b/internal/adapter/smb/handlers/channel_sequence_test.go
@@ -7,6 +7,10 @@ import "testing"
 // for a modifying opcode (WRITE/SET_INFO/IOCTL) and asserts the accept/reject
 // decision matches the upstream expected NT_STATUS at every step. This is the
 // authoritative spec for smb2.replay.channel-sequence.
+//
+// Rows i3 and i8 stand in for two values the test draws at random from ranges
+// that both include 0x7fff while expecting a rejection — the value row i5
+// requires to succeed. That draw demands both outcomes and can never pass.
 func TestVerifyChannelSequence_SambaTable(t *testing.T) {
 	// allow == true means the op is expected to succeed (NT_STATUS_OK);
 	// allow == false means it must be rejected (NT_STATUS_FILE_NOT_AVAILABLE).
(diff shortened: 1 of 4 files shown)


marmos91:
## Empirically confirmed on upstream Samba

The diagnosis no longer rests on reading Samba's source. I ran smbtorture 4.22.6 against **smbd 4.22.6** (`quay.io/samba.org/samba-server:v0.8`, same Samba build as the pinned torture binary), with libc `rand()` forced to 0 via `LD_PRELOAD` so `csn_rand_high` draws exactly `0x7fff` every time instead of 1-in-32768.

**Control — real Samba, unmodified rand():** the first table runs all 16 rows and passes, including
```
Testing write (replay: false) with CSN 0x7fff, expecting: NT_STATUS_OK
```
so Samba does treat `+0x7fff` from tracked `0` as a legal forward step.

**Forced boundary draw — same Samba, same binary:**
```
Testing write (replay: false) with CSN 0x7fff, expecting: NT_STATUS_FILE_NOT_AVAILABLE
WARNING!: replay.c:4627: status was NT_STATUS_OK, expected NT_STATUS_FILE_NOT_AVAILABLE
```

Same file, same line, same expected/actual pair as the DittoFS CI failure in #1923. **Samba fails its own test on that draw.**

## Frequency cross-check

Across the last 60 SMB-conformance runs there were 192 `smbtorture / <profile>` jobs — 145 completed successfully, 1 failed, and that one failure was `smb2.dirlease.unlink_same_set_and_close`, unrelated. So `channel-sequence` failed **0 times in ~1750 draws**, against a model predicting 0.05 expected failures. The rate matches 1-in-32768 per draw, and the single failure in #1923 is explained by selection bias — we examined the run *because* it failed.

Together these close the two things that could have falsified the diagnosis: that the rate is higher than modelled (it isn't), and that Samba somehow passes where we fail (it doesn't).

PR #1965, opened 2026-08-18
fix(block): reach the unplaceable-row guard from every read path, not just cold ones

### LINKED ISSUE #1888: Unplaceable-row guard is unreachable for the rows cold seeding skips, so they still read as silent zeros

marmos91:
Closes #1888.

## The composition

#1881 added a guard so a manifest row whose ID carries no parseable offset ("unplaceable") makes an uncovered read refuse with `ErrManifestInconsistent` instead of being zero-filled. But `SeedColdFromManifest` skips exactly those rows, so they get no cold interval. A range with no interval is a plain hole, and `readAtInternal` returned at `if !cold` without ever consulting the manifest. The rows the guard was written for were the ones it could never see.

## Wider than the issue described

Two things turned up that the issue did not have:

**The engine fix alone is a no-op on badger, the default backend.** `GetFileChunkAtOffset` and `GetFileChunkAtOrAfterOffset` silently `continue` past a malformed index key, so the unplaceable row is invisible to both indexed lookups — verified with a throwaway probe: `resolveCovering` returns `(nil, nil)` and `resolveNextChunkStart` returns no successor. Shipping only the read-path change would have repeated #1881's mistake.

**`DataExtents` had the same blind spot, on an op the guard cannot reach.** It dropped an unplaceable row (`if !ok || off >= fileSize { continue }`), so SEEK_HOLE and READ_PLUS report the range as hole. Neither op calls READ for a hole segment, so a sparse-aware copier (`rsync --sparse`, `cp --sparse=always`) skips the range on SEEK alone and loses the bytes — the #1888 failure mode via a different door. Returning an error from `DataExtents` does not fix it: both callers (`seek.go:110`, `read_plus.go:162`) drop the error and fall back to the CAS block list, which is equally blind.

## The change

- `journal.Store.ReadAt` returns `ReadState{Cold, Hole}` instead of `cold bool`. The interval walk already classified each piece, so the bit is free. Returning a struct rather than a fourth value keeps every `_, _, err :=` call site compiling.
- `readAtInternal` reconciles a hole against the manifest, not just a cold range. A fully warm read still returns without touching the manifest, and a hole on a local-only share is skipped — there is no remote to hydrate from and nothing carves manifest rows there.
- Badger's two indexed lookups report `ErrManifestInconsistent` rather than skipping a malformed index key: the key's suffix is the row ID's suffix verbatim, so it is exactly the unplaceable row. Scoped for coverage (one bad row must not make a whole payload unreadable), unconditional for succession (an unplaceable row may be the true successor) — mirroring the `ListFileChunks` fallback.
- `DataExtents` widens its map to the whole file when it meets such a row. Over-reporting data is the RFC 7862-safe direction and forces the READ that refuses.
- Corrected the comment at `shares/service.go` that asserted the safety property the code did not have, and `dataextents.go`'s "same data view ReadAt reconstructs" claim.

## Verification

Each half is load-bearing, confirmed by reverting it:

- Without the engine change: `TestReadAt_UncoveredOffsetOnUnplaceableRowRefuses` returns `err=nil` with an all-zero buffer.
- Without the badger change: the same test's `OffsetIndexed` subtest fails while the fallback passes.
- Without the `DataExtents` change: `DataExtents = [[0 4096]]`, i.e. a hole SEEK lets a client skip.

`TestReadAt_CoveredOffsetIgnoresUnplaceableRow` also failed before the fix — that is the `DataExtents` ⊃ `ReadAt` asymmetry the issue flags as related, and the leading-zeros signature from #1879, so this closes it too.

Tests run both manifest-store shapes (`ListFileChunks` fallback and badger's offset index); the unplaceable-row contract is pinned in the `storetest` conformance suite so any future indexed backend inherits it. Full suite and `-race` green.

## Not done

Nothing establishes that the #1879 reporter's store actually holds an unplaceable row — that still needs the manifest dump. This closes the mechanism, not that diagnosis.

### DIFF
--- a/docs/internals/architecture.md
+++ b/docs/internals/architecture.md
@@ -310,11 +310,16 @@ dispatcher is suppressed and `Flush`/`SyncNow` are the sole carve drivers.
 
 `ReadAt` resolves by `(payloadID, offset)`. A range present in the journal is
 served straight from its segment. A range that was written but has since been
-evicted returns `cold=true`; the engine then resolves the chunk's remote
-locator from the FileChunk manifest, ranged-GETs its enclosing `blocks/<id>`
-object, decodes the wire frame, recomputes BLAKE3 over the chunk bytes, and —
-on success — hydrates the verified bytes back into the journal so subsequent
-reads are warm.
+evicted reports `ReadState.Cold`, and a range the journal has no interval for
+reports `ReadState.Hole`; the engine reconciles both against the FileChunk
+manifest, since the journal alone cannot tell a never-written range from one
+whose interval it lost. For each covering row it resolves the remote locator,
+ranged-GETs the enclosing `blocks/<id>` object, decodes the wire frame,
+recomputes BLAKE3 over the chunk bytes, and — on success — hydrates the verified
+bytes back into the journal so subsequent reads are warm. A range no row covers
+is a genuine sparse hole and stays zero-filled, unless the payload holds a row
+whose ID carries no offset: that range is unknown, so the read refuses with
+`ErrManifestInconsistent` rather than inventing zeros.
 
 Integrity is fail-closed: every remote fetch is BLAKE3-verified before the
 bytes reach the caller, and a mismatch is returned as an error (and counted),
--- a/docs/internals/implementing-stores.md
+++ b/docs/internals/implementing-stores.md
@@ -462,8 +462,9 @@ type LocalStore interface {
 ```
 
 A write buffers a dirty range and local-acks without fsync; `Commit` is the
-durability point. On a cold read `ReadAt` returns `cold=true` and the engine
-`Hydrate`s the range back from the remote store. `Carve` packs a file's dirty
+durability point. `ReadAt` reports both evicted (`ReadState.Cold`) and uncovered
+(`ReadState.Hole`) ranges; the engine reconciles either against the FileChunk
+manifest and `Hydrate`s whatever the manifest says is remote-resident. `Carve` packs a file's dirty
 ranges into remote blocks via the injected `BlockSink` (see the carve pass in
 [the architecture doc](architecture.md#carve-local--remote)); `Evict` frees
 whole fully-synced segments under pressure.
--- a/docs/internals/implementing-stores.md
+++ b/docs/internals/implementing-stores.md
@@ -442,7 +442,7 @@ and per-method contract; the representative methods are:
 type LocalStore interface {
     // --- Data plane (payloadID + offset keyed) ---
     WriteAt(ctx context.Context, payloadID string, offset int64, data []byte) error
-    ReadAt(ctx context.Context, payloadID string, offset int64, dst []byte) (n int, cold bool, err error)
+    ReadAt(ctx context.Context, payloadID string, offset int64, dst []byte) (n int, st journal.ReadState, err error)
     Hydrate(ctx context.Context, payloadID string, offset int64, data []byte) error // fill from remote on cold read
     Commit(ctx context.Context, payloadID string) error                             // fsync buffered writes
     FileSize(ctx context.Context, payloadID string) (int64, bool)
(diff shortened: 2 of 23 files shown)


marmos91:
Addressed: `ensureAndReadFromLocal` now serves both cold ranges and uncovered holes, so the wrapper said "cold read hydrate failed" for half its failures. Renamed to name the reconcile and carry the payload and offset (047fdee0).

Issue #1936, opened 2026-08-06
develop E2E has been red since 2026-07-28: GC local-disk assertion regressed at #1890, plus a separate NLM rpcbind registration failure


marmos91:
develop's E2E Tests job has been failing continuously since at least 2026-07-28 and nothing currently tracks it — the previous "🔴 develop CI is RED" issues (#1713, #1721, #1723, #1726, #1728) are all closed. E2E does not run on pull requests, only on pushes to develop, so it is not part of any PR's merge gate and the red has gone unnoticed for over a week.

There are two independent failures, not one.

## 1. TestBlocksFlipLifecycle_NFS / _SMB — regressed at a specific commit

These two were **passing** on 2026-07-28 (`b5c01e69`) and 2026-07-29 (`3d659450`), and started failing at `1f95b389` — "fix(block): report the physical segment footprint as local disk_used (#1890)". The only other commit in that range is `7eeec0da` "chore: clean planning files", which cannot affect E2E. Every run since has failed the same way.

The assertion, `test/e2e/blocks_flip_test.go:218`, is the final step of the unlink → GC leg:

```
Error:    Should be zero, but was 1024
Messages: [nfs] GC must free the local block too (local disk used should be 0, got 1024)
```

SMB reports 937 rather than 1024. Everything before this point in the test passes, including the cold read after evict and the remote-side check that `blocks/` is empty with no `cas/` objects introduced — so GC does reclaim the remote tier, and the file's data is gone. Only the local number is non-zero.

This is a direct consequence of what #1890 deliberately changed. `DiskUsed` used to be derived from record payload lengths and now reports the journal's `diskBytes`: the physical footprint of the `.seg` files, which by design includes segment headers and record framing that the old measure ignored. The test asserts exactly `0`, which under the new semantics is only reachable if GC also **retires the now-empty segment file** — not merely the records inside it.

So one of two things is true, and I do not think it can be settled from the log alone:

- **Product**: GC reaps the records but leaves an emptied segment file on disk, so its header/framing keeps charging the share forever. That is a real (small, bounded) leak and the test is right to fail.
- **Test**: retaining the active segment's header is intended and permanent, in which case `require.Zero` encodes the pre-#1890 logical semantics and should become a floor (e.g. "no live payload bytes", or `<= segment header size`).

[...]

PR #1966, opened 2026-08-18
fix(nfs): register with the system rpcbind without an adapter restart

marmos91:
Addresses the two failures in #1936. Both root causes were confirmed on a
Linux VM with the real netns harness, not from reading code.

## 1. NLM never registers with rpcbind — real product bug

`register_with_system` was only consumed when the NFS adapter (re)started.
#1810 stopped restarting an adapter whose bind address and enabled state are
unchanged, so the E2E driver's `adapter settings nfs update
--portmapper-register-with-system` followed by `adapter enable nfs --port 12049`
(the default port, on an already-enabled adapter) became a no-op reload: the
sysreg sidecar never ran and nothing — NFS, MOUNT or NLM — was registered.

Bisected on hardware: the case passes at `9ed1a163` and fails from `ca4761fc`
(#1810) onward, which matches the first red E2E run on 2026-07-21.

`applyNFSSettings` now reconciles the `sysreg` sidecar live, the way it already
does for the mDNS advertiser, so the setting takes effect on a running adapter
whoever toggles it. The transition talks to rpcbind and is bounded by
`systemRegTimeout`, so it runs in the background: the callers are the accept
loop and the settings-watcher goroutine, and neither may block on it.

Verified on a fresh Ubuntu 24.04 VM:

- before (origin/develop): `[nlm] NLM NOT registered in sns rpcbind` — the exact CI failure
- after: `100021 1/3/4 tcp+udp 12049 nlockmgr`, and all four interop cases run

The driver now polls for program 100021 instead of sampling once a second after
enable, since the setting arrives via the settings-watcher poll rather than an
adapter restart.

## 2. blocks-flip GC assertion — test bug

#1890 deliberately changed `local_disk_used` from payload accounting to the
journal's physical segment footprint. The unlink's tombstone record and its
segment header are real bytes that survive GC by design; that PR already relaxed
the equivalent unit assertion in `coldread_delete_reclaim_test.go` to a 64 KiB
framing allowance and could not update the E2E, which does not run on PRs. The
observed residue is 1024 (NFS) / 937 (SMB) bytes against a 16 MiB payload, so no
file data is retained. The assertion now matches the unit test's.

## Still red, tracked separately

With registration fixed, `TestNLMAxisInterop` gets past the rpcbind gate and
fails later, on the lock cases themselves. That is a distinct problem this PR
does not claim to fix, and the hardware runs show the test's expectations are
order-dependent — the same commit reports `BLOCKED` or `ACQUIRED` for the same
case depending on whether the NFSv4 client has taken any earlier lock. Filed
separately with the evidence.

### DIFF
--- a/pkg/adapter/nfs/adapter.go
+++ b/pkg/adapter/nfs/adapter.go
@@ -152,6 +152,15 @@ type NFSAdapter struct {
 	// Stop() may race Serve() (Stop is documented safe to call concurrently).
 	sysregActive atomic.Bool
 
+	// sysregReconciling guards the background registration reconcile so a burst
+	// of settings applies (one per accepted connection) collapses into a single
+	// in-flight transition instead of a goroutine per call.
+	sysregReconciling atomic.Bool
+
+	// sysregAddr overrides the system rpcbind dial address. Empty means the
+	// well-known one; a test sets it to keep the host's rpcbind untouched.
+	sysregAddr string
+
 	// udpConn is the UDP listener serving NLM/NSM/MOUNT when the UDP transport
 	// is enabled (adapters.nfs.udp.enabled). nil when UDP is disabled.
 	udpConn *net.UDPConn
--- a/pkg/adapter/nfs/auxservices.go
+++ b/pkg/adapter/nfs/auxservices.go
@@ -33,13 +33,8 @@ func (s *NFSAdapter) startEnabledAuxServices(ctx context.Context) {
 		logger.Info("Portmapper disabled by configuration")
 	}
 
-	// System rpcbind registration (port 111). Best-effort and time-bounded;
-	// sysregSidecar.Start never returns an error, so the branch is defensive.
-	if s.registerWithSystemEnabled() {
-		if err := s.sidecars.Start(sysregSidecar{s}); err != nil {
-			logger.Debug("System rpcbind registration sidecar failed to start", "error", err)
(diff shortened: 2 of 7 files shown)


marmos91:
Splitting out the second defect this uncovers: #1968. With registration fixed the harness reaches the lock cases and fails there, on a share created after the adapter started — verified on hardware, and independent of this PR's change (it reproduces at the same port with and without a restart).

marmos91:
Copilot review addressed in c3be091a:

- **goroutine per call during the transition window** — real. A `CompareAndSwap` guard now keeps a single transition in flight; the rest return immediately. A flip that arrives mid-transition is picked up by the next apply.
- **mutable package var for the rpcbind address** — real, and it turned out to matter for more than tests: the background transition was reading a config field `applyNFSSettings` writes, which `-race` caught. The address now lives on the adapter, and the desired state is read once on the caller's goroutine so the transition reads nothing shared.
- **`grep -q 100021` false positives** — anchored to the program-number column, for both the wait loop and the logging.
- **log message wording** — leaving as is. `Group.Reconcile` documents that it returns the *Start* error and swallows stop errors, so a non-nil error here is always a failed start.

Re-verified on a Linux VM after these changes: NLM still registers (`100021 1/3/4 tcp+udp 12049 nlockmgr`) and the remaining failure is #1968.

PR #1972, opened 2026-08-18
fix(metadata): close five MED write-path audit findings

marmos91:
Closes a batch of MED findings from the data-plane audit (`.planning/audits/2026-08-05-dataplane/report.md`). Each finding was re-verified against current `develop` before any change.

- **`createEntry` dropped the parent's `..` link-count bump on mkdir** (`pkg/metadata/file_create.go`) — FIXED. A failed `tx.GetLinkCount(parent)` was discarded with `if err == nil`, so the transaction still committed the new directory and its parent edge while the parent's nlink stayed one short — permanently, and silently. On postgres/sqlite the same read collapses errors into `(0, nil)`, so the parent was set to 1 instead. The read is now tx-critical and propagates, matching how `Move` and `RemoveDirectory` already treat it.

- **`Move` had no in-transaction recheck of the source/destination entries** (`pkg/metadata/file_modify.go`) — FIXED. Both names were resolved with `store.GetChild` *outside* the transaction; the transaction body re-read only the two directory inodes and then acted on the stale handles. No lock covers a file rename (`lockCreateName` is create-only, `lockParentLinks` only fires for cross-parent directory renames), so two concurrent renames onto the same destination both `SetChild` — last writer wins and the loser's inode is orphaned. Badger's SSI could not see it either, because the child key was never read through the transaction. Both edges are now re-resolved via `tx.GetChild` inside the transaction and the rename aborts with `ErrConflict` if either changed; that also enters the child keys in the transaction's read set so an optimistic backend detects the conflict.

- **`WaitForGraph` check-then-act pair** (`pkg/metadata/lock/deadlock.go`) — ALREADY-FIXED, no change. The separate `WouldCauseCycle` + `AddWaiter` pair no longer exists: `adb8bf02` (#1935) collapsed them into `TryAddWaiter`, which holds one write lock across the cycle check and the edge insert, and the only caller (`internal/adapter/smb/handlers/lock_async.go:55`) uses it. This is exactly the fix the report recommended.

- **Share-lifecycle writes bypassed the store's own SSI-conflict retry** (`pkg/metadata/store/badger/shares.go`) — FIXED. `CreateShare`, `UpdateShareOptions`, `DeleteShare` and `CreateRootDirectory` called `s.db.Update` directly, so a single `badger.ErrConflict` aborted the operation and escaped unmapped. `DeleteShare` is the worst case — `deleteShareFiles` reads every `f:` row into the transaction's read set, the highest conflict probability in the package, unretried. All four now go through the existing `s.updateWithConflictRetry`. Their closures assign (never accumulate into) the post-commit `freedBytes`/`quotaFreed`/`rootFile` outputs, so a retry cannot double-count. `deleteShareFiles`' doc comment, which asserted a retry that did not exist on this path, is corrected rather than deleted — it is now true.

- **Link-count-with-type-default lookup reinvented 6 times + a real manifest bug** (`pkg/metadata/store/badger/`) — FIXED. The `Get(keyLinkCount)` → nil/`ErrKeyNotFound` → default 2 (dir) / 1 block was open-coded at six sites while `fileLinkCountTxn` implemented exactly it; all six now call the helper (~100 lines deleted). The bug hiding in the duplication: the transaction-scoped `GetFileByPayloadID` legacy full-scan branch built its result inline and returned **without `loadManifest`**, so a row not present in the `pl:` index resolved inside a transaction came back with an empty `Blocks` slice — a silently truncated manifest for `runtime/shares/service.go`, `shares/coordinator.go` and `metadata/block_record_store.go`, all of which read `File.Blocks` off it. The store-level twin did call it. The scan branch now routes through `loadEnrichedFileByID`, the same path the index fast path uses.

## Tests

Each behavioural fix has a test that fails when the fix is reverted (verified):

- `TestCreateDirectory_AbortsOnParentLinkCountFailure`
- `TestMove_AbortsWhenSourceEntryChangedInTransaction`
- `TestMove_AbortsWhenDestinationAppearedInTransaction`
- `TestPayloadIndex_TxScanFallbackReturnsManifest` — fails pre-fix with `"[]" should have 4 item(s), but has 0`

The existing `faultyStore`/`faultyTx` harness in `tx_integrity_test.go` was extended with `GetLinkCount` and `GetChild` interception rather than adding a second harness, and its fixture now builds on `newTestFixture` instead of hand-rolling a second one (which also de-duplicated the pre-existing `TestMove_RollsBackOnPutFileFailure` setup).

The store-level `GetFileByPayloadID` scan fallback was routed through `loadEnrichedFileByID` as well, so both surfaces share one enrichment implementation — the drift between them is what produced the missing-manifest bug in the first place.

## Verification

`gofmt -l .` clean · `go build ./...` · `go test ./pkg/metadata/...` · `go test -race ./pkg/metadata/ ./pkg/metadata/store/badger/ ./pkg/metadata/lock/` · `golangci-lint run ./pkg/metadata/ ./pkg/metadata/store/badger/` — all green.

### DIFF
--- a/pkg/metadata/file_create.go
+++ b/pkg/metadata/file_create.go
@@ -477,11 +477,16 @@ func (s *Service) createEntry(
 		// mkdirs in one directory serialize on it — acceptable, mkdir is rare
 		// relative to file create, which is fully disjoint above.
 		if fileType == FileTypeDirectory {
+			// The read is tx-critical: a failed GetLinkCount must roll the
+			// whole create back rather than commit the new directory with the
+			// parent's ".." count left un-incremented, which permanently
+			// under-reports the parent's nlink.
 			parentLinkCount, err := tx.GetLinkCount(ctx.Context, parentHandle)
-			if err == nil {
-				if err := tx.SetLinkCount(ctx.Context, parentHandle, parentLinkCount+1); err != nil {
-					return err
-				}
+			if err != nil {
+				return err
+			}
+			if err := tx.SetLinkCount(ctx.Context, parentHandle, parentLinkCount+1); err != nil {
+				return err
 			}
 		}
 		return nil
--- a/pkg/metadata/file_modify.go
+++ b/pkg/metadata/file_modify.go
@@ -800,14 +800,13 @@ func (s *Service) Move(ctx *AuthContext, fromDir FileHandle, fromName string, to
 	now := time.Now()
 	txErr := withRelaxedTransaction(store, ctx.Context, func(tx Transaction) error {
 		// The GetChild lookups above ran outside this transaction and are
-		// advisory only: a concurrent rename or unlink can retarget or remove
-		// either name in the gap, and no lock covers a file rename. Re-resolve
-		// both namespace edges through the transaction and abort when they no
-		// longer match what was read, so two renames onto the same destination
-		// cannot both commit and orphan an inode. Reading the child keys
-		// through the transaction also enters them in its read set, which lets
-		// an optimistic backend detect the write-write conflict it otherwise
-		// cannot see.
+		// advisory only: no lock covers a file rename, so a concurrent rename
+		// or unlink can retarget either name in the gap. Re-resolve both
+		// namespace edges here and abort when they no longer match, so two
+		// renames onto the same destination cannot both commit and orphan an
+		// inode. Reading the child keys through the transaction also enters
+		// them in its read set, which is what lets an optimistic backend see
+		// the write-write conflict at all.
 		txSrcHandle, srcErr := tx.GetChild(ctx.Context, fromDir, fromName)
 		if srcErr != nil {
 			return srcErr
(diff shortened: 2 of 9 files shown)


marmos91:
On the `string(txSrcHandle) != string(srcHandle)` comparison — declining this one, the premise is wrong.

Go's compiler recognises `string(a) == string(b)` where both operands are byte slices and compares them in place without materialising either string. Measured:

```
BenchmarkStringCmp-10   395719687   3.005 ns/op   0 B/op   0 allocs/op
```

So there is nothing to save. On top of that, `file_modify.go` already uses this exact idiom at lines 614, 678, 777, 907 and 974 — switching only the two new comparisons to `bytes.Equal` would make the new code the odd one out in its own file for no measurable gain.

Taking the suppressed comment instead: the `faultyStore` doc comment does need updating now that the wrapper injects failures for `GetLinkCount` and `GetChild` as well as `PutFile`.

PR #1978, opened 2026-08-18
perf: sweep the LOW perf findings on the data-plane hot paths

marmos91:
Sweep of the 21 LOW `perf` findings from the data-plane audit. Each was verified against current `develop` first — these carried no independent verification, so several turned out not to be worth a diff. Bar applied: the path must be genuinely per-request / per-block / per-directory-entry, and the fix must be small and obviously behaviour-preserving.

## Changed

- `internal/adapter/nfs/v3/handlers/read.go:116` — **FIXED**. Handle hex was formatted on every READ/WRITE/COMMIT regardless of level; routed through a new `xdr.LazyHandle` slog.LogValuer so the string is built only when a record is emitted.
- `internal/adapter/nfs/v4/handlers/read_plus.go:216` — **FIXED**. Reply buffer now pre-sized to the exact wire size, matching `encodeRead4resok`.
- `internal/adapter/smb/handlers/write.go:161` — **FIXED**. Same treatment for the SMB READ/WRITE entry logs via a `lazyFileID` LogValuer.
- `pkg/block/compression/frame.go:45` — **FIXED**. The frame header is reserved in the buffer the codec streams into, so the framed body is no longer allocated and copied a second time. Measured on a 4 MiB semi-compressible block: 20.6 MB → 14.7 MB allocated per Put. The frame-vs-plaintext decision is numerically unchanged.
- `pkg/block/encryption/frame.go:107` — **FIXED**. Header is built first and `aead.Seal` appends the ciphertext in place, dropping the standalone ciphertext allocation plus its copy.
- `pkg/block/engine/readahead.go:48` — **FIXED**. `Load` before `LoadOrStore`, so a steady-state read no longer builds a throwaway `raState`.
- `pkg/block/local/memory/memory.go:114` — **FIXED**. Zero only the tail the copy does not cover.
- `pkg/block/remote/memory/store.go:69` — **FIXED**. Dropped the redundant second copy; `io.ReadAll` already returns a private buffer and every read path copies on the way out.
- `pkg/metadata/cookies.go:73` — **FIXED**. FNV-1a inlined as a plain loop, dropping all 3 allocations per directory entry. Pinned to `hash/fnv`'s output by a new test — cookies must stay stable across restarts.
- `pkg/metadata/store/sqlite/transaction.go:496` — **FIXED**. The three ad-hoc `Debug` logs on GetFile/PutFile/GetChild are gated on `logger.Enabled`, so the id formatting is skipped when Debug is off.

## Not changed

- `pkg/block/engine/warm.go:95` — **REFUTED**. The code already carries a comment explaining the probe is impossible: the journal is not hash-keyed. `WarmAll` is also an operator-triggered one-shot.
- `pkg/block/encryption/keyprovider/local.go:239` — **WONTFIX**. Real, but `cipher.NewGCM` is dwarfed by the per-chunk AEAD seal it sits next to, and a cached AEAD would outlive `Close()`'s key zeroization.
- `pkg/block/engine/compaction.go:237` — **WONTFIX**. Background GC pass, not a request path; a pre-built locator snapshot would also change the freshness the move decision relies on.
- `pkg/metadata/auth_permissions.go:489` — **WONTFIX**. The loop only evaluates the bits actually requested (normally one). Folding masks into one `EvaluateGranted` risks DENY / owner-implicit semantics drift for no measurable gain.
- `pkg/metadata/store/badger/durable_handles.go:324` — **WONTFIX**. SMB durable-handle reconnect and the expiry janitor, not a per-request path.
- `pkg/metadata/store/badger/transaction.go:438` — **WONTFIX**. A transient "manifest externalized" flag that goes stale means a manifest silently not written — data loss risk far outweighing one point-Get.
- `pkg/metadata/store/badger/transaction.go:795` — **WONTFIX**. Real N+1, but an attrs-optional variant means changing the store interface and the conformance suite across every backend.
- `pkg/metadata/store/memory/files.go:362`, `store.go:557`, `transaction.go:87` — **WONTFIX**. The memory metadata store clones the entire store on every mutation by design; one O(n) among many does not move a real workload.
- `pkg/controlplane/runtime/metrics.go:27` — **WONTFIX**. Per-scrape, N = number of shares; batching needs a new store API.

### DIFF
--- a/internal/adapter/nfs/v3/handlers/commit.go
+++ b/internal/adapter/nfs/v3/handlers/commit.go
@@ -90,15 +90,15 @@ func (h *Handler) Commit(
 	// Extract client IP for logging
 	clientIP := xdr.LazyClientIP(ctx.ClientAddr)
 
-	logger.DebugCtx(ctx.Context, "COMMIT", "handle", fmt.Sprintf("0x%x", req.Handle), "offset", req.Offset, "count", req.Count, "client", clientIP, "auth", ctx.AuthFlavor)
+	logger.DebugCtx(ctx.Context, "COMMIT", "handle", xdr.LazyHandle(req.Handle), "offset", req.Offset, "count", req.Count, "client", clientIP, "auth", ctx.AuthFlavor)
 
 	if ctx.isContextCancelled() {
-		logger.WarnCtx(ctx.Context, "COMMIT cancelled", "handle", fmt.Sprintf("0x%x", req.Handle), "offset", req.Offset, "count", req.Count, "client", clientIP, "error", ctx.Context.Err())
+		logger.WarnCtx(ctx.Context, "COMMIT cancelled", "handle", xdr.LazyHandle(req.Handle), "offset", req.Offset, "count", req.Count, "client", clientIP, "error", ctx.Context.Err())
 		return &CommitResponse{NFSResponseBase: NFSResponseBase{Status: types.NFS3ErrIO}}, nil
 	}
 
 	if err := validateCommitRequest(req); err != nil {
-		logger.WarnCtx(ctx.Context, "COMMIT validation failed", "handle", fmt.Sprintf("0x%x", req.Handle), "offset", req.Offset, "count", req.Count, "client", clientIP, "error", err)
+		logger.WarnCtx(ctx.Context, "COMMIT validation failed", "handle", xdr.LazyHandle(req.Handle), "offset", req.Offset, "count", req.Count, "client", clientIP, "error", err)
 		return &CommitResponse{NFSResponseBase: NFSResponseBase{Status: err.nfsStatus}}, nil
 	}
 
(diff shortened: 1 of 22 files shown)


marmos91:
Declining this one — the premise is measurably wrong, same as the `string(a) == string(b)` claim on #1972.

Go's compiler elides the conversion in `for _, b := range []byte(s)` when the result does not escape; it iterates the string's bytes in place. Measured, with an identical FNV-1a body either way:

```
BenchmarkRangeBytes-10   65962302   16.66 ns/op   0 B/op   0 allocs/op
BenchmarkIndex-10        73381030   17.40 ns/op   0 B/op   0 allocs/op
```

Zero allocations both ways, and the `range` form is marginally faster. The digests are identical (asserted in the same harness), which is unsurprising since both walk the same UTF-8 bytes.

The allocation-free goal of that change is already met — the three allocations it removed were the `hash/fnv` interface value and its internal state, not the string walk. Leaving the `range` form: it reads better and there is nothing to reclaim.

PR #1982, opened 2026-08-18
fix(metadata): copy every FileAttr field in CopyFileAttr

### LINKED ISSUE #1981: CopyFileAttr silently drops 10 of FileAttr's 25 fields and 5 of ACL's 6
`CopyFileAttr` (`pkg/metadata/auth_identity.go:344`) documents itself as "a deep copy of a FileAttr structure … to prevent external modification of internal state". It hand-enumerates the fields it copies, and the enumeration has fallen behind the struct.

`FileAttr` has 25 fields. `CopyFileAttr` copies 15. Silently dropped:

`EAs`, `IdempotencyToken`, `Blocks`, `BlocksDirty`, `BlocksDirtyOffsets`, `NewInode`, `ObjectID`, `DeletedAt`, `OriginalPath`, `DeletedBy`

It also builds its ACL copy as `&acl.ACL{ACEs: aces}` — one of `acl.ACL`'s six fields. Dropped: `Source`, `Protected`, `AutoInherited`, `SACL`, and **`NullDACL`**.

`NullDACL` is the dangerous one, because zeroing it inverts the meaning rather than degrading it. A null DACL means "no DACL present — everyone gets full access"; a zeroed `NullDACL` with an empty `ACEs` slice reads as an explicit empty DACL, which means deny-everyone. A copy that flips a share from open to closed (or the reverse, depending on which side consumes it) is not a partial copy, it is a wrong one.

**Not reachable today.** All 20 call sites feed WCC before/after attributes, which only serialize size/mtime/ctime. So this is latent, not live — hence LOW.

**But the drift mechanism is already proven.** `Blocks` and `IdempotencyToken` were added to `FileAttr` after this function was written, and the copy never followed. The function is exported, so the next caller that needs a genuinely complete copy gets a quietly lossy one with no compile error to warn them.

Fix: copy by value (`c := *attr`) so every current and future field rides along by default, then explicitly clone only the reference-typed members — the `ACEs` slice, the `EAs` map, the `Blocks` slice, `BlocksDirtyOffsets`, the `DeletedAt` pointer, and the `ACL` pointer itself (copying the pointed-to struct by value first, so its five other fields survive).

Found by the LOW `structure` triage of the 2026-08-05 data-plane audit, under a finding that only claimed the function was hand-written boilerplate.

marmos91:
Closes #1981.

`CopyFileAttr` documents itself as a deep copy but hand-enumerated its fields, and the enumeration had fallen ten behind the struct: `EAs`, `IdempotencyToken`, `Blocks`, `BlocksDirty`, `BlocksDirtyOffsets`, `NewInode`, `ObjectID`, `DeletedAt`, `OriginalPath`, `DeletedBy`. It also rebuilt the ACL as `&acl.ACL{ACEs: aces}`, dropping five of six fields.

`NullDACL` is the one that matters. Zeroing it does not degrade the copy, it inverts it — "no DACL present, everyone has full access" becomes an explicit empty DACL, which denies everyone.

Not reachable today: all 20 call sites feed WCC before/after attributes, which serialize only size/mtime/ctime. But the function is exported, and `Blocks` and `IdempotencyToken` were added to `FileAttr` after it was written and never followed — the drift is not hypothetical, it already happened twice.

**Change:** copy the struct by value, then explicitly clone the reference-typed members (the `ACEs` and `SACL` slices, the `EAs` map and its values, `Blocks`, `BlocksDirtyOffsets`, and the `DeletedAt` pointer), copying the pointed-to `acl.ACL` by value first so its other five fields survive. A field added to `FileAttr` from now on is carried over without touching this function.

**On the existing test.** `authentication_test.go` already had a subtest named "copies all fields" that passed throughout. It populated only the 15 fields the function copied, then asserted on those — tautological, and the reason the drift went unnoticed. Renamed to "copies the scalar fields" with a note pointing at the real one.

**New coverage** in `copy_file_attr_test.go`:
- `TestCopyFileAttr_CopiesEveryField` walks the struct reflectively instead of by hand, so any future field is covered the moment it is declared. Verified to fail against the old implementation (`field ACL was not carried over by CopyFileAttr`).
- `TestCopyFileAttr_IsDeep` mutates every reference-typed member of the copy and asserts the source is untouched, including the `NullDACL` flag.
- `TestCopyFileAttr_Nil` pins the nil passthrough.

Found by the LOW `structure` triage of the 2026-08-05 data-plane audit, under a finding that only claimed the function was hand-written boilerplate.

### DIFF
--- a/pkg/metadata/auth_identity.go
+++ b/pkg/metadata/auth_identity.go
@@ -5,8 +5,6 @@ import (
 	"net"
 	"regexp"
 	"slices"
-
-	"github.com/marmos91/dittofs/pkg/metadata/acl"
 )
 
 // AuthContext contains authentication information for access control checks.
@@ -346,34 +344,40 @@ func MatchesIPPattern(clientIP string, pattern string) bool {
 //
 // Useful when returning file attributes to callers to prevent
 // external modification of internal state.
+//
+// The struct is copied by value so a field added to FileAttr is carried over
+// without touching this function; only the reference-typed members below need
+// explicit clones. Enumerating fields by hand here silently drops whatever the
+// list has not kept up with.
 func CopyFileAttr(attr *FileAttr) *FileAttr {
 	if attr == nil {
 		return nil
 	}
 
-	// Deep copy ACL if present
-	var aclCopy *acl.ACL
+	c := *attr
+
 	if attr.ACL != nil {
-		aces := make([]acl.ACE, len(attr.ACL.ACEs))
-		copy(aces, attr.ACL.ACEs)
-		aclCopy = &acl.ACL{ACEs: aces}
+		aclCopy := *attr.ACL
+		aclCopy.ACEs = slices.Clone(attr.ACL.ACEs)
+		aclCopy.SACL = slices.Clone(attr.ACL.SACL)
+		c.ACL = &aclCopy
 	}
 
-	return &FileAttr{
-		Type:         attr.Type,
-		Mode:         attr.Mode,
-		UID:          attr.UID,
-		GID:          attr.GID,
-		Nlink:        attr.Nlink,
-		Size:         attr.Size,
-		Atime:        attr.Atime,
-		Mtime:        attr.Mtime,
-		Ctime:        attr.Ctime,
-		CreationTime: attr.CreationTime,
-		PayloadID:    attr.PayloadID,
-		LinkTarget:   attr.LinkTarget,
-		Rdev:         attr.Rdev,
-		Hidden:       attr.Hidden,
-		ACL:          aclCopy,
+	if attr.EAs != nil {
+		eas := make(map[string][]byte, len(attr.EAs))
+		for k, v := range attr.EAs {
+			eas[k] = slices.Clone(v)
+		}
+		c.EAs = eas
 	}
+
+	c.Blocks = slices.Clone(attr.Blocks)
+	c.BlocksDirtyOffsets = slices.Clone(attr.BlocksDirtyOffsets)
+
+	if attr.DeletedAt != nil {
+		deletedAt := *attr.DeletedAt
+		c.DeletedAt = &deletedAt
+	}
+
+	return &c
 }
--- a/pkg/metadata/authentication_test.go
+++ b/pkg/metadata/authentication_test.go
@@ -473,7 +473,10 @@ func TestCopyFileAttr(t *testing.T) {
 		assert.Nil(t, result)
 	})
 
-	t.Run("copies all fields", func(t *testing.T) {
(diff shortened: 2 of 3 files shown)


marmos91:
CI is green now (21 pass / 4 skipping). The two earlier reds were both pre-existing smbtorture flakes, not this diff:

- run 1 — `smb2.dirlease.unlink_same_initial_and_close`, `lease_break_info.count got 0x2` → the #1701 fingerprint (reopened, with new evidence: the surplus break is dispatched to the *first* directory handle immediately after its grant, with no `OnDirChange` line).
- run 2 — `smb2.rename.rename_dir_bench` + `smb2.rename.close-full-information`, both `NT_STATUS_SHARING_VIOLATION` → the #1658 fingerprint.
- run 3 — green.

Rotating test names across runs, the dirlease subtest passes locally in isolation against this build, and #1972/#1979 pass the same job on a comparable base. `CopyFileAttr`'s output is consumed only by NFS WCC fields and `PreWriteAttr`, never written back to a store and never aliasing its source, so it cannot reach SMB share-mode or lease dispatch.

## One thing worth changing before this lands

`pkg/metadata/io.go:178` calls `CopyFileAttr` on every `PrepareWrite`, i.e. once per WRITE request:

```go
preWriteAttr := CopyFileAttr(&file.FileAttr)
```

The old hand-written copy dropped `Blocks`. The new struct copy deep-clones it (`c.Blocks = slices.Clone(attr.Blocks)`), and `GetFile` on that path does load the manifest. `ChunkRef` is 48 bytes (32-byte `ContentHash` + `Offset` + `Size`), so a 10 GiB file at 1 MiB chunks now allocates ~480 KiB per write op — to build a snapshot whose only consumers read `Mode`, `Size`, `Mtime` and `Ctime` (`io.go:240`, `io.go:274`, `write.go:237`, `CaptureWccAttr`). The same clone now rides every WCC snapshot in `file_create.go`, `file_remove.go`, `directory.go` and `file_modify.go`.

The by-value copy is the right call — dropping fields silently is exactly the bug being fixed. It is the unconditional manifest clone on the hottest path that is new cost. Options, roughly in order of how much I like them:

1. Leave `CopyFileAttr` complete, and give the write path a cheaper snapshot (the WCC consumers want scalars only) rather than a full attr clone.
2. Share `Blocks` by reference in the copy — it is treated as immutable once assigned, and `BlocksDirty`/`BlocksDirtyOffsets` already gate mutation — with a comment saying why that one field is aliased.

Either way `TestCopyFileAttr_CopiesEveryField` should keep asserting the field is carried, so the completeness property this PR adds does not regress.

PR #1993, opened 2026-08-19
perf: close four round-trip MED findings from the data-plane audit

marmos91:
Closes four MED findings from the data-plane audit, each of the "this hot path does N round trips where it needs 1" shape. All four still held against current `develop`.

- **Per-chunk `DeleteSynced`+`MarkSynced` round-trips** (`pkg/metadata/block_record_store.go`) — `DefaultCommitBlock` issued two sequential store calls per chunk, so a block object packing hundreds of chunks cost 2N round trips inside one txn. `metadata.Transaction` gains `PutSyncedLocators`, which records the whole commit's markers at once: Postgres pipelines the upserts through a single `pgx.Batch`, SQLite drops to one upsert per chunk, and the embedded backends keep the existing delete-then-mark pair. Last-wins locator semantics are unchanged; the conformance suite's fault injection now fires on the batched call, and a new case pins the batched result — fresh chunk, overwritten standalone marker, and a hash repeated within one batch — against the per-chunk sequence.

- **Per-block sequential metadata round-trips on Truncate/PunchHole/Delete reap loops** (`pkg/block/engine/readwrite.go`) — each dropped chunk went through its own coordinator call, and at the store level each of those opened its own metadata transaction, so unlinking a 4 GiB file was ~32k sequential transactions. `MetadataCoordinator` and `metadata.Transaction` gain a batched reap; the SQL backends resolve the whole id set with one `UPDATE` plus one reap-at-zero `DELETE`. The engine dedupes offsets before handing them over (each offset names exactly one row). Delete's reap is now all-or-nothing rather than partially applied on failure — the remote sweep still runs unconditionally, and GC reclaims whatever a failed reap left behind. New conformance cases assert the batched result matches the per-id form and that a rollback restores every row.

- **`checkFilePermissionsFile` does an uncached `store.GetShareOptions` round trip on every permission check** (`pkg/metadata/auth_permissions.go`) — Badger already served this from a generation-guarded cache; SQLite and Postgres ran an uncached `SELECT` plus JSON decode per read/write/create/setattr for a record that changes only on an admin reconfigure. The cache moves to `pkg/metadata/store/internal/sharecache` and both SQL backends adopt it. Invalidation is the load-bearing half: every committed share-record write drops the cache — the store-level `UpdateShareOptions`, the four transactional writers (via a `sharesDirty` flag applied after commit, matching Badger's post-commit discipline), `Reset`, and `RestoreSnapshot`. A new `ShareOptionsOps` conformance group pins freshness for **every** backend: update, transactional update, delete, and rolled-back update.

- **GETXATTR scans the parent directory instead of resolving in O(1)** (`pkg/metadata/xattr.go`) — `findStreamChild` paged the entire parent through `ListChildren` on every GETXATTR. It now tries the indexed exact child name first and falls back to the scan only on a miss, which is what still resolves a stream whose name differs from the request in case. Conformance extended with that mixed-case case.

### Verification

`gofmt -l .` clean · `go build ./...` · `go vet ./...` · `go test ./pkg/metadata/... ./pkg/block/... ./pkg/controlplane/... ./internal/adapter/common/...` · `-race` on the touched packages · `golangci-lint run` → 0 issues.

The two new SQL statement shapes (the `pgx.Batch` upsert and the `= ANY($1)` reap) were exercised against a real PostgreSQL 16 through the `integration`-tagged conformance suite, not only against the embedded backends.


marmos91:
Superseded — this branch's content already merged as #1983.

The agent that opened this PR had stalled on its watchdog; I picked its worktree up, rebased it onto a develop that had moved four merges, ran the simplifier and reviewer passes, fixed a Copilot finding and merged it as #1983. This PR points at the pre-rebase state of the same branch, so its diff against current develop reads as -2118 lines — merging it would revert the rebase.

Verified on develop: `findStreamChild`, `PutSyncedLocators`, `DecrementRefCountAndReapMany` and `pkg/metadata/store/internal/sharecache` are all present, along with the "sync.Map, not lock-free" doc correction.

The agent's own notes are worth keeping, so recording them here: the reap loop was worse than the audit reported (the store-level single-id form opened its own transaction per row); engine `Delete`'s reap changed from partially-applied-on-error to all-or-nothing in one transaction, with the remote sweep still unconditional and GC reclaiming the remainder; and the share cache is process-local, so it extends the existing single-instance assumption from badger to postgres.

PR #2003, opened 2026-08-19
ci(smb): stop the Windows compat job failing on a precondition it measures as unmet

### LINKED ISSUE #1995: SMB Client Compatibility workflow is permanently red on develop: Windows job cannot pass by design
The `SMB Client Compatibility` workflow has failed on every develop run I can see (back to 2026-08-05 at least, most recently `d3fbb5e9`). The macOS and Linux jobs pass; the **Windows (net use)** job fails, and it cannot pass on a hosted runner as currently written.

Not caused by any recent merge — it predates today's audit-sweep PRs, and I checked before assuming.

**Two separate defects, both in the harness, neither in DittoFS.**

**1. The job asserts something the workflow already knows is impossible.**

`.github/workflows/smb-client-compat.yml:310` ("Free TCP 445 for DittoFS (best-effort)") carries an excellent write-up of the real constraint: the Windows redirector always connects to TCP 445, UNC syntax has no port field, and freeing 445 means unloading the kernel SMB listener (`srvnet.sys`, PID 4), which hosted runners cannot do without a reboot. The step is explicitly best-effort and logs its result:

```
TCP 445 freed for DittoFS: False (false => hosted-runner kernel listener; net use cannot reach DittoFS without a reboot)
```

The next step then runs `net use` anyway and hard-fails:

```
System error 64 has occurred.
net use failed with exit code 2
```

So the job fails on a precondition it just measured and reported as unmet. The `$freed` result should gate the assertions — skip them with a clear message when 445 is kernel-owned, and let the diagnostics still upload. The comment block already argues DittoFS's SMB path is correct here (macOS `mount_smbfs` and Samba `smbclient` 4.19 both authenticate against it, negotiating identically to the Windows redirector).

**2. The Cleanup step has a PowerShell bug and fails independently.**

`.github/workflows/smb-client-compat.yml:513`:

```powershell
$pid = $env:DFS_PID
if ($pid) { Stop-Process -Id $pid -Force -ErrorAction SilentlyContinue }
```

`$pid` is a PowerShell **automatic read-only variable** (the current process ID). Assigning to it throws:

```
Cannot overwrite variable PID because it is read-only or constant.
##[error]Process completed with exit code 1.
```

So cleanup never kills the `dfs` process, and the step fails even when everything before it succeeded. Rename to `$dfsPid`. This one is a two-line fix and worth doing regardless of (1).

**Why it matters beyond tidiness:** a workflow that is always red trains everyone to ignore it, which means a genuine SMB client regression on macOS or Linux — the two jobs that *do* work — would go unnoticed. Either make the Windows job skip honestly, or move it to a self-hosted/VM runner where 445 can actually be claimed.

marmos91:
Closes #1995.

`SMB Client Compatibility` has been red on develop since at least 2026-08-05. macOS and Linux pass; the Windows job cannot pass as written. Two independent defects, both in the harness — DittoFS itself is fine, and the workflow's own comments already argue that (macOS `mount_smbfs` and Samba `smbclient` 4.19 both authenticate against the same server, negotiating identically to the Windows redirector).

### 1. The job asserted a precondition it had already measured as unmet

The "Free TCP 445" step is explicitly best-effort. The Windows redirector always connects to TCP 445 and UNC syntax has no port field, so reaching DittoFS on 12445 means owning 445 — which means unloading the kernel SMB listener (`srvnet.sys`, PID 4). A hosted runner cannot do that without a reboot. The step says so:

```
TCP 445 freed for DittoFS: False (false => hosted-runner kernel listener; net use cannot reach DittoFS without a reboot)
```

The next step then ran `net use` regardless and threw:

```
System error 64 has occurred.
net use failed with exit code 2
```

Now the probe publishes `freed` as a step output and the assertions gate on it. When 445 is kernel-owned the job skips them and prints why; when a runner *can* free 445 the assertions run exactly as before. A new "Report skipped net use assertions" step makes the skip visible in the log rather than silent, and the step summary renders `:fast_forward: skipped` instead of five green ticks for tests that never ran — the old summary claimed five passes on a job that had just failed.

### 2. Cleanup assigned to `$pid`

```powershell
$pid = $env:DFS_PID
```

`$pid` is a PowerShell **read-only automatic variable** (the current process id). Assigning to it throws:

```
Cannot overwrite variable PID because it is read-only or constant.
```

So the step failed on its own, and the `dfs` process was never stopped. Renamed to `$dfsPid`.

### Notes

This does not paper over a product failure. It stops the job asserting something the environment makes impossible, so a genuine regression in the macOS or Linux jobs — which do exercise the server end to end — becomes visible again instead of drowning in a permanently red check.

If you'd rather actually exercise `net use`, that needs a self-hosted or VM runner where 445 can be claimed; worth a follow-up if Windows redirector coverage matters. I've left the pktmon capture and log-dump diagnostics running unconditionally so that path stays debuggable.

Verified the YAML parses; the PowerShell changes are inspection-only since there is no Windows runner locally — the CI run on this PR is the real check.

### DIFF
--- a/.github/workflows/smb-client-compat.yml
+++ b/.github/workflows/smb-client-compat.yml
@@ -308,6 +308,7 @@ jobs:
           go build -o dfsctl.exe cmd/dfsctl/main.go
 
       - name: Free TCP 445 for DittoFS (best-effort)
+        id: free445
         shell: pwsh
         run: |
           # #879 ROOT CAUSE: the Windows redirector (`net use \\host\share`)
@@ -342,6 +343,10 @@ jobs:
             Start-Sleep -Seconds 2
           }
           Write-Host "TCP 445 freed for DittoFS: $freed (false => hosted-runner kernel listener; net use cannot reach DittoFS without a reboot)"
+          # Publish the result so the net use assertions can skip honestly
+          # instead of failing on a precondition this step just measured as
+          # unmet. Without the gate the job can never pass on a hosted runner.
+          "freed=$($freed.ToString().ToLower())" >> $env:GITHUB_OUTPUT
           (Get-NetTCPConnection -LocalPort 445 -State Listen -ErrorAction SilentlyContinue) | Format-Table -AutoSize
           # sc.exe returns nonzero when a service is already stopped, which pwsh
           # would otherwise propagate as a step failure. This step is
@@ -396,6 +401,7 @@ jobs:
           (Get-NetTCPConnection -LocalPort 445 -State Listen -ErrorAction SilentlyContinue) | Format-Table -AutoSize
 
       - name: Test net use operations
+        if: steps.free445.outputs.freed == 'true'
         shell: pwsh
         run: |
           # Mount SMB share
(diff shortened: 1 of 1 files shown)


marmos91:
Duplicate of #2002, opened 23s apart from the same work on a second branch. #2002 is a superset: same $freed gate and the same $dfsPid cleanup fix, plus it also replaces the hardcoded all-green summary table, which would otherwise report passes for assertions that never ran. Continuing in #2002.

Issue #1967, opened 2026-08-18
Break up the god-functions and duplicated algorithms the data-plane audit surfaced


marmos91:
The 2026-08-05 data-plane audit confirmed a set of MED findings that are pure structural debt — no behavioural defect, no fix that fits in a batch PR. They are collected here so the audit can close without pretending they were addressed.

Seven sibling findings (pool-path vs transaction-path CRUD duplication across badger/postgres/sqlite, the three copy-pasted generation-guarded caches, and the ~180-line config-reconciliation body duplicated between postgres and sqlite) are **not** listed here — they are subsumed by #1828.

What remains:

**God-functions**
- `pkg/block/journal/recovery.go:60` — `recover()` is ~375 lines mixing 8+ unrelated concerns.
- `pkg/metadata/lock/manager.go:701` — `Manager` spans 8+ unrelated responsibilities.
- `pkg/metadata/lock/leases.go:259` — `requestLeaseImplWithMode` is a 440-line function mixing 8+ concerns.
- `pkg/metadata/lock/leases.go:902` — `acknowledgeLeaseBreakImpl` is ~200 lines mixing ack validation, tombstone/timeout classification, and progressive multi-stage dispatch.
- `pkg/controlplane/runtime/shares/service.go:471` — 49 methods, ~3300 LOC.

**Duplicated algorithms**
- `pkg/metadata/acl/evaluate.go:170` — `Evaluate()` and `EvaluateGranted()` duplicate the entire DACL-walk.
- `pkg/metadata/auth_permissions.go:441` and `:1159` — `acl.EvaluateContext` construction is built three times (one inline plus two near-identical builders), each self-described as mirroring the others.

Full evidence for each is in `.planning/audits/2026-08-05-dataplane/report.md` under the matching `### [MED]` heading; every one carries a CONFIRMED verification paragraph.

The lock-package entries are the ones with real risk attached — the duplication there is in concurrency-sensitive code where the two copies can drift silently.

marmos91:
Adding one more, deferred here after verification rather than by triage:

- `pkg/metadata/lock/manager.go:702` — a single share-wide `lm.mu` serializes every lock-manager operation across synchronous per-op store I/O.

[...]

PR #2015, opened 2026-08-19
refactor(journal): decompose recover() into named phases

marmos91:
One of the god-function findings from the 2026-08-05 data-plane audit.
Pure structural refactor: no behaviour change.

## What was wrong

`Store.recover()` was a ~350-line single body threading a dozen mutable locals
(`actives`, `sealedByShard`, `indexByShard`, `emptyPool`, `orphans`, `tombstones`,
`truncations`, `unsynced`, `missingIdx`, `opened`, `maxSegID`, `maxVersion`) across eight
unrelated concerns:

1. segment directory scan and header classification (active / sealed / empty / orphan),
2. torn-tail truncate + fsync,
3. `.idx` sidecar rebuild,
4. record replay into the interval index, including tombstone/truncate marker collection,
5. cold-log replay,
6. cold-log compaction,
7. byte-counter reconciliation (`unsynced`, per-segment `deadBytes`),
8. active-segment assignment and the orphan sweep.

This is the replay-correctness path — the one that decides whether a range comes back as
data or as a hole reading zeros — and none of those phases could be exercised on its own.

## What changed

The working set moves into a `recoveryState` struct, and each phase becomes a method:

| method | phase |
| --- | --- |
| `scanSegments` / `loadSegment` | 1-3 |
| `replayRecords` | 4 |
| `applyColdLog` | 5 |
| `compactColdLog` | 6 |
| `reconcileByteCounters` | 7 |
| `assignActiveSegments` | 8 |

`recover()` is now a 55-line orchestrator calling them in exactly the order the inline body
ran. `recoveryState.opened` keeps the existing all-or-nothing fd cleanup: the same
`ok`-guarded defer closes every segment opened by any phase if a later phase fails.

Mechanical notes:

- the per-segment `continue`s in the scan loop became `return nil` from `loadSegment`,
- `unsynced` is now `reconcileByteCounters`'s return value instead of a shared local,
- `liveColdEntries` was already evaluated unconditionally (it was the `if` statement's init
  clause), so hoisting it into `compactColdLog` changes nothing.

No logic, ordering, or error path was altered.

## Evidence

```
go build ./... && go vet ./pkg/block/journal/          # clean
go test -race ./pkg/block/journal/                     # ok (84s)
```

The journal package has the crash/replay suites; they are the check that matters here and
they pass unchanged under `-race`.

Refs #1967

### DIFF
--- a/pkg/block/journal/recovery.go
+++ b/pkg/block/journal/recovery.go
@@ -53,337 +53,47 @@ func scanSegmentIDs(dir string) ([]uint64, error) {
 // its torn tail truncated. All valid records feed a fresh interval index — order
 // does not matter because insert resolves overlaps by Version — and the global
 // LSN resumes at max(observed Version)+1.
+//
+// The phases run in the order below; each is a method on recoveryState, which
+// carries the tables being rebuilt between them.
 func (s *Store) recover() error {
-	segIDs, err := scanSegmentIDs(s.dir)
-	if err != nil {
-		return err
-	}
-	// Deterministic, ascending replay. Version still decides newest-wins, but a
-	// stable order keeps recovery reproducible and eases debugging.
-	sort.Slice(segIDs, func(i, j int) bool { return segIDs[i] < segIDs[j] })
+	r := newRecoveryState(s)
 
-	n := s.cfg.ShardCount
-	actives := make([]*segmentMeta, n) // data-bearing unsealed segment per shard
-	sealedByShard := make([]map[uint64]*segmentMeta, n)
-	indexByShard := make([]map[FileID]*fileIndex, n)
-	for i := 0; i < n; i++ {
-		sealedByShard[i] = make(map[uint64]*segmentMeta)
-		indexByShard[i] = make(map[FileID]*fileIndex)
-	}
-
-	var (
-		emptyPool   []*segmentMeta // empty unsealed segments, reusable as any shard's active
-		orphans     []uint64       // unattachable segment ids, candidates for the age-gated sweep
-		maxSegID    uint64
-		maxVersion  uint64
-		unsynced    int64
-		missingIdx  int
-		opened      []*segmentMeta           // every fd we opened, closed on error
-		tombstones  = map[FileID]uint64{}    // deleted file -> highest tombstone version
-		truncations = map[FileID]truncMark{} // truncated file -> highest truncate marker
-		ok          bool
-	)
+	ok := false
 	defer func() {
 		if !ok {
-			for _, m := range opened {
-				_ = m.close()
-			}
+			r.closeOpened()
 		}
 	}()
 
-	for _, id := range segIDs {
-		if id > maxSegID {
-			maxSegID = id
-		}
-		path := s.segPath(id)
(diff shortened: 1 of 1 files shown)


marmos91:
CI note: `smbtorture / badger-fs` went red with two new failures, `smb2.rename.rename_dir_bench` and `smb2.rename.close-full-information`, and seven suites (maxfid, aio_delay, compound_find, lease, …) were cut short by their per-suite timeout on the same run.

This diff touches exactly one file — `pkg/block/journal/recovery.go`, the crash-recovery replay path — and nothing in it is reachable from an SMB rename or from any of the timed-out suites. Seven suites timing out in one run is the signature of a loaded runner rather than a product change; six other branches were building concurrently. The rename pair is the known base flake (#1652 / #1658), and rotating test names across runs has previously been established as base flake rather than PR-under-test.

Not retried, per the fix-flakes-never-retry rule — reporting rather than papering over. Worth re-running once the runner queue drains before treating it as signal.

marmos91:
## Reproduced locally on badger-fs — the suite is green on this branch

Ran the full smbtorture suite against the exact config CI used, on this branch:

```
cd test/smb-conformance/smbtorture && ./run.sh --profile badger-fs
```

```
Total 575 | Passed 497 | Failed 38 | Known 38 | New Failures 0
Skipped 40 | Inconclusive 7 | Suites cut short 7
RESULT: All failures are known. CI green.
```

Both tests CI graded as new failures pass here. A targeted rerun of just that suite is 13/13:

```
./run.sh --profile badger-fs --filter smb2.rename
Total 13 | Passed 13 | Failed 0 | New Failures 0
```

### The timeout signature is baseline, not a symptom

The CI run's "7 suites cut short by timeout" and "7 tests with no verdict" looked like the server hanging under the badger-fs journal path. It is not: the **green local run produces the same 7 cut-short suites and the same 7 inconclusive tests** (maxfid, aio_delay, compound_find, lease, multichannel, notify, oplock). That is how this profile behaves on a passing run, so it carries no signal about this diff. The only real delta between the CI run and a green run is the two `smb2.rename` tests — the known rename flake neighbourhood (#1652 / #1658).

### Why recover() cannot affect steady-state SMB latency

`Store.recover()` has exactly one caller, `journal.Open` (store.go:278), and it is gated on `len(ids) > 0`. It runs once, at journal open, before any client I/O. It is not reachable from an SMB rename, from maxfid's handle churn, or from any per-operation path — so "recovery got slower or holds a lock longer" cannot produce these symptoms even if it were true.

It is also not true. Beyond the green suite:

- Every code line is preserved. Comment-stripped, whitespace-stripped multiset diff of `recovery.go` between develop and this branch yields **only** `x` -> `r.x` field plumbing, the new method signatures, and the `continue` -> `return nil` conversions from hoisting the per-segment body into `loadSegment`. No conditional, no ordering, no lock scope changed — there is no lock in `recover()` to change; it runs before `s.shards` is published.
- `go test -race ./pkg/block/journal/` passes (84s), which is where the replay/crash coverage lives.

Not adding anything to KNOWN_FAILURES.md, and not asking for a blind rerun — this is a reproduction that came back green, which is the opposite of papering over it. Re-running the job when the runner queue drains should confirm.

PR #2027, opened 2026-08-19
fix(nfs4): release open state when a v4.1 client's lease expires

chessman:
# Description

## Type of change

## Changes Made

Reap open owners, open states and lock states when the lease expires in NFS v4.1, same as in v4.0.

## Testing

## Testing Instructions

Purely theoretical, the following steps should reproduce the issue:
1. One client locks a file using v4.1.
2. Cut off the connection.
3. Try to lock the file with another client from another machine after the lease has expired (in reality, a bit longer).

## Checklist

### DIFF
--- a/internal/adapter/nfs/v4/state/manager.go
+++ b/internal/adapter/nfs/v4/state/manager.go
@@ -725,6 +725,43 @@ func (sm *StateManager) RemoveClient(clientID uint64) {
 		"client_id_str", record.ClientIDString)
 }
 
+// removeClientOpenStateLocked drops every open-owner belonging to clientID
+// together with its open states, its lock states, and the locks those hold in
+// the unified lock manager.
+//
+// It scans sm.openOwners by ClientID rather than walking a per-client owner
+// list: V41ClientRecord carries no such list, and the v4.0 ClientRecord one
+// goes stale because freeStateidLocked removes owners from sm.openOwners
+// without removing them there.
+//
+// Caller must hold sm.mu.
+func (sm *StateManager) removeClientOpenStateLocked(clientID uint64) {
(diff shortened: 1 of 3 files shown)


marmos91:
@copilot review

marmos91:
Nice catch, and thanks for the test. I verified this: disabling the new call in `purgeV41Client` makes your test fail on all three assertions, and the full `internal/adapter/nfs/v4/...` tree is green under `-race`. The share reservation assertion is the good one, since it proves the symptom instead of just map hygiene. I also checked that v4.0 and v4.1 client IDs come from the same counter and that owners are only dropped from `sm.openOwners` once their open states are empty, so the scan is neither ambiguous nor lossy.

Two things:

- **Typo in the docstring.** `freeStateidLocked` should be `freeOpenStateidLocked` (`stateid.go:496`). The reasoning is correct, that really is the site which leaves `ClientRecord.OpenOwners` behind while `CloseFile` cleans both, so it is worth having the right name on it.

- **Test misses the lock state half.** The title mentions lock states but the test only asserts `openOwners` and `openStateByOther`. Could you take a byte range lock before the lease lapses and assert `lockStateByOther`, `lockOwners` and `ListUnifiedLocks` all drain? That is the part that unblocks the second client in your scenario.

- One note, no change needed: the v4.0 path used to walk the per record list and now scans all of `sm.openOwners` and `closedOwnerByOther`. Right call, since the per record list is exactly what goes stale, and invisible at our sizes. Just worth remembering if a reap across many clients ever shows up in a profile.

chessman:
@marmos91 Thanks, I've fixed the comments.

marmos91 (approved):
LGTM

Issue #1715, opened 2026-07-15
Metadata store cleanup — simplify, de-duplicate, cut hot-path write cost (3 backends)


marmos91:
Consolidated tracker from three read-only audits of the metadata store. **Goal: make it simpler, cleaner, and easier for a human to maintain** — fewer types, fewer tables, one source of truth. Lower hot-path write cost (and #1687 relief) is the co-benefit.

**Findings docs** (`.planning/perf/`): `2026-07-15-metadata-schema-inventory.md` (badger), `-metadata-entity-model.md` (Go structs), `-metadata-sql-schema.md` (postgres+sqlite).
**Rendered:** [badger schema](https://claude.ai/code/artifact/e7501e85-e19b-4c3a-a861-e9ae593e8931) · [entity model](https://claude.ai/code/artifact/5f27af76-4358-480f-b1ea-9e77c2c2e0ff) · [SQL](https://claude.ai/code/artifact/60731911-afcd-423e-85c9-53ee612e8120) · [cross-findings](https://claude.ai/code/artifact/ffe1245a-a955-430e-af47-a11284fc895b)

## Root cause the three audits converge on
Every backend **rewrites the whole block manifest on an attr-only change** — badger re-serializes `Blocks[]` in the `f:` blob; SQL runs `DELETE + INSERT×M` on `file_block_refs`. A `chmod` rewrites the file's entire chunk list. Shared cause: **`PutFile(File)` conflates attribute mutation with manifest mutation.** Fixing that seam is the biggest win and the cleanest simplification.

## Wave 1 — independent, ship now (no journal dependency)
- [x] **Separate attr-write from manifest-write** — badger: split `Blocks[]` → `fb:<uuid>` written only at sync finalization. SQL: gate `putFileChunkRefs` on blocks-dirty (keep it when truncate pruned). *Biggest write-amp + complexity cut; universal.* (pg PutFile:476 / sqlite:407, file_block_refs.go)
- [x] **Drop dead/redundant SQL indexes** — `idx_inodes_updated_at` (both) + pg `update_inodes_updated_at` trigger, `idx_inodes_has_acl` (pg), `idx_parent_child_map_parent` (both) + `_parent_name` (pg), `idx_file_block_refs_file_id` (sqlite). 1–2 fewer btrees per namespace write, zero read regression. One migration per backend.
[...]

PR #2058, opened 2026-08-21
refactor(metadata): split PutFile into UpdateAttrs and SetManifest

marmos91:
## What

Splits the single `Files.PutFile(ctx, file)` method into two:

```go
UpdateAttrs(ctx, file)   // attrs only; stored manifest untouched
SetManifest(ctx, file)   // attrs + replace the stored block manifest from file.Blocks
```

and deletes the transient `File.BlocksDirty` flag that used to carry that choice
as data. `File.BlocksDirtyOffsets` is renamed `File.ManifestDirtyOffsets`, since
the field it qualified no longer exists; its semantics are unchanged.

This is the interface half of #1715 Wave 3. The *data* half already shipped as
the `BlocksDirty` gate in #1820/#1841 — this only moves the signal from a struct
field to the method the caller picks.

## Why now

#1828 stage 1 of 7. The later `store/sql` merge (stages 2-3) collapses the sqlite
and postgres query bodies into one copy. Doing the reshape first means that merge
inherits an already-clean shape instead of porting the `BlocksDirty` convention
into the new package and reshaping it afterwards — a smaller total diff.

Independently, a flag on a struct is the wrong place for "which write is this".
It is set at one call site and read at another, four packages away, and nothing
in the type system stops a caller from forgetting it — the exact failure mode
that produces a silently-stale `file_block_refs` table. The method name cannot be
forgotten.

## Behaviour

No change, by construction. Each backend keeps one body and gains two entry
points:

```go
func (tx *sqliteTransaction) UpdateAttrs(ctx, file) error { return tx.putFile(ctx, file, false) }
func (tx *sqliteTransaction) SetManifest(ctx, file) error { return tx.putFile(ctx, file, true) }
```

and the gate that read `file.BlocksDirty` now reads the `writeManifest`
parameter. Same for postgres and badger. `memory` keeps the block list inline in
the stored `FileAttr`, so its `SetManifest` is `UpdateAttrs` — it already ignored
`BlocksDirty`.

Badger's extra fallback is preserved verbatim: an attr-only write on a file that
carries blocks but has no `fm:<uuid>` key yet still writes the manifest, which is
what migrates a legacy `f:` blob off its embedded manifest.

## Call sites converted to `SetManifest`

Every site that previously set `BlocksDirty = true`, and only those:

- `metadata/block_record_store.go` `putProjection` — carve/rollup commit, reap+re-carve, `ReprojectBlocks`
- `metadata/file_modify.go` — size-down truncate that prunes the block list
- `metadata/sparse.go` — punch-hole
- `adapter/common/clone.go`, `adapter/common/copy_payload.go` — wholesale manifest replacement on the destination
- `controlplane/runtime/shares/coordinator.go` — post-quiesce ObjectID write

In `file_modify.go` the flag was set inside the truncate branch and read at an
unconditional commit further down, so it becomes a `blocksPruned` local that
selects the method at the commit. Everything else was adjacent enough to convert
in place.

## Verification

- `go build ./...`, `go vet ./...` clean
- `go test ./pkg/metadata/... ./pkg/block/engine/... ./pkg/controlplane/... ./internal/adapter/common/...` — 26/26 packages pass
- `pkg/metadata/storetest/` conformance passes for all four backends; the manifest-gate
  assertions there (`truncate_block_refs.go`, `blockref_roundtrip.go`,
  `objectid_*.go`, `snapshot_conformance.go`) now exercise the two entry points
  rather than the flag, so the write-amplification guarantee is still under test.
- `manifest_projection_test.go`'s fake transaction asserts the projection path
  goes through `SetManifest`, replacing the old `BlocksDirty` assertion.

No on-disk format change: `BlocksDirty` was `json:"-"` and the SQL backends write
explicit column lists, so it never reached a row.

Refs #1828

### DIFF
--- a/docs/internals/architecture.md
+++ b/docs/internals/architecture.md
@@ -1371,7 +1371,7 @@ Production call chain (per-write, on quiesce):
             └─ persistFileChunksAfterFlush
                 └─ ComputeObjectID(blocks)
                 └─ coordinator.PersistFileChunks(blocks, objectID)
-                    └─ runtime coordinator: WithTransaction(GetFileByPayloadID + PutFile)
+                    └─ runtime coordinator: WithTransaction(GetFileByPayloadID + SetManifest)
                         // FileAttr.Blocks AND FileAttr.ObjectID
                         // written in one metadata txn
 ```
--- a/internal/adapter/common/clone.go
+++ b/internal/adapter/common/clone.go
@@ -25,7 +25,7 @@ import (
 // Everything is atomic in one metadata transaction:
 //   - engine.CopyPayload's per-hash IncrementRefCount UPDATEs are bound to the
 //     txn (via metadata.WithTx) so they commit/roll back together with the
-//     destination PutFile. On any error nothing is committed — no partial
+//     destination UpdateAttrs. On any error nothing is committed — no partial
 //     dstFileAttr, no leaked RefCount bumps.
 //   - cache.InvalidateFile (if cache != nil) runs POST-txn, after the commit.
 //
@@ -74,7 +74,7 @@ func CloneWholeFile(
 	err := metadataStore.WithTransaction(ctx, func(tx metadata.Transaction) error {
 		// Bind the active txn into the context so the per-share coordinator's
 		// RefCount UPDATEs (driven by engine.CopyPayload) join the same txn as
-		// the destination PutFile and commit/roll back together.
+		// the destination UpdateAttrs and commit/roll back together.
 		txCtx := metadata.WithTx(ctx, tx)
 
 		// Re-read the source INSIDE the txn, AFTER the drain, so the copy uses the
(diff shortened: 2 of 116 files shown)


marmos91:
All seven Copilot findings addressed in 3c2b2d1d.

Six were comments the blanket `PutFile` → `UpdateAttrs` rename walked over where the text describes *manifest* persistence — `blocksink.go`, `coordinator.go` (engine), `readwrite.go`, `filechunk.go` (x2), `shares/coordinator.go` (x2). Those now say `SetManifest`, which is what those paths actually call. Same for three more the review did not flag: `drain_persist_test.go:22`, `blocksink_ssi_test.go:62`, `api_blockref_test.go:384`.

The one real code finding was `drain_persist_test.go:66`: `testCoordinator.PersistFileChunks` is a deliberate mirror of the production `metadataCoordinator.PersistFileChunks`, and production now writes through `SetManifest`. The fake had drifted before this PR too — production set `BlocksDirty = true` and the fake never did — so it could pass while a SQL backend left `file_block_refs` stale. It now calls `SetManifest`. `go test ./pkg/block/engine/` passes.

The `store.go` doc was also wrong as written: badger keeps its manifest in a separate `fm:<uuid>` key, so it is not "inline in the file record". Only `memory` is. Reworded to say the manifest-rewriting method is the only one that touches a separately-stored manifest — SQL's `file_block_refs` and badger's `fm:` alike — and that memory alone treats it as `UpdateAttrs`.

Issue #2070, opened 2026-08-24
Carve drain convergence: land #1872 in stages, serial-runs first


marmos91:
Tracking issue for the staged work that closes #1872. Splits it so the risky half (the flip chain) comes last and may turn out not to be needed.

Supersedes the approach in #1875 (draft, closed): that PR fixed both defects in one diff by hoisting the block builder *and* the dispatcher to file scope, which forces `flipUpTo`'s ordered chain to span runs. It regressed on hardware — blocks committed continuously while `UnsyncedBytes` stayed flat at 644 MiB, i.e. records were never flipped synced and the carve re-carved the same ranges forever. Root cause was never found. The obvious watermark-lag theory is wrong: `fileOff` is correctly reset per run.

The two defects behind #1872 are separable, which #1875 assumed they were not:

- **serial**: `carveRun` ends with `disp.wait()`, so runs upload one after another (~21 serial PUTs/s measured)
- **small objects**: a block is cut at every run boundary, so a 4 KiB run commits its own remote object

Only the second one touches the flip path. Fixing the first alone makes the drain *converge*, which is the actual user-visible bug — a randomly-written share's drain never finishes, so `dfsctl system drain-uploads` and eviction both fail.

## There is no reproduction of the stall

The unit level does not reproduce it. `carve_scatter_test.go` on the closed branch writes 200 non-contiguous 4 KiB runs, carves, and asserts `post-carve unsynced want 0` — it **passes on the branch that stalls on hardware** (`ok github.com/marmos91/dittofs/pkg/block/journal 0.905s`).

So the gap between the unit case and the rig case is where the bug lives:

| | unit test | rig |
| --- | --- | --- |
| files / shards | one / one | many |
| record shape | one interval per record | records carrying several fragments |
| prior state | fresh store | interleaved synced + cold intervals |
| row straddling | none | `extendRunToRowEnd` in play |
| dedup oracle | empty fake | real manifest lookups |
| invocation | one `Carve(Force:true)` | `Force:true` in a drain retry loop |
| scale | 800 KiB | 644 MiB |

Building a reproduction is step 3's entry price, not a warm-up for it. Steps 1 and 2 are unaffected — that is the argument for doing them first.

## Steps

[...]

PR #2071, opened 2026-08-24
fix(engine): hold the per-file commit lock while reaping superseded manifest rows

marmos91:
Part of #2070 (which stages the fix for #1872). Self-contained — it does not depend on the carve-concurrency work and can land on its own.

## The defect

`CommitBlock` takes a striped per-file mutex (`carveCommitLocks.forKey`) before writing manifest rows, because that write ends in `ProjectManifestToBlocks` — a read-modify-write of the file's `File.Blocks` row — and badger's SSI aborts concurrent writers to the same row. That lock is what keeps a file's concurrent carve commits off each other; without it the retry budget drains and the conflict surfaces as EDEADLK.

Both `ReapSupersededManifest` implementations performed the same `File.Blocks` write **without** taking that lock.

It was unreachable while carve runs were serial: a run reaps only after its own `disp.wait()` has drained every commit for that run, and no sibling run has started. Making runs concurrent (the next change in #2070) exposes it two ways — reap racing a sibling reap, and reap racing a sibling run's still-in-flight `CommitBlock`.

## Measured

The new test drives 8 goroutines at one file, alternating `CommitBlock` and `ReapSupersededManifest`. Every reap genuinely re-projects the row: `metadata.ReapSupersededManifest` calls `ProjectManifestToBlocks` on any non-empty span whether or not a row matched.

```
before the fix:  TransactionConflictsForTest() = 9
after the fix:   TransactionConflictsForTest() = 0
```

The failure is on the conflict counter rather than an error return — badger's retry loop absorbs the conflict, which is exactly why this would have degraded quietly under the workload the concurrency change targets.

## The change

Both sinks call one `reapSupersededManifest` helper that takes the stripe and holds it across the whole transaction, mirroring `commitManifestRows`. Badger's retry loop runs inside that single `WithTransaction` call, so every attempt is covered. `localBlockSink`'s nil-committer no-op is checked before the lock; `forKey`'s nil-receiver path keeps fixtures that wire no stripes working.

Not touched: `ManifestRowEndAfter`, the run-extension straddle lookup, is a read outside the lock. It is not a conflict source, but concurrent runs will each call it — tracked on #2070 rather than widened into this PR.

## Verified

`go test -race -count=1 ./pkg/block/engine/` green, including the pre-existing `TestLocalBlockSink_ConcurrentSameFileCommit_NoSSIConflict`. `go build ./...` and `go vet` clean.

### DIFF
--- a/pkg/block/engine/blocksink.go
+++ b/pkg/block/engine/blocksink.go
@@ -155,6 +155,10 @@ func (s localBlockSink) ReapSupersededManifest(ctx context.Context, id journal.F
 	if s.committer == nil {
 		return nil
 	}
+	if mu := s.commitLocks.forKey(string(id)); mu != nil {
+		mu.Lock()
+		defer mu.Unlock()
+	}
 	return s.committer.WithTransaction(ctx, func(tx metadata.Transaction) error {
 		return metadata.ReapSupersededManifest(ctx, tx, string(id), runStart, runEnd, newOffsets)
 	})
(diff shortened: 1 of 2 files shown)


marmos91:
@copilot review

Issue #1956, opened 2026-08-18
PunchHole: cold read of a punched range does not read back as zeros


marmos91:
## Summary

`PunchHole`'s contract is that `[offset, offset+length)` reads back as zeros (NFSv4.2 DEALLOCATE, RFC 7862). After a punch, a **cold** read of the punched range can return the pre-punch bytes instead.

Reproduced on unmodified `develop` (fcf560ac + #1942), memory metadata store + memory remote.

## Repro

A randomized mutation soak: seed a 6 MiB file, then repeatedly apply one of write / truncate / punch-hole, carve after each (`Flush` + `DrainRollups` + `DrainAllUploads`), `DrainLocalSynced` to force the file cold, then read the whole file back and compare against a model buffer. Call pattern matches production — `nil` currentBlocks to `WriteAt`, the live manifest refs to `Truncate`/`PunchHole`.

5 of 24 seeds fail, always immediately after a punch, always at a byte **inside** the punched range:

```
seed-16 iter-4  punch(off=1847777,len=7524221): cold read mismatch at byte 6134749 (size 12190873)
seed-23 iter-40 punch(off=3422258,len=3010972): cold read mismatch at byte 3919182 (size 12582912)
seed-21 iter-18 punch(off=2966012,len=7112756): cold read mismatch at byte 9860476 (size 12582912)
```

The model expects zeros there; the read returns non-zero, so pre-punch bytes survive into a cold read.

## Not the covering-row lookup

The same failures occur with the `resolveCovering` fallback returning the first covering row (develop), and with it returning the greatest-start row (#1953). So this is independent of which row the lookup picks — the punched bytes are still reachable through whichever row wins.

Likely direction: `PunchHole` zeros the range through the local append log, but a cold read hydrates the covering chunk from the **remote**, and the remote object still holds the pre-punch content. The zero-overwrite is durable locally but the manifest row for that range still addresses the old remote chunk, so once the local bytes are evicted the hydrate resurrects them.

## Severity

Silent wrong data, no error — same family as #1879, #1888, #1894, #1909, #1911. Requires a remote-backed share plus eviction (or a restart) between the punch and the read.

## Repro harness

The soak used is attached below; it is not in the tree. Drop it in `pkg/block/engine/` as `manifest_soak_test.go` and run `go test ./pkg/block/engine/ -run TestManifestSoak`.

<details><summary>manifest_soak_test.go</summary>

```go
package engine_test

import (
	"bytes"
	"context"
[...]

PR #2072, opened 2026-08-24
perf(journal): carve a file's dirty runs concurrently

marmos91:
Carves a file's dirty runs concurrently instead of one at a time. Part of #2070.

## The defect

`carveFile` split a file's live dirty intervals into *runs* — maximal contiguous spans, since a hole resets FastCDC chunking — then carved each run serially, waiting on its own dispatcher before starting the next.

On a scattered file that means each run holds a single block, so the per-run dispatcher has nothing to overlap and effective concurrency pins at **1**. `CarveUploadConcurrency` buys nothing at all, and a drain converges far too slowly.

Measured on a harness of 200 scattered 4 KiB runs (concurrency 8, 2 ms stand-in commit latency):

| | max concurrent commits | wall clock |
| --- | --- | --- |
| before | **1** | 644 / 534 / 559 ms |
| after | **8** | 70 / 91 / 80 ms |

~7x, and the `1` is the defect stated directly.

## What changed

Five commits, smallest first:

1. **`CarveResult` per run** — it was a shared `*CarveResult`, a genuine data race the moment runs overlap.
2. **One semaphore per file, not per run** — a single `sem` bounds in-flight blocks across all of a file's runs, so peak arena memory stays `cap(sem) x (CarveBlockSize + one overhang chunk)` however many runs the file has.
3. **Runs carve concurrently** under an errgroup, with `splitRuns` extracted.
4. **A run's row extension is bounded by the next run's start.** See below — this is the load-bearing correctness commit.
5. **The regression test fails loudly when its ordering gate is skipped.**

`flipUpTo` is untouched in every round. It holds `sh.mu` across its whole body, and that atomicity is the entire safety argument for letting runs flip concurrently.

## The correctness problem concurrency exposed

`extendRunToRowEnd` may extend a run forward to a manifest row boundary. It decides what to absorb using `warmAt`/`warmTail`, which test `iv.synced`.

But `flipUpTo` *sets* `iv.synced` in memory as each run's watermark advances. So once runs overlap, a sibling run that has already flipped its head intervals is indistinguishable from pre-existing warm data, and a run can extend forward over a sibling's territory.

That is corruption, not wasted work. `ReapSupersededManifest` deletes every manifest row whose start falls in `[runStart, runEnd)` and is not in that run's `newOffsets` — and deletes it *whole*. An over-extended run A deletes rows sibling B just committed; B then reaps A's fresh rows; the range ends up covered by nothing and cold-reads as silent zeros. Same class as #1879 / #1956.

Serially this was unreachable, because run *i+1* had not started when run *i* extended, so its intervals were still unsynced and `warmTail` bailed on them. That safety was a property of the schedule, and concurrency removes it.

**The fix is refusal, not truncation.** Each run receives the next run's start as a `limit`, and refuses to extend at all when the row reaches past it:

```go
if rowEnd > limit {
    // The row reaches past a later run: leave this one as snapshotted rather
    // than half extended, which is what a serial carve does when it finds that
    // range still dirty.
    return run, nil
}
```

This is provably serial-equivalent rather than merely safe: `splitRuns` guarantees run *i+1*'s first interval starts exactly at `limit`, so any `rowEnd > limit` forces `warmTail`'s walk through an unsynced interval. Serial refuses iff `rowEnd > limit` — the identical predicate. No geometry where serial extends and this refuses, none where serial refuses and this permits.

Truncating to `limit` instead was tried and rejected: it breaks `TestCarveRunDoesNotExtendIntoDirtyRange` (#2049), whose contract is "left as snapshotted rather than half extended", and it is unsafe on its own terms — since the reap deletes straddling rows whole, a run tiling only to `limit` still deletes the row straddling it, trading one uncovered range for a smaller one without changing the bug's class.

Run reap ranges are now pairwise disjoint and ordered, so one run's reap can never delete a row a sibling just wrote.

## Tests

- `TestCarveScatteredRunsConvergeConcurrently` — 200 scattered runs converge to `UnsyncedBytes() == 0`, per-offset committed bytes and on-disk `synced` flags verified. A `-race` vehicle and regression guard; it does not itself prove concurrency.
- `TestCarveSharedRecordAcrossRuns` — two runs holding fragments of one physical record, separated by a synced gap. The shape was found by fuzzing 300 seeds and delta-debugging to five plain writes; the test asserts via `runsSharingARecord` that the shape was actually constructed *before* carving, so it cannot silently degenerate.
- `TestCarveRunDoesNotExtendPastNextRun` — pins the fix. Gates on *ordering*, not on a return value: the first run's row lookup blocks until the second run's interval is observed `synced`, then returns a row end past it. Reverted, it prints `reap range [0,32768) covers the run starting at 16384`. It also fails loudly if the gate is never reached, rather than passing vacuously.

## Ordering

Depends on #2071 (merged), which puts `ReapSupersededManifest` under the per-file commit stripe lock. Without it, concurrent same-file runs turn badger SSI retry churn into `EDEADLK` at the client under exactly this PR's target workload. Rebased onto it.

## Known and deliberately unchanged

- `ManifestRowEndAfter` remains an unlocked read; the refusal removes the consequence that mattered, but its answer is a snapshot.
- The pre-existing "run ends mid-row, reap deletes the straddling row" behaviour from #2049 is preserved exactly.
- On error, counters are summed across all runs including cancelled siblings, where serially they were the prefix up to the failure. Counters only — every production caller discards `CarveResult`.
- The documented carve-RAM bound now names the per-run `chunker.MaxChunkSize` scratch buffer, which is not under `sem`, with a `ponytail:` marker for the `ChunkParams.Max` upgrade path.

### DIFF
--- a/.planning/perf/2026-08-24-1872-carve-drain-convergence-PLAN.md
+++ b/.planning/perf/2026-08-24-1872-carve-drain-convergence-PLAN.md
@@ -1,0 +1,104 @@
+# #1872 carve drain convergence — staged plan (tracking: #2070)
+
+A randomly-written share's drain does not converge: `dfsctl system drain-uploads` times out and
+eviction fails with it. Measured tail on the bench VM was ~2 MiB/30s (~68 KiB/s), network TX
+matching, about 21 serial PUTs/s.
+
+#1875 fixed both contributing defects in one diff and **regressed on hardware** — blocks committed
+continuously while `UnsyncedBytes` stayed flat at 644 MiB, meaning records were never flipped
+synced and the carve re-carved the same ranges forever. That root cause is still unknown. The
+plan below exists to avoid paying that price again.
+
+## The two defects, and why they are separable
+
+`pkg/block/journal/carve.go`, current develop:
+
+| # | Defect | Site | Touches the flip chain? |
+| --- | --- | --- | --- |
+| 1 | Runs upload serially — `carveRun` ends by waiting for its own commits | `carve.go:448` | **no** |
+| 2 | A block is cut at every run boundary, so a 4 KiB run commits its own remote object | `carve.go:424`, tail flush at `:447` | **yes** |
+
+#1875 assumed they were one change, because it fixed both by hoisting the block builder *and* the
+dispatcher to file scope — and a block spanning runs forces `flipUpTo` (`carve.go:574`) to walk the
+whole snapshot with one shared `flipIdx`. That is the part that broke.
+
+Defect 1 alone can be fixed with `flipUpTo` untouched, because each run already carries its own
+dispatcher and its own `flipIdx` over its own slice (`carve.go:273`, `:281`). Defect 1 is also the
+one that makes the drain *converge*, which is the user-visible bug. Defect 2 only makes it faster.
+
+## Step 0 — the regression test that was missing
+
+`pkg/block/journal/carve_scatter_test.go` (new; the same file name exists on the closed
+`perf/1872-carve-pack-across-runs` branch and its two tests are worth salvaging as-is).
+
+Assert that after carving a scattered dirty set, `UnsyncedBytes() == 0`, under a sink that mimics
+the real commit path rather than a fake that always succeeds. #1875's tests asserted block count
+and peak commit concurrency; both passed while the drain looped forever, which is exactly the
+failure this must catch.
+
+Bar: this test fails on `perf/1872-carve-pack-across-runs` and passes on develop. Verify both before
+moving on — a test that passes on the known-bad branch is worth nothing here.
+
+## Step 1 — parallelise runs (no flip changes)
+
+**`carve.go:246`** — the run-splitting loop calls `carveRun` serially. Replace with a bounded
+errgroup over the run slices. Runs are disjoint by construction, and each already owns its
+`flipIdx`, its dispatcher and its `newOffsets`.
+
+**`carve_dispatch.go:60`** — `sem` is built per dispatcher, so N concurrent runs would multiply
+peak carve RAM by N. Hoist it: build one `chan struct{}` in `carveFile` and pass it to
+`newCarveDispatcher`. This *restores* the documented contract — `store.go:67` already says
+`CarveUploadConcurrency` bounds how many of **one file's** blocks are in flight, which per-run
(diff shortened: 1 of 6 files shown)


marmos91:
Addressed the re-review finding by filing it as #2073 with a validated design rather than landing it here. Reasoning, since "not fixing it" deserves an argument:

**The finding is real** — verified end to end. `ctx.Err()` at `carve.go:416` sets `packErr`, and the abort block returns on both exits without reaping, after `commitAndFlip` has already flipped everything committed before the abort.

**But the suggested fix is unsafe.** Reaping only the committed prefix opens a worse hole than the overlap it closes. `ReapSupersededManifest` deletes a row *whole* when `runStart <= rowStart < runEnd`, regardless of `rowEnd` — the narrowing branch fires only for `rowStart < runStart` and explicitly refuses when `rowEnd > runEnd`, precisely to avoid a tail gap. The only thing normally preventing a straddler delete is `runEnd` being a manifest row boundary, which `extendRunToRowEnd` guarantees and a prefix watermark does not.

The consequence splits, and one half is not self-healing:

| straddler | uncovered range | outcome |
| --- | --- | --- |
| `rowEnd <= runEnd` | still-dirty local data | reads serve from the journal, next carve re-tiles — invisible |
| `rowEnd > runEnd` | warm and durable, not dirty | nothing revisits it — **permanent**, cold-reads as zeros |

That second row is reachable on purpose: `extendRunToRowEnd` deliberately declines to extend when the tail is not warm, when the sink has no `manifestRowEnder`, or when the row reaches past the next run's start.

Clamping to a committed offset does not rescue it either — `newOffsets` holds the *fresh* rows' starts, while the straddler is a *stale* row with unrelated boundaries.

**There is a sound fix** (guard with `ManifestRowEndAfter(watermark) <= watermark`, ~15 lines, no new config or interface method) and it is written up in #2073. It is not in this PR because it is new behaviour on the error path where the fakes do not implement `manifestRowEnder` at all, it depends on three subtleties at once, and botching it yields a hole rather than an overlap.

**Scope:** pre-existing on develop, which already hits this whenever a run's own commit fails. This PR adds sibling cancellation as a second trigger — same failure mode and severity, higher frequency. Not a regression this PR introduces, unlike the first reap finding, which was and is fixed here in `5f5511fd`.

Thanks — both reap findings in this review were real, and the first one was a genuine regression that two prior reviews missed.

Issue #1891, opened 2026-07-30
Field evaluation for a tier-0 sovereign-storage use case: metadata scale, offline resilience, multi-client — seeking your perspective (+ partner POC?)


raffaelegiustini-create:
## Context

We're evaluating DittoFS for a **large enterprise / tier-0 sovereign-storage use case** (telco-scale, Windows/AD environment, strict compliance, "keeps serving if the cloud is unreachable" as a core requirement). Reliability and performance have improved a lot recently — the corruption (#1879) and startup-scan (#1885) fixes in v0.29.2 made the box solid, and everyday drive usage now feels production-grade.

Before proposing a structured POC, I want your perspective on **which validation gaps are highest-risk** for a deployment of this size. Three areas stand out. I ran quick, informal probes on our single box (Hetzner CX33, 4 vCPU / 7.6 GB, Cubbit DS3 remote); the numbers below are indicative, not rigorous.

## 1. Metadata scale (millions of files)

The concern is directly the #1885 class: what happens at 1M–10M+ files, not 11k.

Informal probe (generating many tiny identical files via the NFS loopback):
- **Create rate ~260 files/s single-stream.** For a bulk ingest of millions of files that's slow per-stream — presumably it scales with concurrent streams, but I haven't measured the aggregate.
- **Metadata ~1.3 KB/file** → ~1.3 GB badger at 1M files. Linear, fine.
- Dedup works as expected (identical content → no disk growth).
- I could **not** cleanly capture a startup-time-vs-file-count curve — an operational issue got in the way (see below). Since v0.29.2 writes the `pl:` index at write time, I'd expect new-file startup to stay fast; the open question is the remaining O(N) `usedBytes` scan at 500k–1M+.

**Q:** Is ~260 creates/s single-stream expected? Any known ceiling (startup, enumeration, badger recovery) as file count reaches the millions?

## 2. Offline resilience / sovereignty (the differentiator)

This is *the* reason the account is interested — it has to keep serving during a remote (Cubbit) outage. We validated the basics in #1843 (theme 4): warm working set serves offline, writes store-and-forward, cold archive fails fast (#1845). The first-class offline-capable mode is tracked in #1848.

**Q:** Is #1848 (warm-on-add + pin + a "N blocks remote-only / not offline-safe" health signal) the right thing to build the sovereign story on, and where is it on your roadmap? For a tier-0 pitch I'd want an operator-visible "this box would survive an outage" signal.

## 3. Multi-client concurrency

[...]

PR #2074, opened 2026-08-24
docs(durability): document relaxed_durability and the namespace loss window

marmos91:
## What was wrong

`relaxed_durability` is documented nowhere. It defaults to `true` on both the
badger and postgres metadata stores (`runtime/init.go:96`), and it defers the
fsync for every namespace operation — `create`, `unlink`, `rename`, `mkdir`,
`rmdir`, attribute-only `setattr` — to a 100 ms background syncer
(`durabilitySyncInterval`).

`docs/guide/durability.md` compounded this: it describes the default `local`
tier as surviving "process/node crash" with no qualification. That is true for
file data and the metadata paired with it, but not for namespace ops, which the
`durability` enum does not govern at all — they are controlled by a separate
metadata-store setting the reader is never told about.

This surfaced from a field evaluation (#1891) asking precisely what the
durability contract is under abrupt termination. We could not point at an
answer.

## Evidence

Measured on a 4 vCPU / 7 GB box, scoring acknowledged creates against an
fsync'd witness log and counting how many survive:

| failure | `relaxed_durability: true` (default) | `false` |
|---|---|---|
| SIGKILL / OOM-kill | 0 lost (0/321, 0/332, 0/331) | 0 lost (0/218, 0/233, 0/226) |
| hard power loss (`sysrq-b`) | 10/804, 6/1463 | 0/913 |

Two things worth stating in the docs came out of that:

- **Killing the process loses nothing at either setting.** Badger's memtable WAL
  is a `z.MmapFile`, so an acknowledged op is already in the page cache, and the
  page cache outlives the process. Only an event that takes the kernel down can
  lose the un-fsync'd tail. Operators conflate these two constantly — the
  evaluation that prompted this did.
- **The loss window is real but bounded**, and the ~100 ms measured matches
  `durabilitySyncInterval` exactly.

## What this changes

Docs only; no behaviour change.

- `durability.md` gains a "Namespace durability" section: what the setting
  covers, the failure-mode table above, why it is bounded loss and never
  corruption, and the opt-out.
- `configuration.md` gains the `relaxed_durability` key in the metadata-store
  reference, where an operator would actually look for it.

Relates to #1891.

### DIFF
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -976,6 +976,26 @@ These global sizes apply to every BadgerDB metadata store on the node. A single
 store can be overridden via its config-map keys when it is created (see below):
 `--config '{"path":"...","block_cache_mb":2048,"index_cache_mb":1024}'`.
 
+#### `relaxed_durability`
+
+Applies to the `badger` and `postgres` metadata stores. **Defaults to `true`.**
+
+Namespace operations (`create`, `unlink`, `rename`, `mkdir`, `rmdir`,
+attribute-only `setattr`) commit without an inline `fsync`; a background syncer
+makes them durable within 100 ms. Writes paired with file data commit
+synchronously either way, so this is bounded loss, never corruption.
+
+```bash
+# Strict: fsync every namespace commit (roughly a third of the create throughput)
+./dfsctl store metadata add --name badger-strict --type badger \
+  --config '{"path":"/var/lib/dittofs/metadata","relaxed_durability":false}'
+```
+
+Only an event that takes the kernel down — power loss, kernel panic, hypervisor
+reset — can lose that 100 ms window. Killing the `dfs` process (`SIGKILL`,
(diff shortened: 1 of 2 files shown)


marmos91:
Both findings were correct, thanks.

- **Postgres generalization**: fixed. Postgres has no background syncer at all; relaxed namespace commits run with `SET LOCAL synchronous_commit = off` (`postgres/transaction.go:139`), so the window is set by the server's own `wal_writer_delay`, not our ~100 ms. The doc now gives a per-backend table instead of one number. I've also marked the measured failure-mode figures as badger, while noting the page-cache mechanism behind them is backend-independent.
- **Config form**: fixed. `relaxed_durability` is a per-store config key set through `dfsctl store metadata add/edit --config`, not a server-config YAML block. Replaced the snippet with the CLI form.

Pushed in d06f7fbf.

Issue #1848, opened 2026-07-25
Explicit offline-capable share mode (warm-on-add + pin + offline-safety health signal)


marmos91:
## Motivation

Field use (see #1843) wants "sovereign storage keeps serving if the cloud is unreachable" as a first class, observable guarantee. The primitives already exist: `dfsctl share warm` pulls a share fully local, and the `pin` retention policy never evicts what's local, so `warm` + `pin` already gives full offline read availability today. But it's a convention operators have to assemble by hand, and there's no way to *see* whether a box would actually survive an outage.

The offline read hang that made this worse is already fixed (#1845, v0.29.0): a cold remote-only read now fails fast and cleanly instead of wedging the mount. This issue is about the positive side, making the box genuinely offline-safe and letting operators confirm it.

## Proposed shape

An explicit offline-capable mode for a share, built on the existing warm/pin machinery:

1. **Warm on add** — new remote-only blocks are pulled local automatically as they appear, not only on first read. So the local copy stays complete as data is written or drained, without an operator re-running `warm`.
2. **Pin** — reuse the existing `RetentionPin` policy so nothing local is ever evicted. Warm-on-add without pin would just churn.
3. **Health signal** — surface "N blocks remote-only / not offline-safe" (metric + `dfsctl share status`) so an operator can tell at a glance whether the share would keep serving with the remote unreachable. Zero remote-only blocks = provably offline-safe.

## Open questions

- Per-share flag/mode vs. composing `pin` + a new "warm on add" toggle.
- Where warm-on-add hooks in (drain completion? carve?) and its backpressure interaction with the local capacity cap (#1844).
- What the health signal counts (remote-only block count is the obvious one; whether to also expose unsynced/dirty bytes separately).

## Not in scope

Offline *writes* already work (store and forward, drains on reconnect). This is read availability only.

marmos91:
Shipped in #2082 (merged to develop as `ca01f58e`).

**The warm+pin half was already done** — `dfsctl share warm` + `--retention pin` already give full offline read availability, so nothing there was rebuilt. What was missing was purely observability: no remote-only block count existed anywhere in the tree, no metric, no share status field. That is what shipped.

**Why the obvious sources could not answer it**, recorded so nobody retries them:
- `BlocksLocal` / `BlocksCached` are **dead fields, never incremented** — `classifyBlocks` buckets by sync state, so a fully warm share reports 100% `BlocksRemote`.
[...]

PR #2080, opened 2026-08-24
feat(integrity): run the structural manifest scan on a schedule

### LINKED ISSUE #2076: Integrity audit never runs on its own: no scheduled manifest scan
## Summary

`CheckManifests` and `dfsctl store block audit-refcounts` (#1882) can detect a manifest that disagrees with a file's own block list, and `manifest_repair.go` can repair it. Nothing ever runs them. There is no scheduler, no config key, and no default cadence, so the detection only fires when an operator remembers to invoke it by hand.

## Why this matters

A range that has no manifest row but that the file's block list still claims is, at read time, indistinguishable from a legitimate sparse hole. That is deliberate and correct: sparse files depend on absent rows meaning zeros, and the read path has no honest basis to refuse.

The consequence is that this one damage shape stays silent by design. The unplaceable-row case is loud (#1881, #1965 make the read refuse), and the unknown-hash case is loud. The absent-row-with-a-live-claim case is only ever visible to the audit, which nothing schedules.

That is the exact shape reported in #1879: five files reading back with a clean zero prefix, correct total size, no error at any layer, stable checksum across reads. Those files sat damaged for over a month before anyone noticed, and only because a human opened a PDF.

## What exists today

- `CheckManifests` walks a payload's rows against its claim and reports holes, unknown hashes and unplaceable rows.
- `dfsctl store block audit-refcounts <share>` runs it and persists a summary at `<localStore>/audit-state/last-inv02.json`.
- `planPayloadRepairs` proposes `RepairReplaceRow` / `RepairRecreateRow` for what it finds.
- The CLI help suggests `audit-refcounts <share> || alert`, i.e. it already assumes someone wires up cron.

## Proposed

A per-share scheduled structural scan, off the read path, in the shape of the existing GC and syncer loops:

- An interval config key with a sane default, and an explicit off switch.
- Metrics for last-run time, payloads scanned, and findings by kind, so "when was this share last verified" is answerable without reading a JSON file off disk.
- Findings surfaced as a health signal rather than only a log line, so a damaged share is visible in `share status`.

Metadata-only, so no S3 egress. Distinct from #1490, which re-hashes block content for bit rot; this one checks structure and would have caught #1879 while the evidence still existed.

Context: raised while auditing what #1879 does and does not close, and it overlaps the proactive-integrity question a tier-0 evaluation is asking in #1891.

marmos91:
## What was wrong

The detection layer for structural manifest damage has been complete for a
while: `CheckManifests` walks a share's manifest against every file's own
block list, `dfsctl store check` runs it, and `manifest_repair.go` can put
the missing rows back. Nothing ever ran any of it. There was no scheduler,
no config key, no default cadence — the CLI help even suggested wiring
`|| alert` into cron, i.e. the product shipped the detection and left the
schedule to the operator.

That matters for exactly one damage shape, and it is the worst one. A range
with no manifest row that the file's block list still claims is, at read
time, indistinguishable from a legitimate sparse hole. Absent rows are how
sparse files are represented, so the read path has no honest basis to
refuse: it returns zeros and reports success. The unplaceable-row case is
loud and the unknown-hash case is loud; this one is silent by design, and
the audit was the only thing that could see it — the thing nothing
scheduled.

## What this adds

A per-share scheduled structural scan, shaped after the existing auto-GC
loop in `blockgc_scheduler.go`:

- **`integrity.auto_enabled` / `integrity.auto_interval`** — on by default,
  24h, with an explicit off switch. The interval floor is 5m rather than
  GC's 1m because a scan is a full metadata walk of every share; below that
  it never stops running.
- **The loop** (`pkg/controlplane/runtime/integrity_scheduler.go`) scans
  shares one at a time, synchronously, so runs never overlap and the walk
  never competes with itself for the metadata store. It calls the existing
  `Runtime.CheckManifests` with no repair options set, so it writes nothing.
  Metadata-only — no block is fetched, no remote object touched, no S3
  egress.
- **Metrics**: `dittofs_integrity_last_scan_timestamp_seconds`,
  `_last_scan_duration_seconds`, `_files_scanned`, and
  `dittofs_integrity_findings{kind}` split by `payloads_with_findings`,
  `damaged_payloads`, `claimed_uncovered_ranges`, `unplaceable_rows`,
  `unknown_hash_rows`. A share never scanned emits a last-scan timestamp of
  0 rather than nothing, so an alert on scan age fires for it instead of
  silently having no series to evaluate.
- **A health signal**: `dfsctl share show` reports when the share was last
  scanned and what was found, and a share with damaged payloads reads as
  `degraded` even though every subsystem probe is healthy — which is the
  point, since the probes cannot see this damage.

## The status shape

`ShareResponse.Status` and `GET /api/v1/shares/{name}/status` change type
from `health.Report` to a new `health.ShareStatus`, which **embeds**
`health.Report`. The JSON is a strict superset: the same top-level
`status` / `message` / `checked_at` / `latency_ms` keys, plus an optional
`integrity` object. Nothing that reads the old fields is affected.

The struct is deliberately shaped to take more than one section. The
offline-safety counter in #1848 is the next one, and it wants the same
thing this does — a per-share number an operator charts, sitting next to
the health verdict. Folding numbers into `Report.Message` would have worked
for one of them and collapsed as soon as there were two, since a share can
have several independent things worth reporting at once and prose cannot
be parsed back apart.

Note the downgrade rule only ever worsens a report: damage degrades a
healthy share and leaves an already-degraded or unhealthy one alone.

## Known ceiling

The schedule is a plain ticker, so it restarts from zero on every server
start and a box restarted more often than the interval never completes a
scan. The fix is to persist the last-run time per share and scan at startup
when it is stale — but scanning at every boot puts a full metadata walk on
the startup path, which is the cost this scheduler exists to keep off it.
Marked with a `ponytail:` comment naming the trigger, and called out in the
docs.

## Verification

- `go build ./... && go vet ./... && go test ./...` green.
- `TestStartScheduledIntegrityScan_TicksAndStops` runs the real scan against
  a real memory metadata store — no fake sink — and asserts the recorded
  status appears where the status report reads it, and that the loop stops
  on cancel.
- `TestWithIntegrity_DowngradeRule` pins the four cases of the downgrade
  join, including "never upgrades".
- `TestLoad_IntegrityAutoEnv_RoundTrip` proves the `*bool` and `Duration`
  knobs decode from `DITTOFS_INTEGRITY_*` through `Load`.

Closes #2076

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -310,6 +310,18 @@ func runStart(cmd *cobra.Command, args []string) error {
 		logger.Info("auto-GC disabled by config (gc.auto_enabled=false)")
 	}
 
+	// Background structural manifest scan: verifies that each share's
+	// manifest still covers what its files claim to hold. Metadata-only, so
+	// it fetches no block and touches no remote object. The damage it looks
+	// for is invisible at read time — an uncovered range a file still claims
+	// is indistinguishable from a sparse hole, so the read returns zeros and
+	// reports success — which is why nothing else can report it.
+	if cfg.Integrity.AutoScanEnabled() {
+		rt.StartScheduledIntegrityScan(ctx, cfg.Integrity.AutoInterval)
+	} else {
+		logger.Info("integrity scan disabled by config (integrity.auto_enabled=false)")
+	}
+
 	// Configure runtime
 	rt.SetShutdownTimeout(cfg.ShutdownTimeout)
 	// Seed an operator-pinned machine SID (if configured) BEFORE Serve so the
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -310,12 +310,10 @@ func runStart(cmd *cobra.Command, args []string) error {
 		logger.Info("auto-GC disabled by config (gc.auto_enabled=false)")
 	}
 
-	// Background structural manifest scan: verifies that each share's
-	// manifest still covers what its files claim to hold. Metadata-only, so
-	// it fetches no block and touches no remote object. The damage it looks
-	// for is invisible at read time — an uncovered range a file still claims
-	// is indistinguishable from a sparse hole, so the read returns zeros and
-	// reports success — which is why nothing else can report it.
+	// Background structural manifest scan: periodically checks that each
+	// share's manifest still covers what its files claim to hold — damage the
+	// read path cannot report, because an uncovered range reads back as a
+	// sparse hole. Enabled by default; bound to ctx so it stops on shutdown.
 	if cfg.Integrity.AutoScanEnabled() {
 		rt.StartScheduledIntegrityScan(ctx, cfg.Integrity.AutoInterval)
 	} else {
--- a/cmd/dfsctl/commands/share/list.go
+++ b/cmd/dfsctl/commands/share/list.go
@@ -175,7 +175,7 @@ func runList(cmd *cobra.Command, args []string) error {
 
 // shareStatusString renders the server's status, or empty when the server sent
 // none, so an absent status is never shown as a status of "".
-func shareStatusString(r *health.Report) string {
+func shareStatusString(r *health.ShareStatus) string {
 	if r == nil {
 		return ""
 	}
--- a/cmd/dfsctl/commands/share/show.go
+++ b/cmd/dfsctl/commands/share/show.go
@@ -206,15 +206,15 @@ func integrityRows(in *health.IntegrityStatus) [][]string {
 	if in == nil {
 		return [][]string{{"Integrity Scan", "not yet run"}}
 	}
+	at := in.LastRunAt.Local().Format("2006-01-02 15:04:05")
 	if in.Error != "" {
 		return [][]string{
-			{"Integrity Scan", "failed " + in.LastRunAt.Local().Format("2006-01-02 15:04:05")},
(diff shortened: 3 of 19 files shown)


marmos91:
All three Copilot comments are one defect, and they are right that it is real — but documenting both meanings of `0` would have preserved the problem rather than fixed it.

A scan that fails on every tick publishes a zero last-scan timestamp forever, which is byte-identical to a share whose scanner was never switched on. That is silent-by-design failure of the exact shape this scan exists to detect, so the fix is a second series, not a caveat.

Added `dittofs_integrity_last_scan_failed{share}` — 1 when the most recent scan errored, 0 when it completed or never ran. Paired with the timestamp: 0-timestamp + failed=0 is "never scanned", 0-timestamp + failed=1 is "scans are erroring", and there is now an alert for each.

The docs and comments at all four sites are corrected too, since "0 = never scanned" was incomplete regardless. The zeroing of the counters on a failed scan is deliberate and unchanged: a failed scan has no counts worth publishing, and stale ones would read as current.

`dfsctl share show` already distinguished the two cases correctly ("not yet run" vs "failed <time>" plus the error); the gap was metrics-only, which is the surface an operator has when the issue's own goal is answering "when was this last verified" without reading a JSON file off disk.

`TestIntegritySnapshot_FailedScanIsNotNeverScanned` pins it.

PR #2082, opened 2026-08-24
feat(offline): report per-share remote-only bytes and offline read safety

### LINKED ISSUE #1848: Explicit offline-capable share mode (warm-on-add + pin + offline-safety health signal)

marmos91:
Stacked on #2080 — review that first; this branch reuses the status shape it
introduces.

## What was missing

The warm+pin half of this issue was already done: `dfsctl share warm` pulls a
share fully local and the `pin` retention policy never evicts what is local,
so `warm` + `pin` already gives full offline read availability today. What
was missing was any way to *see* it. A repo-wide grep for remote-only /
cold-bytes residency returned nothing: no counter, no metric, no share-status
field. An operator could assemble the guarantee by hand and had no way to
confirm it held, or to notice when eviction quietly broke it again.

## Where the answer already lived

The residency bit exists and is exact — it just was not exported.
`journal.interval.cold` marks a range that was written, is durable on the
remote, and whose local bytes have been evicted. That is definitionally
"remote-only": a read of it fetches, and would fail with the remote
unreachable. Everything else in the local index is resident.

Nothing else in the tree answers this, and it is worth saying why, because
each near-miss looks usable:

- `BlockStoreStats.BlocksLocal` / `BlocksCached` are **dead fields, never
  incremented**. `classifyBlocks` (pkg/block/engine/stats.go) classifies by
  *sync state* only, so every non-dirty row lands in `BlocksRemote` — a fully
  warm share reports 100% `BlocksRemote`.
- `FileChunk.State` is upload lifecycle, not residency. `BlockStateRemote`
  means "carved and uploaded", and eviction never writes back to metadata.
  So **metadata alone cannot answer this at any cost**, and adding a
  residency column would put a DB write on every evict and every hydrate.
- `DataExtents` deliberately reports cold ranges as data (it exists to make
  SEEK correct), so it cannot distinguish them.
- `share warm` does not compute a remote-only set either — it re-fetches
  every row on purpose, for the same reason.

## What this adds

- `journal.Store.ColdExtents()` — an in-memory walk of the interval index,
  one shard lock at a time, tallying cold bytes and ranges. No metadata, no
  disk, no S3. It reuses the traversal shape `Stats()` and `liveColdEntries`
  already use.
- `engine.Store.OfflineReadiness()` — the gating, plus `Safe()`.
- `dfsctl share show`: `Offline Safe: yes` / `no (12.4 GiB remote-only across
  431 ranges)`.
- `dittofs_offline_safe{share}`, `dittofs_offline_remote_only_bytes{share}`,
  `dittofs_offline_remote_only_ranges{share}`.

## It refuses to answer rather than answer zero

The whole thing turns on one asymmetry: reporting zero remote-only bytes for
a share whose residency is not actually known reads as *provably offline
safe*, and would say that about exactly the shares whose data is most likely
to be remote-only. So three cases report unknown instead:

- the local tier does not track residency (in-memory backend),
- **the local tier has not been seeded from the manifest** — an unseeded tier
  holds no interval at all for ranges living only on the remote, so its index
  reports them absent rather than cold and the raw count comes back zero on
  the worst case the measurement exists to catch,
- the block store is closed.

An unknown share reports `dittofs_offline_safe = 0` and publishes no byte
counts, so a dashboard cannot mistake it for a clean fully-local one.

## Two deliberate calls

**Bytes, not blocks.** The issue asks for a block count. Journal intervals are
byte ranges that split on partial overwrite and clamp on hydrate, so they do
not stand in one-to-one for manifest chunk rows. A true block count means
joining cold extents against `ListFileChunks`, which drags a metadata walk
back in — for a number that answers "how much would break offline" less
directly than the bytes do. Reporting extents alongside the bytes gives the
fragmentation picture without the lie.

**This does not degrade the share's health.** A share that is not offline-safe
is not unwell — most remote-backed shares are not, by design, because the
local tier evicts what it has mirrored. That is the reason the status shape
#2080 introduces carries structured counters rather than prose: #2076's
finding *should* downgrade a share and this one should not, and a single
`Report.Message` string cannot express both at once.

## Known ceiling

`ColdExtents` is O(live intervals) per call, on the metrics scrape path and
the share-status path. A running counter would be O(1), but `cold` is set and
cleared across insert, split, clamp, evict, hydrate and repack, and a counter
that drifts on any one of those reports a share offline-safe when it is not.
Marked with a `ponytail:` comment naming the trigger: take the scan until a
scrape actually shows up in a latency profile.

## Verification

- `go build ./... && go vet ./... && go test ./...` green.
- `TestColdExtents_CountsEvictedRangesOnly` drives the **real** evict path —
  no fake — and asserts the tally is zero before eviction, positive after,
  and never exceeds the bytes the eviction actually freed on disk.
- `TestColdExtents_HydrateClearsIt` re-hydrates every cold range and asserts
  the figure returns to zero, so the number an operator watches during a warm
  actually falls.
- `TestOfflineReadinessOf_Gating` pins each refusal, including that a refusal
  always carries a reason and an answer never does.

Closes #1848

### DIFF
--- a/cmd/dfsctl/commands/share/show.go
+++ b/cmd/dfsctl/commands/share/show.go
@@ -103,6 +103,7 @@ func (sd ShareDetail) Rows() [][]string {
 			rows = append(rows, []string{"Status Detail", s.Status.Message})
 		}
 		rows = append(rows, integrityRows(s.Status.Integrity)...)
+		rows = append(rows, offlineRows(s.Status.Offline)...)
 	}
 
 	// Only show Retention TTL when a TTL is set
@@ -226,4 +227,21 @@ func integrityRows(in *health.IntegrityStatus) [][]string {
 			in.ClaimedUncoveredRanges, in.UnplaceableRows, in.UnknownHashRows)})
 	}
 	return rows
+}
+
+// offlineRows renders whether the share would keep serving reads with its
+// remote unreachable. A share whose residency the server could not determine
+// says so rather than claiming either answer.
+func offlineRows(o *health.OfflineStatus) [][]string {
+	if o == nil {
+		return nil
+	}
+	if o.Unknown != "" {
+		return [][]string{{"Offline Safe", "unknown (" + o.Unknown + ")"}}
+	}
+	if o.Safe {
+		return [][]string{{"Offline Safe", "yes"}}
+	}
+	return [][]string{{"Offline Safe", fmt.Sprintf("no (%s remote-only across %d ranges)",
+		bytesize.ByteSize(o.RemoteOnlyBytes), o.RemoteOnlyRanges)}}
 }
--- a/docs/guide/configuration.md
+++ b/docs/guide/configuration.md
@@ -642,6 +642,53 @@ Findings surface in three places:
 Run `dfsctl store check <share>` for the per-file detail behind a finding,
 and `--repair` to act on it.
 
+#### Offline read safety
+
+A remote-backed share's local tier is a cache: once a range is mirrored to the
+remote it becomes evictable, and an evicted range is served by fetching it
+back. Reads of such a range fail while the remote is unreachable. Whether a
+box would keep serving through an outage therefore depends on how much of its
+data is currently remote-only, and that number moves with every eviction and
+every warm.
+
+The server reports it per share:
+
+- `dfsctl share show <name>` prints `Offline Safe: yes`, or
+  `no (12.4 GiB remote-only across 431 ranges)`.
+- Prometheus: `dittofs_offline_safe{share}` (1/0),
+  `dittofs_offline_remote_only_bytes{share}` and
+  `dittofs_offline_remote_only_ranges{share}`.
+
+Zero remote-only bytes is a provably offline-safe share. To get there, warm
+the share and stop it evicting again:
+
+```bash
+dfsctl share warm /archive
+dfsctl share edit /archive --retention pin
+```
+
+The measurement never guesses. Three cases report **unknown** rather than a
+number, because a zero would read as "provably safe" for exactly the shares
+whose data is most likely to be remote-only:
+
+- the share's local tier does not track residency (the in-memory backend),
+- the local tier has not been seeded from the manifest yet — it holds no
+  record of ranges that live only on the remote, so they would count as
+  absent rather than remote-only,
+- the block store is closed.
+
+An unknown share reports `dittofs_offline_safe = 0` but publishes no byte
+counts, so a dashboard cannot mistake it for a clean fully-local share.
+
+The figure is **bytes, not blocks**. The local tier tracks byte ranges, which
+split and merge independently of manifest chunk rows; a block count would
+need a metadata walk to produce and would not answer "how much would break
+offline" any more precisely.
+
+Note this is read availability only. Offline **writes** already work — writes
+are stored locally and drain when the remote returns.
+
+
 The schedule restarts from zero on server start, so a box restarted more
 often than `auto_interval` never completes a scan. Lower the interval on a
 host that reboots frequently.
@@ -2262,6 +2309,8 @@ spec:
 | ⭐ `dittofs_integrity_last_scan_timestamp_seconds{share}` | Unix time the structural manifest scan last completed. 0 means it has never run since process start **or** its last attempt failed. |
 | ⭐ `dittofs_integrity_last_scan_failed{share}` | `1` when the most recent structural manifest scan failed, `0` when it completed or has never run. Pairs with the timestamp above to tell "never scanned" from "scans are erroring". |
 | `dittofs_integrity_last_scan_duration_seconds{share}` / `dittofs_integrity_files_scanned{share}` | Cost and reach of the last structural manifest scan. |
+| ⭐ `dittofs_offline_safe{share}` | `1` when every byte the share holds can be served with the remote unreachable, else `0`. A share whose residency cannot be determined reports `0` and publishes no byte counts. |
+| `dittofs_offline_remote_only_bytes{share}` / `dittofs_offline_remote_only_ranges{share}` | Data the local tier no longer holds and would have to fetch from the remote to serve, and how many ranges it spans. |
 
 ### Example alert expressions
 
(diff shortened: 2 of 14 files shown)


marmos91:
Review caught a real hole in the gating, now fixed on this branch.

`offlineReadinessOf` had a fast path: no remote configured → `Known: true, 0 bytes` → `Safe`, on the reasoning that a share with no remote has nothing to be cut off from. That is wrong for one reachable case, and it fails in the worst direction the whole design is built to avoid.

`dfsctl share edit <name> --remote ""` clears the binding and hot-reloads through `RebindShareBlockStore`, which rebuilds the block store over the same local dir with `remote == nil`. Journal recovery replays `cold.log` unconditionally, so intervals marked cold while the remote was still bound come back as cold in a store that now has no remote. And `read_internal.go:56` reconciles on `st.Cold` regardless of `HasRemoteStore()` — so those ranges are not merely unsafe during an outage, they never serve at all. The old code reported that share `dittofs_offline_safe = 1`.

Fix is to stop shortcutting: ask the local tier for its cold count even with no remote, and report it. The seed gate still applies only when a remote is configured, since seeding never runs without one — otherwise every genuinely local-only share would report unknown forever, which was the other way to get this wrong.

The reported figure understates that case (it says "not offline-safe" where the truth is "those ranges are unreadable"), but it is non-zero and honest, and distinguishing the two is a separate concern from this issue.

Covered by two new cases in `TestOfflineReadinessOf_Gating` — `no remote but cold ranges remain` and `no remote and no residency tracking` — plus a blanket assertion in the table that nothing ever reports `Safe()` with non-zero remote-only bytes. Documented in the configuration guide.

marmos91:
Retargeted the base from `feat/2076-audit-schedule` to `develop`: CI here is gated on `pull_request: branches: [develop, main]`, so a PR pointed at another feature branch gets no checks at all and can never go green.

The branch is still stacked on #2080 and should be merged after it. Until #2080 lands, the diff shown here includes its commits; it collapses to just the offline-safety change once #2080 is in. The commits belonging to this PR are the last four:

- `feat(offline): report per-share remote-only bytes and offline read safety`
- `test(offline): cover the cold tally against the real evict path and the gating rules`
- `docs(offline): document per-share offline read safety and its metrics`
- `fix(offline): never report a share with orphaned cold ranges as offline-safe`

marmos91:
All four Copilot comments addressed.

**1. `offline.go:70` — `!hasRemote` fast path reports Safe for an unreadable share.** Correct, and already fixed in `3cd50ec8` — Copilot reviewed `c834be68`, one commit earlier. Independently found by a reviewer pass on the same code, which is worth noting: two passes reached the same rebind-to-local-only scenario from different directions. Fix asks the local tier for its cold count even with no remote; the seed gate now applies only when a remote is configured, since seeding never runs without one and gating unconditionally would have made every genuinely local-only share report unknown forever.

**2. `coldstats.go:34` — cold bit set without a persisted marker.** Real, and worse than a gap in this metric. Traced and filed as **#2084**.

`Invalidate` (store.go:1150) sets `iv.cold` with no `appendCold`. `evictSegment` (reclaim.go:238) builds its cold-log entries from intervals that are **not** already cold — right for the case it was written for, since an already-cold interval owns no record in that segment, but `Invalidate` is the one path that sets `cold` on an interval whose record is still live. So no entry is logged, `retireSegment` unlinks the only copy, and recovery finds neither record nor marker: the range becomes a hole and reads as zeros. The comment above `evictSegment` states the invariant this violates verbatim.

Reachable via `healCorruptWarmRead` when the heal's re-hydrate fails (unreachable remote), which leaves the interval cold-in-memory indefinitely. Not fixed here — it belongs with journal durability rather than an observability change, and `reclaim.go` is adjacent to the #1872/#2073 rewrite. Filed as a code-read finding with a note to reproduce on the device-loss rig before fixing, since "recovery drops the interval" is the load-bearing claim.

Note this is invisible to the metric either way: once the interval is gone there is nothing left to count. No gate here can cover it.

**3. `offline.go:20` — context not propagated.** Real, fixed. `ColdExtents` now takes a ctx and gives up between shards, so a caller's deadline is honoured after at most one shard's slice and the check never lands while a lock is held. A cancelled or failed walk reports **unknown**, not the partial total — a truncated count would read as a share holding less remote-only data than it does, which is the wrong direction. `TestColdExtents_HonoursContext` and `TestOfflineReadinessOf_CancelledScanIsUnknown` cover it.

**4. `collectors.go:179` — `Known` branch untested.** Correct, added `TestRuntimeCollector_UnknownOfflineShareOmitsByteGauges` and `TestRuntimeCollector_KnownOfflineShareEmitsByteGauges`, asserting an unknown share emits `offline_safe=0` and no byte gauges, and a known one emits both.

PR #2087, opened 2026-08-24
perf(seed): batch the manifest cold seed into one durable append

### LINKED ISSUE #2083: perf(seed): SeedColdFromManifest does one fsync per payload — dominant on KV backends, ~30% on postgres
## Summary

`SeedColdFromManifest` spends **87–98% of its wall clock in one fsync per payload**, not in metadata lookups. On a 3,500-file share that is ~20–25s of pure fsync, and it is the actual reason a large-store migration "looks like a hang" (#1843).

This was found while measuring #1868, whose premise — that the per-chunk `IsSynced` round-trip was the driver — is measured false on badger and sqlite (≤1%, below the run-to-run noise floor). The cost is a layer down.

## The path

```
shares.SeedColdFromManifest   (pkg/controlplane/runtime/shares/lifecycle.go:529)
  └─ engine.Store.SeedCold    (pkg/block/engine/flush.go:113)      once per payload
       └─ journal.Store.SeedCold (pkg/block/journal/store.go:503)
            └─ appendCold     (pkg/block/journal/cold.go:121)      Stat + Write + Sync
```

`appendCold` fsyncs before returning and holds `s.coldMu` across the whole call, so the seed is a serialized chain of ~5.5–6.2ms fsyncs, one per payload.

## Measurements

Apple M1 Max. Harness `pkg/controlplane/runtime/shares/cold_seed_bench_test.go` on branch `1868-seed-measure` (not pushed) — four attribution modes over one manifest: `full` (production), `baseline` (no `IsSynced`), `scanonly` (enumerate + list only), `lookuponly` (scan + `IsSynced`, no `SeedCold`).

6,400 chunks / 400 payloads, 5 iterations:

| backend | full | scanonly | fsync share |
| --- | --- | --- | --- |
| memory | 2.169s | 53ms | ~97% |
| badger | 2.192s | 27ms | ~98% |
| sqlite | 2.869s | 358ms | ~87% |

56,000 chunks / 3,500 payloads (badger): `full` 18.7–25.3s, `scanonly` 255–320ms.

**The knob is file count, not chunk count.** Same 56,000 chunks spread over only 100 large files — so only 100 fsyncs — collapses badger's `full` from 18.7s to **0.879s**. `SeedCold` scales with payloads; `IsSynced` scales with chunks.

## Why this is safe to batch on the seed path specifically

`appendCold` already takes a `[]coldEntry` slice — it is batched at the entry level and simply called once per payload. Batching across payloads needs no new API.

The durability comment on `appendCold` is written for its **other** caller:

> the caller (eviction) is about to unlink the only local copy of those bytes, and a lost entry means silent zeros, not a slow read

That reasoning is correct and must be preserved for `reclaim.go:261`. It does **not** transfer to the seed path: the seed is idempotent — an interrupted seed leaves no `ColdSeeded` marker and simply repeats on next start. So a crash mid-seed costs a repeat, not silent zeros.

**Any fix must keep eviction's per-call fsync and relax only the seed's.** Collapsing both is a silent-data-loss bug of the #1850 / #1879 class.

## Not in scope

- The `IsSynced` question is #1868 and is closed as refuted.
- Postgres is a separate story (~35% in lookups at ~213µs each); that is a prescan question, not a fsync one.

## Note on sequencing

`pkg/block/journal/` is being actively rewritten for #1872 (carve block packing). This should land after that, not concurrently.

marmos91:
## Problem

`SeedColdFromManifest` called `engine.SeedCold` once per payload, and every one of
those calls fsynced the cold log. That fsync — not the metadata lookups — is where
the seed spends its time:

| backend | share of seed wall clock |
| --- | --- |
| memory | ~97% |
| badger | ~98% |
| sqlite | ~87% |
| postgres | 29.3% |

The knob is **file count, not chunk count**: `SeedCold` fsyncs per payload while
`IsSynced` runs per chunk. The same 56,000 chunks spread over 100 large files
instead of 3,500 small ones takes badger's seed from 18.7s to 0.879s. A 3,500-file
share pays ~20-25s of pure fsync, which is a large part of why a big migration
"looks like a hang" (#1843).

## Fix

`appendCold` already takes a `[]coldEntry`. It was batched at the entry level and
merely *called* once per payload, so batching across payloads needs no API change
and — critically — no change to `appendCold` itself.

- `journal.SeedColdBatch` takes many files and makes one durable append for the
  batch. `SeedCold` is now a one-element call into it.
- `engine.SeedCold` becomes `engine.SeedColdBatch`.
- `SeedColdFromManifest` buffers seeds across payloads and flushes every 64Ki
  extents, plus once at the end.

**`appendCold` is unchanged except for its comment**, so eviction's per-call fsync
cannot have been affected by this diff.

`engine.Store.SeedCold` is removed rather than kept as a wrapper: it had one
caller and no behaviour of its own. `journal.Store.SeedCold` stays, as a
one-element call into `SeedColdBatch`.

## Why only the seed may batch

`appendCold` has two callers with opposite requirements, and its comment read like
a general contract while having been written for only one of them:

- **eviction** (`reclaim.go`) is about to unlink the only local copy of the bytes
  its entries describe. A lost entry there is silent zeros on the next start. Its
  per-call fsync is load-bearing.
- **seed** (`store.go`) takes nothing away. An interrupted seed leaves no
  `ColdSeeded` marker and simply repeats on the next start, so a crash mid-seed
  costs a repeat, not data.

The first commit is comment-only and says exactly this at the definition and at
both call sites, so the distinction is visible where the decision gets made rather
than only in issue text. This project has lost data twice on this confusion
(#1850, #1879); the expensive part both times was diagnosis.

## Also fixed here

`engine.Store.SeedCold` never took the store's `closeMu` lifecycle gate, though
every other public data op on `Store` does. A seed racing `Close` could pass the
journal's own `closed` check and then have `appendCold` reopen `cold.log` behind a
torn-down store, leaking the fd. Pre-existing, but batching lengthens the call, so
it is gated here and added to `TestStore_OpAfterCloseReturnsErrStoreClosed`, which
enumerates the gated ops. Batching in bounded chunks rather than one call for the
whole manifest is also what keeps a concurrent `Close`'s wait short.

## Tests

Two guards in `cold_test.go`, both verified by mutation (the stated regression was
introduced, the test failed, the mutation was reverted):

- `TestEvictAppendsColdEntriesBeforeReturning` — the important one. Reads the cold
  log back from the filesystem, not from the store's index, right after `Evict`
  returns. Fails if eviction's append is ever deferred or batched as a "symmetry
  cleanup". It cannot see the fsync itself: a write without `Sync` still reads back
  out of page cache, so no in-process test distinguishes them — losing the `Sync`
  alone stays the device-loss rig's job, and the test comment says so.
- `TestSeedColdBatchIsDurableAndMatchesPerFileSeeding` — the batch must be on disk
  when it returns, and must write byte-for-byte what per-file seeding wrote.

The controlplane's final flush is already covered end-to-end by
`TestSnapshotByteVerify_RemoteColdSeed_Matrix`, whose manifests are small enough to
live entirely in the tail batch.

Closes #2083

### DIFF
--- a/pkg/block/engine/close_gate_test.go
+++ b/pkg/block/engine/close_gate_test.go
@@ -182,6 +182,11 @@ func TestStore_OpAfterCloseReturnsErrStoreClosed(t *testing.T) {
 			t.Fatalf("want ErrStoreClosed, got %v", err)
 		}
 	})
+	t.Run("SeedColdBatch", func(t *testing.T) {
+		if err := bs.SeedColdBatch(ctx, []ColdSeed{{PayloadID: "p", Extents: [][2]int64{{0, 4096}}}}); !errors.Is(err, ErrStoreClosed) {
+			t.Errorf("SeedColdBatch after Close = %v, want ErrStoreClosed; a seed that slips past Close reopens the cold log behind a torn-down store", err)
+		}
+	})
 	t.Run("EvictLocal", func(t *testing.T) {
 		if err := bs.EvictLocal(ctx, "p"); !errors.Is(err, ErrStoreClosed) {
 			t.Fatalf("want ErrStoreClosed, got %v", err)
--- a/pkg/block/engine/flush.go
+++ b/pkg/block/engine/flush.go
@@ -118,9 +118,18 @@ type ColdSeed struct {
 // of a manifest seed's wall clock on a share with many small files. Every entry
 // is held in memory until that write, so callers bound their own batches.
 //
+// The batch is one lifecycle-gated op, so a Close waits out the append in flight
+// and the seeds after it fail fast rather than reopening the cold log behind a
+// torn-down store. Seeding in batches rather than one call for a whole manifest
+// is what keeps that wait short.
+//
 // No-op when the local store has no remote-hydration support (e.g. the in-memory
 // test store), which the caller only hits on non-remote paths anyway.
 func (bs *Store) SeedColdBatch(ctx context.Context, seeds []ColdSeed) error {
+	if err := bs.enter(); err != nil {
+		return err
+	}
+	defer bs.closeMu.RUnlock()
 	type coldSeeder interface {
 		SeedColdBatch(ctx context.Context, seeds []journal.ColdSeed) error
 	}
--- a/pkg/block/engine/flush.go
+++ b/pkg/block/engine/flush.go
@@ -101,24 +101,38 @@ func (bs *Store) DrainRollups(ctx context.Context) error {
 	return err
 }
 
-// SeedCold marks payloadID's extents — each an {offset, length} pair — as
-// remote-durable-but-not-local so a subsequent read faults them in from the
-// remote store rather than zero-filling. Callers pass a whole file's extents at
-// once: the local tier persists the markers, and one durable write per file beats
-// one per extent when a caller is seeding an entire manifest. Snapshot restore
-// and the pre-journal upgrade both use it (remote-backed shares only) to arm cold
-// reads over a local tier that holds none of the bytes. No-op when the local store
-// has no remote-hydration support (e.g. the in-memory test store), which the
-// caller only hits on non-remote paths anyway.
-func (bs *Store) SeedCold(ctx context.Context, payloadID string, extents [][2]int64) error {
+// ColdSeed is one payload's worth of work for SeedColdBatch: the payload ID and
+// the {offset, length} extents of it to mark cold.
+type ColdSeed struct {
+	PayloadID string
+	Extents   [][2]int64
+}
+
+// SeedColdBatch marks the seeds' extents as remote-durable-but-not-local so a
+// subsequent read faults them in from the remote store rather than zero-filling.
+// Snapshot restore and the pre-journal upgrade both use it (remote-backed shares
+// only) to arm cold reads over a local tier that holds none of the bytes.
+//
+// Callers pass many payloads at once because the local tier makes the markers
+// durable once per call, not once per payload — an fsync per file is nearly all
+// of a manifest seed's wall clock on a share with many small files. Every entry
+// is held in memory until that write, so callers bound their own batches.
+//
+// No-op when the local store has no remote-hydration support (e.g. the in-memory
+// test store), which the caller only hits on non-remote paths anyway.
+func (bs *Store) SeedColdBatch(ctx context.Context, seeds []ColdSeed) error {
 	type coldSeeder interface {
-		SeedCold(ctx context.Context, id journal.FileID, extents [][2]int64) error
+		SeedColdBatch(ctx context.Context, seeds []journal.ColdSeed) error
 	}
 	cs, ok := bs.local.(coldSeeder)
 	if !ok {
 		return nil
 	}
-	return cs.SeedCold(ctx, journal.FileID(payloadID), extents)
+	js := make([]journal.ColdSeed, 0, len(seeds))
+	for _, sd := range seeds {
+		js = append(js, journal.ColdSeed{ID: journal.FileID(sd.PayloadID), Extents: sd.Extents})
+	}
+	return cs.SeedColdBatch(ctx, js)
 }
 
 // RestoreToVersion rewinds the local journal to a snapshot's version watermark
--- a/pkg/block/journal/cold.go
+++ b/pkg/block/journal/cold.go
@@ -115,9 +115,17 @@ func decodeColdEntry(buf []byte) (coldEntry, int, error) {
 }
 
 // appendCold durably records entries so a restart still knows the ranges are
-// remote-resident rather than holes. It fsyncs before returning: the caller
-// (eviction) is about to unlink the only local copy of those bytes, and a lost
-// entry means silent zeros, not a slow read.
+// remote-resident rather than holes.
+//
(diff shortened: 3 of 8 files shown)


marmos91:
Do not merge before #1872 lands — this touches `pkg/block/journal/store.go`, which that branch also edits. Rebase onto its merge first.

marmos91:
Thanks — going through the three:

**1. Hold `closeMu` for the entire batch operation — agreed, fixed in `ebed4068`.**
This was a real gap, and a pre-existing one: `engine.Store.SeedCold` never took the
gate on develop either, so a seed racing `Close` could pass the journal's own
`closed` check and then have `appendCold` reopen `cold.log` behind a torn-down
store, leaking the fd. `SeedColdBatch` now opens with `bs.enter()` /
`defer bs.closeMu.RUnlock()` like every other public data op in that file, and
`TestStore_OpAfterCloseReturnsErrStoreClosed` — which enumerates the gated ops —
now has a `SeedColdBatch` case. Batching in bounded chunks rather than one call for
a whole manifest is also what keeps a concurrent `Close`'s wait short.

**2. Serialize planning, append and index updates, or revalidate against concurrent
deletion/truncation — real, pre-existing, and deliberately left; recorded as a
`ponytail:` ceiling in `44351e9b`.**
Concurrent *writes* are already safe: `fileIndex.insert` resolves overlaps by
version, and anything landing after planning carries a higher global version, so it
shadows the pending cold entry rather than losing to it. The genuine gap is
`Delete`, which fences on a tombstone version and sweeps once — a cold entry
planned below that version and inserted after the sweep recreates the range. That
gap is not introduced here: a single-file `SeedCold` drops the shard lock across
`appendCold` too. Batching holds it open for the batch instead of one fsync.

Not closed because neither caller can reach it — `seedColdIfNeeded` runs during
share add, before the share serves, and the snapshot path runs against a local tier
that was just reset — and the two available fixes are both worse than the race:
holding shard locks across an fsync serializes the journal on disk latency, and
revalidation needs a per-file delete watermark that has no other user. The comment
names the watermark as the upgrade path if a seed ever has to run against a live
share.

**3. Preserve an exported one-file `SeedCold` on `engine.Store` — intentionally not
kept.**
`pkg/block/engine` is internal to this binary, `engine.Store.SeedCold` had exactly
one caller (`SeedColdFromManifest`), and it was a pass-through with no behaviour of
its own. A wrapper kept only so the removal isn't a removal is surface this repo
prefers to delete. `journal.Store.SeedCold` *is* kept, as a one-element call into
`SeedColdBatch` — it's the natural single-file entry point and the journal tests use
it. Called out in the PR description.

marmos91:
Second pass — three of the four are the first pass's findings re-raised against
the full PR diff rather than the current tree; the fourth is new and correct.

**`lifecycle.go:608` — the 64Ki bound is not enforced for a single oversized
payload. Correct, and the comment overstated it. Fixed in `277d7347`.**
The comment now says what is actually true: the bound is on what buffering
*across* payloads adds, not on peak heap. A payload whose own extents exceed the
bound goes in whole and flushes on the very next check.

Not splitting it, though, because splitting buys nothing here: `ListFileChunks`
has already materialized that payload's rows and the `extents` slice built from
them in full, before the batch ever sees them, and they stay live for the rest of
the callback. Chopping the slice into threshold-sized `SeedColdBatch` calls would
add flushes without freeing a byte. Peak is one payload's extents (unchanged from
before this PR) plus at most one bound's worth of buffer.

**`flush.go:133` — lifecycle gate. Already fixed in `ebed4068`**, before this
review ran. `SeedColdBatch` now opens with `bs.enter()` / `defer
bs.closeMu.RUnlock()`, and `TestStore_OpAfterCloseReturnsErrStoreClosed` has a
`SeedColdBatch` case. Good catch on the first pass — it was a real gap, and a
pre-existing one, since `engine.Store.SeedCold` never took the gate either.

**`flush.go:128` — removing exported `engine.Store.SeedCold`.** As before:
`pkg/block/engine` is internal to this binary, the method had one caller and no
behaviour of its own beyond a string→FileID conversion, so a wrapper kept only so
the removal isn't a removal is surface this repo deletes. Noted in the PR
description. `journal.Store.SeedCold` is kept because it has actual users.

**`store.go:582` — plan/append/index gap.** As before: real, pre-existing (a
single-file `SeedCold` drops the shard lock across `appendCold` too), and left
deliberately with a `ponytail:` marker in `44351e9b` naming the ceiling and the
upgrade path. Concurrent *writes* are already safe by version order; only `Delete`
can slip through, and neither caller can reach it — the seed runs against a share
that is not serving yet, or one whose local tier was just reset. Both available
fixes are worse than the race: holding shard locks across an fsync serializes the
journal on disk latency, and revalidation needs a per-file delete watermark with no
other user. The comment names that watermark as the trigger if a seed ever has to
run against a live share.

marmos91:
Ordering constraint **lifted**. The session that owns `pkg/block/journal` has released this to merge ahead of #1872: its rig run stalled the drain (badger SSI conflicts in `CommitBlock`) and that branch is now behind a stall investigation of unknown length, so holding a finished seed-side PR for it is the wrong trade. The overlap is `store.go` only and distant from its `CarveRunConcurrency` region; that branch will rebase over this.

This merges once CI is green, in blast-radius order behind #2089 and #2088.

Issue #1877, opened 2026-07-28
Cold barrier intermittently fails to evict the local tier on the sqlite backend


marmos91:
Seen once during the 292-cell QoS matrix run, on `dittofs-sqlite-s3-nfs3`:

```
FAIL dittofs-sqlite-s3-nfs3: evict: cold barrier failed: local disk only fell
5453MiB→4665MiB (want ≥80% drop); the cold pass would measure locally-served
reads, not S3
```

The drain succeeded — this is the evict half. A 14% drop at 5.4 GiB, well above the 64 MiB floor below which the barrier declines to judge.

**Not reproduced.** Re-running the same tier and cells immediately afterwards passed cleanly (`4096MiB→0MiB` and `4399MiB→42MiB`), and the published cell is from that retry. The other four DittoFS tiers passed the identical barrier in the same run, first time.

So this is one unexplained occurrence rather than a characterised defect, filed so it is not lost. Worth noting evict can only drop blocks it knows are remote-durable, so a plausible direction is the synced-hash oracle disagreeing about what is durable at that moment — but that is a guess, not a diagnosis, and the sqlite-only appearance may equally be coincidence given a single sample.

If it recurs, the useful evidence is `dfsctl store block stats -o json` before and after the evict plus the server log for that window.

marmos91:
Context update: the 292-cell matrix this came out of is published (#1876, closed #1743). This remains the only DittoFS-side gap in it — the cell was produced from an immediate retry, which passed cleanly (`4096MiB→0MiB`), and the other four DittoFS tiers cleared the identical barrier in the same run.

Still not reproduced, so still one unexplained occurrence rather than a characterised defect. Leaving it open rather than closing as unreproducible, because the direction noted in the body — the synced-hash oracle disagreeing about which blocks are remote-durable, so evict declines to drop them — is the same subsystem that #1879 and #1882 are circling. If a manifest/synced-hash inconsistency turns out to be behind that report, this is a plausible second symptom of it and should be re-checked then.

marmos91:
Diagnostics landed in #2079 (merged to develop as `059844cd`). **Keeping this open deliberately** — that PR instruments the failure, it does not explain it.

What the barrier now captures on *any* error out of it (the drain half included, not just the evict half), printed to the run log that `--remote` already pulls back:

- raw `store block stats -o json` at three points — barrier entry, **after the drain but before the evict**, and after the evict;
[...]

PR #2088, opened 2026-08-24
fix(dfsbench): judge the cold-barrier drain residue by the segments it pins

### LINKED ISSUE #2078: dfsbench cold barrier tolerates 32 MiB unsynced, but eviction's unit is a 256 MiB segment
Found while wiring diagnostics into the cold barrier for #1877. **A code-read fit, not a diagnosis** — no reproduction behind it, filed so the arithmetic is on record and the barrier diagnostics landing in that PR can confirm or kill it on the next occurrence.

## The mismatch

`dittofsDrainUntilSynced` (`internal/dfsbench/backend/dittofs.go`) stops draining once the store reports **≤32 MiB unsynced** (`dittofsDrainDriftFloorBytes`), on this reasoning:

> A few MiB out of a multi-GiB file is <0.1% and does not meaningfully warm the cold pass (the post-evict ≥80%-drop check is the real cold gate).

That reasoning is byte-denominated. Eviction is not.

- The journal evicts **whole segments**. `Config.SegmentSize` defaults to 256 MiB and `ShardCount` to 16 (`pkg/block/journal/store.go`), and `pkg/block/local/fs/fs.go` sets neither — so both defaults hold for every bench share.
- A segment qualifies for eviction only when **every** record in it is synced: `claimColdestEvictable` for the sealed set, `sealableActive` (`syncedRecords == records`) for the actives that `sealSyncedActives` force-seals during a `DrainLocalSynced`.

So a single straggler unsynced record pins its whole 256 MiB segment. 32 MiB of tolerated residue, scattered one straggler at a time, pins up to **18 segments = 4608 MiB** — and the tolerance is by design, so the barrier that follows fails on bytes the drain deliberately left behind.

## Why it fits the #1877 report

The one field occurrence was `5453MiB→4665MiB`. 4665 MiB is ~18.2 segments. The mechanism is intermittent in exactly the reported way — it turns on *where* the residue lands, not how much of it there is, which is why immediate reruns of the same tier passed cleanly (`4096MiB→0MiB`, `4399MiB→42MiB`).

It is not the only mechanism that fits those two numbers, though: a remote-health blip, accumulated `failed_syncs`, or a snapshot pinning segments all fit equally well from the outside. The barrier diagnostics in the #1877 PR capture the post-drain `unsynced_bytes` and the shape of the surviving file set, which separate them in one line of log.

## What a fix would look like, if it is confirmed

Options, cheapest first:

1. Have the barrier's drain loop require `unsynced_bytes == 0` when the store is idle, rather than a byte floor. The floor exists because a sub-GiB file's final bytes stay pinned in the unsealed append log indefinitely without another write — so this needs a way to force those out, not just a tighter threshold.
2. Have the barrier judge the drift against eviction's actual granularity rather than a fixed 32 MiB.
3. Report evictable-vs-pinned bytes from the store so the harness can ask the question directly instead of inferring it from a drop ratio.

Not proposing one until there is evidence which mechanism is real.

## Not a product bug (probably)

Refusing to drop a segment holding an unsynced record is correct — it is the only copy of those bytes. If this is confirmed, the defect is the harness's drift floor being denominated in the wrong unit, not the journal's eviction rule.

marmos91:
## What was wrong

`dittofsDrainUntilSynced` stopped draining once the store reported `unsynced_bytes <= 32 MiB`,
on the reasoning that a few MiB out of a multi-GiB file is <0.1% and cannot meaningfully warm
the cold pass. That reasoning is denominated in **bytes**. Eviction is denominated in
**segments**, and the two do not convert.

Verified against the tree, not assumed:

- `journal.Config.SegmentSize` defaults to `256 << 20` and `ShardCount` to `16`
  (`pkg/block/journal/store.go:100,104`).
- `pkg/block/local/fs/fs.go:152` builds its `journal.Config` with only `MaxLocalBytes`,
  `EvictMaxWait`, `ChunkParams` and `DirtyExpiry` — both defaults hold for the bench share.
- `evictable` (`pkg/block/journal/reclaim.go:207`) requires `syncedRecords == records`, and
  `sealableActive` the same for a force-seal during `DrainLocalSynced`. One unsynced record
  therefore pins its entire 256 MiB segment.
- `unsynced_bytes` in `dfsctl store block stats` is the journal's own dirty-record counter
  (`engine/stats.go` → `syncer.UnsyncedBytes()` → `journal.Store.UnsyncedBytes`), so the
  residue the harness tolerates is exactly the residue that pins segments.

So the tolerated residue's cost is not its size but the segments it lands in: 32 MiB sitting
in one segment costs 256 MiB of pinned local disk, while the same 32 MiB spread one straggler
at a time costs 32 segments — 8 GiB. The cold barrier that runs immediately after then fails
its ≥80%-drop check on bytes the drain deliberately left behind, and it does so intermittently
in exactly the way that depends on *where* the residue landed rather than how much of it there is.

## The fix

Re-denominate the tolerance. `dittofsDrainResidueOK(unsyncedBytes, localResidentBytes)` accepts
a residue only when its worst case — every straggler alone in its own segment — still leaves the
local tier able to shed the ≥80% the barrier goes on to demand. Below
`dittofsColdBarrierFloorBytes` the barrier never verifies the drop, so nothing a residue does
there can invalidate a measurement that was never going to be checked, and it is accepted.

A tolerance survives, as intended — it just scales with what eviction can free instead of
being a flat number:

| local tier | old floor | new tolerance |
| --- | --- | --- |
| 1 GiB | 32 MiB | 0 (256 MiB clears no 20% budget this small) |
| 4 GiB | 32 MiB | 3 MiB |
| 8 GiB | 32 MiB | 6 MiB |
| 64 GiB | 32 MiB | 51 MiB |

A genuine append-log trickle still passes; 32 MiB against a multi-GiB tier no longer does. Note
that the case it now refuses is one the old code did not really pass either — 32 MiB spread over
four shards' tails pins 1 GiB, so a 4 GiB tier failed the barrier a moment later, with a message
about local disk rather than about the residue that caused it.

The drain and the barrier now take their budget from one `dittofsColdBarrierBudgetBytes` helper,
so the tolerance cannot drift away from the assertion it exists to satisfy.

Two secondary changes:

- The straggler floor is an estimate (1 MiB, the carver's minimum chunk) because the store does
  not report how many records or segments the residue occupies. It is marked `ponytail:` with the
  upgrade path — report unsynced records or pinned segments from `dfsctl store block stats` and
  the whole derivation collapses into reading that field. The estimate is deliberately on the
  pessimistic side: a false drain failure is loud, whereas a tolerated pin produces a warm number
  labelled "cold". The known ceiling is named in the marker: several stragglers sharing one
  segment — one file's un-carved tail — are counted as if each held a segment, so a residue can
  be refused that eviction would have freed.
- An intolerable residue that sits unchanged for 3 rounds now fails immediately instead of
  riding out all 60. Each round costs a 15m `drain-uploads` timeout, so the tighter tolerance
  could otherwise have turned a failed barrier into a many-hour one. A residue that is still
  shrinking keeps draining as before.

The failure message now spells out the segment arithmetic, since the residue is small and what
actually fails is the local disk it pins.

## Scope

This is a harness fix on its own merits — the arithmetic was wrong in a unit that eviction
cannot honour. It is **not** a diagnosis: refusing to drop a segment holding an unsynced record
is correct journal behaviour, and no reproduction ties this to any observed barrier failure.
Nothing in `pkg/block/` changes.

## Verification

```
go build ./... && go vet ./...
go test -race ./internal/dfsbench/...
```

`TestDittofsDrainResidueOK` pins the pathological shape rather than the benign one: 32 MiB
against an 8 GiB tier must be rejected (the old flat floor accepted it), 1 MiB against the same
tier accepted, and 32 MiB against a 128 GiB tier accepted because that tier can absorb the pin.
Reverting the tolerance to a byte floor fails the third case.

Closes #2078
Refs #1877

### DIFF
--- a/internal/dfsbench/backend/dittofs.go
+++ b/internal/dfsbench/backend/dittofs.go
@@ -580,9 +580,14 @@ const dittofsEvictSegmentBytes int64 = 256 << 20
 // has to assume the finest spread it can plausibly see — the carver's minimum
 // chunk (chunker.MinChunkSize).
 //
-// ponytail: an estimate standing in for a count the store already tracks; report
-// unsynced records (or pinned segments) from `dfsctl store block stats` and the
-// estimate collapses into reading that field.
+// ponytail: an estimate standing in for a count the store already tracks, and a
+// deliberately pessimistic one — several stragglers sharing a segment (one file's
+// un-carved tail) are counted as if each held a segment of its own, so a residue
+// can be refused that eviction would in fact have freed. That direction is the
+// cheap one to be wrong in: a refused drain is loud, while a tolerated pin is a
+// warm number labelled cold. Report unsynced records (or pinned segments) from
+// `dfsctl store block stats` and both the estimate and its pessimism collapse
+// into reading that field.
 const dittofsDrainStragglerBytes int64 = 1 << 20
 
 // dittofsDrainResidueOK reports whether a drain residue is small enough to leave
--- a/internal/dfsbench/backend/dittofs.go
+++ b/internal/dfsbench/backend/dittofs.go
@@ -561,25 +561,81 @@ func dittofsWaitSettled(ctx context.Context) error {
 	}
 }
 
-// dittofsDrainDriftFloorBytes is the residual unsynced size the drain loop
-// tolerates as "done". A freshly-written file's final bytes sit in the append log
-// until the next rollup boundary, so a few chunks stay un-carved/un-synced
-// indefinitely without another write — requiring EXACTLY 0 loops forever. A few
-// MiB out of a multi-GiB file is <0.1% and does not meaningfully warm the cold
-// pass (the post-evict ≥80%-drop check is the real cold gate).
-const dittofsDrainDriftFloorBytes = 32 << 20 // 32 MiB
+// dittofsEvictSegmentBytes is the granularity at which eviction actually frees
+// local disk. The journal reclaims whole segments and refuses any segment still
+// holding an unsynced record, so one straggler keeps its entire segment resident.
+// pkg/block/local/fs leaves Config.SegmentSize unset, so the bench share runs on
+// the journal's own default.
+const dittofsEvictSegmentBytes int64 = 256 << 20
 
-// dittofsDrainUntilSynced loops `dfsctl system drain-uploads` until the store's
-// unsynced bytes fall to the drift floor, stable across two polls. A single drain
-// is not enough: rollup is async and carveFlush is snapshot-at-claim, so one pass
-// misses chunks that roll up mid-drain — leaving them locally resident and
-// un-evictable, so the "cold" pass silently reads them from local disk (the
-// confound that made several cold-read A/Bs meaningless). The short settle between
-// rounds lets the async rollup produce the next batch of CAS chunks for the
-// following drain to upload.
+// dittofsDrainStragglerBytes is the smallest residue assumed able to pin a
+// segment of its own. Unsynced bytes are spread over an unknown number of
+// records, and only that count decides how many segments they pin, so the drain
+// has to assume the finest spread it can plausibly see — the carver's minimum
+// chunk.
+//
+// ponytail: an estimate standing in for a count the store already tracks; report
+// unsynced records (or pinned segments) from `dfsctl store block stats` and the
+// estimate collapses into reading that field.
+const dittofsDrainStragglerBytes int64 = 1 << 20
+
+// dittofsDrainResidueOK reports whether a drain residue is small enough to leave
+// behind. Its size alone does not answer that: eviction frees segments, not
+// bytes, so what a residue costs is the segments it pins. 32 MiB sitting in one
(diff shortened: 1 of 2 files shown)


marmos91:
Both Copilot points addressed in 5a73f1d1.

**Constant drift** — `dittofsDrainStragglerBytes` now *is* `chunker.MinChunkSize` rather than a literal with a comment pointing at it. `pkg/block/chunker` is a leaf with no DittoFS dependencies, so the import costs nothing and the claim is now checkable instead of asserted.

**Rising residue read as a stall** — correct as filed: `>= prevUnsynced` counted a round where the residue ticked up as "not progressing", and a drain round can legitimately end higher than the one before it. Strict `==` is not the fix either, since a byte of jitter then lets a genuinely wedged residue ride out all 60 rounds at 15m each. Progress is now measured against the *lowest* residue seen so far: a transient rise is not a stall, and only three consecutive rounds that all fail to beat the best is. The comment says that rather than "unchanged".

Issue #2104, opened 2026-08-24
CodeQL: triage of the 88 open code-scanning alerts


marmos91:
Triage of every open alert on the [code-scanning tab](https://github.com/marmos91/dittofs/security/code-scanning). 88 alerts, six rules. I read the flagged code rather than classifying by rule name, so each verdict below has a reason.

**One is a real bug** (share name `..` escapes its storage directory, verified empirically). One class is worth fixing mechanically. The rest are false positives that should be dismissed so the tab becomes usable.

| Rule | Count | Verdict |
| --- | --- | --- |
| `go/path-injection` | 39 | 1 real, rest false positive |
| `actions/missing-workflow-permissions` | 24 | **Real — fix** |
| `go/incorrect-integer-conversion` | 15 | Latent, cheap central fix |
| `go/weak-cryptographic-algorithm` | 7 | False positive — dismiss |
| `go/weak-sensitive-data-hashing` | 2 | False positive — dismiss |
| `go/clear-text-logging` | 1 | False positive — dismiss |

---

## 1. REAL: a share named `..` escapes its per-share storage directory

Found while checking the `go/path-injection` alerts. `deriveLocalStoreDir` in
`pkg/controlplane/runtime/shares/blockstore_config.go:1147` builds the share's
block-store root as:

```go
sanitized := sanitizeShareName(shareName)
shareDir := filepath.Join(expanded, "shares", sanitized)
```

and the sanitizer (`shares/service.go:355`) is:

```go
func sanitizeShareName(name string) string {
	name = strings.TrimPrefix(name, "/")
	return url.PathEscape(name)
}
```

`url.PathEscape` escapes `/`, but **`.` is an unreserved character and is left
alone**, so `".."` passes through unchanged. Verified by running the sanitizer
verbatim:

```
name="normal"  sanitized="normal"      join="/var/lib/dittofs/shares/normal"   ESCAPES=false
name=".."      sanitized=".."          join="/var/lib/dittofs"                 ESCAPES=true
name="."       sanitized="."           join="/var/lib/dittofs/shares"          ESCAPES=true
name="/.."     sanitized=".."          join="/var/lib/dittofs"                 ESCAPES=true
name="../.."   sanitized="..%2F.."     join="/var/lib/dittofs/shares/..%2F.."  ESCAPES=false
```

Nothing upstream rejects it. `ValidateShareName` (`pkg/metadata/validation.go:135`)
only checks for empty, a `:`, and the length budget.

[...]

PR #2116, opened 2026-08-25
fix(shares): reject share names that cannot address their own directory

marmos91:
Fixes the one real bug from the #2104 CodeQL triage (`go/path-injection`).

## The bug

A share's local data directory is built in `deriveLocalStoreDir`:

```go
return filepath.Join(expanded, "shares", sanitizeShareName(shareName))
```

and the sanitizer is:

```go
func sanitizeShareName(name string) string {
	name = strings.TrimPrefix(name, "/")
	return url.PathEscape(name)
}
```

`url.PathEscape` escapes `/`, but **`.` is unreserved and passes through
untouched**. So `".."` survives the sanitizer intact and `filepath.Join` then
resolves the share's directory to `base` — outside the `shares/` tree that keeps
shares isolated from one another. Verified by running the sanitizer verbatim:

```
name="normal"  sanitized="normal"      join=".../shares/normal"    ESCAPES=false
name=".."      sanitized=".."          join="/var/lib/dittofs"     ESCAPES=true
name="/.."     sanitized=".."          join="/var/lib/dittofs"     ESCAPES=true
name="../.."   sanitized="..%2F.."     join=".../shares/..%2F.."   ESCAPES=false
```

Traversal is capped at one level, because the sanitizer *does* escape `/`.

Nothing upstream rejected it: `ValidateShareName` checked only for empty, a `:`,
and the length budget. The contrast is right next door — `ValidateName`, for
filenames, has always rejected `"."`, `".."`, `/` and NUL, and its doc comment
explains why that guard is centralized. Share names never got the same guard.

## A second case, found by the test rather than by reading

Writing the property test surfaced one I had missed: **`"/"`**.

`normalizeShareName` trims all leading slashes and re-adds one, returning `"/"`
for an empty or all-slash input — and the REST handler normalizes *before* it
validates. So `POST /shares {"name": ""}` produced a share named `"/"`, which
sanitizes to `""` and derives `<base>/shares` — the shares tree **itself**, the
container for every other share.

That is why the fix rejects `""`, `"."` and `".."` rather than just the dots.

## The fix

Three cases added to `ValidateShareName`, the function whose doc already says
"call this at every seam that creates a share". Both such seams — the REST
create handler and `AddShare` — already call it, and I confirmed every path that
reaches `deriveLocalStoreDir` / `deriveGCStateRoot` funnels through share
creation, so there is no bypass to patch separately.

## Impact and safety

Creating a share is an authenticated control-plane operation, so this is an
isolation bug rather than an unauthenticated exploit — a privileged caller could
put a share's blocks outside its own directory, or make two shares collide.

For an existing deployment carrying such a share, `init.go` treats a failed
`AddShare` as **warn-and-skip**, not a boot abort (only `ErrFutureFormat` /
`ErrLegacyLocalFormat` stop the daemon), so this cannot brick a running install.
I also checked no legitimate share is named `"/"`: the only hits are `"/" + name`
constructions in tests.

## Tests

- `TestValidateShareName` — there was no test for this function at all. Table
  covers the new rejections plus the pre-existing rules, and names that merely
  *contain* dots (`/my.share.v2`, `/share.`, `/...`) which must still be accepted.
- `TestAcceptedShareNamesStayInsideSharesTree` — asserts the **derived path**,
  not the validator's verdict. The validator and the sanitizer live in different
  packages, so a later change to either one alone would silently reopen this; a
  validator-only test would not catch it. This is the test that found the `"/"`
  case.

Both verified to fail against the pre-fix tree — 5 subtests fail there, 0 after.
Notably `"//.."` passes both ways, confirming it was already inert because the
sanitizer strips only one leading slash and escapes the rest.

`go test ./...` 134 packages green, `go vet` and `golangci-lint` clean.

### DIFF
--- a/pkg/controlplane/runtime/shares/share_name_isolation_test.go
+++ b/pkg/controlplane/runtime/shares/share_name_isolation_test.go
@@ -1,0 +1,52 @@
+package shares
+
+import (
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/marmos91/dittofs/pkg/metadata"
+)
+
+// TestAcceptedShareNamesStayInsideSharesTree pins the property that makes the
+// per-share data directory an isolation boundary: every share name that
+// ValidateShareName accepts must derive a directory strictly beneath
+// <base>/shares/.
+//
+// deriveLocalStoreDir builds that path as Join(base, "shares", sanitized), and
+// sanitizeShareName escapes '/' but not '.'. So the guarantee rests on the two
+// functions agreeing, and they live in different packages — a later change to
+// either one alone would silently reopen a name like ".." resolving the share's
+// directory to base. Asserting the derived path here, rather than only the
+// validator's verdict, is what catches that.
+func TestAcceptedShareNamesStayInsideSharesTree(t *testing.T) {
+	t.Parallel()
+
+	const base = "/var/lib/dittofs"
+	cfg := &fakeBlockStoreConfig{cfg: map[string]any{"path": base}}
+	sharesRoot := filepath.Join(base, "shares")
+
+	names := []string{
+		"data", "/data", "/my.share.v2", "/share.", "/...", "/a/b",
+		".", "/.", "..", "/..", "//..", "/../..", "/a/../..",
+		"", "/", strings.Repeat("a", 300),
+	}
+
+	for _, name := range names {
+		t.Run(name, func(t *testing.T) {
+			t.Parallel()
+
+			if err := metadata.ValidateShareName(name); err != nil {
+				t.Skipf("rejected before any directory is derived: %v", err)
+			}
+
+			got := deriveLocalStoreDir(cfg, name)
+			if got == "" {
+				t.Fatalf("deriveLocalStoreDir(%q) = %q, want a path", name, got)
+			}
+			if got == sharesRoot || !strings.HasPrefix(got, sharesRoot+string(filepath.Separator)) {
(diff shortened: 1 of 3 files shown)


marmos91:
Thanks — two of the three land, the third does not.

**Doc wording (`validation.go`) — fixed.** Correct, it is the sanitized
component that reaches `Join`, not the raw name. Reworded.

**Error message for `/` and `///` — fixed.** Also correct, and the better catch
of the two: those are the spellings that actually arrive, since the handler
normalizes before validating. The message now says the leading slashes are
trimmed first, so it reads accurately for slash-only input.

**Empty subtest name will panic — not correct.** `testing.T.Run` accepts an
empty name; it renders as `#00`. Checked directly rather than by reading:

```go
func TestEmptySubtestName(t *testing.T) {
    t.Run("", func(t *testing.T) { t.Log("ran with an empty name") })
}
```
```
=== RUN   TestEmptySubtestName/#00
--- PASS: TestEmptySubtestName (0.00s)
    --- PASS: TestEmptySubtestName/#00 (0.00s)
```

The suite also passed on macOS and Linux with that case present, and the earlier
Windows failure was a hardcoded POSIX base path, not a panic.

The suggestion is still worth taking on its own merit, though — `#00` says
nothing about which input failed. The case is now labelled `<empty>`, so a
failure names it. Adopted for readability, not because it would have panicked.

marmos91:
@copilot review

PR #2118, opened 2026-08-25
fix(block): bound parsed chunk offsets to int64 at the parse

marmos91:
Addresses the 15 `go/incorrect-integer-conversion` alerts from #2104.

## The defect

`ParseChunkOffset` (and its stricter sibling `ChunkOffsetFor`) parse a chunk
ID's trailing component with `strconv.ParseUint(..., 10, 64)` and return it
unchecked. Every one of the ~24 call sites then converts to `int64` to compare
against a file offset or a run boundary — so a value above `MaxInt64` arrives
there **negative**, ordering before every real offset rather than after them.

CodeQL flags the conversion sites, including six in
`sqlite/postgres/file_block_refs.go` that contain no parse of their own — it is
tracing those `uint64`s back through dataflow to this parse. That is why the
guard belongs here: it sits on every path, so one barrier covers all fifteen
rather than two dozen call sites each guarding themselves.

## Reachability — latent, not exploitable

The parsed component comes from chunk IDs **we generate** (`<payloadID>/<offset>`),
where the offset is a file offset and cannot exceed `int64`. So nothing we
produce is newly rejected, and a negative offset would be a corrupted-metadata
symptom rather than an attack. This is a robustness fix.

## A pinned test I had to change

`TestParseChunkOffset` carried `{"max-digits", "p/18446744073709551615", ..., true}`
— asserting that max-uint64 is *accepted*. My change makes it fail, so the
question is whether that pin is load-bearing.

It is not: it asserts the range of `ParseUint` itself, not a property any caller
relies on. Every caller narrows to `int64`, and an 8-exabyte file offset is not
representable in Go's file APIs regardless. The case is now the boundary pair —
`max-int64` accepted, `above-int64-max` and `max-uint64` rejected.

## Scope

- `ChunkOffsetFor` gets the same bound. Fixing only one of two functions with
  identical shape would leave the fix half-applied.
- `ChunkOffsetFor` had **no test coverage at all**; it gains one covering the
  bound plus its documented payload-boundary strictness (`p/q/42`, `pp/42`).
- The third `ParseUint` in the file keys a sort in `uint64` space and never
  narrows, so it is correct as-is and left alone.

## Testing

`go test ./...` 134 packages green, `go vet` and `golangci-lint` clean.

One caveat worth stating: CodeQL runs as GitHub's default setup, so I cannot
confirm locally that all 15 alerts clear. The code is correct either way; if any
survive, they will show a path I have not accounted for and I will follow up.

---

## Added after review: the index scan shares the ceiling

Copilot flagged that `scanChunkIndexOffsets` (badger) still parsed its `fb-file`
key suffixes unbounded, so index and parser now disagreed about what is
placeable: the scan would keep a high-bit suffix as a candidate offset, then the
covering guard would refuse the row and the offset would fall through as a
**sparse hole the reader zero-fills** — the outcome that function's contract
exists to prevent.

The seam is real and is now fixed: both parses share one ceiling, so the index
places exactly the rows the guard can resolve, and such a row is reported
unplaceable so `GetFileChunkAtOffset` refuses through the existing
`errUnplaceableRow` path.

**The hole is not a regression from this PR**, which I measured rather than
inferred — storing a row `p/9223372036854775808` and reading offset 100:

| tree | `ParseChunkOffset` | result |
| --- | --- | --- |
| develop (unbounded) | `ok=true` | `(nil, nil)` — sparse hole |
| this PR, before the scan fix | `ok=false` | `(nil, nil)` — sparse hole |
| this PR, after the scan fix | `ok=false` | `ErrManifestInconsistent` |

Unbounded, the parse produced an offset so large the covering guard's
`off >= abs` failed anyway, reaching the same fall-through. The bound changed
which line rejects the row, not the outcome. So this closes a standing gap
rather than one the offset bound introduced.

Two regression tests against the real badger path: one pins the refusal (fails
on develop), one pins that a payload with a covering row still reads normally,
since one unreadable row must not make a whole payload unreadable.

### DIFF
--- a/pkg/block/ids.go
+++ b/pkg/block/ids.go
@@ -1,6 +1,7 @@
 package block
 
 import (
+	"math"
 	"sort"
 	"strconv"
 	"strings"
@@ -10,14 +11,21 @@ import (
 // ID of the form "<payloadID>/<chunkOffset>" and returns
 // (chunkOffset, true) on success. Returns (0, false) for malformed IDs
 // (no slash, trailing slash, non-digit characters after the slash, or an
-// offset that does not fit in a uint64).
+// offset that does not fit in an int64).
+//
+// The int64 ceiling, rather than the uint64 one the parse itself permits, is
+// what callers need: every one of them converts the result to int64 to compare
+// it against a file offset or a run boundary, and a value above MaxInt64 would
+// arrive there negative and order before every real offset. Rejecting it here
+// keeps that conversion total for all callers instead of asking each to guard
+// its own.
 func ParseChunkOffset(id string) (uint64, bool) {
 	slash := strings.LastIndexByte(id, '/')
 	if slash < 0 || slash == len(id)-1 {
 		return 0, false
 	}
 	v, err := strconv.ParseUint(id[slash+1:], 10, 64)
-	if err != nil {
+	if err != nil || v > math.MaxInt64 {
 		return 0, false
 	}
 	return v, true
@@ -59,7 +67,7 @@ func ChunkOffsetFor(id, payloadID string) (uint64, bool) {
 		return 0, false
 	}
 	v, err := strconv.ParseUint(rest, 10, 64)
-	if err != nil {
+	if err != nil || v > math.MaxInt64 {
 		return 0, false
 	}
 	return v, true
--- a/pkg/block/ids_test.go
+++ b/pkg/block/ids_test.go
@@ -11,7 +11,7 @@ func TestParseChunkOffset(t *testing.T) {
 	}{
 		{"simple", "share/file/0", 0, true},
 		{"non-zero", "share/file/12345", 12345, true},
-		{"max-digits", "p/18446744073709551615", 18446744073709551615, true},
+		{"max-int64", "p/9223372036854775807", 9223372036854775807, true},
 		{"nested-slashes", "a/b/c/42", 42, true},
 
 		{"no-slash", "noslash", 0, false},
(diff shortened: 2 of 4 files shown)


marmos91:
Good catch on the seam — the inconsistency is real and I have fixed it. The
stated consequence is not, though, and the difference matters for how the commit
is described.

**The gap is real.** `scanChunkIndexOffsets` classified a suffix as unplaceable
only when `ParseUint` failed outright, so a high-bit suffix was kept as a
candidate offset while the covering guard — now bounded — refused the row. Index
and parser disagreed about what is placeable. Fixed by applying the same
`MaxInt64` ceiling in the scan, so a high-bit suffix is reported unplaceable and
`GetFileChunkAtOffset` refuses via the existing `errUnplaceableRow` path.

**But this PR did not introduce it.** I measured rather than reasoned, storing a
row with ID `p/9223372036854775808` and calling `GetFileChunkAtOffset(p, 100)`:

| tree | `ParseChunkOffset` | result |
| --- | --- | --- |
| develop (unbounded) | `ok=true` | `(nil, nil)` — **sparse hole** |
| this PR, before the scan fix | `ok=false` | `(nil, nil)` — **sparse hole** |
| this PR, after the scan fix | `ok=false` | `ErrManifestInconsistent` |

Identical before and after. Unbounded, the parse returned an offset so large the
covering guard's `off >= abs` failed anyway, reaching the same fall-through and
the same hole. So the zero-fill path was already there; the bound changed which
line rejects the row, not the outcome.

That distinction is why the commit says this closes a standing gap rather than
one the offset bound created — a fix described as repairing its own regression
sends the next reader looking for a break that was never there.

Two regression tests, both against the real badger path: one pinning the refusal
(fails on develop), one pinning that a payload with a covering row still reads
fine, since one unreadable row must not make a whole payload unreadable.

marmos91:
@copilot review

Issue #2095, opened 2026-08-24
nfs4.1: OPEN admits the wire open_owner4 clientid instead of the session's


marmos91:
Follow-up to #2063.

`handleOpen` (`internal/adapter/nfs/v4/handlers/open.go`) takes the client ID from the
`open_owner4` field on the wire and admits it, for v4.0 and v4.1 alike.

In v4.1 that field is meant to be ignored: the server takes the client ID from the
session the COMPOUND arrived on, and the opaque owner alone identifies the owner
within that client (RFC 8881 §18.16.3). A conformant client may send `clientid = 0`.

That used to be survivable — `OpenFile` did a plain table lookup and tolerated a miss.
#2063 turned the lookup into an admission check, so a client sending zero now gets
`NFS4ERR_STALE_CLIENTID` on every OPEN. Linux and macOS both fill the field in, so
nothing in our test suite sends a zero and CI stays green.

### Fix

Put the session's client ID on `CompoundContext` and admit that when
`MinorVersion >= 1`, ignoring the wire field. The v4.1 dispatch path already resolves
the session in order to handle SEQUENCE, so the ID is at hand — it just isn't passed
down. Note `ClientState.ClientID` is not it: that comes from SETCLIENTID_CONFIRM and
only exists for v4.0.

LOCK's `lock_owner4` and the owners in OPEN_CONFIRM / OPEN_DOWNGRADE read the same
field the same way, so they're worth changing in the same pass.

### Test

A v4.1 OPEN carrying `clientid = 0` over an established session, expecting success.
Every v4.1 handler test today passes a real registered ID, which is why this went
unnoticed.

To be clear, this is about dropping the dependency on the wire field — the check
#2063 added should stay.

marmos91:
CC @chessman

marmos91:
Fixed in #2146 (thanks @chessman).

PR #2146, opened 2026-08-25
fix(nfs4): key v4.1 open- and lock-owners by the session's client

chessman:
# Description

## Type of change

Fixes: https://github.com/marmos91/dittofs/issues/2095

## Changes Made

Use the session's client ID in NFS v4.1.

## Testing

## Checklist

### DIFF
--- a/internal/adapter/nfs/v4/handlers/compound.go
+++ b/internal/adapter/nfs/v4/handlers/compound.go
@@ -479,6 +479,9 @@ func (h *Handler) dispatchV41(compCtx *types.CompoundContext, tag []byte, numOps
 
 	// SEQUENCE succeeded -- set v4.1 bypass for per-owner seqid
 	compCtx.SkipOwnerSeqid = true
+	if sess != nil {
+		compCtx.SessionClientID = sess.ClientID
+	}
 
 	// Ensure slot is released and response is cached via defer
 	var responseBytes []byte
--- a/internal/adapter/nfs/v4/handlers/compound_test.go
+++ b/internal/adapter/nfs/v4/handlers/compound_test.go
@@ -12,6 +12,8 @@ import (
 	"github.com/marmos91/dittofs/internal/adapter/nfs/v4/pseudofs"
 	"github.com/marmos91/dittofs/internal/adapter/nfs/v4/types"
 	"github.com/marmos91/dittofs/internal/adapter/nfs/xdr/core"
+	"github.com/marmos91/dittofs/pkg/metadata"
+	"github.com/marmos91/dittofs/pkg/metadata/lock"
 )
 
 // buildCompoundArgs encodes a COMPOUND4args structure for testing.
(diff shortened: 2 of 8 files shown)


marmos91:
Holds up. `ValidateAndRenewClient` falls through to `v41ClientsByID`, so the EXCHANGE_ID client is admitted, and reverting the three lines makes the new test fail. Skipping OPEN_CONFIRM and OPEN_DOWNGRADE is right despite what #2095 says: neither carries an owner.

One ask: nothing covers the LOCK/LOCKT hunks or `compCtx.SessionClientID = sess.ClientID`. `createTestSession` already builds a real session, so SEQUENCE plus OPEN with `clientid = 0` then a LOCK would cover all three at once.

marmos91:
Heads-up on two things that affect this PR, neither of them a review comment.

**1. A history rewrite is planned before the next release.** `k8s/dittofs-operator/manager-linux-amd64` is an 18 MB packed blob — the largest in the repository, about 10% of what every clone downloads — and it is being removed from history with `git filter-repo` (#2186). That rewrites every commit SHA after `915deff59`, so **this branch will need to be recreated on top of the rewritten history**, and its current signature will not carry over.

Your PR is a single commit (`a147a82d2`), so this should be a cherry-pick rather than anything painful. But I did not want it to happen under you without warning. If you would rather land this first, say so and I will hold the rewrite until it merges — that is a perfectly reasonable ask and the rewrite is not urgent.

**2. `handlers/io_test.go` moved underneath you.** #2173 (NFSv4 client-ID principal binding, merged) added a variadic parameter to `testClientID` near line 327; this PR appends near line 662. Disjoint, so the rebase should be trivial — flagging it so a conflict does not come as a surprise.

The review ask from before still stands and is the only thing blocking from my side: nothing covers the LOCK/LOCKT hunks or `compCtx.SessionClientID = sess.ClientID`. `createTestSession` already builds a real session, so SEQUENCE + OPEN with `clientid = 0` + a LOCK would cover all three at once.

CC @chessman

chessman:
@marmos91 Added a test.

btw, I ran https://github.com/kofemann/pynfs and the results are:
```
NFSv4.0: 354 passed, 97 skipped, 143 fail
NFSv4.1: 105 passed,  1 skipped, 92 fail
```
Probably worth fixing.

marmos91:
Merged, thanks. I checked the new test is load bearing and not just green: reverting any one of the four production lines turns it red, each for the right reason (`SessionClientID` or the OPEN hunk gives `NFS4ERR_STALE_CLIENTID`, either lock hunk gives `NFS4ERR_DENIED`). The v4.0 subtest keeps the wire clientid honest too.

On pynfs, agreed. We do test the NFS surface today, but only for POSIX semantics: `test/posix/` runs pjdfstest over both v3 and v4 with its own known failures lists. What nothing exercises is the v4 protocol itself, which is where state, sessions, owners and locking live, and it is exactly the class of bug this PR fixed.

I opened #2204 to bring pynfs in under `test/nfs-conformance/`, shaped the way `test/posix/` and `test/smb-conformance/` already are (they share the table format and the `test/common/known-failures.sh` parser): per test verdicts, a baseline run against Linux nfsd so "no server passes this" is measured rather than assumed, then a `KNOWN_FAILURES.md` with a reason per row and the real gaps filed as their own issues. Without the baseline the known failures list just turns into somewhere to park bugs.

If you still have the run output, dropping it on #2204 would save the first pass.

Issue #2132, opened 2026-08-25
Deleting a watched directory never completes its pending CHANGE_NOTIFY (93% of smb2.notify's runtime)


marmos91:
## Summary

When a watched directory is marked for deletion, DittoFS never completes the
CHANGE_NOTIFY requests pending on it. Per [MS-FSA], those requests must be
completed with `STATUS_DELETE_PENDING`. The client instead waits until its own
transport gives up, roughly four minutes later.

This is the **seventh** instance of one shape — *the watch's subject goes away
and nothing answers it* — and the only one reached by this trigger. #2109 fixed
six of them (cancel racing registration two ways, close and teardown, the
interim-sent signal, and supersede). All six live in the notify registry or the
paths that call into it; this one does not, which is why routing more callers
through `NotifyRegistry.CloseByFileID` does not reach it.

Affects `smb2.notify.rmdir1` through `rmdir4` — four tests, each burning about
four minutes of dead air.

## What the test requires

`source4/torture/smb2/notify.c`, `torture_smb2_notify_rmdir`
("TESTING NOTIFY CANCEL FOR DELETED DIR"):

```c
req = smb2_notify_send(tree1, &(notify.smb2));
status = smb2_util_rmdir(tree2, BASEDIR_RMD);      /* delete the watched dir */
CHECK_STATUS(status, NT_STATUS_OK);
...
status = smb2_notify_recv(req, torture, &(notify.smb2));
CHECK_STATUS(status, NT_STATUS_DELETE_PENDING);
```

`rmdir1`/`rmdir2` use one tree, `rmdir3`/`rmdir4` two; the second variant of
each pair re-issues the notify first.

## Observed

Graded result (`memory` profile, no interference):

```
(../../source4/torture/smb2/notify.c:2466) Incorrect status
    NT_STATUS_CONNECTION_DISCONNECTED - should be NT_STATUS_DELETE_PENDING
```

`CONNECTION_DISCONNECTED` here is the client's transport abandoning the request
after ~4 minutes of silence, not a server crash — the server is alive and
serving other connections throughout. So this is a hang that self-terminates
rather than one that blocks forever, which is why it grades as a failure at all.

Server log at the trigger:

```
CHANGE_NOTIFY: registered watch path=test_notify_RMD ... messageID=7 asyncId=111950
Sent SMB2 response command=CHANGE_NOTIFY status=STATUS_PENDING messageID=7
[...]

Issue #2129, opened 2026-08-25
smb2.notify: only one outstanding CHANGE_NOTIFY per handle; a second replaces the first


marmos91:
## Summary

`NotifyRegistry` keys pending watches by `FileID`, so a directory handle can
carry only **one** outstanding `CHANGE_NOTIFY`. A second request on the same
handle replaces the first. SMB2 clients are entitled to keep several
outstanding, and smbtorture's `smb2.notify.double` requires exactly that.

Surfaced while fixing #2109. That PR stops the replaced request from *hanging* —
it is now completed with `STATUS_CANCELLED` instead of being dropped with no
response at all — so `notify.double` fails cleanly rather than blocking the
suite forever. This issue is the remaining capability gap, which is separate
work and deliberately **not** filed as a known failure on the strength of that.

## What the test requires

`source4/torture/smb2/notify.c`, `torture_smb2_notify_double`:

```c
smb2_util_mkdir(tree2, BASEDIR_NDOH "\\subdir-name");
req1 = smb2_notify_send(tree1, &(notify.smb2));
req2 = smb2_notify_send(tree1, &(notify.smb2));   /* same handle h1 */

status = smb2_notify_recv(req1, torture, &(notify.smb2));
CHECK_STATUS(status, NT_STATUS_OK);
CHECK_VAL(notify.smb2.out.num_changes, 1);

smb2_util_mkdir(tree2, BASEDIR_NDOH "\\subdir-name2");
status = smb2_notify_recv(req2, torture, &(notify.smb2));
CHECK_STATUS(status, NT_STATUS_OK);
CHECK_VAL(notify.smb2.out.num_changes, 1);
```

Two requests outstanding on one handle, each completing with **one** change, in
order. So the registry needs a queue of waiters per handle and one-event-per-
waiter delivery, not a single slot.

## Observed (server DEBUG log, `memory` profile)

```
CHANGE_NOTIFY: registered watch path=test_notify_NDOH ... messageID=10 asyncId=110343
Sent SMB2 response command=CHANGE_NOTIFY status=STATUS_PENDING messageID=10
NotifyRegistry: registered watch path=test_notify_NDOH ... totalWatches=1     <- messageID=10 evicted
CHANGE_NOTIFY: registered watch path=test_notify_NDOH ... messageID=9 asyncId=110344
Sent SMB2 response command=CHANGE_NOTIFY status=STATUS_PENDING messageID=9
```

[...]

Issue #2127, opened 2026-08-25
SMB2: a request's lease break can be sent after a later request on the same connection is answered


marmos91:
Split out of #2114. That issue's rename-specific half is fixed in PR #2124; this is what is left, and it is not rename-specific.

### What happens

A client pipelines two requests on one connection: a SET_INFO rename (message 8) and a CREATE (message 9). DittoFS reads them in order and then answers the *second* one before it has started the *first*:

```
SMB2 request  command=SET_INFO messageID=8      <- read first
SMB2 request  command=CREATE   messageID=9      <- read second
Dispatching   command=CREATE   messageID=9      <- dispatched first
Sent response command=CREATE   ...OBJECT_NAME_NOT_FOUND messageID=9
Dispatching   command=SET_INFO messageID=8      <- only now
SMBBreakHandler: dispatching lease break notification leaseKey=adbeed…
```

The reader hands each request to its own goroutine, so message 9's handler can run start to finish before message 8's is scheduled at all. Message 8 owed a lease break notification; the client received a response that logically comes *after* that break without the break having been sent.

### Why it is a conformance problem, not just a curiosity

Samba and Windows both give this ordering for free — smbd processes a connection's requests on one thread, so request N's side effects, break notifications included, are on the wire before request N+1 is even read. `smb2.lease.rename_wait` depends on it: it reads the break out of the following CREATE's I/O and acknowledges it, so when the break has not been sent yet it acknowledges an all-zero lease key and the server refuses it with `STATUS_INVALID_PARAMETER`.

Nothing in the rename path can fix this. PR #2124 removed the systematic handicap — a case-insensitive destination lookup that enumerated the whole directory before the break was dispatched — and after that change the break is dispatched on the very next log line after the handler is entered. `badger-fs` now passes; `memory` still fails, because the CREATE can still be answered before the rename handler starts at all. The remaining window is pure goroutine scheduling and does not close by making the rename faster.

### Scope beyond this test

Any operation that owes a break has the same exposure: CREATE-triggered breaks, WRITE-triggered breaks, SET_INFO delete-on-close, and the directory-lease breaks. `rename_wait` is simply the case that asserts on it, because it is the one that reads a break out of a later request's I/O.

### Shape of a fix

The invariant to restore is per-connection: a request's break notifications reach the wire before a later request on the same connection is answered. Options, roughly in increasing cost:

[...]

PR #2151, opened 2026-08-25
fix(smb): give buffered notify events to the earliest outstanding request

### LINKED ISSUE #2147: smb2.notify.double hangs when two CHANGE_NOTIFYs on one connection are dispatched out of order
## Summary

`smb2.notify.double` **hangs** on `develop` `1dedf7db`, depending on the order
in which the server dispatches two CHANGE_NOTIFY requests that the client sent
back to back on one connection. It was a graded `failure` before #2138 / #2133
and is now an intermittent hang, which is worse: a hang burns the suite budget
and leaves every test after it ungraded.

**Introduced by #2138 (#2133's fix).** Not to assign blame — the next person
needs to know which tree it appeared in. That PR made a CHANGE_NOTIFY arriving
with events already buffered answer synchronously instead of registering a
watch. That is correct and fixed `smb2.notify.tree`; it also made *which* of two
concurrent requests gets the buffered event decide whether the client makes
progress.

## Observed

Server DEBUG log, verbatim:

```
SMB2 request  command=CHANGE_NOTIFY messageID=9      <- req1, sent first
SMB2 request  command=CHANGE_NOTIFY messageID=10     <- req2
Dispatching   command=CHANGE_NOTIFY messageID=10     <- dispatched FIRST
CHANGE_NOTIFY: answered from buffered events without going pending ... messageID=10
Sent SMB2 response command=CHANGE_NOTIFY status=STATUS_SUCCESS messageID=10
Dispatching   command=CHANGE_NOTIFY messageID=9      <- only now
NotifyRegistry: registered watch path=test_notify_NDOH ... totalWatches=1
Sent SMB2 response command=CHANGE_NOTIFY status=STATUS_PENDING messageID=9
<nothing further — client blocked>
```

## Why it deadlocks

`torture_smb2_notify_double` (`source4/torture/smb2/notify.c`) does the mkdir
**first**, then sends both requests, then blocks on the **first** one:

```c
smb2_util_mkdir(tree2, BASEDIR_NDOH "\\subdir-name");
req1 = smb2_notify_send(tree1, &(notify.smb2));
req2 = smb2_notify_send(tree1, &(notify.smb2));

status = smb2_notify_recv(req1, torture, &(notify.smb2));   /* blocks here */
CHECK_VAL(notify.smb2.out.num_changes, 1);
smb2_util_mkdir(tree2, BASEDIR_NDOH "\\subdir-name2");      /* never reached */
status = smb2_notify_recv(req2, torture, &(notify.smb2));
```

The client does **not** wait symmetrically. It blocks on `req1` and only issues
the second mkdir once `req1` returns. When the server serves msgID 10 first,
req2 consumes the buffered event and req1 is left pending on an event the client
will never generate. Deadlock.

Whether it hangs is decided purely by which goroutine wins, so it passes some
runs and hangs others.

## This is not #2129

Worth stating because it looks like it at a glance. #2129 is *"only one
outstanding CHANGE_NOTIFY per handle; a second replaces the first"* —
`Register`'s replacement path. **Nothing was replaced here**: msgID 10 answered
synchronously and never registered, msgID 9 registered and stayed, and
`totalWatches=1` throughout. Only one watch ever existed. The defect is not that
a second request evicted a first; it is that **the wrong one was answered**.

## Root cause is shared with #2127

#2127 describes the same underlying property from a different symptom — *"The
reader hands each request to its own goroutine, so message 9's handler can run
start to finish before message 8's is scheduled at all."* That issue is about a
**break notification** owed by request N not reaching the wire before request
N+1's response; this is about **which request is serviced** when both are
eligible. Different symptom, one cause: per-connection dispatch is unordered.

Both are covered by #2127's own second fix option — *"serialize per-connection
request processing up to the point where a request either completes or goes
async (STATUS_PENDING)"* — which is what smbd does and what SMB2's async model
is built around.

**A notify-registry-local fix is not available.** To answer correctly the server
must know that an earlier MessageID on the same connection is received but not
yet dispatched. The registry cannot know that — at the moment msgID 10 is
handled, msgID 9 has not reached the registry at all. That information exists
only in the dispatcher, so this is dispatcher work, not notify work.

## Effect on the suite budget

`double` is the **14th** test in `smb2.notify`; `rmdir1`-`rmdir4` are 19th-22nd.
So a hang here ungrades everything after it, including the `rmdir` tests #2132
is about. Tests 1-13 take ~84s (`mask` alone is 81s), so under the 120s CI budget
the suite now stops at `double`, where it previously reached `rmdir1`.

Fixing #2132 alone will **not** bring `smb2.notify` back inside its budget,
because `double` hangs first. Both this and #2132 are required.

## Reproduction

```
cd test/smb-conformance/smbtorture
./run.sh --profile memory --filter smb2.notify.double --timeout 300
```

Run it several times — it is order-dependent, and a single green run means
nothing. Note single-filter mode drops the family prefix `KNOWN_FAILURES.md`
patterns match on, so grade on the `success:` / `failure:` verdict line only.

marmos91:
Fixes the **hang** #2147 names. `smb2.notify.double` no longer hangs: 5/5 runs
graded in 7-9s.

## It does NOT make `double` pass, and the suite still times out

Stating both first so nobody has to find them later.

**`double` remains order-dependent.** The hang is gone in both dispatch orders,
but the outcome still is not:

| dispatch order | before this PR | after |
| --- | --- | --- |
| earliest request wins | `success` | `success` |
| later request wins | **hang** | graded `failure` in 7-9s |

When the later request wins it now declines the buffered events (correct) and
registers — and then the earlier request registers too, which trips **#2129**:
`Register` holds one watch per FileID, so the second registration evicts the
first and completes it with `STATUS_CANCELLED`. The client fails with

```
Incorrect status NT_STATUS_CANCELLED - should be NT_STATUS_OK
(../../source4/torture/smb2/notify.c:1808)
```

My five local runs all hit the losing order; CI hit the winning one and reports
`smb2.notify.double PASS`. Both are real — that is what order-dependent means.
`double` needs #2129 as well before it passes deterministically. If you run it
locally and see a failure, a pass is equally expected — that is what
order-dependent means, and neither observation is the whole picture.

**#2129 is now reproducible deterministically in the losing order specifically**
— the later request registers, the earlier registers behind it, and `Register`
evicts one. Whoever takes #2129 can rely on that rather than hoping for the
right draw. #2129 has been updated with the evidence.

**`smb2.notify` still times out at 120s**, from this PR's own CI run:

```
smbtorture timed out after 120s on filter: smb2.notify — tests it had not
reached are UNGRADED (inconclusive)
```

A `pass` on the `smbtorture` job means no new failures outside
`KNOWN_FAILURES.md`; it does not mean the suite ran to completion. What this PR
buys is the four tests between `double` and `rmdir` — `file`, `tcp`, `rec`,
`overflow` — which a hang left ungraded and which now grade. Getting the suite
inside its budget needs #2132, which removes `rmdir1`-`rmdir4`'s ~1200s.

## The bug

A client that pipelines two CHANGE_NOTIFYs on one handle blocks on the first and
does not generate the next event until it returns:

```c
req1 = smb2_notify_send(tree1, &(notify.smb2));
req2 = smb2_notify_send(tree1, &(notify.smb2));
status = smb2_notify_recv(req1, torture, &(notify.smb2));   /* blocks here */
smb2_util_mkdir(tree2, BASEDIR_NDOH "\\subdir-name2");      /* not reached */
```

The connection read loop is sequential but hands each request to its own
goroutine, so the second could be dispatched first, take the buffered events,
and strand the first on an event that would never arrive:

```
Dispatching   CHANGE_NOTIFY messageID=10     <- dispatched first
CHANGE_NOTIFY: answered from buffered events without going pending ... messageID=10
Sent SMB2 response CHANGE_NOTIFY status=STATUS_PENDING messageID=9
<nothing>
```

## The fix

The read loop already knows the answer: it reads in wire order, so a notify
recorded there was received before any later one, however the goroutines are
then scheduled. Record CHANGE_NOTIFY arrivals per handle before dispatch, and
have a request decline the buffered events when an earlier one on the same
handle is still unanswered.

**Scoped to CHANGE_NOTIFY on the same handle.** A broader "any earlier request"
test would make a notify decline events because some unrelated operation was in
flight — which is exactly what `smb2.notify.tree` depends on not happening.

**`BeginHandleOp` is untouched.** The new bookkeeping is gated on
`hdr.Command == types.CommandChangeNotify`, so every other command's path
through the read loop is unchanged: no lock, no map, no allocation. Storage is a
slice under the registry's existing mutex, `ponytail:`-marked with the
pipelining depth that would justify an ordered structure.

## Relationship to #2127 — not addressed

Spelled out mechanically so nobody has to infer it:

- changes **which request claims already-buffered events**
- does **not** change the order requests are dispatched in
- does **not** change the order responses are written in
- involves no notification ordering; the dispatcher is unmodified

#2127 is a break notification owed by request N not reaching the wire before
N+1's response. Untouched. It stays open.

## Verification

Ran on the merged tree, `--profile memory`:

| test | runs | result |
| --- | --- | --- |
| `double` | 5 | graded `failure` in 7-9s every time — **no hang** |
| `tree` | 3 | `success` 3/3 — the regression check |
| `mask` | 1 | `success` — synchronous path undisturbed |

`double` × 5 because it is order-dependent and one green run proves nothing —
that is how this regression shipped in the first place. `tree` × 3 because the
scoping condition above is the only thing stopping this fix from trading
`double` for `tree`. `mask` because "an argument that a test cannot be affected"
is not the same object as "the test passed".

Unit test `TestHasEarlierInFlightNotify` encodes the hazard directly: read 9
then 10, assert 10 declines and 9 takes, a different handle is unaffected, and
the tracking does not leak after both release.

Closes #2147

### DIFF
--- a/internal/adapter/smb/handlers/change_notify.go
+++ b/internal/adapter/smb/handlers/change_notify.go
@@ -249,10 +249,13 @@ type NotifyRegistry struct {
 
 	// inFlightNotify tracks CHANGE_NOTIFY requests that the connection's read
 	// loop has read off the wire but whose handlers have not finished, keyed
-	// by handle. The read loop is sequential, so a lower MessageID here was
-	// necessarily received first even when its goroutine has not run yet.
-	// That is what lets a later request tell it is later. See
-	// HasEarlierInFlightNotify.
+	// by handle. The read loop is sequential and appends here in arrival
+	// order, so element 0 is the earliest-received request still outstanding.
+	//
+	// Order is the slice's, NOT the MessageID's. SMB2 clients may consume
+	// MessageIDs out of order within the credit sequence window, so a numeric
+	// comparison would call 100 "later" than 50 even when 100 arrived first.
+	// See HasEarlierInFlightNotify.
 	//
 	// ponytail: linear scan of a slice. N is one client's CHANGE_NOTIFY
 	// pipelining depth on one handle — 1 or 2 in practice — so a map costs
@@ -1609,22 +1612,22 @@ func (r *NotifyRegistry) MarkNotifyInFlight(fileID [16]byte, messageID uint64) f
 }
 
 // HasEarlierInFlightNotify reports whether another CHANGE_NOTIFY on the same
-// handle was received before messageID and is still unanswered.
+// handle arrived before messageID and is still unanswered.
 //
 // A request that answers yes must not claim buffered events: they belong to the
 // earlier request, which is the one the client is blocked on. Handing them to
 // the later request leaves the earlier one waiting for an event the client will
 // never generate, because the client does not send it until the earlier request
 // returns.
+//
+// "Earlier" means earlier on the wire, which is this slice's append order —
+// never a MessageID comparison. MessageIDs may be consumed out of order within
+// the credit sequence window, so 100 can arrive before 50.
 func (r *NotifyRegistry) HasEarlierInFlightNotify(fileID [16]byte, messageID uint64) bool {
 	r.mu.RLock()
 	defer r.mu.RUnlock()
-	for _, id := range r.inFlightNotify[string(fileID[:])] {
-		if id < messageID {
-			return true
-		}
-	}
-	return false
+	ids := r.inFlightNotify[string(fileID[:])]
+	return len(ids) > 0 && ids[0] != messageID
 }
 
 // UnregisterAllForSession unregisters all pending watchers for a session AND
--- a/internal/adapter/smb/handlers/change_notify.go
+++ b/internal/adapter/smb/handlers/change_notify.go
@@ -247,6 +247,19 @@ type NotifyRegistry struct {
 	byMsgKey  map[notifyMsgKey]*PendingNotify
 	byAsyncId map[uint64]*PendingNotify // asyncId -> pending request (for async CANCEL)
 
+	// inFlightNotify tracks CHANGE_NOTIFY requests that the connection's read
+	// loop has read off the wire but whose handlers have not finished, keyed
+	// by handle. The read loop is sequential, so a lower MessageID here was
+	// necessarily received first even when its goroutine has not run yet.
+	// That is what lets a later request tell it is later. See
+	// HasEarlierInFlightNotify.
+	//
+	// ponytail: linear scan of a slice. N is one client's CHANGE_NOTIFY
+	// pipelining depth on one handle — 1 or 2 in practice — so a map costs
+	// more than it saves. Swap for an ordered structure only if a client
+	// appears that pipelines notifies deeply enough to measure.
+	inFlightNotify map[string][]uint64
+
 	// armed tracks directory handles that have at some point issued a
 	// CHANGE_NOTIFY. Mirrors Samba `notify_buffer`: once a watcher has been
 	// "armed" on a handle, subsequent matching filesystem events MUST be
@@ -374,6 +387,7 @@ func NewNotifyRegistry() *NotifyRegistry {
 		byFileID:         make(map[string]*PendingNotify),
 		byMsgKey:         make(map[notifyMsgKey]*PendingNotify),
 		byAsyncId:        make(map[uint64]*PendingNotify),
+		inFlightNotify:   make(map[string][]uint64),
 		armed:            make(map[string]*armedHandle),
 		cancelTombstones: make(map[notifyMsgKey]time.Time),
 		closeTombstones:  make(map[string]time.Time),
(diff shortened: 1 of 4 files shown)


marmos91:
Copilot's MessageID finding is fixed in 919963ee — ordering now uses the read loop's append order, which *is* arrival order, rather than numeric MessageID comparison. Test added for the 100-before-50 case it named.

Follow-up in 67935fa4: `ids[0]` is only the earliest arrival while removal preserves order, so the invariant is now stated next to the removal and pinned by a test that releases the earliest of three and asserts the *second* arrival is promoted. Verified it fails under a swap-remove and passes under the shift — a check that cannot fail pins nothing.

marmos91:
Third comment addressed in 25c11291 — and it is reachable, not theoretical. Checked rather than assumed:

- `negotiate` advertises `CapMultiChannel` for SMB 3.0 (asserted in `negotiate_test.go`), so clients do open extra channels.
- `h.files` is a single `sync.Map` keyed by FileID alone, not per-connection, so a handle resolves on any channel of the session.

Two CHANGE_NOTIFYs on one handle can therefore arrive on different connections carrying the same MessageID. Stored bare they are indistinguishable: the second matches the first's number, concludes it is earliest, and takes events belonging to the other connection's request — the same starvation, by another route. Release was equally ambiguous about which entry to remove.

Now keyed on the `(ConnID, MessageID)` pair the cancel tombstones already use. Append order still means arrival order across connections: the read loops are separate but append under one registry lock, so `ids[0]` stays the earliest arrival globally.

Test added for two connections sharing MessageID 10 on one handle. Verified it FAILS against a build that compares `ids[0].MessageID` only, and passes with the pair — a check that cannot fail pins nothing.

Issue #2155, opened 2026-08-25
Audit MS-FSCC and MS-FSA spec citations in the SMB adapter — 66 sites cite sections that do not exist


marmos91:
Split out of #2152, which audits **MS-SMB2** citations in `internal/adapter/smb/`. While sweeping
that spec I found the same class of defect in the **MS-FSCC** and **MS-FSA** citations in the same
tree. Filing separately so #2152 stays reviewable — folding a second audit of comparable size into
it would bury the corrections.

## Exact counts

Verified against the published specs (MS-FSCC v2025, 243pp; MS-FSA v2025, 287pp; MS-ERREF v2024,
497pp; MS-DTYP v2024, 155pp), downloaded from the Microsoft Open Specifications site.

| spec | citations | distinct sections | **cited sections that do not exist** | sites |
| --- | --- | --- | --- | --- |
| MS-FSCC | 196 | 62 | **5** | **12** |
| MS-FSA | 124 | 33 | **15** | **54** |
| MS-ERREF | 7 | 3 | 0 | 0 |
| MS-DTYP | 37 | 10 | 0 | 0 |

MS-ERREF and MS-DTYP are **clean** — every cited section resolves.

**How "does not exist" was established.** Not by absence from a table of contents. My extracted TOC
for MS-FSA was visibly the weakest of the four (an early parse produced page-footer lines as
sections), so every number below is confirmed by **zero occurrences of the section string anywhere
in the complete extracted body text** of the spec. A section that is cross-referenced even once
anywhere in the document would register; these register nothing.

### MS-FSCC — 5 sections, 12 sites

| cited | sites | why it cannot be right |
| --- | --- | --- |
| `2.1.4.4` | 4 | `query_directory.go:942,954,992,1011` |
| `2.4.11.2` | 2 | §2.4.11 is `FileDispositionInformation`, no subsections — `set_info.go:1261,1269` |
| `2.4.21.2` | 4 | §2.4.21 is `FileIdAllExtdDirectoryInformation`, no subsections — `set_info.go:1580,2214,2229`, `set_info_link_info_test.go:36` |
| `2.4.51` / `2.4.51.1` | 2 | §2.4 does not reach 51 — `stub_handlers.go:1102,1127` |

### MS-FSA — 15 sections, 54 sites

The bulk is a fabricated `2.1.5.14.x` family (43 sites) and a fabricated `2.1.5.9.x` family (9 sites).

[...]

PR #2156, opened 2026-08-25
docs(smb): correct 60+ wrong MS-SMB2 spec citations in the SMB adapter

### LINKED ISSUE #2152: Audit MS-SMB2 spec citations in the SMB adapter — three found wrong in one day
## Summary

Three MS-SMB2 citations in the SMB adapter have been found wrong in one day, by two
different people, each while working on something else. A wrong spec citation is a
uniquely durable defect: it reads as settled fact, so the next reader stops thinking
rather than checking, and it gets more credible with age.

This is a bounded audit task, not a bug report.

## Confirmed wrong

**1. `internal/adapter/smb/handlers/close.go` — corrected in #2151.**
Cited "MS-SMB2 3.3.4.1 and 3.3.5.16.1" for the close-trigger `STATUS_NOTIFY_CLEANUP`
behaviour. The correct section is **3.3.4.17** ("Server Application Requests Closing an
Open"), which says verbatim that the server MUST send a CHANGE_NOTIFY response with
`STATUS_NOTIFY_CLEANUP` for any pending CHANGE_NOTIFY associated with the Open being
closed. §3.3.5.16 is "Receiving an SMB2 CANCEL Request".

**2. `NotifyRegistry.NotifyRmdir`'s doc comment — corrected in #2151.**
Cited "Per MS-SMB2 3.3.5.15" for completing a deleted directory's pending CHANGE_NOTIFY
with `STATUS_NOTIFY_CLEANUP`. §3.3.5.15 is "Receiving an SMB2 IOCTL Request". CHANGE_NOTIFY
server processing is §3.3.5.19, whose status list contains neither `NOTIFY_CLEANUP` nor
`DELETE_PENDING`. The claim was real but attached to the wrong trigger — `NOTIFY_CLEANUP`
belongs to *closing an Open* (§3.3.4.17), not to *deleting a directory*.

**3. `internal/adapter/smb/types/status.go:211` — NOT yet corrected.**

```go
// StatusFileLockConflict indicates an I/O operation (READ/WRITE) conflicts
// with an existing byte-range lock held by another session.
// Per MS-SMB2 3.3.5.15 (Read) and 3.3.5.16 (Write).
```

Both halves are wrong. Verified against the spec:

| cited | actually is | should be |
| --- | --- | --- |
| 3.3.5.15 (Read) | Receiving an SMB2 IOCTL Request | **3.3.5.12** — Receiving an SMB2 READ Request |
| 3.3.5.16 (Write) | Receiving an SMB2 CANCEL Request | **3.3.5.13** — Receiving an SMB2 WRITE Request |

## It is not a systematic offset — checked

Both wrong numbers are exactly +3 from the correct ones, which looked like a shifted
read of the spec. **That hypothesis is refuted.** `status.go:216` and `:229` cite
"MS-SMB2 3.3.5.14 (Lock)", and §3.3.5.14 *is* "Receiving an SMB2 LOCK Request" — correct.
Since READ=12, WRITE=13, LOCK=14, a uniform +3 shift would have made the Lock citation
wrong too, and it is not. So line 211 is a localized error, and the rest of the file
should be checked individually rather than mass-corrected by an offset.

## Unverified, listed rather than trusted

These were noted in passing while correcting the above and have **not** been checked.
Listing them so they are not mistaken for audited: §3.3.5.19, §3.3.5.2.5, §3.3.5.22,
§3.3.5.22.1, §3.3.5.22.2, §3.3.5.9. The ~20 other mentions of §3.3.5.15 across
`response.go`, `ioctl_dispatch.go`, `stub_handlers.go` and `ioctl_copychunk.go` are
mostly about IOCTL, which §3.3.5.15 genuinely is, and are probably fine — but "probably"
is the state this issue exists to remove.

## Suggested approach

Sweep every `MS-SMB2 x.y.z` reference in `internal/adapter/smb/`, confirm each section
number against its actual title, and either correct it or delete it. A citation nobody
can check is not documentation.

Two rules learned the hard way today, worth applying:

- **Quote the section title, not just the number.** `// Per MS-SMB2 3.3.4.17 ("Server
  Application Requests Closing an Open")` is checkable at a glance; a bare number is not.
- **A wrong statement is not a location, it is a string.** The same false claim was found
  in two separate files today, and the second was only caught because someone grepped for
  the wording rather than fixing the site they were editing.

marmos91:
Closes #2152.

Comment-only. Every changed line is a `//` comment — verified mechanically, not asserted:
`git diff -U0 | grep '^[+-]'` with comment and blank lines stripped returns nothing.

## What was checked, and how

`microsoft_docs_fetch` on the MS-SMB2 landing page gave the current published revision
(**v87.0, 2026-07-14**). I pulled that PDF and extracted the complete section-number → title map
(527 sections) plus the full 499-page body text, so every verdict here is checked against the
authoritative document rather than inferred from surrounding code — which is how the three
originally-reported wrong citations survived in the first place.

**The map was validated before being trusted 800+ times:** it reproduces all seven independently
verified titles (§3.3.4.17, §3.3.5.12–.16, §3.3.5.19) exactly. Where it disagreed with the code I
confirmed against the spec *body*, not just the table of contents.

## A parser bug looks exactly like a finding

Worth recording, because it cuts both ways and the second direction is the dangerous one.

**Pass 1 invented findings.** My first extractor reported 9 cited sections as absent from the spec.
Seven were wrapped-TOC-line artifacts of my own parser. Only three survived re-parsing and a
full-body grep. The three are credible *because* the seven were killed.

**Pass 2 hid findings.** My sweep regex was `MS-SMB2[^0-9]{0,18}?(\d+(?:\.\d+)+)`, which cannot
reach a *second* section number on the same line — the first number's digits break the `[^0-9]` run.
So `write.go`'s `§3.3.4.18 / §3.3.5.16` was counted once and its second half was invisible.
**I only noticed because a later grep returned a site my own dataset did not contain.**

**Pass 3.** Fixing that surfaced a third blind spot: bare `§N.N.N` with no spec token on the line.

## The completeness argument rests on the shape-sweep, not on the keyword passes

Three keyword passes, three parser faults, and the count grew every time — 803, +25, +67. That is
not evidence of completeness; it is only evidence that no fourth fault had been tripped over yet.
**The keyword passes demonstrably missed 93 references between them**, so they cannot support the
claim. So the sweep was redone by *shape*: every `\d+(\.\d+)+` token in `internal/adapter/smb/`,
with no requirement for `MS-`, `§`, or any keyword, then every match classified.

| category | count |
| --- | ---: |
| **Total shape-matches** (263 files) | **1861** |
| − SMB dialect versions (`3.1.1`, `2.0.2`, …) | 291 |
| − IP addresses in string literals | 162 |
| − code literals outside comments (`0.25`, `1.5`) | 51 |
| − comment prose numbers (8.3 short name, Windows 8.1, DCE 1.1) | 23 |
| − ASN.1 OIDs (`1.2.840.113554.1.2.2`) | 11 |
| **= spec citations** | **1323** |

Of those 1323: MS-SMB2 814, MS-FSCC 156, MS-FSA 121, MS-NLMP 50, MS-DTYP 41, multi-spec lines 33,
RFC 11, C706 8, MS-ERREF 7, MS-SRVS 5, others 6, and 71 written as `§`/`Section` with the spec named
on a nearby line.

**The shape-sweep found a fourth class, exactly as predicted** — 67 references the keyword passes
never saw, almost all of them second-and-later section numbers on multi-section lines, plus RFC
`Section N.N` word-form. Most resolve correct, but one real cluster fell out of it: seven citations
of **§3.2.4.1.4, a CLIENT section**, used for server response behaviour (commit 7). That cluster
would have merged undetected on the strength of "I looked three times".

## Corrections

**Three fabricated sections** — zero occurrences across all 499 pages:

| cited | correct |
| --- | --- |
| `§3.1.1.2` | **§2.2.3.1.7 "SMB2_SIGNING_CAPABILITIES"** (§3.1.1 has only §3.1.1.1; the 5 textual matches are cross-refs to MS-KILE / MS-LSAD) |
| `§3.3.4.20.13` | **deleted** — §3.3.5.15.11 does not specify the `MaxOutputResponse` case at all; smbtorture `bug14788` is the real authority |
| `§3.3.5.18.1 ¶6` | **§3.3.5.18** — the quote was also wrong twice: the ADM variable is `Share.DoAccessBasedDirectoryEnumeration`, and the spec requires `GENERIC_READ` **and** `FILE_LIST_DIRECTORY` |

**Three wrong clusters** (36 sites):

| cited | actually is | correct |
| --- | --- | --- |
| `§3.3.4.18` ×13 | "Server Application Queries a Session" (MS-SRVS `SESSION_INFO_502`) | **§3.3.7.1** persist gate / **§3.3.4.7** break-driven purge |
| `§3.3.4.20` ×13 | "Server Application Queries an Open" (MS-SRVS `FILE_INFO_3`) | **deleted** — see below |
| `§3.3.4.4` ×11 | "Sending an Error Response" | **§3.3.4.2 "Sending an Interim Response for an Asynchronous Operation"** |

**Nine individual corrections** and **four label fixes** — see the per-commit messages.

## Two findings that are not renumberings

**The parent-key suppression rule is not in MS-SMB2 at all.** I traced every `ParentLeaseKey`
occurrence in the spec: it appears in the wire structures (§2.2.13.2.10, §2.2.14.2.10), in
§3.3.5.9.11 where `Lease.ParentLeaseKey` is *established* from the CREATE, and in §3.2.4.3.8 — a
**client** section. §3.3.4.7 hands the break decision to the object store. So the 13 `§3.3.4.20`
citations have no correct replacement and are deleted, leaving Samba `dirlease_should_break` as the
binding reference, with the delegation stated once on the `ParentLeaseKey` field so nobody re-adds a
server-section number for a rule the server sections do not contain.

**There is no offset to apply.** The wrong numbers are off by +1, +3, −14 and more depending on the
site. `read.go` cited §3.3.5.13 for READ (+1) while `status.go` cited §3.3.5.15 for Read (+3) — two
different errors for the same operation. Every site was matched by the behaviour it describes.

## Verified correct — left untouched deliberately

Per review, titles were added **only to citations actually changed**; the verified-correct set is
recorded here instead, to avoid hundreds of no-op lines burying the real corrections. After this PR,
151 distinct sections across 792 sites resolve correctly. The largest:

| section | title | sites |
| --- | --- | --- |
| §3.3.5.9 | Receiving an SMB2 CREATE Request | 44 |
| §3.3.5.5.2 | Reauthenticating an Existing Session | 40 |
| §3.3.5.5.3 | Handling GSS-API Authentication | 22 |
| §3.3.4.7 | Object Store Indicates a Lease Break | 22 |
| §3.3.5.14 | Receiving an SMB2 LOCK Request | 21 |
| §2.2.13 | SMB2 CREATE Request | 20 |
| §3.3.5.5 | Receiving an SMB2 SESSION_SETUP Request | 19 |
| §3.3.5.9.7 | Handling the SMB2_CREATE_DURABLE_HANDLE_RECONNECT Create Context | 19 |
| §2.2.10 | SMB2 TREE_CONNECT Response | 17 |
| §3.3.5.16 | Receiving an SMB2 CANCEL Request | 13 |

Notably §3.3.5.16 splits: 13 genuine CANCEL citations are correct and untouched, while 6 sites used
the same number for WRITE semantics and are corrected. Bulk-retitling the correct set is a separate
mechanical change with a different risk profile.

## Out of scope, split out

The same defect class exists in the **MS-FSCC** (12 sites) and **MS-FSA** (54 sites) citations in
this tree, including a fabricated `§2.1.5.14.x` family of 43 sites. Filed as **#2155** rather than
folded in. **MS-ERREF (7 sites) and MS-DTYP (37 sites) are clean.**

`close.go`'s `3.3.4.1` / `3.3.5.16.1` pair belongs to #2151 and arrives via develop.

### DIFF
--- a/internal/adapter/smb/compound.go
+++ b/internal/adapter/smb/compound.go
@@ -155,7 +155,8 @@ func ProcessCompoundRequest(ctx context.Context, firstHeader *header.SMB2Header,
 
 	result, fileID, handlerCtx := ProcessRequestWithFileIDAndCallback(ctx, firstHeader, firstBody, firstRaw, connInfo, isEncrypted, asyncNotifyCallback)
 
-	// Per MS-SMB2 §3.3.4.4 and smbtorture compound_async.getinfo_middle:
+	// Per MS-SMB2 §3.3.4.2 ("Sending an Interim Response for an Asynchronous Operation")
+	// and smbtorture compound_async.getinfo_middle:
 	// When a compound command returns STATUS_PENDING with an AsyncId AND
 	// there are remaining commands, send the interim response standalone and
 	// defer the remaining compound commands to the async completion callback.
@@ -207,7 +208,7 @@ func ProcessCompoundRequest(ctx context.Context, firstHeader *header.SMB2Header,
 		// on the wire. The callback was swapped above (ReplaceCallback) so the
 		// goroutine picks up the continue-compound wrapper, and MarkStarted runs
 		// post-interim so the final compound response can never overtake the
-		// interim (MS-SMB2 §3.3.4.4 ordering; same class as the standalone path
+		// interim (MS-SMB2 §3.3.4.2 (interim async response) ordering; same class as the standalone path
 		// in response.go that smb2.bench.oplock1 trips). Skipping the release
 		// would deadlock the CREATE after the wait drains (smbtorture
 		// compound.compound-break IO_TIMEOUT race).
@@ -800,7 +801,7 @@ type compoundLoopState struct {
 	// AsyncIds of parked CREATEs whose interim STATUS_PENDING is buffered into
 	// responses (not yet on the wire). MarkStarted MUST fire only after the
 	// compound frame is written, so the resume goroutine's final response can
-	// never overtake the interim (MS-SMB2 §3.3.4.4; same ordering class as the
+	// never overtake the interim (MS-SMB2 §3.3.4.2 (interim async response); same ordering class as the
 	// standalone path that smb2.bench.oplock1 trips).
 	markStartedAsyncIDs []uint64
 }
@@ -1057,7 +1058,8 @@ func (s *compoundLoopState) processRemaining(
 // STATUS_PENDING. It builds a compound response containing the CREATE's final
 // result and all subsequent commands, then sends it as a single compound frame.
 //
-// Per MS-SMB2 §3.3.4.4: the CREATE's interim STATUS_PENDING was already sent
+// Per MS-SMB2 §3.3.4.2 ("Sending an Interim Response for an Asynchronous Operation"):
+// the CREATE's interim STATUS_PENDING was already sent
 // standalone by ProcessCompoundRequest. completeCompoundAfterAsyncCreate delivers
 // the remaining responses (the CREATE completion + GETINFO + CLOSE etc.) as a compound.
 //
--- a/internal/adapter/smb/compound.go
+++ b/internal/adapter/smb/compound.go
@@ -299,7 +299,7 @@ func ProcessCompoundRequest(ctx context.Context, firstHeader *header.SMB2Header,
 
 // sendCompoundResponses sends all compound responses in a single NetBIOS frame.
 //
-// Per MS-SMB2 3.3.5.2.7 — Sending Compounded Responses:
+// Per MS-SMB2 3.3.4.1.3 ("Sending Compounded Responses"):
 //   - Each non-last response is padded to 8-byte alignment
 //   - NextCommand in the header points to the next command's offset
 //   - Per MS-SMB2 3.3.4.1.1: each command is signed individually over its own
@@ -630,7 +630,7 @@ func ParseCompoundCommand(data []byte) (*header.SMB2Header, []byte, []byte, erro
 
 // VerifyCompoundCommandSignature verifies the signature of a compound sub-command.
 //
-// Per MS-SMB2 3.3.5.2.7.2 — Handling Compounded Requests:
+// Per MS-SMB2 3.3.5.2.7.2 ("Handling Compounded Related Requests"):
 // Each command in a compound request is signed individually over its own bytes
 // (from its SMB2 header to NextCommand offset, or end for the last command).
 // The signature covers ONLY that command's bytes, not the entire compound.
--- a/internal/adapter/smb/compound.go
+++ b/internal/adapter/smb/compound.go
@@ -80,8 +80,9 @@ func ProcessCompoundRequest(ctx context.Context, firstHeader *header.SMB2Header,
 		return
 	}
 
-	// Per MS-SMB2 3.2.4.1.4: compound-level credit accounting.
-	// The first command's CreditCharge covers the entire compound.
+	// Per MS-SMB2 §3.3.5.2.5 ("Verifying the Credit Charge and the Payload Size"):
+	// compound-level credit accounting. The first command's CreditCharge
+	// covers the entire compound.
 	// CreditCharge size validation is skipped for exempt commands; sequence
 	// window Consume still runs for NEGOTIATE and first SESSION_SETUP but is
 	// skipped for CANCEL — see response.go for rationale (#378).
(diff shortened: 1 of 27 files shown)


marmos91:
Addressed in e7a26cde1 — with a correction to the suggestion itself.

The `§3.3.5.12` reference was attached to the phrase *"the symmetric contract to the READ clamp"*, so it was pointing at the READ section on purpose and was correct as written. Repointing it at `§3.3.5.13` as suggested would have produced "the READ clamp (§3.3.5.13 Receiving an SMB2 WRITE Request)" — a new wrong citation, and precisely the defect class this PR exists to remove.

The underlying point stands though: a WRITE test should lead with the WRITE section. So the comment now reads:

```go
// ... is rejected with STATUS_INVALID_PARAMETER, per MS-SMB2 §3.3.5.13
// ("Receiving an SMB2 WRITE Request") — the symmetric contract to the READ
// clamp above (§3.3.5.12).
```

Both sections named, each pointing at what it actually describes.

Issue #2160, opened 2026-08-25
A reflink onto a destination that still holds local bytes serves the destination's old content


marmos91:
A whole-file reflink replaces the destination's manifest wholesale — new
`FileChunk` rows, new `FileAttr.Blocks`, new size — and never touches the
destination's own journal intervals. Those intervals are still there, still
warm, still holding the destination's pre-clone bytes.

`readAtInternal` asks the local tier first and only reconciles against the
manifest when the tier reports a hole or a cold range. A covered warm range
reports neither, so the read never reaches the new manifest: the destination
serves its **old** content, indefinitely, with nothing logged.

### Reproduced on a real engine

Journal-backed local tier, memory remote, both files carved:

```go
bs.WriteAt(ctx, src, nil, srcData, 0)   // 4 MiB, seed 1
bs.WriteAt(ctx, dst, nil, dstData, 0)   // 4 MiB, seed 2
carve(src); carve(dst)
bs.CopyPayload(ctx, src, dst, srcRefs)  // reflink src -> dst
bs.ReadAt(ctx, dst, back, 0)
```

```
RESULT: destination reads its OWN PRE-CLONE bytes (stale)
```

`bytes.Equal(back, dstData)` — not `srcData`, not zeros. The manifest says one
thing and every read says another.

### Why it has not been noticed

`cp --reflink` opens the destination `O_TRUNC`, which issues a SETATTR size=0
first; the journal truncate clears the intervals and the clone then lands over
holes, which do reconcile. The bug needs a CLONE onto a destination that still
holds bytes — `xfs_io -c "reflink ..."` over an existing file, or any client
that does not truncate first. Both protocols reach it: NFSv4.2 CLONE drives
`common.CloneWholeFile`, and `FSCTL_DUPLICATE_EXTENTS_TO_FILE` is documented as
adopting the same primitive.

Local-only shares are unaffected — `materializeLocalClone` writes real bytes
through the normal write path, which supersedes the old intervals by version.

### Where

`engine.CopyPayload` (`pkg/block/engine/readwrite.go`) writes the destination's
rows; `common.CloneWholeFile` / `common.CopyPayload`
(`internal/adapter/common/`) persist `FileAttr.Blocks` and invalidate the read
cache. Nothing in that sequence invalidates the destination's journal intervals,
and the read cache is not the tier that serves these bytes.

### Shape of a fix

The destination's local ranges for the replaced span have to stop being
[...]

Issue #2110, opened 2026-08-24
OfflineReadiness reports Safe() for a range whose interval was lost, not evicted


marmos91:
`OfflineReadiness` is built from `ColdExtents`, which counts an interval only when `iv.cold` is set:

```go
for _, iv := range fi.ivs {
    if !iv.cold {
        continue
    }
    bytes += iv.length
    extents++
}
```

`Safe()` is then `Known && RemoteOnlyBytes == 0`.

That makes the check blind to a range that is **missing** rather than cold. A lost cold marker leaves no interval at all: it contributes zero, `RemoteOnlyBytes` stays `0`, `Known` stays `true`, and `Safe()` returns `true` — green-lighting exactly the detach that cannot serve those bytes.

The docstring claims more than the mechanism delivers:

> Zero remote-only bytes is a provable offline-safe share.

It is provable only against ranges the index still describes. A range the index has forgotten is indistinguishable from a range that was never there.

### How this was found

#2084 produced a concrete instance. `Invalidate` set `cold` in memory without appending to `cold.log`; the reclaim paths then unlinked the segment holding the only copy, and after a restart `applyColdLog` had nothing to replay. The interval was gone, not cold — so the share reported `RemoteOnlyBytes: 0` and `Safe() == true` while the range read back as a hole.

**#2103 fixes that loss vector**, so this specific route closes with it. The detection gap does not: it is structural, and any future path that drops a marker is invisible to the check the same way. A guard built on markers cannot see a missing marker.

### Why it is worth tracking separately

This is the shape from [a guard that has never refused anything is unverified] — the check accrues confidence in proportion to how long it is silently useless, and it is most likely to be wrong precisely on the shares whose data is most at risk. `Known == false` was designed to stop zero being read as safe when residency is unknown; the case here is worse, because residency is confidently *believed* and wrong.

### Possible directions

Not proposing an implementation — noting that the honest answers likely come from outside the interval index, since that is the thing being doubted:

- Cross-check against the manifest: reconcile the ranges the index claims to cover against the `FileChunk` rows for the same file, and report a shortfall as indeterminate rather than safe.
- Tie into the manifest audit (`manifest_repair.go`) — related to #2076, which records that the audit never runs on its own.
- At minimum, soften the docstring so the guarantee matches the mechanism.

### Related

- #2084 / #2103 — the loss vector that exposed it
- #2076 — the audit that could supply the cross-check never runs unprompted
[...]

PR #2161, opened 2026-08-25
fix(block): seed the local tier for a server-side copy's destination

### LINKED ISSUE #2123: A server-side copy leaves its destination invisible to the local tier's residency accounting
`CopyPayload` (`pkg/block/engine/readwrite.go`) — the block-store side of NFSv4.2
CLONE and SMB server-side copy — writes one `FileChunk` row per source block for
the destination payload and creates **no interval in the local journal tier** for
any of them. The clone's bytes are the source's, already on the remote, and a
read of the destination faults them in through the hole-reconcile path, so reads
are correct.

What is missing is the local tier's account of them. Verified on a real engine
(`TestOfflineReadiness_CloneIsIndeterminate`, added in #2121): after cloning
4 MiB, `local.DataExtents(dst, size)` returns zero extents while the manifest
places the full 4 MiB.

That leaves the destination's ranges invisible to everything that reasons about
residency from the index:

- `OfflineReadiness` cannot classify them. Since #2121 it cross-checks the index
  against the manifest, and a clone is indistinguishable there from a range whose
  interval was lost, so the share reports **indeterminate** instead of "N bytes
  remote-only". The verdict is not wrong — those bytes do need the remote — but
  it is the coarse answer, and it lands on any remote-backed share holding a
  clone nobody has read back yet.
- The remote-only byte count an operator would act on (`dfsctl share show`,
  the offline metrics) does not include them.

### Suggested fix

Seed cold intervals for the destination extents as part of the copy — the same
thing `SeedColdFromManifest` does for a tier that opened empty over remote data.
`engine.SeedColdBatch` already exists and takes exactly this shape.

The ordering needs care rather than a one-liner: `CopyPayload` runs inside the
caller's metadata transaction, and the journal write is not part of it. Seeding
before the transaction commits would leave cold intervals pointing at rows that a
rolled-back clone never created — a read would then try to hydrate a chunk the
manifest does not have. The seed belongs after the commit, or behind something
that reconciles if the commit does not land.

### Effect on #2121

Once this lands, a clone reports as a remote-only count and the manifest
shortfall is left meaning a genuine loss — which is the discrimination #2110
asked for. `TestOfflineReadiness_CloneIsIndeterminate` is written to fail when
the destination gains index extents, so it is the marker for updating that
expectation.

### Related

- #2110 / #2121 — the cross-check that surfaced this
- #1784 — CLONE reads zeros on local-only shares, the same missing-interval seam
  seen from the read side

marmos91:
A server-side copy hands the destination the source's chunk refs and moves no
bytes. `engine.CopyPayload` writes one manifest row per source block for the
destination and creates no interval in the local tier for any of them, so the
destination's ranges exist in the manifest and nowhere in the index.

Reads were never affected — a hole on a remote-backed share reconciles against
the manifest and hydrates exactly as a cold range does (`readAtInternal`:
`st.Cold || (st.Hole && bs.HasRemoteStore())`). What was affected is everything
that reasons about residency from the index: the remote-only byte count an
operator acts on, and `OfflineReadiness`, which since #2121 cross-checks the
index against the manifest and could only report **indeterminate** for a share
holding a copy nobody had read back — indistinguishable there from a range whose
interval was lost.

## Premise

Confirmed on a real engine before any change. On unmodified `develop`,
`TestOfflineReadiness_CloneIsIndeterminate` (added in #2121) passes with:

```
local index does not describe 4194304 bytes in 1 ranges the manifest places,
so whether they are remote-only or lost cannot be determined
```

after cloning 4 MiB: `local.DataExtents(dst, size)` returns zero extents while
the manifest places the full 4 MiB.

## The fix

The copy now seeds cold intervals for the destination's ranges, **after** its
metadata transaction commits.

- `engine.Store.SeedColdRefs` derives the extents a payload's `[]ChunkRef`
  places and hands them to the local tier. It has nothing to do on a share with
  no remote — a cold range there has nowhere to hydrate from and would fail its
  reads closed where the same range left as a hole reads as the zeros it is, and
  a local-only copy materializes real bytes the index already describes.
- `internal/adapter/common` calls it post-commit from both copy helpers
  (`CloneWholeFile`, which NFSv4.2 CLONE drives, and the `CopyPayload` sibling),
  through one shared `seedClonedRanges`.

**The ordering is the whole difficulty, not a detail.** Seeding inside the
transaction — the obvious place, since `engine.CopyPayload` is where both
callers converge — is wrong: a rolled-back copy would leave cold intervals over
rows it never wrote, and a read of the destination's sparse ranges would then
fail closed where it should serve zeros. So the engine primitive stays
interval-free by construction and its doc comment says so; the seed is the
caller's post-commit step. A crash in the window between commit and seed lands
in exactly today's state — the coarse answer, not a wrong one.

A seed failure is logged and the copy still succeeds. The copy is durable and
its reads are correct either way; failing a committed copy to report an
accounting gap would be the worse trade.

## The seed had to become live-file-safe first

`journal.SeedColdBatch` carries a `ponytail:` note naming this exact condition:
it takes an entry's version while planning, drops the shard lock for the fsync,
and inserts after — so a `Delete` landing in that gap sweeps the index above the
pending entry and the insert recreates the range it buried. Its two callers seed
a share that is not serving yet (share add) or one whose local tier was just
reset (snapshot restore). *"Build the watermark if a seed ever runs against a
live share."*

A copy destination is a live file, so this would have been the first such
caller. Rather than build the watermark, `journal.Store.SeedCold` — previously a
one-line delegation to the batch, with no production caller — is now the
single-file variant that holds the shard lock across its own fsync, the same
trade `Invalidate` already takes for the same reason. A concurrent `Delete` or
`Truncate` then lands entirely before the seed (which plans against what it
left) or entirely after (where its version fences the seeded intervals away like
any other write). The batch keeps its split and its note, now scoped to batches.
The planning loop is shared between the two.

## Effect on #2121 / #2110

A copy now reports as a remote-only count, and a manifest shortfall is left
meaning a genuine loss — the discrimination #2110 asked for.
`TestOfflineReadiness_CloneIsIndeterminate` was written to fail when the
destination gains index extents; it is replaced by
`TestOfflineReadiness_CloneIsRemoteOnly`, which pins both halves: the copy alone
still describes nothing (it is a manifest operation), and the seed is what gives
the index its account. The store is closed and reopened over the same journal
before the readiness answer, so the assertion also proves the seed reached the
cold log rather than only memory.

That engine test also reads the destination back and compares bytes, so the
"seeding changes no read" claim is asserted rather than argued.
`TestCloneWholeFile_SeedsDestinationRanges` and
`TestCloneWholeFile_SelfCloneSeedsNothing` pin the same thing at the adapter seam
a protocol CLONE actually goes through; the first fails on `develop` with
`the index describes 0 bytes of the clone's destination, want 10240`.

## Residue

Fix-forward only. Destinations copied before this change keep no intervals and
keep reporting indeterminate until something reads them back or the share is
re-seeded from its manifest; no detection or repair path is added for them.

## Found on the way, not fixed here

A reflink onto a destination that **still holds local bytes** serves the
destination's pre-clone content: the copy replaces the manifest and never
invalidates the destination's journal intervals, and a covered warm range never
reaches the manifest on read. Reproduced on a real engine and filed as **#2160**.

It is pre-existing and untouched by this PR — the seed only fills holes, so it
skips exactly the covered ranges that go stale. The two compose: once the
destination's intervals stop being authoritative, the whole cloned span becomes
a hole and this seed covers all of it.

Closes #2123

### DIFF
--- a/pkg/block/engine/flush.go
+++ b/pkg/block/engine/flush.go
@@ -2,6 +2,7 @@ package engine
 
 import (
 	"context"
+	"math"
 
 	"github.com/marmos91/dittofs/pkg/block"
 	"github.com/marmos91/dittofs/pkg/block/journal"
@@ -142,6 +143,66 @@ func (bs *Store) SeedColdBatch(ctx context.Context, seeds []ColdSeed) error {
 		js = append(js, journal.ColdSeed{ID: journal.FileID(sd.PayloadID), Extents: sd.Extents})
 	}
 	return cs.SeedColdBatch(ctx, js)
+}
+
+// SeedColdRefs gives the local tier an account of the ranges a payload's chunk
+// refs place, marking each one remote-durable-but-not-local. A server-side copy
+// hands the destination the source's refs without moving a byte, so the
+// destination's ranges exist in the manifest and nowhere in the index; this is
+// what puts them there, and it is the caller's post-commit step — seeding before
+// the manifest lands would describe rows a rolled-back copy never wrote.
+//
+// It changes no read: a hole on a remote-backed share already reconciles against
+// the manifest and hydrates exactly as a cold range does. What it changes is
+// everything that reasons about residency from the index — the remote-only byte
+// count an operator acts on, and OfflineReadiness, which can tell a copy nobody
+// has read back from a range whose interval was lost only once the copy has one.
+//
+// Only the parts of each ref the index does not already describe are seeded, so
+// a copy over a destination that still holds local bytes leaves those alone, and
+// a repeat call costs nothing. Refs with no hash place no bytes (a sparse hole
+// carries no chunk) and are skipped, as are empty ones.
+//
+// One file at a time, under its shard lock, because the destination is live —
+// see journal.Store.SeedCold.
+func (bs *Store) SeedColdRefs(ctx context.Context, payloadID string, refs []block.ChunkRef) error {
+	if err := bs.enter(); err != nil {
+		return err
+	}
+	defer bs.closeMu.RUnlock()
+	// Nothing to record on a share with no remote: a cold range there has
+	// nowhere to hydrate from and fails its reads closed, where the same range
+	// left as a hole reads as the zeros it is. Local-only copies materialize real
+	// bytes into the destination's own journal, which describes them already, so
+	// there is no work here rather than work being refused — the same shape as
+	// the no-remote-hydration case below.
+	if !bs.HasRemoteStore() {
+		return nil
+	}
+	type coldSeeder interface {
+		SeedCold(ctx context.Context, id journal.FileID, extents [][2]int64) error
+	}
+	cs, ok := bs.local.(coldSeeder)
+	if !ok {
+		return nil
+	}
+	extents := make([][2]int64, 0, len(refs))
+	for _, r := range refs {
+		if r.Size == 0 || r.Hash.IsZero() {
+			continue
+		}
+		// The index addresses its ranges in int64, so a ref out past that could
+		// only be described as a negative offset. Skipping it leaves the range a
+		// hole, which still reconciles against the manifest on read.
+		if r.Offset > math.MaxInt64-uint64(r.Size) {
+			continue
+		}
+		extents = append(extents, [2]int64{int64(r.Offset), int64(r.Size)})
+	}
+	if len(extents) == 0 {
+		return nil
+	}
+	return cs.SeedCold(ctx, journal.FileID(payloadID), extents)
 }
 
 // RestoreToVersion rewinds the local journal to a snapshot's version watermark
--- a/internal/adapter/common/clone.go
+++ b/internal/adapter/common/clone.go
@@ -6,6 +6,8 @@ import (
 	"io"
 	"time"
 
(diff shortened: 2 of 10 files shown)


pkg/block/engine/flush.go
```diff
@@ -144,6 +145,66 @@ func (bs *Store) SeedColdBatch(ctx context.Context, seeds []ColdSeed) error {
...
+		return err
+	}
+	defer bs.closeMu.RUnlock()
+	// Nothing to record on a share with no remote: a cold range there has
+	// nowhere to hydrate from and fails its reads closed, where the same range
+	// left as a hole reads as the zeros it is. Local-only copies materialize real
+	// bytes into the destination's own journal, which describes them already, so
+	// there is no work here rather than work being refused — the same shape as
+	// the no-remote-hydration case below.
+	if !bs.HasRemoteStore() {
```
marmos91, line 179: Both fixed in ec0270300.

The local-only branch was always a no-op by design and the comment overstated it; it now says there is no work rather than work being refused, and the PR body is corrected to match. The warning is now prefixed `server-side copy:` and the helper's doc says both reflink helpers call it.

PR #2173, opened 2026-08-26
fix(nfs4): bind the client ID to the authenticating principal where the RFCs require it

### LINKED ISSUE #2096: nfs4: client ID is not bound to the authenticating principal
Follow-up to #2063, from @chessman's note there:

> it doesn't validate client ID actually belongs to authenticated user (bound
> principal), so there still is a gap where one user can use the client ID of
> another user.

#2063 made OPEN reject a client ID that was never registered, never confirmed, or
whose lease has lapsed. It doesn't check that the ID belongs to whoever authenticated
the RPC. So a client that learns a live client ID can act as that client: open its
files, take or break its share reservations and byte-range locks. On an AUTH_UNIX
export the ID isn't even secret — it's in the clear in every OPEN the real client
sends.

The spec binds a client ID to the principal that created it, at SETCLIENTID for v4.0
(RFC 7530 §9.1.1) and EXCHANGE_ID for v4.1, and expects `NFS4ERR_PERM` when a state
operation arrives under a different principal.

### Fix

Store the authenticating principal on `ClientRecord` / `V41ClientRecord` when the
client registers, then compare it in `StateManager.ValidateAndRenewClient`. #2063 made
that function the one place every OPEN passes through, so the check has a single home
rather than one copy per handler.

The open question is what to compare for AUTH_UNIX. For `sec=krb5` it's the RPCSEC_GSS
principal and straightforward. AUTH_UNIX only gives a uid, which any client can claim,
so it's worth deciding whether comparing it buys anything real or just makes the gap
harder to see.

### Test

Register a client under one credential, then send an OPEN with that client ID under a
different one and expect `NFS4ERR_PERM`. The `sec=krb5` mount job already in CI is a
place to exercise the Kerberos side.

marmos91:
Closes #2096.

## The issue's premise holds; its proposed fix does not

The symptom is real and reproduces. `TestOpen_ClientIDSpansPrincipals` shows a caller
authenticated as uid 1001 presenting uid 1000's clientid and open-owner and being handed
uid 1000's live open stateid — `StateManager.OpenFile` returns the *existing* `OpenState`
when the owner key already has that filehandle open, bumping only the seqid. From there
CLOSE releases the victim's open state and its share reservations.

The fix the issue proposes — store the principal on `ClientRecord` / `V41ClientRecord`,
compare it in `ValidateAndRenewClient`, return `NFS4ERR_PERM` — is wrong in three separate
ways, and shipping it would have broken every multi-user mount on day one:

**A client ID is per-client, not per-principal.** RFC 7530 §3.1 has one TCP connection
multiplexing every user on a machine; §9.1.1 gives each of them their own open-owner under
the one client ID. RFC 8881 §2.10.8.3 lists "the physical client has multiple users, but
the client implementation has a unique client ID for each user" as one of only three
scenarios for `SP4_MACH_CRED` — i.e. the default assumption is one client ID shared by
many principals. The Linux client is built exactly this way: `nfs4_get_clid_cred()` returns
the machine credential for SETCLIENTID/EXCHANGE_ID, while `nfs4_run_open_task()` and
`_nfs4_do_setlk()` send OPEN and LOCK under `data->owner->so_cred`, and state owners are
looked up *by credential*.

**No production server enforces it on OPEN.** In Linux nfsd, `same_creds()` is called from
exactly six places — `nfsd4_exchange_id`, `nfsd4_create_session`, `replay_matches_cache`,
`nfsd4_setclientid`, and three sites in `nfsd4_setclientid_confirm`. Zero on OPEN or LOCK;
`struct nfs4_stateowner` has no credential field at all. `nfs4_op_open.c` and
`nfs4_op_lock.c` in nfs-ganesha contain no credential comparison either, and illumos'
`creds_ok()` is a literal `return (TRUE);`. RFC 8881 §18.35.3 warns clients off putting
filehandle-carrying operations under principal enforcement, "because the server's access
policies MAY conflict with the client's choice, and thus the client would then be unable
to access a subset of the server's namespace" — OPEN and LOCK carry filehandles.

**`NFS4ERR_PERM` is not the code.** RFC 7530 §13.1.6.2 scopes it to "the requester is not
the owner" of the *file*, and Table 12 lists it as valid only for CREATE, OPEN and SETATTR.
The code for a principal that may not touch a piece of state is `NFS4ERR_WRONG_CRED`
(RFC 8881 §15.1.6.4), which does not exist in NFSv4.0 and is not valid for OPEN in v4.1.

What *does* protect the file in the reproduction is the per-OPEN permission check: the same
foreign caller asking for `SHARE_ACCESS_BOTH` on a file it cannot write is refused with
`NFS4ERR_ACCESS` before any of this. The test pins that, because it is the load-bearing
block and nothing else stands behind it.

## What is actually missing, and is fixed here

Two spec-mandated principal bindings, both absent:

**RENEW had no principal check at all** (`RenewLease`, `manager.go`). RFC 7530 §16.28.5:

> The client that issues RENEW MUST choose the principal, RPC security flavor, and, if
> applicable, GSS-API mechanism and service via one of the following algorithms:
> - The client uses the same principal [...] used when the client ID was established via
>   SETCLIENTID_CONFIRM.
> - The client uses any principal [...] that currently has an OPEN file on the server.
>
> The server MUST reject a RENEW that does not use one of the aforementioned algorithms,
> with the error NFS4ERR_ACCESS.

Both algorithms are implemented — the second is what keeps a multi-user mount working, so
`OpenOwner` now records the principal that created it and `renewPrincipalAllowed` scans
them. One caller the RFC does not name is admitted: a record with no stored principal,
established under AUTH_NONE, where there is nothing to compare against. The check runs
*before* the renewal, so a refused RENEW leaves the lease exactly where it was.

Root gets no exemption. FreeBSD's `nfsrv_notsamecredname()` has one ("allow the same uid or
root", reached only on its non-GSS path), and I copied it before working out that it does not
survive being lifted out of that structure here: `Principal()` renders a GSS-resolved uid and
a client-asserted AUTH_SYS uid identically, and RENEW carries no filehandle for an export's
`sec=` policy to judge — so the exemption handed any AUTH_SYS peer writing `uid=0` into its
credential the lease of a client established under Kerberos, which is precisely the case
where this check has teeth. Nothing needs it: the Linux client renews under the same machine
credential it established the client ID with (`nfs4_get_clid_cred()` and
`nfs4_get_renew_cred()` both return `rpc_machine_cred()`), so plain equality already matches,
and the documented fallback to a state owner's credential is covered by the open-holder scan.
`root_is_refused_against_another_principal's_record` pins it.

**The SETCLIENTID / EXCHANGE_ID takeover guards failed open on an identity-less caller.**
All three sites tested `stored != "" && incoming != "" && stored != incoming`, so a request
carrying no principal matched everything. A client ID established under Kerberos could be
taken over by an AUTH_NONE caller sending SETCLIENTID with a different verifier, which then
evicts the previous confirmed record and every open, lock and delegation hanging off it.
Neither SETCLIENTID nor SETCLIENTID_CONFIRM carries a filehandle, so an export's `sec=`
policy never sees the request and cannot refuse it first. RFC 7530 §19 calls the check on
these two operations "imperative"; §9.1.1 states it as a MUST NOT on cancelling another
principal's leased state; RFC 8881 §18.35.4 case 9 is the same rule for EXCHANGE_ID. The
three sites now share `principalHijacks`, which no longer treats "no credential offered" as
a match. Records with no stored principal stay open — there is nothing to check against.

### The RENEW check does not make a lease unreachable, and is not meant to

RFC 7530 §9.5 grants implicit renewal to five other operations — "Any operation made with a
valid clientid (DELEGPURGE, LOCK, LOCKT, OPEN, RELEASE_LOCKOWNER, or RENEW) [...] informs the
server to renew all of the leases for that client" — none of which carries a principal
restriction, and `ValidateAndRenewClient` implements that on the OPEN path today. So a caller
refused at RENEW can still hold a lease open by sending an OPEN that fails afterwards. That is
the RFC's own design, not a hole this PR leaves: §16.28.5's MUST is scoped to RENEW alone, and
FreeBSD reads it the same way — `nfsrv_notsamecredname()` is reached only via `CLOPS_RENEWOP`,
passed from exactly one place, `nfsrvd_renew()`. The value here is conformance with a MUST that
was simply absent, not a new perimeter.

### What this is and is not worth under AUTH_SYS

`Principal()` is `"uid:N"`, so under AUTH_SYS the identity is whatever the client claimed and
an attacker who knows the victim's uid can still forge it. The binding has real teeth under
`sec=krb5`, where the uid comes from a Kerberos principal the server verified; under AUTH_SYS
it closes the identity-less case and the accidental one, and that is all it can close — which
is also why no AUTH_SYS uid, root included, gets a standing exemption. That is a property of
AUTH_SYS, not of this check — RFC 7530 §19's answer to it is RPCSEC_GSS. The
GSS principal string itself is still not threaded onto `CompoundContext`, so two Kerberos
principals mapping to one uid are one identity here; threading it is a separate change.

## Evidence

Every new assertion was run against a deliberately broken build first:

| Break | Test that went red |
| --- | --- |
| RENEW principal check removed | `stranger_is_refused`, `refused_renew_does_not_extend_the_lease` |
| "any principal holding an OPEN" alternative removed | `principal_holding_an_open_is_allowed` |
| uid 0 exemption re-added | `root_is_refused_against_another_principal's_record` |
| `record.Principal == ""` clause removed | `record_with_no_principal_accepts_any_caller` |
| fail-open on empty incoming principal restored | `TestSetClientID_IdentitylessCallerCannotTakeOver`, `TestExchangeID_Case2_PrincipalMismatch` |
| per-OPEN permission check neutered | `TestOpen_ClientIDSpansPrincipals` |

`TestExchangeID_Case2_PrincipalMismatch` changes: its last assertion pinned the fail-open
(an idempotent EXCHANGE_ID with no principal succeeding). It now asserts the refusal. Its
original point — that such a call must never *clear* the stored principal — is kept.

`go build ./... && go vet ./... && gofmt -l` clean; `go test ./...` green.

## Deliberately not in this PR

- **`SP4_MACH_CRED` / `SP4_SSV`** (RFC 8881 §2.10.8.3) are the spec's actual answer to
  "an attacker forges a client ID and destroys the victim's locking state". They are
  rejected as out of scope at `v41/handlers/exchange_id.go:28`, and implementing them is a
  feature, not a bug fix. nfs-ganesha does not implement them either.
- **The open-owner handoff itself** is left as-is: refusing it is what no other server does
  and what §18.35.3 advises against, and the per-OPEN permission check is what bounds it.
- Auditing the client-ID plumbing turned up a wider unauthenticated surface that is not
  principal-related and belongs in its own issues — `BIND_CONN_TO_SESSION` binds any
  connection to any known sessionid with no check at all (`manager.go` `BindConnToSession`),
  `DESTROY_CLIENTID` fails open for an unidentifiable caller, a v4.1 COMPOUND can reach
  OPEN/LOCK/CLOSE/LOCKU without SEQUENCE when the first op is session-exempt, and
  `types.V4ClientState.ClientID` is read in two places but never written. I have filed
  nothing for these; say the word and I will.

## Relationship to #2146

Nothing here changes a line #2146 changes. #2146 changes which *field* the clientid is read
from in v4.1 (`open.go` around line 103, plus `lock.go`) and adds
`CompoundContext.SessionClientID`; this PR touches `RenewLease`, the SETCLIENTID/EXCHANGE_ID
guards, and the four `OpenFile` call sites further down `open.go`. Branched off
`origin/develop`, not off #2146.

The one file both touch is `handlers/io_test.go`: #2146 appends tests near line 662, this PR
adds a variadic `principal` parameter to `testClientID` near line 327. Whichever lands second
gets a trivial rebase; happy for that to be me.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/client_principal_test.go
+++ b/internal/adapter/nfs/v4/handlers/client_principal_test.go
@@ -9,26 +9,8 @@ import (
 	xdr "github.com/marmos91/dittofs/internal/adapter/nfs/xdr/core"
 )
 
-// principalClientID registers a confirmed v4.0 client whose record carries
-// principal, the identity SETCLIENTID_CONFIRM was issued under.
-func principalClientID(t *testing.T, sm *state.StateManager, name, principal, addr string) uint64 {
-	t.Helper()
-
-	result, err := sm.SetClientID(name, [8]byte{1, 2, 3, 4, 5, 6, 7, 8}, state.CallbackInfo{
-		Program: 0x40000000,
-		NetID:   "tcp",
-		Addr:    "127.0.0.1.8.1",
-	}, addr, principal)
-	if err != nil {
-		t.Fatalf("SetClientID(%s): %v", name, err)
-	}
-	if err := sm.ConfirmClientID(result.ClientID, result.ConfirmVerifier); err != nil {
-		t.Fatalf("ConfirmClientID(%s): %v", name, err)
-	}
-	return result.ClientID
-}
-
-func openResult(fx *ioTestFixture, ctx *types.CompoundContext,
+// doOpen drives one OPEN through the handler and returns its raw result.
+func doOpen(fx *ioTestFixture, ctx *types.CompoundContext,
 	seqid uint32, clientID uint64, owner []byte, filename string, openType, shareAccess uint32,
 ) *types.CompoundResult {
 	args := encodeOpenArgs(seqid, shareAccess, types.OPEN4_SHARE_DENY_NONE,
@@ -43,7 +25,7 @@ func openStateid(t *testing.T, fx *ioTestFixture, ctx *types.CompoundContext,
 ) types.Stateid4 {
 	t.Helper()
 
-	result := openResult(fx, ctx, seqid, clientID, owner, filename, openType, shareAccess)
+	result := doOpen(fx, ctx, seqid, clientID, owner, filename, openType, shareAccess)
 	if result.Status != types.NFS4_OK {
 		t.Fatalf("OPEN status = %d, want NFS4_OK", result.Status)
 	}
@@ -80,7 +62,7 @@ func TestOpen_ClientIDSpansPrincipals(t *testing.T) {
 	fx := newIOTestFixture(t, "/export")
 	sm := fx.handler.StateManager
 
-	clientID := principalClientID(t, sm, "shared-client", "uid:1000", "10.0.0.1:1000")
+	clientID := testClientID(t, sm, "shared-client", "uid:1000")
 	owner := []byte("open-owner-1000")
 
 	// uid 1000 establishes the client ID and opens a file it owns.
@@ -92,13 +74,15 @@ func TestOpen_ClientIDSpansPrincipals(t *testing.T) {
 		t.Fatalf("ConfirmOpen: %v", err)
 	}
 
+	// A second user of the same client, under its own credential.
 	other := newRealFSContext(1001, 1001)
-	other.ClientAddr = "10.0.0.9:9999"
 	setCurrentFH(other, fx.rootHandle)
 
 	// The permission check is load-bearing: uid 1001 cannot write the file, and
 	// naming uid 1000's client ID and owner does not get it write access.
-	if status := openResult(fx, other, 3, clientID, owner, "victim.txt",
+	// checkOpenAccess refuses before the state layer sees the request, so seqid
+	// 3 is still unused when the read OPEN below presents it.
+	if status := doOpen(fx, other, 3, clientID, owner, "victim.txt",
 		types.OPEN4_NOCREATE, types.OPEN4_SHARE_ACCESS_BOTH).Status; status != types.NFS4ERR_ACCESS {
 		t.Errorf("OPEN for write by a uid without write permission: status = %d, want NFS4ERR_ACCESS", status)
 	}
@@ -119,7 +103,7 @@ func TestOpen_ClientIDSpansPrincipals(t *testing.T) {
 func TestRenew_PrincipalBinding(t *testing.T) {
 	t.Run("establishing principal is allowed", func(t *testing.T) {
 		fx := newIOTestFixture(t, "/export")
-		clientID := principalClientID(t, fx.handler.StateManager, "c", "uid:1000", "10.0.0.1:1000")
+		clientID := testClientID(t, fx.handler.StateManager, "c", "uid:1000")
 
 		if status := renewStatus(fx, newRealFSContext(1000, 1000), clientID); status != types.NFS4_OK {
 			t.Errorf("RENEW by the establishing principal: status = %d, want NFS4_OK", status)
@@ -128,7 +112,7 @@ func TestRenew_PrincipalBinding(t *testing.T) {
 
 	t.Run("stranger is refused", func(t *testing.T) {
 		fx := newIOTestFixture(t, "/export")
-		clientID := principalClientID(t, fx.handler.StateManager, "c", "uid:1000", "10.0.0.1:1000")
+		clientID := testClientID(t, fx.handler.StateManager, "c", "uid:1000")
 
 		if status := renewStatus(fx, newRealFSContext(1001, 1001), clientID); status != types.NFS4ERR_ACCESS {
 			t.Errorf("RENEW by a stranger: status = %d, want NFS4ERR_ACCESS", status)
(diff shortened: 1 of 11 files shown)


marmos91:
Copilot's log-noise point was right, and it caught a duplicate I had added: the refusal was logged in `RenewLease` *and* again by `handleRenew`, so every refused RENEW produced two Info lines. Removed the one in `RenewLease` (c9a0d65) — `handleRenew`'s existing line already carries the client ID and the error text ("renew principal not permitted for this client ID").

I left `handleRenew`'s level at Info rather than dropping it to Debug. It is pre-existing and covers every RENEW failure, not just this one, so changing it would quietly change how STALE_CLIENTID and EXPIRED are reported too — and a principal refusal is the one RENEW failure an operator actually wants to see, since it means someone presented a clientid they have no standing on. Happy to move it if you would rather.

Also worth flagging for the human review Copilot asked for: the reviewer pass caught a real hole in my own first cut. I had copied FreeBSD's `nfsrv_notsamecredname()` root exemption without noticing it only works there because it sits on that function's non-GSS path. Lifted out, it meant any AUTH_SYS peer writing `uid=0` into its credential could renew the lease of a client established under Kerberos — exactly the case the check exists for. Dropped in 8a4277f, with `root_is_refused_against_another_principal's_record` pinning it.

PR #2175, opened 2026-08-26
fix(smb): answer every CHANGE_NOTIFY the notify path owes, one operation at a time

### LINKED ISSUE #2129: smb2.notify: only one outstanding CHANGE_NOTIFY per handle; a second replaces the first

### LINKED ISSUE #2132: Deleting a watched directory never completes its pending CHANGE_NOTIFY (93% of smb2.notify's runtime)

### LINKED ISSUE #2164: notify: a re-issued CHANGE_NOTIFY does not observe the mask armed on the same handle; an empty filter is refused
## Summary

With the rmdir defect ([#2140](https://github.com/marmos91/dittofs/issues/2140)) fixed, `smb2.notify.rec` and `smb2.notify.mask-change` both advance past their rmdir assertions and stop on what looks like one shared behaviour: **a CHANGE_NOTIFY re-issued on a directory handle does not observe the completion filter armed by an earlier request on that same handle.**

Observed on `smbtorture / memory` in the run for #2140's fix (PR #2159):

| test | assertion | status | expected |
| --- | --- | --- | --- |
| `smb2.notify.rec` | `notify.c:642` | `NT_STATUS_INVALID_PARAMETER` | `NT_STATUS_OK` |
| `smb2.notify.mask-change` | `notify.c:834` | `NT_STATUS_CANCELLED` | `NT_STATUS_OK` |

Both rmdirs and the following unlink now return `NT_STATUS_OK` in both tests (`notify.c:631`/`:634`/`:636` and `:827`/`:829`/`:831`), so this is strictly the next failure, not a regression.

## The two arms

**`notify.rec`** arms a recursive CHANGE_NOTIFY with `completion_filter = 0` at `notify.c:624-627` and requires the receive to succeed. DittoFS answers `STATUS_INVALID_PARAMETER`. The test is relying on the filter armed by the earlier request on the same handle still applying — an empty filter is not meant to be an error.

**`notify.mask-change`** arms a *different* filter at `notify.c:774-778` and requires the earlier mask to win. DittoFS cancels the request.

## What is and is not being proposed here

`smb2.notify.mask-change` is already classified as permanently unimplementable (bucket 9) on the grounds that MS-SMB2 §3.3.5.19 does not specify mask-coalescing across separate CHANGE_NOTIFY requests on one handle. **That classification is not being re-litigated** — and its own note asked for the rationale to be re-tested once #2140 landed, which has now happened: the test sits on exactly the assertion the rationale describes, so the placement is now supported by observation rather than assumed.

The open question is narrower and belongs to `notify.rec`: **is refusing `CompletionFilter = 0` with `STATUS_INVALID_PARAMETER` correct?** A zero filter arguably means "match nothing" — a request that simply never completes until cancelled — rather than a malformed request. That is a separate decision from mask stickiness and worth settling on its own.

Note that Samba also carries `smb2.notify.rec` as a known failure (`^samba3.smb2.notify.rec` in `selftest/knownfail`), so passing it end-to-end may not be achievable regardless; its `num_changes == 9` assertion at `notify.c:644` is still further on.

Filed so the `KNOWN_FAILURES.md` rows have a live tracking link once #2140 closes.

### LINKED ISSUE #2169: smb2.notify.valid-req intermittently returns two change records for one change
`smb2.notify.valid-req` fails intermittently with the server returning **two** change
records where the test requires one:

```
(../../source4/torture/smb2/notify.c:122) wrong value for n.out.num_changes  0x2 should be 0x1
```

The test (`TESTING VALIDITY OF CHANGE NOTIFY REQUEST`) makes a single change and expects
exactly one record back. Getting two means an event is delivered twice, or a stale
buffered event is drained into the same response as the fresh one.

### Why this is not either PR's fault, and not a base flake either

It surfaced on two PRs at once — #2161 (`pkg/block` local-tier residency) and #2162
(`pkg/adapter/smb` response ordering) — both running on base `60d4e61ca`.

The two commits between the last green develop conformance run (`cce71902a`) and that
base cannot change SMB behaviour, and this was checked rather than assumed:

- `60d4e61ca` (#2156) — filtering the diff for non-comment Go lines returns **nothing**;
  it is comment-only.
- `92ff2de39` (#2158) — one file, `.github/workflows/smb-client-compat.yml`, +21/-0.

So the failing bases are behaviourally identical to a base that passed. That rules out
a regression introduced in that window, and it equally rules out "just a flake we can
ignore": the same test is listed under **Passing** in `KNOWN_FAILURES.md` (Change Notify
section), and it has no row in the live table — which is why it fails the job as new.

Current tally on behaviourally-identical bases: **2 failures, 1 pass**. Three draws is
not a rate, and the point of stating it is that it is not one — a deterministic bug does
not pass at all, so this is a race, and a single green run is not evidence it is gone.

### Where it probably lives

`num_changes == 2` is a delivery-multiplicity bug, which puts it in the same area as the
recent change-notify wave (#2131, #2138, #2151) and adjacent to #2129 (several waiters
per handle) and #2164 (a re-issued notify not observing the armed mask). The buffered
event path is the obvious suspect: an event that has already been handed to one request
being drained again into the next.

### What would settle it

Repeat `smb2.notify.valid-req` against current develop head — five runs, not one — and
if it reproduces, discriminate between *the same event delivered twice* and *two distinct
events generated for one change* by logging what the response carries, not just how many.
Those have different fixes.

marmos91:
Four defects on one code path — `NotifyRegistry` and the handlers that feed it.
They are one PR because they overlap line-for-line: three of the four change how
a buffered change reaches a request, and splitting them produces PRs that each
pass CI alone and conflict on merge.

Stacked as four commits, one per issue, in the order they were worked, plus a
fifth that folds the duplication the first four created.

Closes #2132
Closes #2129
Closes #2164
Closes #2169

---

## #2132 — a directory marked for deletion never answered its watchers

**Symptom.** `smb2.notify.rmdir1`–`rmdir4` each burned a 300s client transport
timeout — 1201s, 93% of the whole `smb2.notify` suite — and graded
`NT_STATUS_CONNECTION_DISCONNECTED` where `NT_STATUS_DELETE_PENDING` was
required.

**Root cause.** Nothing in DittoFS completed a pending CHANGE_NOTIFY when the
directory it watched was marked for deletion. The watcher is normally a
*different* handle from the one setting the disposition, so neither the close
path (which answers only the closing handle's own watch, with
`STATUS_NOTIFY_CLEANUP`) nor `NotifyRegistry.CloseByFileID` — which is keyed by
handle — could reach it. The request simply stayed registered until the client
gave up.

**What the spec requires.** [MS-FSA] 2.1.5.14.3 is explicit, and it is a MUST:

> If `InputBuffer.DeletePending` is TRUE: … Set `Open.Link.IsDeleted` to TRUE.
> If `Open.Stream.StreamType` is DirectoryStream: Set
> `Open.ChangeNotifyDirectoryMarkedDeleted` to TRUE. **For each
> `ChangeNotifyEntry` in `Volume.ChangeNotifyList` where
> `ChangeNotifyEntry.OpenedDirectory.File` is equal to `Open.File`**: Remove the
> entry; **Complete the ChangeNotify operation with status
> STATUS_DELETE_PENDING.**

Two things follow that were not obvious from the issue:

- The trigger is **the mark, not the unlink.** The directory is still on disk
  until the last handle closes — in the torture test the watcher's own handle is
  still open when the assertion runs — so a fix keyed to the actual removal
  would never fire. Samba agrees: `set_delete_on_close_lck`
  (`source3/locking/locking.c`) broadcasts `MSG_SMB_NOTIFY_CANCEL_DELETED` to
  every share-mode entry the moment the token is added, and each recipient runs
  `remove_pending_change_notify_requests_by_fid(fsp, NT_STATUS_DELETE_PENDING)`.
- The match is on the **file**, not the Open — hence path+share here, not FileID.
  Watchers on *ancestor* directories are deliberately untouched even when
  recursive: their own subject still exists.

[MS-FSA] also settles the open question the issue raised about the disposition
being cleared again: `ChangeNotifyDirectoryMarkedDeleted` "**Once set TRUE it is
never reset back to FALSE, even if the deletion state of the directory is
removed**". Completion is one-way; there is nothing to undo.

**Fix.** `NotifyRegistry.CompleteWatchersForDeletePending(shareName, dirPath)`,
called from the two points where DittoFS commits the shared mark:

- `set_info.go`, the `FileDispositionInformation` branch;
- the delete-on-close election, which is where a handle carrying
  `FILE_DELETE_ON_CLOSE` from CREATE commits it (DittoFS deliberately defers that
  promotion to close — see `doc_election.go`). Both close paths sweep: the CLOSE
  handler and the LOGOFF / tree-disconnect / transport-drop teardown in
  `closeFilesWithFilter`. The sibling teardown path is included on purpose; it
  had the same hole.

Both sweeps run **after** the path that answers the closing handle's own watch,
so a closing handle still gets the `STATUS_NOTIFY_CLEANUP` MS-SMB2 3.3.4.1 owes
it and the "CHANGE_NOTIFY responses MUST be the last responses sent for the
FileId" ordering is unchanged.

`NotifyRmdir` is deleted. It had no production caller (only two tests), and it
answered this trigger with `STATUS_NOTIFY_CLEANUP`, which is the *close* trigger
from MS-SMB2 3.3.4.17 — the wrong status for this one.

**Deliberately not done.** [MS-FSA] 2.1.5.11 also says a CHANGE_NOTIFY *arriving*
on an Open that is already marked must be answered `STATUS_DELETE_PENDING`
immediately. No test in the suite reaches that ordering (only `notify.rmdir1-4`
use delete-on-close, and they send the notify first), so it is left out rather
than added on speculation. It would need a sticky per-Open flag, since
`DeletePending` can be cleared and this mark cannot.

## #2129 — only one outstanding CHANGE_NOTIFY per handle

**Root cause.** `byFileID` was `map[string]*PendingNotify` — one slot per handle.
A second CHANGE_NOTIFY on the same handle evicted the first and completed it with
`STATUS_CANCELLED`. [MS-FSA] 2.1.5.11 gives a directory **one** ChangeNotifyEntry
and lets any number of ChangeNotify operations wait on it; Samba queues them on
`fsp->notify->requests` (`DLIST_ADD_END`). Real Windows clients pipeline them.

**Fix.** `byFileID` is now an arrival-ordered queue. An event goes to the **head
only** and completes just that request — Samba's `notify_fsp` replies to the list
head and removes it, and the rest keep waiting for the changes after it. Copying
the event to every waiter would answer requests the client has not read yet with
the same change and leave the next change with nobody to go to.

Three consequences that are not bookkeeping:

- `unregisterLocked` removed its entry from the path list **by FileID**. With
  several waiters that drops whichever happened to be first rather than the one
  asked for, leaving the caller's notify registered and answering a request the
  client is still waiting on. Removal is by identity now, and order-preserving —
  a swap-remove would move the newest waiter into the oldest slot and hand it the
  next change. The regression test releases the **head** of a three-deep queue
  for exactly this reason: releasing the middle produces the same list under both
  implementations and proves nothing.
- `Register`'s replay of the armed handle's buffered events needed the same
  earliest-arrival gate #2151 gave the handler's `TakeBufferedEvents` fast path.
  The request that reaches `Register` first is not necessarily the one that
  arrived first on the wire, and those events belong to the earliest arrival.
  Without the gate, the queue turns `notify.double` from a wrong answer into a
  hang.
- A `(ConnID, MessageID)` collision now always unregisters the old holder. It
  used to skip that when the FileID matched, because the replacement path covered
  it; with that path gone the old entry would keep a registration no CANCEL could
  ever reach.

`MaxPendingWatches` bounds outstanding requests rather than distinct handles.

**Honest note on coverage.** No smbtorture subtest demonstrates this end to end
any more. `notify.double` reached the replacement only because `Register`
replayed buffered events into the later arrival; with that gated (above), the
earliest arrival is answered synchronously and never registers, so only one watch
is ever outstanding. `notify.mask-change` is the one test that genuinely holds two
at once, and it is a documented permanent known-failure for an unrelated reason.
The queue is therefore pinned by unit tests, each of which was run against a
deliberately broken build first (see Verification).

## #2164 — an empty CompletionFilter was refused

`smb2.notify.rec` re-issues a CHANGE_NOTIFY with `completion_filter = 0` and
requires it to succeed; DittoFS answered `STATUS_INVALID_PARAMETER`.

[MS-FSA] 2.1.5.11 makes the filter a property of the directory's
ChangeNotifyEntry, established by the first CHANGE_NOTIFY on the Open and reused
by later ones. Neither it nor MS-SMB2 3.3.5.19 validates the field at all, and
Samba does not either — `smbd_smb2_notify_send` calls `change_notify_create` only
while `fsp->notify` is nil and never inspects the filter. DittoFS already
implemented the sticky mask; the check just ran against the request's own field
instead of the one the request would actually run with.

The check now runs on the effective filter, so an empty filter is refused only
when the handle has no mask to inherit. Reserved and unrecognised bits stay
accepted — they never match an event, which is what `smb2.notify.mask` walking
all 32 bit positions requires. `CaptureNotifyCompletionFilter` no longer stores
an empty filter: doing so would arm the handle with a mask matching nothing for
the rest of its life, including for the later requests that *do* name one.

`smb2.notify.rec` still fails afterwards — further on, at its
`num_changes == 9` assertion, which is where Samba fails it too
(`^samba3.smb2.notify.rec` in `selftest/knownfail`). The `mask-change`
classification is not touched.

## #2169 — two change records for one change, intermittently

**The discrimination first**, because the count alone does not decide it: the two
records are **distinct events**, not one record delivered twice.
`FILE_ACTION_ADDED` from the CREATE and `FILE_ACTION_MODIFIED` from the WRITE
that follows it, in `torture_setup_simple_file` (`smb2_util_unlink` →
`torture_smb2_testfile_access` → `smb2_util_write` → close). Both match
`FILE_NOTIFY_CHANGE_ALL` and both are correct in isolation — the *third*
assertion of the same test requires that MODIFIED.

**Root cause is the boundary, not the event.** Buffered changes were flushed on a
100µs wall-clock window opened by the first event, so a reply carried whatever
landed inside it. The window exists to hold together the several records **one**
operation emits — a CREATE that overwrites fires REMOVED + ADDED + MODIFIED — but
it also absorbed the *next* operation's whenever the client's next request
happened to arrive fast enough. The reply's contents were a function of client
timing, which is exactly why the same code passed on some runs and failed on
others.

**Fix.** Changes carry the id of the operation that produced them; a reply
carries the leading run belonging to one operation and the rest stay buffered for
the client's next request. The two places where one operation emits several
events — a CREATE that overwrites, and `SET_REPARSE_POINT`'s replacement — now
emit them as one batch through `NotifyChanges`, so their grouping no longer rests
on the clock either. This matches Samba, which answers a pending request on the
first event batch (`notify_fsp` → `change_notify_reply`) and gives a request that
arrives with a backlog the whole backlog (`change_notify_fsp_has_changes`); both
behaviours are preserved.

**One defect in the first cut of this fix, found in review and fixed here.**
`NotifyChanges` allocates its emission id under `r.mu` and releases the lock
between events, so two operations emitting concurrently interleave into the same
watcher's buffer. Counting every entry whose id matches the first made the run 2
for `[A B A]` and shipped `B` inside `A`'s response — precisely the merge this
accounting exists to prevent. The run is now a contiguous prefix.
`TestBufferEventLocked_InterleavedEmissionsDoNotMerge` pins it.

**A pinning test asserted the bug.** `TestNotifyChange_ValidReq_RemovedAddedModifiedSequence`
asserted that three *separate* operations arrive in one reply, and justified it
with `valid-req`'s third assertion (`num_changes == 3`). That assertion is
actually served by the armed handle's backlog, not by the flush window: the
request before it advertises `buffer_size = 0`, so it is answered synchronously
with `STATUS_NOTIFY_ENUM_DIR` and no watch is pending while those three events are
produced. The test was rewritten to pin the correct boundary in both directions.

---

## Verification

`go build ./...`, `go vet ./...`, `gofmt`, and `go test -race ./internal/...` are
green.

**Every new test was run against a deliberately broken build before being
trusted.** Each of the 17 mutations below was applied to the final code and the
suite confirmed to go red; none passed:

| Mutation | Result |
| --- | --- |
| answer the sweep with `STATUS_NOTIFY_CLEANUP` instead of `STATUS_DELETE_PENDING` | RED |
| sweep ignores the share name | RED |
| sweep every watch path, ancestors included | RED |
| remove the close-path sweep entirely (revert #2132) | RED |
| sweep only when this handle performs the unlink | RED |
| deliver an event to every waiter (drop the head-only rule) | RED |
| swap-remove instead of an order-preserving shift | RED |
| `CloseByFileID` answers only the oldest waiter | RED |
| `Register` replays buffered events unconditionally | RED |
| `Register` gates on queue head only, ignoring wire order | RED |
| `Register` evicts the existing watch again (revert #2129) | RED |
| deliver the whole window (revert #2169) | RED |
| split at one record instead of one operation | RED |
| break a batched operation into separate emissions | RED |
| validate the request's filter instead of the effective one (revert #2164) | RED |
| capture an empty filter as the handle's sticky mask | RED |
| count emission-id matches anywhere instead of a contiguous prefix | RED |

The #2169 test covers both delivery paths — the test-only `FlushAll` and the
production flush timer — because a mutation of `flushWatcher` alone was initially
invisible to a `FlushAll`-only test.

## Protocol verification — measured in CI, on both profiles

`smb2.notify` is run by CI on `memory` and `badger-fs`. It was **not** giving these
issues any coverage: `smb2.notify.mask` alone eats ~78s of the shared 120s budget,
so the tail of the suite has been ungraded for as long as the budget has existed.
This PR runs the suite per subtest, so the tail is graded for the first time.
Identical results on both profiles (run 32945955526): **21 subtests, 20 pass, 1 fail.**

- **#2132 — verified.** `rmdir1`, `rmdir2`, `rmdir3`, `rmdir4` all pass, on two
  independent metadata backends. They were never failing in CI; they were never
  *reached*.
- **#2164 — verified.** `rec` moved from `notify.c:642`
  (`NT_STATUS_INVALID_PARAMETER`) to `notify.c:644`
  (`num_changes 0x5 should be 0x9`) — the empty filter is accepted, the test
  advances, and it now fails where Samba fails it
  (`^samba3.smb2.notify.rec` in `selftest/knownfail`). Its `KNOWN_FAILURES.md`
  row is rewritten from that graded output, not from the prediction.
- **#2169 — `valid-req` graded `success` on every run of this branch.** It is a
  race, so the count is what matters, not the verdict: memory ×2, badger-fs ×2 at
  the time of writing, with a further run in flight. Reported as a count
  deliberately.
- **#2129 — still not demonstrated end-to-end.** `double` passes, but it passes on
  develop too, so it is not evidence for the queue. See the note in the #2129
  section above. It rests on unit tests plus the mutation table.

### A regression this PR caused, and the harness change that fixes it

The first CI run graded **6** notify subtests against develop's **18** — twelve
lost — while the job still reported PASS, because the harness classifies a filter
timeout as *ungraded*, not as a failure.

Cause: `mask-change` issues two CHANGE_NOTIFYs on one handle. Before this PR the
second evicted the first with `STATUS_CANCELLED`, so the test failed in 32ms. That
eviction is the defect #2129 removes. With both requests legitimately outstanding
neither is ever answered — correctly, because the handle's sticky mask matches none
of the events the test generates, which is precisely why the test is classified
permanently unimplementable — so it blocks until the budget expires and starves
everything after it. The fast failure was an accident of the bug.

`smb2.notify` is therefore run per subtest with `mask-change` skipped, the same
workaround already used for `scan.scan` and `dirlease.oplocks`. The split is
load-bearing independently of `mask-change`, since it is what gives `rmdir1-4` a
budget they can finish in.

`smb2.notify.security` is omitted too: it exists in Samba master but not in the
pinned smbtorture 4.22.6, which answers `Unknown torture operation`.

No `KNOWN_FAILURES.md` row was moved on a prediction — `rec` was rewritten from a
graded run, and `mask-change` records the skip.


marmos91:
Copilot's finding on `NotifyChanges` is right and is now fixed in 38bf85ea5.

The batch took `r.mu` once per event, so a concurrent operation could land between two events of the same batch. The run that batch occupies in a watcher's buffer would then not be contiguous, and `drainFirstEmissionLocked` would cut the group in half — the group being split across two responses is the same defect this batching exists to prevent, reached from the other side. My contiguity guard stopped a batch absorbing a *foreign* event; it did not stop a foreign event splitting *this* one.

Every event of a batch is now buffered under one acquisition of the lock. The byte accounting still runs after the unlock, because `chargeArmedBuffer` takes the lock itself and fires `OnOverflow` outside it.

The contiguity guard is kept and its comment reworded to say what it now does: with every emitter holding the lock across its batch nothing interleaves today, but counting emission-id matches anywhere in the buffer would make "one response, one operation" a property of caller discipline rather than of the accounting itself. It is pinned directly by `TestBufferEventLocked_InterleavedEmissionsDoNotMerge`, which drives `bufferEventLocked` with an interleaved `[A B A]` sequence.

All 17 mutations in the PR body's table were re-run against the restructured code and are still red, including a new one — emitting each event of a batch under its own emission id — which `TestNotifyChange_BatchedOperationStaysInOneResponse` catches.

marmos91:
## The first CI run exposed a regression this PR caused. Fixed in b5797c238.

Recording it precisely, because the job reported **PASS** both before and after and the difference is invisible from the check status.

### What happened

`smb2.notify` graded **6** subtests on this branch against **18** on develop (`ce37eaa70`, this branch's base). Twelve tests stopped being graded, including `smb2.notify.double`, which develop grades as **success**.

| | develop | this branch (before b5797c238) |
| --- | --- | --- |
| graded | 18 (16 pass, 2 fail) | 6 (6 pass) |
| `mask-change` | `failure` in **32ms** | started, never finished |
| everything after it | graded | **UNGRADED** |

The job stayed green because the harness classifies a filter timeout as *ungraded / inconclusive*, not as a failure. A green check was hiding twelve tests that had stopped running.

### Root cause — and it is this PR's own fix working correctly

`mask-change` issues two CHANGE_NOTIFYs on one handle. Before #2129 the second **evicted** the first and completed it with `STATUS_CANCELLED`, so `smb2_notify_recv` returned immediately and the test failed in 32ms. That eviction is exactly the defect #2129 removes.

With both requests now legitimately outstanding, neither is ever answered — correctly, because the handle's sticky mask (`FILE_NOTIFY_CHANGE_ATTRIBUTES`, fixed by the first CHANGE_NOTIFY on that handle per [MS-FSA] 2.1.5.11) matches none of the name/rename events the test generates. That is precisely the mask-coalescing behaviour the test's own permanent-known-failure classification says DittoFS does not implement. So the test now blocks in `smb2_notify_recv` until the 120s budget expires and starves the rest of the suite.

The old fast failure was an accident of the bug. Removing the bug removed the accident.

### Fix

`smb2.notify` is now run per subtest with `mask-change` skipped — the same workaround shape already used for `scan.scan` and `dirlease.oplocks`, both skipped for the same class of reason (a test that aborts or hangs the rest of its suite).

The split is load-bearing a second time, independently of `mask-change`: **`smb2.notify.mask` alone takes ~78s of the shared 120s budget**, so `rmdir1-4` at the tail of the suite were never reached *even on develop*. That is why #2132's four tests have never appeared in any CI grading — not because they fail, because they are never run. Per subtest, each gets its own budget.

### What this means for the four issues

The first run gave exactly one piece of evidence: `smb2.notify.valid-req` graded **success** (#2169). That is **one draw, not five**, and #2169 is a race — I am not calling it fixed on it.

`rmdir1-4` (#2132), `double` (#2129) and `rec` (#2164) were never reached, so they still have **no** end-to-end evidence. This run is the first that can produce any.

I have not touched the `rec` row in `KNOWN_FAILURES.md`. I expect it to keep failing, further along at its `num_changes == 9` assertion where Samba also fails it — but that is a prediction, and the row moves on measurement or not at all.

marmos91:
## Measured, on both profiles: `rmdir1-4` pass, and they had never been graded before

With the per-subtest split, `smb2.notify` grades **21** subtests instead of 6. Identical results on `memory` and `badger-fs` (run 32945955526):

| subtest | memory | badger-fs |
| --- | --- | --- |
| **rmdir1** | success | success |
| **rmdir2** | success | success |
| **rmdir3** | success | success |
| **rmdir4** | success | success |
| double | success | success |
| valid-req | success | success |
| rec | failure | failure |
| the other 14 | success | success |

### #2132 — verified

All four `rmdir` tests pass, on two independent metadata backends. Worth being precise about what changed: **these were never failing in CI, they were never *reached*.** `smb2.notify.mask` alone consumes ~78s of the shared 120s budget, so the tail of the suite has been ungraded for as long as the budget has existed. They are graded for the first time here, and they pass.

### #2164 — verified, and the row is re-measured

`rec` moved from `notify.c:642` — `NT_STATUS_INVALID_PARAMETER` — to `notify.c:644`:

```
(../../source4/torture/smb2/notify.c:644) wrong value for notify.smb2.out.num_changes  0x5 should be 0x9
```

Exactly the predicted outcome: the empty-filter refusal is gone, the request is accepted, and the test advances to the assertion Samba also fails (`^samba3.smb2.notify.rec` in `selftest/knownfail`). Its `KNOWN_FAILURES.md` row is updated from that graded output rather than from the prediction.

### #2169 — four draws so far, two more in flight

`valid-req` has graded `success` on every run of this branch: memory ×2, badger-fs ×2. Since it is a race I am holding to the five-run minimum; the run now in flight brings it to six. I will report the count rather than declare it fixed on a partial one.

### #2129 — still not demonstrated end-to-end, and that is unchanged by this run

`double` passes here, but it also passes on develop, so it is not evidence for the queue. As noted in the PR body, once `Register`'s replay is correctly gated, the earliest arrival is answered synchronously and never registers, so only one watch is ever outstanding. `mask-change` is the one test that genuinely holds two, and it is skipped for the reason in the previous comment. #2129 rests on unit tests plus 17 mutations, not on the suite.

### One more harness fix

`smb2.notify.security` is in Samba master but not in the pinned smbtorture 4.22.6, which answered `Unknown torture operation`. Removed, with a note to add it when the pin moves.

marmos91:
Informational, on merged work — flagging a citation, not disputing the fix.

This PR's body quotes **[MS-FSA] 2.1.5.14.3** and calls it "explicit, and it is a MUST". That section does not exist: 2.1.5.14 is "Server Requests a Query of Security Information" and has no subsection 3.

The rule this PR implements — `CompleteWatchersForDeletePending` sweeping the directory's watchers when the delete disposition is committed, answering each `STATUS_DELETE_PENDING` — is **[MS-FSA] 2.1.5.15.3 `FileDispositionInformation`** step 3.2.3.2:

```
```

`2.1.5.14.3` → `2.1.5.15.3`: one digit in the middle component. **The MUST is real and this PR implements it correctly** — only the number pointing at it was wrong, which is why nothing here needs revisiting.

Two things that follow, both already handled elsewhere:

- The same typo reached six code comments; #2269 corrects them to 2.1.5.15.3 (and 2.1.5.15.4 for the `FileDispositionInformationEx` class `set_info.go` serves in the same `case`).
- The half this PR deliberately deferred — a CHANGE_NOTIFY *arriving* on an already-marked directory — genuinely is **2.1.5.11**, and #2199 tracks it. That pairing is right; only the sweep half was misnumbered.

Raising it because this PR is the premise other work cites. Someone tracing #2199 back to here would find a MUST attached to a section they cannot look up, and the natural repair — reaching for 2.1.5.11, the other section in the story — is the exact mistake I made in #2269 before a peer caught it.

Issue #2177, opened 2026-08-26
A repeated server-side copy counts the source's hashes up every time


marmos91:
`engine.CopyPayload` bumps `RefCount` once per unique source hash
(`pkg/block/engine/readwrite.go`, the `seen`-set loop) and has no notion of
having done it before. Everything else the copy does is idempotent: the
destination's rows are `Put` by exact `{payloadID}/{offset}` ID and rewrite in
place, the stale-row reap computes an empty set the second time round, and
`SetManifest` writes the same manifest. The increments are the one part that
counts up on every run.

So a server-side copy that is *repeated* — same source, same destination —
leaves the source's hashes counted one higher than they are referenced, once
per repeat. Nothing pairs those bumps with a decrement: the reclaim paths
(`Delete`, `Truncate`) decrement a payload's **own** rows by ID, so the
destination's later unlink decrements its rows once, not once per attempt.
The count never returns to 0 and the chunk is never swept.

### When a copy gets repeated

- The reply is lost after the server committed, and the client reissues on a
  connection where the duplicate-request cache no longer has the entry.
- The operation reports an error *after* its transaction committed and the user
  reruns it. #2160's PR adds one such path deliberately: the destination's
  stale local ranges are discarded post-commit, because dropping them before
  the commit would destroy the destination's real content if the copy rolled
  back, and a failure there is reported rather than swallowed.

This is a defect with a benign failure mode, not acceptable drift. The
*immediate* failure is the mild one — a chunk nothing references is kept rather
than one something references being reclaimed, the trade `Truncate`'s reap
comment already states — but a count that only ever rises is still a count that
permanently blocks reclamation of blocks nothing references. That is a real
resource bug on a slow fuse: it never corrupts anything and it never stops
growing, and on a share where copies are routine it ends as storage that cannot
be recovered without rebuilding the counts. It is also invisible — nothing
reports a count that no longer matches the rows — so the first symptom is a
share that will not shrink, long after the copies that caused it.

### Why the obvious fix is not obviously right

Skipping the increment for a hash the destination already holds a row for makes
a repeat free, but it also skips the increment when the destination genuinely
held that hash before the copy (cross-file dedup on a first run), which counts
*down* relative to today. Under-counting is the dangerous direction: a row
reaped one decrement early is data the file still needs. Any fix has to tell
"this copy already ran" apart from "these two files happened to share a chunk",
which the current row set alone does not say.

Worth weighing against how load-bearing `RefCount` still is here:
[...]

Issue #2176, opened 2026-08-26
A local-only clone leaves the destination describing bytes past its new size


marmos91:
`materializeLocalClone` is the local-only (no remote store) clone path: it
copies the source's bytes into the destination's own journal over
`[0, srcSize)`, then sets `dstFile.Size = srcFile.Size`. Those writes supersede
the destination's own intervals by version, so the copied span reads correctly
— which is why local-only shares are not affected by #2160's copied-span
defect.

What it never touches is the destination **past the new size**. When the
destination was larger than the source, its intervals and its own `FileChunk`
rows over `[srcSize, oldDstSize)` survive the clone. Nothing clips them: the
path calls no truncate, and the final transaction is `UpdateAttrs`, which
writes the size and leaves the manifest alone.

Every protocol read clamps to the file size, so this is invisible until the
destination is grown again — a `SETATTR`/`ftruncate` up, or a write past the
new size. POSIX says the grown region reads as zeros; the destination serves
the bytes it held before the clone instead.

### Verified on a real engine

Journal-backed local tier, no remote, source 4096 bytes of `0x11`, destination
8192 bytes of `0x22`, then `CloneWholeFile`:

```
dst extents after local-only clone (size now 4096): [[0 8192]]
RESULT: the tail past the clone's new size still serves the destination's old bytes
```

The index still describes the full 8192 bytes the destination held, not the
4096 the clone gave it.

### Where

`internal/adapter/common/clone.go`, `materializeLocalClone`. The remote-backed
reflink path has the equivalent problem fixed by #2160's PR — it discards the
destination's local ranges wholesale and reaps the destination's rows the copy
does not take over — but that fix is scoped to the manifest-only reflink and
does not run for a local-only share.

### Shape of a fix

The destination has to stop describing anything past the new size. Clipping the
local tier alone is not enough: the destination's own `FileChunk` rows past
`srcSize` also survive, and a cold path resolves through those. The existing
`engine.Truncate` does both — it reaps the rows past the size by exact
`{payloadID}/{offset}` identity and clips the local tier — so driving it for the
destination at `srcFile.Size` is the obvious candidate, subject to the same
ordering care the reflink path needed: it must not run before the copy is known
to have landed.

### Related

- #2160 — the same contract on the remote-backed reflink path
- #1784 — CLONE reads zeros on local-only shares

marmos91:
Fixed in #2188.

PR #2178, opened 2026-08-26
fix(block): stop a reflink's destination serving the content it replaced

### LINKED ISSUE #2160: A reflink onto a destination that still holds local bytes serves the destination's old content

marmos91:
A whole-file reflink replaces the destination's content, and until now the
destination went on serving the content it replaced. The manifest said one
thing and every read said another, with nothing logged.

Confirmed on a real engine before any change — journal-backed local tier,
memory remote, both files written and carved, then `CopyPayload(src -> dst)`:

```
RESULT: destination reads its OWN PRE-CLONE bytes (stale)
```

`bytes.Equal(back, dstData)` — not `srcData`, not zeros.

## Two independent halves, and fixing only one is worse than fixing neither

The copy leaves the destination two kinds of stale state, and each on its own
is enough to serve the wrong bytes.

**The local tier.** `readAtInternal` asks the local tier first and only
reconciles against the manifest when the tier reports a hole or a cold range. A
covered warm range reports neither, so a destination that still holds its own
bytes never reaches the new manifest at all. This is the half the issue
describes.

**The manifest.** `engine.CopyPayload` writes one row per source block and
never removes the destination's own rows. `SetManifest` does not remove them
either: it rewrites the `Blocks` projection (`file_block_refs` on SQL,
`fm:<uuid>` on badger), which is a different structure from the per-chunk rows
a cold read resolves through (`ListFileChunks` reads `file_blocks`).
`CopyPayload`'s own doc-comment says so — "the cold-read path resolves a
payload's bytes via ListFileChunks(dstPayloadID), NOT via FileAttr.Blocks".

The two files chunk their own content, so their offsets rarely line up. The
destination's surviving rows interleave with the copied ones and
`findRowCoveringOffset` takes whichever starts latest — so a cold read of the
destination walks a **mosaic** of both files' content. Today that is reachable
by evicting the destination and reading it back; and it is exactly what fixing
only the local-tier half produces, which is why both are in this PR.

## The fix

- `engine.Store.DiscardLocalContent` drops the local tier's whole account of a
  payload, through the tier's existing durable, version-fenced clip. Both
  reflink helpers in `internal/adapter/common` call it **post-commit**: until
  the commit lands, those bytes are the destination's real content, and a
  rolled-back copy that had already dropped them could not get them back. A
  self-clone/self-copy is guarded — it replaced nothing.

- `engine.CopyPayload` reaps the destination's rows the copy does not take over
  at the same offset, by exact `{payloadID}/{offset}` identity, the way
  `Truncate` already reaps a dropped tail. It runs in the caller's transaction
  so a rolled-back copy leaves the destination's manifest as it found it.

**Holes, not cold.** What the discard leaves behind is holes. That is
deliberate and it is the distinction the fix turns on: an absent local range
and a stale one are not the same state, and only the manifest can tell them
apart. A hole the manifest covers hydrates from the remote; a hole it does not
cover reads as the zeros a sparse source has there. Marking the copied span
cold instead would fail the sparse case closed. `TestCopy_DestinationSparseRangeReadsAsZeros`
is what says so.

**Residual window, and what it costs.** The discard cannot be moved before the
commit, so a crash between the two leaves the destination serving the replaced
content until the copy is repeated. A discard that *fails* is reported rather
than swallowed, for the same reason — reporting success on a destination that
serves the content the copy replaced is the defect this closes. That makes the
discard the one place in these helpers where an error does not mean nothing
landed, and the doc comments that claimed otherwise are corrected here. It also
means a repeat is not free: `engine.CopyPayload`'s manifest work is idempotent
but its per-hash `RefCount` bumps are not, so each repeat counts the source's
hashes one higher. That drifts in the direction the reclaim paths already
prefer — keep a chunk nothing references rather than reclaim one something does
— and is filed as #2177 rather than fixed here, because the obvious fix
(skip the bump for a hash the destination already holds) cannot tell a repeat
from cross-file dedup and would count *down* on the latter.

## Why it was never noticed

`cp --reflink` opens the destination `O_TRUNC`, which issues a SETATTR size=0
first. That truncate clears both halves — the journal clip drops the intervals
and `engine.Truncate` reaps the rows — so the clone lands on an empty
destination and reads correctly. The bug needs a CLONE onto a destination that
still holds bytes: `xfs_io -c "reflink ..."` over an existing file, or any
client that does not truncate first. A reproduction built on `cp --reflink`
shows a passing system and tells you nothing.

## Tests

Every one of these was run against a deliberately broken build first and
watched go red; the `O_TRUNC` masking makes a self-satisfying test unusually
easy to write here.

- `TestCopy_DestinationServesTheCopiedContent` — byte-level read-back at two
  known offsets, with a **bystander payload** as the control: carved and warm
  exactly like the destination, not part of the copy, and still reading its own
  bytes afterwards. Fails on either half alone (stale content without the
  discard; the mosaic without the reap).
- `TestCopy_DestinationSparseRangeReadsAsZeros` — a sparse source, so the
  destination's middle must read as zeros rather than as its own old bytes or
  an error.
- `TestCopy_SelfCopyKeepsItsOwnRows` — the one destination whose surviving rows
  are not stale. The blocks handed in are the caller's snapshot of the
  projection, not the authority on which rows are live, so a row missing from
  them must not be reaped out from under a live file.
- `TestCloneWholeFile_DropsTheDestinationsStaleLocalRanges` /
  `TestCloneWholeFile_SelfCloneKeepsItsLocalRanges` /
  `TestCopyPayload_SelfCopyKeepsItsLocalRanges` — the wiring and the self-copy
  guard.
- `DiscardLocalContent` joins the close-gate table.

## Related

- #2161 is the adjacent half of this path, seen from residency accounting, and
  it is now merged. **The two compose, and the order between them is
  load-bearing**: the discard takes the destination's replaced ranges back to
  holes, and #2161's seed then accounts for the ones the copy gave it. Seeding
  first is not merely redundant — `SeedCold` skips a range the index already
  describes, so it would skip the whole copy and the discard would then drop
  what little it recorded, leaving the tier describing none of the copy. That
  ordering is pinned by #2161's own `TestCloneWholeFile_SeedsDestinationRanges`,
  which goes red ("the index describes 0 bytes of the clone's destination") when
  the two calls are swapped — verified by actually swapping them, not asserted.

  The rebase onto merged #2161 also changed what this PR's adapter-level test
  can assert. It used to check the destination's index was empty; with the seed
  in place the index is populated again, as **cold** ranges. A destination that
  dropped nothing and one that dropped everything now both come back described,
  so the index cannot tell them apart. That test now asserts the bytes instead —
  which is what the contract was always about.

- #2177 — `engine.CopyPayload`'s `RefCount` bumps are not idempotent; a repeated
  copy counts up and never comes back down. Pre-existing, and this PR adds one
  more way to reach it. Not "acceptable drift": the immediate failure is benign,
  but a count that only rises permanently blocks reclamation of blocks nothing
  references.
- #2176 is a narrower sibling this PR does not touch: `materializeLocalClone`
  (the local-only path) materializes real bytes over `[0, srcSize)`, which
  supersede by version, but leaves the destination's intervals and rows **past**
  the new size in place, where a later grow exposes them. Verified on a real
  engine and filed separately rather than folded in here — it is a different
  code path with a different trigger.

Closes #2160

### DIFF
--- a/internal/adapter/common/clone.go
+++ b/internal/adapter/common/clone.go
@@ -24,12 +24,15 @@ import (
 // either file produces new CAS blocks under a new hash, leaving the other side
 // untouched.
 //
-// Everything is atomic in one metadata transaction:
+// The manifest side is atomic in one metadata transaction:
 //   - engine.CopyPayload's per-hash IncrementRefCount UPDATEs are bound to the
 //     txn (via metadata.WithTx) so they commit/roll back together with the
-//     destination UpdateAttrs. On any error nothing is committed — no partial
-//     dstFileAttr, no leaked RefCount bumps.
-//   - cache.InvalidateFile (if cache != nil) runs POST-txn, after the commit.
+//     destination UpdateAttrs. An error from inside the txn commits nothing —
+//     no partial dstFileAttr, no leaked RefCount bumps.
+//   - discardStaleDestination and cache.InvalidateFile run POST-txn, after the
+//     commit. An error out of the discard is therefore an error on a copy whose
+//     manifest is already durable — see its own comment for why it is still
+//     reported rather than swallowed.
 //
 // CLONE copies the source's CAS block manifest (FileAttr.Blocks). A freshly
 // written source whose bytes are still in the append log / in-memory buffer has
@@ -165,10 +168,16 @@ func CloneWholeFile(
 // dropped them could not get them back. What is left is the window between the
 // two, where a read still serves the replaced content.
 //
-// A failure fails the copy, which has already committed. The alternative is
-// reporting success on a destination whose every read serves the content the
-// copy replaced, and a caller that retries a copy it was told failed re-runs an
-// operation that lands on the same manifest and gets another chance at this.
+// A failure fails the copy, which has already committed — the one place in
+// these helpers where an error does not mean nothing landed. Reporting it is
+// still the better trade: the alternative is reporting success on a destination
+// whose every read serves the content the copy replaced, which is the defect
+// this exists to close. It is not free, though. A caller that retries a copy it
+// was told failed re-runs a copy whose manifest work is idempotent but whose
+// per-hash RefCount bumps are not, so each retry leaves the source's hashes
+// counted one higher than they are referenced. That errs toward keeping a
+// chunk nothing references rather than reclaiming one something does, which is
+// the direction the reclaim paths already prefer.
 func discardStaleDestination(ctx context.Context, blockStore *engine.Store, dstPayloadID metadata.PayloadID) error {
 	if err := blockStore.DiscardLocalContent(ctx, string(dstPayloadID)); err != nil {
 		return fmt.Errorf("discard the copy destination's replaced local ranges: %w", err)
--- a/internal/adapter/common/clone.go
+++ b/internal/adapter/common/clone.go
@@ -135,8 +135,11 @@ func CloneWholeFile(
 	// Order matters between these two. The discard takes the destination's
 	// replaced ranges back to holes; the seed then gives the local tier its
 	// account of the ranges the copy put there. Seeding first would find those
-	// ranges already described and skip them, leaving the tier holding the
-	// content the copy replaced.
+	// ranges still described by the content the copy replaced and skip them all,
+	// and the discard would then drop what little it had recorded — reads would
+	// still be right, resolving against the manifest, but the tier would end up
+	// describing none of the copy and nothing could report its remote-only
+	// bytes.
 	if err := discardStaleDestination(ctx, blockStore, dstPayloadID); err != nil {
 		return err
 	}
--- a/internal/adapter/common/clone.go
+++ b/internal/adapter/common/clone.go
@@ -122,12 +122,20 @@ func CloneWholeFile(
 	if err != nil {
 		return err
 	}
-	// A self-clone left the destination exactly as it was: it inherited no
-	// ranges the local tier has to account for, and its cached entries still
-	// describe the content it holds. Everything below is about a destination
-	// whose content changed.
+	// A self-clone left the destination exactly as it was: it holds the content
+	// its manifest describes, it inherited no ranges the local tier has to
+	// account for, and its cached entries still describe what it holds.
+	// Everything below is about a destination whose content changed.
 	if selfClone {
 		return nil
+	}
+	// Order matters between these two. The discard takes the destination's
+	// replaced ranges back to holes; the seed then gives the local tier its
+	// account of the ranges the copy put there. Seeding first would find those
+	// ranges already described and skip them, leaving the tier holding the
+	// content the copy replaced.
+	if err := discardStaleDestination(ctx, blockStore, dstPayloadID); err != nil {
+		return err
 	}
 	seedClonedRanges(ctx, blockStore, dstPayloadID, copied)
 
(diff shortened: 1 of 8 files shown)


marmos91:
Good catch on the empty-\`srcBlocks\` early return — that path is reachable and it is the same defect, so it is fixed here rather than deferred.

A source that places no blocks is a fully sparse one (sized, never written): it carves no rows, so it hands the destination an empty manifest and the destination should read as the zeros the source has. \`CopyPayload\` returned before the reap for it, and with the local tier already dropped the destination's own rows were then the only thing a read could resolve through — the exact stale content this PR closes, just reached through a manifest the copy never touched.

The empty-source return moved past the reap. The coordinator-not-wired check is now scoped to the case that needs it, the reap is guarded on the coordinator being present, and the trailing deep copy yields the same nil list for an empty source that the early return did, so \`TestCopyPayload_EmptySource\` and the close-gate case are unchanged.

Pinned by \`TestCopy_SparseSourceLeavesNothingOfTheDestination\`: sized-but-never-written source, destination written and carved, read back at two known offsets. Verified against a deliberately broken build first — restoring the early return makes it fail with "the destination at offset 0 served the content the copy replaced".

marmos91:
Rebased onto current `develop` (`41cab93ff`, now carrying merged #2161 plus #2162/#2168/#2170/#2173/#2174/#2175). Six commits became eight, all signed, force-pushed with `--force-with-lease`. Seven files conflicted; the resolutions were not mechanical, so here is what actually had to be decided and what was re-verified rather than assumed.

**The ordering between the discard and #2161's seed is load-bearing, and it is now pinned by #2161's own test.** I claimed before the merge that the two compose — discard makes the span a hole, the seed then covers it. That was a prediction against a branch; it is now testable against the merged code, so I tested it by swapping the two calls. `TestCloneWholeFile_SeedsDestinationRanges` goes red: *"the index describes 0 bytes of the clone's destination, want 10240"*. `SeedCold` skips a range the index already describes, so seeding first skips the entire copy, and the discard then drops what little it recorded. Reads stay correct either way — they resolve against the manifest — but the tier ends up describing none of the copy and nothing can report its remote-only bytes. The comment I first wrote at that call site claimed a different (worse) failure mode; the test disagreed with me and the comment is corrected to say what actually happens.

**The rebase invalidated one of this PR's assertions, which is the part worth flagging.** `TestCloneWholeFile_DropsTheDestinationsStaleLocalRanges` asserted the destination's index was empty after the clone. With #2161 merged that is no longer the contract: the seed puts the copied ranges back into the index as **cold** ones. A destination that dropped nothing and one that dropped everything now both come back described, so `DataExtents` cannot tell them apart — the test was passing on develop and failing here for a correct reason. It now asserts the **bytes**: the destination must not serve the content the clone replaced, with a pre-clone read proving it held those bytes to begin with, and the bystander control still reading its own back so a fixture broken enough to fail every read cannot pass. Re-verified red against a build with the discard call removed.

Re-verified on the rebased tree, each against a deliberately broken build:

- discard disabled → destination serves the replaced content at every offset, including the sparse range. #2161's seed alone does **not** save it.
- reap disabled → *"served neither the copied content nor its own"* at 2 MiB (the mosaic), and the sparse-source case fails outright.
- order inverted → #2161's test red, as above.

Local `go test ./...` and `go test -race ./pkg/block/... ./internal/adapter/common/` both clean. Everything #2161 added is intact: both its helpers, all three of its tests, and `newCopyTestEngineWithLocal` defined exactly once (both PRs had added it).

Issue #2182, opened 2026-08-26
SMB rename reverts ctime, discarding the update MS-FSA 2.1.5.15.12 requires


marmos91:
Split out of #2172 (item 3) after investigation. The claim there was that rename preserving both
mtime and ctime contradicts the spec. **Investigation narrowed it: the mtime half is correct and is
already a no-op. Only the ctime half is a defect** — and it is a discard, not the deferral that
MS-FSA licenses.

## The distinction that decides it

MS-FSA 2.1.5.15.12: *"The object store MUST update Open.File.LastChangeTime.`<187>`"*

Product note `<187>` verbatim:

> Section 2.1.5.15.12: The file system only updates LastChangeTime if no user has explicitly set
> LastChangeTime. The NTFS and ReFS file systems defer setting LastChangeTime until the handle is
> closed.

That licenses two things: skipping the update when the user explicitly set LastChangeTime, and
**deferring** it to close. It does not license discarding it. #2172 flagged this as the weakest of
its six items on the grounds that `<187>` might make DittoFS match Windows in practice — it does
not, because DittoFS discards rather than defers.

## What the code does

`internal/adapter/smb/handlers/set_info.go`, `saveTimestamps()` snapshots Mtime **and** Ctime and
writes both back after the rename (four call sites across the file and stream rename paths).

Two facts make the ctime half a straightforward bug:

1. `pkg/metadata/file_modify.go` already sets `srcFile.Ctime = now` inside `Move`'s transaction, and
   leaves Mtime untouched. **The metadata layer already implements the MUST correctly.** So the
   mtime half of the save/restore is restoring a value nothing changed, and the ctime half exists
   solely to revert a spec-mandated update.
2. Nothing defers it. `close.go` does no ctime work for a renamed handle — the only CLOSE-time
   timestamp work is the pending-atime flush and `restoreFrozenTimestamps`, which restores only
   fields explicitly frozen by a `-1` sentinel. There is no deferred-rename-ctime state anywhere.

## Observable

Close the handle, reopen, QUERY_INFO FileBasicInformation: `ChangeTime` is the pre-rename value,
permanently. Windows reports the rename time once the handle closes. It is also cross-protocol
inconsistent — an NFS client renaming through the same `Move` gets the correct ctime bump; the
identical rename over SMB has it reverted.

## Suggested shape

Drop `Ctime` from `saveTimestamps` and let `Move` stand. The `<187>` user-set case is already
[...]

Issue #2181, opened 2026-08-26
SET_INFO FileBasicInformation bumps LastWriteTime on an attribute-only change


marmos91:
Split out of #2172 (item 2) after investigation. Not fixed alongside the three clear gates in that
issue, so a red conformance run points at one change rather than four.

## What the spec says

MS-FSA 2.1.5.15.2 `FileBasicInformation`, the `If InputBuffer.FileAttributes != 0` branch, touches
exactly one timestamp:

> If Open.UserSetChangeTime is FALSE and InputBuffer.ChangeTime != -1, the object store MUST set
> Open.File.LastChangeTime to CurrentTime.

`LastModificationTime` appears in 2.1.5.15.2 only under `If InputBuffer.LastWriteTime != 0`, i.e.
when the client explicitly sets it. There is no SHOULD and no MAY for bumping it on an
attribute-only change. The branch's one outbound call, 2.1.4.19 (Updating Duplicated Information),
only *copies* `Link.LastModificationTime` from the file; it never sets a time to now. 2.1.4.17
("noting that a file modification time is updated") is not invoked from this section. The section's
only product note, `<179>`, is about FAT32 ignoring ChangeTime.

## What the code does

`internal/adapter/smb/handlers/set_info.go`:

```go
// ... the object store SHOULD also update LastWriteTime. ...
if fileAttrs != 0 && setAttrs.Mtime == nil && mtimeFT == 0 && !openFile.MtimeFrozen {
    now := time.Now()
    setAttrs.Mtime = &now
}
```

**The "SHOULD also update LastWriteTime" clause quoted in that comment is not in the cited
section.** It arrived in `bb07b48a1` ("resolve 20+ WPTS known failures") citing the fabricated
`2.1.5.14.2`, and #2174 renumbered the citation to the real `2.1.5.15.2` without re-checking the
claim underneath — a retarget of a known-bad citation onto a section that does not support it. That
is worth recording as a limitation of a pure renumbering pass.

## Why it is user-visible

Setting a DOS attribute (Hidden, ReadOnly) advances mtime, so every backup and sync tool treats the
file as content-modified. POSIX `chmod` updates ctime only, which is what the metadata layer already
does correctly for the NFS path — so the two protocols disagree about the same operation.

## Risk if fixed

Low. The two tests in `set_info_timestamp_preservation_test.go` pin the *skip guards*, not the bump
— both assert `Mtime == <explicit value>`, which still holds with the bump deleted. No
[...]

Issue #2179, opened 2026-08-26
Lint spec citations in CI: fail when a comment names a structure and cites a different section


marmos91:
## The problem

#2156 and #2174 together corrected roughly 300 wrong specification citations across the SMB
adapter. Both audits reached the same conclusion about which errors matter: a citation naming a
section that **does not exist** is annoying but self-announcing, while a citation naming a section
that **exists and says something else** survives every check a reader is likely to run. In #2174
the second category was 174 of 311 — more than twice the first.

Nothing in CI notices either kind, so the tree drifts back the moment someone writes
`[MS-FSCC] 2.4.x` from memory.

## The proposal

A large fraction of these are mechanically checkable, because **the comment names the structure
right next to the number it cites**. If a line says `FileEaInformation` and cites a section whose
spec title is `FileDispositionInformationEx`, that is decidable without reading the spec prose.

Concretely, for every `MS-FSCC`/`MS-FSA` citation in the tree:

1. If the cited section does not exist in that spec's section map → fail.
2. If the comment on that line (or the line above) names exactly one `File*Information`,
   `FILE_*_INFORMATION`, or `FSCTL_*` identifier, and exactly one section's title matches that
   identifier, and it is not the cited one → fail.
3. Otherwise → pass. Family references (`§2.4`, `§2.6`), prose claims, and anything ambiguous are
   deliberately out of scope. **This check only fires when the source has already told us the
   answer and disagrees with itself.**

## Worked example — what it caught in #2174

`internal/adapter/smb/handlers/requests.go` defines the info-class constants. Every citation in the
block was wrong, and the check flagged all eleven because each line names its own class:

| constant | cited | that section actually is | correct |
| --- | --- | --- | --- |
| `FileStandardInformation` | 2.4.41 | FileQuotaInformation | **2.4.47** |
| `FileInternalInformation` | 2.4.20 | FileIdAllExtdBothDirectoryInformation | **2.4.27** |
| `FileEaInformation` | 2.4.12 | FileDispositionInformationEx | **2.4.13** |
| `FileNameInformation` | 2.4.24 | FileIdFullDirectoryInformation | **2.4.32** |
| `FileRenameInformation` | 2.4.34 | FileNetworkOpenInformation | **2.4.42** |
| `FilePositionInformation` | 2.4.32 | FileNameInformation | **2.4.40** |
| `FileModeInformation` | 2.4.24 | FileIdFullDirectoryInformation | **2.4.31** |
[...]

PR #2184, opened 2026-08-26
fix(smb): apply the three SET_INFO access gates MS-FSA states as MUSTs

### LINKED ISSUE #2172: MS-FSA/MS-FSCC behaviour divergences surfaced by the #2155 citation audit
Surfaced while auditing MS-FSCC/MS-FSA citations for #2155. **None are fixed there** — that PR is
comment-only. Each of these is a place where the specification section that genuinely covers the
behaviour says something the code does not do, so repointing the comment at a section that matched
the code instead would have buried a real divergence.

**Verification status for everything below: code reading plus spec reading. Nothing here was
reproduced against a client or a torture case.** Confirm the premise before designing a fix — a
few of these are likely deliberate Samba-parity choices rather than defects.

Filed as one issue because they were found together and share that caveat; split it if any one
turns out to be worth its own fix.

---

### 1. `FileDispositionInformationEx` honours only bit 0

`internal/adapter/smb/handlers/set_info.go`. MS-FSCC §2.4.12 defines five flags; the code reads
`deletePending := (flags & 0x01) != 0` and discards the rest.

- `FILE_DISPOSITION_IGNORE_READONLY_ATTRIBUTE (0x10)` — spec: *"If set, allows files with the
  READ_ONLY attribute to be deleted anyway. Without this flag, deleting a read-only file MUST
  return STATUS_CANNOT_DELETE."* The read-only gate is applied unconditionally, so `0x11` against a
  read-only file returns `STATUS_CANNOT_DELETE` where the delete should be permitted. The MUST half
  is implemented; the escape hatch is not.
- `FILE_DISPOSITION_ON_CLOSE (0x08)` — the `STATUS_NOT_SUPPORTED` case is never checked.
- `FILE_DISPOSITION_POSIX_SEMANTICS (0x02)` — accepted and silently ignored.

### 2. An attribute-only SET_INFO bumps LastWriteTime

`set_info.go`. MS-FSA §2.1.5.15.2's attribute path touches exactly one timestamp: *"If
Open.UserSetChangeTime is FALSE and InputBuffer.ChangeTime != -1, the object store MUST set
Open.File.LastChangeTime to CurrentTime."* LastWriteTime appears nowhere in it. The code sets
`setAttrs.Mtime` when `fileAttrs != 0`, which NTFS does not do.

### 3. Rename restores Ctime, which §2.1.5.15.12 says MUST be updated

`set_info.go`, three sites. `saveTimestamps()` snapshots Mtime *and* Ctime and writes both back.
Preserving Mtime is spec-consistent — §2.1.5.15.12 never touches `Open.File.LastModificationTime`.
Preserving Ctime contradicts *"The object store MUST update Open.File.LastChangeTime."*

**Weakest of the set.** Product note `<187>` says NTFS and ReFS defer setting LastChangeTime until
the handle closes, so the observable behaviour may well match Windows. Worth measuring before
touching.

### 4. No `Open.Link.IsDeleted` guard on the rename path

`set_info.go`. MS-FSA §2.1.5.15.12: *"If Open.Link.IsDeleted is TRUE, the operation MUST be failed
with STATUS_ACCESS_DENIED."* No such pre-check exists; the rename succeeds and then clears
`openFile.DeletePending`.

`openFile.DeletePending` is the right analogue for `Open.Link.IsDeleted` — it is set by SET_INFO
`FileDispositionInformation`, while the `FILE_DELETE_ON_CLOSE` create option is tracked separately
as `InitialDeleteOnClose` and only promoted at CLOSE, matching §2.1.5.5 Phase 1. So this is a
missing guard rather than a create-option/disposition conflation.

### 5. `FileAllocationInformation` has no `FILE_WRITE_DATA` gate

`set_info.go` exempts `FileAllocationInformation` from the FILE_WRITE_ATTRIBUTES gate on the
grounds that it has its own check, and the handler arm never performs one — the
`WriteAuthorizedByHandle` assignment happens only on the truncating branch. MS-FSA §2.1.5.15.1:
*"If Open.GrantedAccess does not contain FILE_WRITE_DATA, the operation MUST be failed with
STATUS_ACCESS_DENIED."* A non-truncating SetAlloc on a read-only handle currently succeeds and
still fires the Level-II lease break.

### 6. `FILE_READ_ATTRIBUTES` is granted unconditionally

`pkg/metadata/auth_permissions.go`. MS-FSA §2.1.5.1.2.1 conditions the grant: *"…MUST set
Open.GrantedAccess.FILE_READ_ATTRIBUTES **if AccessCheck(SecurityContext,
Open.Link.ParentFile.SecurityDescriptor, FILE_LIST_DIRECTORY) returns TRUE**."* The code grants it
with no parent check, on both the explicit and MAXIMUM_ALLOWED paths.

#2155 updates the comment there to state the spec's condition and note that the code does not apply
it, rather than leaving the comment claiming the spec is unconditional. The `parent != nil` guard
already used for the DELETE override is the natural place if this is worth changing — but it may
be a deliberate `smb2.acls` accommodation.

---

### Also noted, lower confidence — code does more than the spec, comments overstated it

The #2155 comments were reworded to attribute these to their real source; no behaviour change is
implied and they may not be worth acting on.

- **SUPERSEDE injects the wrong bit.** §2.1.5.1.2 maps `FILE_SUPERSEDE` to
  `RemainingDesiredAccess.DELETE` and every other overwrite disposition to `FILE_WRITE_DATA`; the
  code requires `FILE_WRITE_DATA` for SUPERSEDE. That is Samba's `open_access_mask |=
  FILE_WRITE_DATA` on `O_TRUNC`.
- **Delete-on-close without DELETE.** §2.1.5.1 Phase 1 lists it under `STATUS_INVALID_PARAMETER`;
  the code returns `STATUS_ACCESS_DENIED`, following Samba.
- **Parent-directory Handle-lease strip on child create.** §2.1.4.12 with `Flags = PARENT_OBJECT`
  sets `BreakCacheState` to `(READ_CACHING|WRITE_CACHING)` — HANDLE_CACHING is not in that list.
  The H-strip is MS-SMB2/Samba dirlease behaviour.
- **Parent-directory atime bumped on write** (and on copychunk). Nothing in MS-FSA §2.1.5.4,
  §2.1.4.17 or §2.1.4.19 touches parent-directory timestamps on write.
- **Zero-length READ advances CurrentByteOffset.** §2.1.5.3 returns on `ByteCount == 0` before any
  offset or atime update.
- **CurrentByteOffset updated unconditionally.** §2.1.5.3 updates it only when
  `Open.Mode.FILE_SYNCHRONOUS_IO_ALERT` or `FILE_SYNCHRONOUS_IO_NONALERT` is TRUE — correct for
  SMB2, but the spec is not unconditional.

marmos91:
Closes #2172.

Fixes three of the six divergences in #2172. The other three are triaged in a comment there: two
were split out as #2181 and #2182, and one was withdrawn as my own misreading of the spec.

These three are the unambiguous MUSTs with no plausible Samba-parity defence. The conformance-risky
ones are deliberately **not** bundled in, so a red suite here points at one change rather than four.

## 1. `FileDispositionInformationEx` read only bit 0 of its Flags field

MS-FSCC §2.4.12 defines five flags. The handler read `flags & 0x01` and discarded the rest, so:

- **`FILE_DISPOSITION_IGNORE_READONLY_ATTRIBUTE` (0x10) did nothing.** The spec: *"If set, allows
  files with the READ_ONLY attribute to be deleted anyway. Without this flag, deleting a read-only
  file MUST return STATUS_CANNOT_DELETE."* DittoFS implemented the MUST-refuse half and not the
  documented escape hatch, so a client sending `DELETE|IGNORE_READONLY` (`0x11`) against a
  read-only file got `STATUS_CANNOT_DELETE`.
- **`FILE_DISPOSITION_ON_CLOSE` (0x08) was never checked.** The spec requires
  `STATUS_NOT_SUPPORTED` when it is set on a handle not opened `FILE_DELETE_ON_CLOSE`.

`FILE_DISPOSITION_POSIX_SEMANTICS` (0x02) stays accepted-and-unimplemented — honouring it means
keeping data streams readable through other handles after the link is unlinked, which is a feature,
not a gate. It is now stated in the code instead of being silently ignored.

The 1-byte `FileDispositionInformation` (class 13) shares this arm and carries no flags, so it can
never reach the waiver; the read-only refusal is unchanged for it.

## 2. Rename never checked `Open.Link.IsDeleted`

MS-FSA §2.1.5.15.12: *"If Open.Link.IsDeleted is TRUE, the operation MUST be failed with
STATUS_ACCESS_DENIED."* There was no such check — the rename succeeded and then cleared the
disposition.

**The subtle part is which field is the analogue.** It is the disposition set through SET_INFO
(`DeletePending`), not the `FILE_DELETE_ON_CLOSE` create option. That option is held per-handle as
`InitialDeleteOnClose` and promoted to `DeletePending` only during the CLOSE election, matching
§2.1.5.5 phase 1 — `doc_election.go` says so in as many words, and the reason is
`smb2.dirlease.unlink_*_initial_and_close`.

Conflating them would make every delete-on-close handle un-renameable, which is the one plausible
regression this change could cause, so it has its own test
(`TestSetInfo_Rename_AllowedOnDeleteOnCloseHandle`) that fails if the two are OR-ed together.

**Review caught that my first version was under-inclusive.** It read `openFile.IsDeletePending()`
— this open's own flag. But `IsDeleted` belongs to the *link*, and a disposition committed on one
handle is not copied onto its siblings until the CLOSE election runs, so a second open of the same
file still reported false and its rename went through. It now uses `h.isFileDeletePending()`, the
cross-handle scan the CREATE gate already relies on — which is also deliberately blind to
`InitialDeleteOnClose`, so widening the gate does not reintroduce the conflation above.
`TestSetInfo_Rename_RefusedOnSecondHandleWhenDeletePending` pins it, and goes red against my own
original single-handle version.

## 3. `FileAllocationInformation` had no access check at all

The `FILE_WRITE_ATTRIBUTES` gate in `SetInfo` exempts a list of classes on the stated grounds that
they "have specific access checks in their handlers". `FileAllocationInformation` was on that list
and had none — the only authorization in the arm is a `WriteAuthorizedByHandle` assignment on the
truncating branch. MS-FSA §2.1.5.15.1: *"If Open.GrantedAccess does not contain FILE_WRITE_DATA,
the operation MUST be failed with STATUS_ACCESS_DENIED."*

**The gate is placed before the Level-II lease break, not after.** That ordering is the part worth
reviewing: previously a handle with no write access could still fire `BreakReadLeasesOnWrite` and
invalidate another client's cached read state, which is a denial-of-service shape rather than a
tidiness issue.

## Every test was run against a broken build first

Two agents in the current batch shipped tests that passed against the exact bug they were written
to catch, so each gate here was removed and its test observed to fail before the test was trusted:

| mutation | test | result |
| --- | --- | --- |
| read-only waiver forced off | `..._IgnoreReadonlyAttribute` | RED |
| `ON_CLOSE` check → `if false` | `..._OnCloseRequiresDeleteOnClose` | RED |
| rename guard → `if false` | `..._Rename_RefusedWhenDeletePending` | RED |
| `FILE_WRITE_DATA` gate → `if false` | `..._SetAllocation_RequiresWriteData` | RED |
| `IsDeletePending() \|\| InitialDeleteOnClose` (the conflation) | `..._Rename_AllowedOnDeleteOnCloseHandle` | RED |
| rename gate reverted to the single-handle `openFile.IsDeletePending()` | `..._Rename_RefusedOnSecondHandleWhenDeletePending` | RED |

All green with the gates restored. Every test also carries a control case taking the opposite
branch, so none can pass by the gate simply never being reached.

## Also here: one comment correction from the citation audit

#2174 added a comment in `pkg/metadata/auth_permissions.go` describing MS-FSA §2.1.5.1.2.1's
`FILE_READ_ATTRIBUTES` clause as *conditional* on `AccessCheck(parent, FILE_LIST_DIRECTORY)`. It is
not — the clause follows the file-DACL loop, has no `else` and no masking step, and only ever
**adds** the bit. It is an override in the same shape as the DELETE-via-parent clause directly above
it, which this codebase already implements as an override.

That misreading is what produced item 6 of #2172, now withdrawn: the real divergence is confined to
the case where the file DACL denies the bit *and* the parent denies `FILE_LIST_DIRECTORY`, and the
current unconditional grant is load-bearing for `smb2.acls.OWNER`. A mischaracterised spec clause is
the same defect class #2155 existed to remove, so it is corrected here rather than left standing.

### DIFF
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -695,6 +695,18 @@ func (h *Handler) setFileInfoFromStore(
 			return setInfoStatus(types.StatusAccessDenied), nil
 		}
 
+		// Per MS-FSA 2.1.5.15.12 ("FileRenameInformation"): if Open.Link.IsDeleted
+		// is TRUE, the operation MUST be failed with STATUS_ACCESS_DENIED. The
+		// analogue of Open.Link.IsDeleted is the disposition set through SET_INFO
+		// FileDispositionInformation, not the FILE_DELETE_ON_CLOSE create option:
+		// that option is held per-handle as InitialDeleteOnClose and only promoted
+		// at CLOSE, matching 2.1.5.5 phase 1.
+		if openFile.IsDeletePending() {
+			logger.Debug("SET_INFO: rename of a link already marked for deletion",
+				"path", openFile.Name().Path)
+			return setInfoStatus(types.StatusAccessDenied), nil
+		}
+
 		// Before renaming, check that no other open handle on the same file
 		// conflicts with the rename: all other opens must have FILE_SHARE_DELETE
 		// (0x04) in ShareAccess. MS-FSA 2.1.5.15.12 ("FileRenameInformation")
@@ -1288,6 +1300,10 @@ func (h *Handler) setFileInfoFromStore(
 		}
 
 		var deletePending bool
+		// ignoreReadonly is only reachable through the Ex class; the 1-byte
+		// FileDispositionInformation carries no flags and so can never waive the
+		// read-only refusal below.
+		ignoreReadonly := false
 		if class == types.FileDispositionInformationEx {
 			// FileDispositionInformationEx uses a 4-byte Flags field per MS-FSCC 2.4.12 (FileDispositionInformationEx)
 			if len(buffer) < 4 {
@@ -1295,8 +1311,24 @@ func (h *Handler) setFileInfoFromStore(
 			}
 			dispR := smbenc.NewReader(buffer)
 			flags := dispR.ReadUint32()
-			// Bit 0 (FILE_DISPOSITION_DELETE) = delete on close
-			deletePending = (flags & 0x01) != 0
+
+			// Per MS-FSCC 2.4.12: FILE_DISPOSITION_ON_CLOSE updates the
+			// FILE_DELETE_ON_CLOSE state, and "if set and the file is not opened
+			// with FILE_DELETE_ON_CLOSE, STATUS_NOT_SUPPORTED MUST be returned".
+			// The create option is held per-handle as InitialDeleteOnClose.
+			if flags&types.FileDispositionOnClose != 0 && !openFile.InitialDeleteOnClose {
+				logger.Debug("SET_INFO: FILE_DISPOSITION_ON_CLOSE on a handle not opened delete-on-close",
+					"path", openFile.Name().Path)
+				return setInfoStatus(types.StatusNotSupported), nil
+			}
+
+			deletePending = flags&types.FileDispositionDelete != 0
+			ignoreReadonly = flags&types.FileDispositionIgnoreReadonlyAttribute != 0
+
+			// FILE_DISPOSITION_POSIX_SEMANTICS is accepted and not acted on: the
+			// link is removed from the namespace at close either way, and keeping
+			// the data streams readable through other handles after the unlink is
+			// not implemented.
 		} else {
 			deletePending = buffer[0] != 0
 		}
@@ -1324,8 +1356,11 @@ func (h *Handler) setFileInfoFromStore(
 				return setInfoStatus(types.StatusAccessDenied), nil
 			}
 
-			// Read-only files cannot be marked for deletion
-			if !openFile.IsDirectory {
+			// Read-only files cannot be marked for deletion unless the caller sent
+			// FILE_DISPOSITION_IGNORE_READONLY_ATTRIBUTE. Per MS-FSCC 2.4.12 that
+			// flag "allows files with the READ_ONLY attribute to be deleted
+			// anyway"; without it the refusal below is a MUST.
+			if !openFile.IsDirectory && !ignoreReadonly {
 				metaSvc := h.Registry.GetMetadataService()
 				file, fileErr := metaSvc.GetFile(authCtx.Context, openFile.MetadataHandle)
 				if fileErr == nil {
@@ -1537,6 +1572,19 @@ func (h *Handler) setFileInfoFromStore(
 
 	case types.FileAllocationInformation:
 		// FILE_ALLOCATION_INFORMATION [MS-FSCC] 2.4.4.
+		//
+		// Per MS-FSA 2.1.5.15.1 ("FileAllocationInformation"): if Open.GrantedAccess
+		// does not contain FILE_WRITE_DATA, the operation MUST be failed with
+		// STATUS_ACCESS_DENIED. This class is exempt from the FILE_WRITE_ATTRIBUTES
+		// gate above precisely because it carries this check, and the gate runs
+		// before the lease break so a handle that may not write cannot invalidate
+		// another client's cached read state.
+		if !hasAccessRight(openFile.GrantedAccess, uint32(types.FileWriteData)) {
+			logger.Debug("SET_INFO: allocation set without FILE_WRITE_DATA",
+				"path", openFile.Name().Path,
+				"grantedAccess", fmt.Sprintf("0x%x", openFile.GrantedAccess))
+			return setInfoStatus(types.StatusAccessDenied), nil
+		}
 		//
 		// Allocation size is not persisted (DittoFS does not preallocate), but
 		// per MS-FSA 2.1.5.15.1 ("FileAllocationInformation") and Samba `smbd_smb2_setinfo_lease_break_fsp_check`
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -1317,10 +1317,12 @@ func (h *Handler) setFileInfoFromStore(
 			dispR := smbenc.NewReader(buffer)
 			flags := dispR.ReadUint32()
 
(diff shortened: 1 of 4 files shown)


marmos91:
Both review comments were right and are fixed in 338a6093c.

`FILE_DISPOSITION_ON_CLOSE`: the comment said it "updates the FILE_DELETE_ON_CLOSE state", which is what MS-FSCC says the flag *means*, not what this handler does with it. The only thing done with it is the `STATUS_NOT_SUPPORTED` refusal — the disposition itself rides on the DELETE bit read just below, and no separate delete-on-close state is written. Reworded to say that, in both the constant's doc comment and the handler.

`FILE_DISPOSITION_DELETE`: "when clear, no flag is set and the file MUST NOT be deleted" conflated the bit being clear with the all-zero Flags word. MS-FSCC names the latter `FILE_DISPOSITION_DO_NOT_DELETE_FILE`, and that is what cancels a pending deletion; clearing this one bit says nothing about the others. Reworded.

Worth calling out that this is the same defect this PR's parent work exists to remove — a comment asserting behaviour the code does not have — arriving in the fix for it. Caught here by review rather than by anything mechanical, which is part of the argument in #2179 for a linter that can check a quoted clause against the section it cites.

PR #2190, opened 2026-08-26
test(block): the copy's RefCount bump reaches nothing — #2177 refuted, not fixed

### LINKED ISSUE #2177: A repeated server-side copy counts the source's hashes up every time

marmos91:
**The premise of #2177 does not hold, and this PR does not fix the leak —
because there is no leak to fix.** It records why, in code and in two tests, so
that the fix nobody should write does not get written later.

## What I went looking for

#2177 says a repeated server-side copy counts the source's hashes up every time
and never brings them down. `engine.CopyPayload` does bump `RefCount` once per
unique source hash while everything else it does is idempotent, so the shape of
the claim is right.

## What is actually true

That bump reaches nothing, on any backend, and never has.

The coordinator resolves the hash with `GetByHash`, and **every backend scopes
`GetByHash` to rows in the `Remote` state** — memory checks `IsRemote()`, sqlite
and postgres both spell it `state = 2 /* Remote */`, and sqlite's increment is
itself `UPDATE file_blocks SET ref_count = ref_count + 1 WHERE hash = ?1 AND
state = 2`. **No production path ever puts a row in that state.** The codebase
already says so twice, in comments written for other reasons:

> The carve path records synced markers via DefaultCommitBlock but never
> transitions FileChunk.State to BlockStateRemote (the row state remains
> Pending/Syncing for the life of the payload). The authoritative per-hash sync
> signal is therefore SyncedHashStore — not FileChunk.State.

So the lookup resolves nothing, `IncrementRefCount` returns
`ErrFileChunkNotFound`, and `CopyPayload` skips it as the tolerated miss it
already documents. The count never moves.

This is a structural kill, not a numeric one: `bs.coordinator.IncrementRefCount`
has exactly one production caller (this loop), its only implementation is gated
on a state transition no production code performs, and `AddRef` — the other way
a count could rise — has no production caller at all. Corroborated on a real
engine: three repeated copies, every row still `state=Pending refcount=0`.

## Why the obvious fix would have been actively harmful

Skipping the bump for a hash the destination already holds a row for makes a
repeat free. It also skips it when the destination genuinely held that hash
before the copy — cross-file dedup on a *first* run — which counts **down**.
Under-counting is the dangerous direction: a row reaped one decrement early is
data a file still needs. So the fix for a leak that cannot happen would have
traded it for corruption of a counter other things trust.

## What this PR leaves behind instead

- A `ponytail:` marker on the increment naming the ceiling (unreachable because
  rows never reach `Remote`; `SyncedHashStore` is the authoritative signal) and
  the upgrade path (whoever opens that gate owes this loop a way to tell a
  repeat from cross-file dedup, and the destination's rows cannot say which it
  is looking at).
- `TestCopyPayload_RefCountIncrementIsUnreachable` — three repeated copies leave
  every count where it started.
- `TestCopyPayload_RefCountWouldInflateOnceReachable` — moves one row to
  `Remote` by hand, confirms the lookup now resolves, and shows the count going
  up by exactly one per copy with nothing to bring it back. This is what makes
  the first test meaningful rather than vacuous, and it is what will fail the
  day someone turns the transition on without fixing the counting.

Verified against a deliberately broken build: opening the gate for every row in
the fixture makes the unreachability test fail with *"went from RefCount 0 to 3
over three copies; the increment is reachable after all, and it is not
idempotent."*

## Recommendation

#2177 should be closed as refuted rather than fixed, and this PR's marker and
tests are what carry the finding forward. The separate question this raises —
whether an increment that can never fire should exist at all, or whether the
`Remote` transition is itself the missing piece — is a design call I have not
made here.

Closes #2177

### DIFF
--- a/pkg/block/engine/copy_refcount_test.go
+++ b/pkg/block/engine/copy_refcount_test.go
@@ -1,0 +1,140 @@
+package engine_test
+
+import (
+	"context"
+	"math/rand"
+	"testing"
+
+	"github.com/marmos91/dittofs/pkg/block"
+	"github.com/marmos91/dittofs/pkg/block/engine"
+	remotememory "github.com/marmos91/dittofs/pkg/block/remote/memory"
+	metadatamemory "github.com/marmos91/dittofs/pkg/metadata/store/memory"
+)
+
+// refCounts returns each of a payload's rows' RefCount, keyed by row ID.
+func refCounts(t *testing.T, ctx context.Context, ms *metadatamemory.MemoryMetadataStore, payloadID string) map[string]uint32 {
+	t.Helper()
+	rows, err := ms.ListFileChunks(ctx, payloadID)
+	if err != nil {
+		t.Fatalf("ListFileChunks(%s): %v", payloadID, err)
+	}
+	if len(rows) == 0 {
+		t.Fatalf("%s has no rows, so the counts below would say nothing", payloadID)
+	}
+	out := make(map[string]uint32, len(rows))
+	for _, r := range rows {
+		out[r.ID] = r.RefCount
+	}
+	return out
+}
+
+// copyFixture writes and carves one source, and returns the engine, the store,
+// the two payload IDs and the ChunkRef list a copy would hand the destination.
+func copyFixture(t *testing.T) (*engine.Store, *metadatamemory.MemoryMetadataStore, string, string, []block.ChunkRef) {
+	t.Helper()
+	ctx := context.Background()
+	ms := metadatamemory.NewMemoryMetadataStoreWithDefaults()
+	bs, _ := openOfflineEngine(t, t.TempDir(), ms, remotememory.New())
+	t.Cleanup(func() { _ = bs.Close() })
+
+	root := createShare(t, ms, "refcount")
+	src, _ := createRealFile(t, ms, "refcount", "src.bin", root)
+	dst, _ := createRealFile(t, ms, "refcount", "dst.bin", root)
+
+	data := make([]byte, 2*1024*1024)
+	rand.New(rand.NewSource(5)).Read(data) //nolint:gosec // deterministic fixture
+	if _, err := bs.WriteAt(ctx, src, nil, data, 0); err != nil {
+		t.Fatalf("WriteAt: %v", err)
+	}
+	carve(t, bs, ctx, src)
+
+	var refs []block.ChunkRef
+	for _, r := range manifestRefs(t, ms, src) {
+		if !r.Hash.IsZero() {
+			refs = append(refs, r)
+		}
+	}
+	if len(refs) == 0 {
+		t.Fatal("the source carved no placeable rows")
+	}
+	return bs, ms, src, dst, refs
+}
+
+// TestCopyPayload_RefCountIncrementIsUnreachable records why a repeated
+// server-side copy does not inflate any RefCount today, so nobody sets out to
+// fix a leak that cannot happen and reaches for the one fix that would break
+// something real.
+//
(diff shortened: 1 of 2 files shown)


marmos91:
Both comments are the same point and both are right — fixed in `33d9fc61a`.

"The only RefCount increment in the system" was overstated. The stores do expose two ways to raise a count, `AddRef` and a store-level `IncrementRefCount`, and a breadcrumb that denies they exist would rot the moment someone greps for them.

The accurate claim, which is what I actually verified and what the wording now says: this is the only *production call site* that raises a RefCount. Nothing in production calls `AddRef` at all — every hit is a store implementation or the interface declaration — and the coordinator this loop goes through is the only caller of the store-level `IncrementRefCount`. That coordinator is the one gated on `GetByHash`, so the conclusion is unchanged: no production path raises a count today.

Naming the two other mechanisms explicitly is better than leaving them out, because it tells the next reader exactly what to re-check rather than asking them to trust an absolute. Corrected in the `ponytail:` marker and in the test comment that repeated it.

PR #2193, opened 2026-08-26
fix(smb): let rename update ChangeTime, keeping only an explicitly set one

### LINKED ISSUE #2182: SMB rename reverts ctime, discarding the update MS-FSA 2.1.5.15.12 requires

marmos91:
Closes #2182.

MS-FSA 2.1.5.15.12: *"The object store MUST update Open.File.LastChangeTime.`<187>`"*

`Move` already implements this correctly — `file_modify.go` sets `srcFile.Ctime = now` and never
touches the renamed file's `Mtime`. (The `Mtime`/`Ctime` pairs a few lines below it are the source
and destination **directories**, which is separate and correct.)

The handler then wrapped the call in a blanket snapshot-and-restore of both stamps. So:

- the **mtime** half restored a value nothing had changed — a no-op;
- the **ctime** half reverted the update the specification mandates.

## Why `<187>` does not excuse it

> Section 2.1.5.15.12: The file system only updates LastChangeTime if no user has explicitly set
> LastChangeTime. The NTFS and ReFS file systems defer setting LastChangeTime until the handle is
> closed.

That licenses **deferring** the update, and skipping it when the user set ChangeTime explicitly. It
does not license discarding it. DittoFS discarded: nothing anywhere defers a rename ctime to close,
so a client that reopened the file saw the pre-rename ChangeTime permanently. #2172 flagged this as
its weakest item on the theory that `<187>` might make the behaviour match Windows — it does not,
and *discard vs defer* is the distinction that settles it. From inside the handler the two look
identical; to a client that reopens the file they are not.

It was also cross-protocol inconsistent: an NFS rename goes through the same `Move` and reports the
ctime bump correctly, so the identical operation gave different answers per protocol.

## The straight deletion was wrong, and a control test caught it

#2182 was filed — by me — as "drop `saveTimestamps` and both call sites", on the reasoning that once
ctime is out, the mtime restore is a no-op and the whole helper is dead. The first half is right.
The second is not.

I wrote the deletion, and the control case for the `<187>` user-set half failed immediately: a
sentinel-frozen ChangeTime no longer survived a rename. `saveTimestamps` was doing double duty —
reverting the spec-mandated update *and* preserving an explicitly-set one — and nothing else covers
the renamed file, because the `restoreFrozenTimestamps` call already present on that path runs
against the **parent directories**, not the file.

So the blanket save/restore is replaced, not deleted, by `restoreFrozenTimestamps(authCtx,
openFile)` at both rename sites. That writes back only fields pinned by a `-1` sentinel — precisely
`Open.UserSetChangeTime` and nothing else — and lets `Move`'s ctime update stand.

Worth recording because the issue I filed prescribed the wrong fix, and only a test asserting the
behaviour I was about to remove surfaced it.

## Tests, and how they were checked

Both run against the **pre-change `set_info.go` from develop**, not a synthetic mutant:

| test | vs develop | vs this PR |
| --- | --- | --- |
| `..._UpdatesChangeTimeAndLeavesWriteTime` — unfrozen rename advances ctime, leaves mtime | **RED** (ctime reverted to the backdated value) | PASS |
| `..._KeepsFrozenChangeTime` — sentinel-frozen ctime survives a rename | PASS | PASS |

The split is the point: the first isolates the defect, the second pins the behaviour that must not
regress, and it already passed on develop — so it is a genuine guard rather than a restatement of
the fix. Removing the new `restoreFrozenTimestamps` call at the rename site turns it red.

The first test backdates the stamps through the metadata service rather than SET_INFO, deliberately:
an explicit SET_INFO would freeze the field and put the file in the `<187>` exempt case, testing the
opposite branch.

## Risk

No in-tree test pinned ctime across rename. The rename rows in `smbtorture/KNOWN_FAILURES.md` are
`smb2.lease.rename_wait` (a lease-ordering flake) and `smb2.streams.rename2`; neither asserts
timestamps. Torture sources are not available locally, so the conformance suites on this PR are the
check.

### DIFF
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -771,11 +771,6 @@ func (h *Handler) setFileInfoFromStore(
 			oldFileName := oldName.FileName
 			oldParentPath := GetParentPath(oldName.Path)
 
-			// Save mtime/ctime before the rename. MS-FSA 2.1.5.15.12 requires
-			// LastChangeTime to be updated; preserving it matches NTFS, which
-			// defers the update to handle close.
-			restoreTimestamps := h.saveTimestamps(authCtx, openFile.MetadataHandle)
-
 			// Perform the rename
 			metaSvc := h.Registry.GetMetadataService()
 			_, err = metaSvc.Move(authCtx, toDir, oldFileName, toDir, toName)
@@ -787,8 +782,11 @@ func (h *Handler) setFileInfoFromStore(
 				return setInfoStatus(common.MapToSMB(err)), nil
 			}
 
-			// Restore mtime/ctime after rename
-			restoreTimestamps()
+			// Move updates LastChangeTime, which MS-FSA 2.1.5.15.12 requires and
+			// which the old blanket save/restore reverted. Product note <187>
+			// exempts only a ChangeTime the user set explicitly, so restore just
+			// the sentinel-frozen fields rather than every stamp.
+			h.restoreFrozenTimestamps(authCtx, openFile)
 
 			// Clear delete-on-close after rename. Written under the handle
 			// lock: the delete-pending gates and the CLOSE delete-on-close
@@ -1183,12 +1181,6 @@ func (h *Handler) setFileInfoFromStore(
 		oldParentPath := GetParentPath(oldPath)
 		srcParentHandle := oldName.ParentHandle
 
-		// Save mtime/ctime before the rename so we can restore them after.
-		// MS-FSA 2.1.5.15.12 never touches LastModificationTime but does require
-		// LastChangeTime to be updated; preserving both matches NTFS, which
-		// defers that update to handle close.
-		restoreTimestamps := h.saveTimestamps(authCtx, openFile.MetadataHandle)
-
 		// Pre-overwrite the case-mismatched destination: Move's destination
 		// probe is exact-case GetChild(toName), so a destination that exists
 		// under a different casing (e.g. on disk "Foo.txt", client said
@@ -1213,8 +1205,11 @@ func (h *Handler) setFileInfoFromStore(
 			return setInfoStatus(common.MapToSMB(err)), nil
 		}
 
-		// Restore mtime/ctime after rename
-		restoreTimestamps()
+		// Move updates LastChangeTime, which MS-FSA 2.1.5.15.12 requires and which
+		// the old blanket save/restore reverted. Product note <187> exempts only a
+		// ChangeTime the user set explicitly, so restore just the sentinel-frozen
+		// fields rather than every stamp.
+		h.restoreFrozenTimestamps(authCtx, openFile)
 
 		// Per MS-FSA §2.1.5.15.2 ("FileBasicInformation"): Restore frozen timestamps on parent directories.
 		// Move updates both source and destination parent directory timestamps.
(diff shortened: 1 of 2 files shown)


marmos91:
**Marking this draft — the fix as written is wrong, and the conformance suite caught it.**

`smb2.rename.simple_modtime` fails on both backends (memory and badger-fs), deterministically, and
it is not a flake:

```
../../source4/torture/smb2/rename.c:1527:
  c1.out.change_time was Wed Aug 26 13:19:13 2026 UTC, expected Wed Aug 26 13:19:18 2026 UTC
```

`13:19:13` is when the file was created; `13:19:18` is what the server reports after the rename.
The test asserts the two are **equal** — that ChangeTime is unchanged across a rename. The upstream
header comment states its purpose as verifying that modify and change timestamps *survive a rename
while the handles remain open*, and the test does exactly that: it renames through `h1` and then
issues `smb2_getinfo_file` on that same still-open handle.

## This is the distinction I named in #2182 and then failed to implement

I wrote there that product note `<187>` does not excuse the old behaviour because *"DittoFS
discards where NTFS defers"* — and then implemented an **immediate** update, which is neither.
`<187>`:

> The NTFS and ReFS file systems defer setting LastChangeTime until the handle is closed.

While the handle is open, Windows still reports the pre-rename ChangeTime. That is precisely what
this test measures, and it is why the old blanket save/restore passed it. My change makes the new
value visible immediately, so the query through the open handle sees it and the assertion fails.

The diagnosis in #2182 stands — reopening the file after close showed the stale ChangeTime forever,
which is a real defect, and `Move` setting ctime is correct. What is wrong is *when* the update
becomes observable.

## What the fix has to be instead

Defer, don't apply: record on the OpenFile that a rename has occurred, and settle LastChangeTime at
CLOSE — which is also where `restoreFrozenTimestamps` and the pending-atime flush already run, so
there is an existing seam for it. `Open.UserSetChangeTime` (the sentinel-frozen case) must still
win over the deferred update.

That is a larger change than this PR, and it needs its own conformance run rather than being
improvised on top of a red one. **Not doing it here without a steer** — say the word and I will
take it as a fresh branch, or park #2182 with this analysis recorded.

## Note on my own tests

`TestSetInfo_Rename_UpdatesChangeTimeAndLeavesWriteTime` asserts the update is visible immediately
through the same handle. That assertion is wrong in the same way the fix is, and would have to
become "visible after close" under the deferral design. My control test
(`..._KeepsFrozenChangeTime`) is unaffected and still correct.

Worth recording that the unit tests were mutation-checked and still did not catch this: they
verified the change did what I intended, and what I intended was wrong. Only a real client
expectation — Samba's own test — encoded the distinction.

marmos91:
Superseded by #2197, which implements the deferred shape: the rename records what it owes on the handle and CLOSE settles it, so the still-open handle keeps reading the pre-rename value that `smb2.rename.simple_modtime` asserts.

Closing rather than force-pushing over it — the wrong approach and the test run that caught it are worth keeping legible, since the comment above is the clearest record of why an immediate update is not deferral.

Issue #2196, opened 2026-08-26
restoreFrozenTimestamps does not preserve a frozen ChangeTime across rename + CLOSE


marmos91:
Found while building the regression coverage for #2182. **Pre-existing on develop, unrelated to
that change** — reproduced against unmodified develop before being filed.

## The claim that does not hold

`close.go` carries this, and it is the design intent:

> After flushing pending writes (which may overwrite frozen timestamps), restore any timestamps
> that were frozen via SET_INFO -1. The deferred commit flush sets Mtime/Ctime to the WRITE time,
> but if the handle has frozen timestamps, those must be preserved.

That is MS-FSA product note `<187>`'s user-set clause — *"the file system only updates
LastChangeTime if no user has explicitly set LastChangeTime"* — and the `Open.UserSetChangeTime`
guard in 2.1.5.15.2. `restoreFrozenTimestamps` is supposed to deliver it.

It does not, at least for ChangeTime on a handle that closes after a rename.

## Reproduction

Against a block-store-backed fixture (`setupReparseShare`, needed so CLOSE reaches its flush step
rather than bailing out early):

1. open a file, set `CtimeFrozen = true` and `FrozenCtime = <2021-02-03>` on the handle;
2. rename through that handle;
3. CLOSE — asserting `STATUS_SUCCESS`, so the test cannot pass by CLOSE failing early;
4. read the file back.

`Ctime` is the flush time, not the pinned value. No
`restoreFrozenTimestamps: restoring` debug line appears, so the restore is not reaching its write —
either the branch it sits in is not entered on this path, or its `SetFileAttributes` is failing
silently (that call's error is logged, not returned).

Same result with and without the deferred-ChangeTime write added by #2182's replacement PR, and the
same result on unmodified develop — which is what establishes it as pre-existing.

## Why it was not pinned in the #2182 PR

A test asserting the frozen value survives would have been asserting a bug rather than the
behaviour. It is noted in place in `set_info_rename_ctime_test.go` with a pointer here instead.

It also decided a design question there: a `!IsCtimeFrozen()` guard on the new CLOSE-time write was
considered and **deliberately left out**, because the flush already sets Ctime unconditionally on
this path — so the guard could never be observed failing, which is the "guard that never fired"
anti-pattern. Freeze semantics should have one owner, and that owner is `restoreFrozenTimestamps`.
Fixing it here fixes it for every caller rather than for the rename path alone.

## Scope worth checking when this is picked up

[...]

PR #2197, opened 2026-08-26
fix(smb): settle a rename's ChangeTime at CLOSE instead of discarding it

### LINKED ISSUE #2182: SMB rename reverts ctime, discarding the update MS-FSA 2.1.5.15.12 requires

marmos91:
Closes #2182. Replaces #2193, which was the same diagnosis with the wrong remedy.

MS-FSA 2.1.5.15.12 requires a rename to update `LastChangeTime`. `Move` already does — it sets
`Ctime` on the renamed file and never touches `Mtime`. The handler then snapshotted both stamps
around the call and wrote them back, so the update was never persisted: a file reopened after close
kept its pre-rename ChangeTime permanently, while the NFS path renaming through the same `Move`
reported it correctly.

## Why the obvious fix is wrong

Letting `Move`'s write stand — #2193 — fails `smb2.rename.simple_modtime` on both backends:

```
c1.out.change_time was 13:19:13 (creation), expected 13:19:18 (post-rename)
```

The test renames through `h1` and then queries **that same still-open handle**, asserting ChangeTime
is unchanged. That is product note `<187>`:

> The NTFS and ReFS file systems defer setting LastChangeTime until the handle is closed.

So the update must happen, but must not be observable until the handle closes. Discarding it and
applying it immediately are both wrong, in opposite directions.

## The shape

The rename records what it owes on the handle; CLOSE settles it. This mirrors the pending-atime
mechanism sitting immediately above it in `close.go` — `noteSmbRenameChange` / `takeSmbPendingCtime`
against `noteSmbAccess` / `takeSmbPendingAtime`.

One deliberate difference: `SmbPendingAtime` is surfaced by QUERY_INFO through `applySmbPendingAtime`,
and `SmbPendingCtime` is **not**. Overlaying it would defeat the point — while the handle is open the
pre-rename value is exactly what the client must read. That asymmetry is the whole fix, so it is
stated on the field.

A `!IsCtimeFrozen()` guard on the CLOSE-time write was considered and left out. The deferred write
flush already sets Ctime unconditionally on this path, so the guard could never be observed failing
— the "guard that never fired" anti-pattern. Freeze semantics get one owner,
`restoreFrozenTimestamps`, which turns out to be independently broken here → **#2196**, filed
separately and reproduced on unmodified develop.

## Both halves are pinned, and each was run against a build missing it

A fix satisfying only the first is indistinguishable from the original bug; one satisfying only the
second breaks the torture case.

| mutation | test | result |
| --- | --- | --- |
| CLOSE settle removed | `..._ChangeTimeSettlesOnClose` | RED |
| rename writes immediately (restore removed) | `..._ChangeTimeUnchangedWhileHandleOpen` | RED |

Both green with the change intact. The two tests use different fixtures on purpose — the
open-handle half never closes, so it uses the simpler share; the close half needs the
block-store-backed one, without which CLOSE fails its flush step for a regular file and the
assertion would pass vacuously. That second point was found the hard way: an earlier version of the
close test passed while CLOSE was returning `STATUS_INTERNAL_ERROR` unchecked, so close status is
now asserted.

## Method note, because this PR is the argument for it

#2193 was mutation-tested and still shipped the wrong behaviour. Mutation testing proves a test is
not vacuous. It cannot prove the specification the test encodes is the right one — and there the
encoded spec was wrong, so every check agreed with itself. The reasoning was right, the code was
wrong, and the code looked like the reasoning: `<187>` was quoted correctly in the issue, used as
the entire argument for why the old behaviour was wrong, and then implemented as an immediate
update, which is neither deferral nor the status quo.

Only an independent encoding of the behaviour caught it — Samba's `simple_modtime`, which asserts
ChangeTime is *unchanged* across a rename on a still-open handle. The conclusion is not "be more
careful": it is that a spec reading which changes observable protocol behaviour should be run
against conformance **before** it is trusted, not after it is merged. The two halves pinned here
exist so the next person changing this cannot satisfy one and believe they are done.

### DIFF
--- a/internal/adapter/smb/handlers/close.go
+++ b/internal/adapter/smb/handlers/close.go
@@ -290,6 +290,25 @@ func (h *Handler) Close(ctx *SMBHandlerContext, req *CloseRequest) (*CloseRespon
 				}
 			}
 
+			// Settle the LastChangeTime a rename on this handle owed the file.
+			// MS-FSA 2.1.5.15.12 requires the update; product note <187> defers it
+			// to close on NTFS and ReFS, and holding it until here is what lets a
+			// client reading through the still-open handle keep seeing the
+			// pre-rename value.
+			//
+			// No frozen check here. restoreFrozenTimestamps below owns freeze
+			// semantics for every field, and a second guard would duplicate that
+			// ownership while being impossible to observe failing: the deferred
+			// write flush above already sets Ctime unconditionally, so a
+			// sentinel-frozen ChangeTime does not survive this path with or without
+			// one. That is a pre-existing gap in the frozen restore, not something
+			// this write introduces.
+			if pending := takeSmbPendingCtime(openFile); !pending.IsZero() {
+				if _, ctimeErr := metaSvc.SetFileAttributes(authCtx, openFile.MetadataHandle, &metadata.SetAttrs{Ctime: &pending}); ctimeErr != nil {
+					logger.Warn("CLOSE: deferred LastChangeTime flush failed", "path", closePath, "error", ctimeErr)
+				}
+			}
+
 			// Per MS-FSA §2.1.5.15.2 ("FileBasicInformation"): After flushing pending writes (which may overwrite
 			// frozen timestamps), restore any timestamps that were frozen via SET_INFO -1.
 			// The deferred commit flush sets Mtime/Ctime to the WRITE time, but if the
--- a/internal/adapter/smb/handlers/handler.go
+++ b/internal/adapter/smb/handlers/handler.go
@@ -613,6 +613,11 @@ type OpenFile struct {
 	// persisted at CLOSE.
 	SmbAtimeWrittenAt time.Time // when the last READ-driven atime reached the store
 	SmbPendingAtime   time.Time // newest access time not yet written to the store (zero ⇒ none)
+	// SmbPendingCtime is the LastChangeTime a rename on this handle owes the
+	// file, settled at CLOSE (zero ⇒ none). It is deliberately NOT surfaced by
+	// QUERY_INFO: while the handle is open the pre-rename ChangeTime is what
+	// the client must still see.
+	SmbPendingCtime time.Time
 
 	// SmbParentAtimeWrittenAt bounds the parent-directory LastAccessTime bump a
 	// WRITE performs, the way SmbAtimeWrittenAt bounds the file's. A suppressed
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -781,6 +781,20 @@ func (h *Handler) setFileInfoFromStore(
 			// Restore mtime/ctime after rename
 			restoreTimestamps()
 
+			// Move set LastChangeTime and restoreTimestamps has just put the
+			// pre-rename value back, which is what a client reading through this
+			// still-open handle must see. MS-FSA 2.1.5.15.12 still requires the
+			// update, so record it and let CLOSE settle it, matching the NTFS and
+			// ReFS deferral described in product note <187>.
+			noteSmbRenameChange(openFile, time.Now())
+
+			// Move set LastChangeTime and restoreTimestamps has just put the
+			// pre-rename value back, which is what a client reading through this
+			// still-open handle must see. MS-FSA 2.1.5.15.12 still requires the
+			// update, so record it and let CLOSE settle it, matching the NTFS and
+			// ReFS deferral described in product note <187>.
+			noteSmbRenameChange(openFile, time.Now())
+
 			// Clear delete-on-close after rename. Written under the handle
 			// lock: the delete-pending gates and the CLOSE delete-on-close
 			// election read this field under it.
--- a/internal/adapter/smb/handlers/set_info_rename_ctime_test.go
+++ b/internal/adapter/smb/handlers/set_info_rename_ctime_test.go
@@ -1,0 +1,139 @@
+package handlers
+
+import (
+	"testing"
+	"time"
+
+	"github.com/marmos91/dittofs/internal/adapter/smb/types"
+	"github.com/marmos91/dittofs/pkg/metadata"
+)
+
+// MS-FSA 2.1.5.15.12 requires a rename to update LastChangeTime, and product
+// note <187> says NTFS and ReFS defer that update until the handle closes. Both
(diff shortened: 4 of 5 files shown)


marmos91:
**Draft again. The evidence I offered for this PR does not hold, and I am not shipping it.**

Copilot's fourth comment started the unravelling, and pulling that thread took the rest of the PR
with it.

## 1. The record call was in the wrong place — twice

Both `noteSmbRenameChange` calls landed in the **stream-rename** branch. My insertion matched
`\t\trestoreTimestamps()` as a substring of the 3-tab stream site, so the file-rename path — the
common case — recorded nothing. Copilot caught it; I had not.

Fixed, and then the fix exposed the real problem.

## 2. My earlier mutation evidence was a build error I misread as a test failure

The "close-settle-off → RED" line in the PR body was produced by replacing the condition with
`if false {`, which leaves `pending` unused. Go refuses to compile that. `go test` printed `FAIL`,
my check was `grep -q FAIL`, and I recorded a passing mutation test.

**A mutant that does not compile is not a mutant.** Every mutation result in this PR needed a build
check before the test run, and did not have one.

## 3. With build-safe mutants, the close-settlement test is vacuous

Three mutations, each verified to compile first:

| mutation | result |
| --- | --- |
| CLOSE settlement disabled (`&& false`) | **still green** |
| file-path record removed | **still green** |
| stream-path record removed | **still green** |

None of my mechanism is load-bearing for that test. The ChangeTime advance it observes comes from
the deferred-write flush earlier in CLOSE.

## 4. Which puts #2182's premise in doubt

The close-settlement test **passes against unmodified develop**. In that fixture, ChangeTime is
already advanced after rename + close — so the "stale pre-rename ChangeTime forever" behaviour I
described in #2182 is not what happens there.

It may still hold for a file with no deferred writes, where CLOSE never reaches the flush. I could
not test that: the fixture without a block store cannot close a regular file at all
(`STATUS_INTERNAL_ERROR`), which is what pushed me onto the flushing fixture in the first place.

So the honest position is: **I have not demonstrated that the defect exists**, and the one
observation I have points the other way.

## What survives

- The open-handle half (`..._ChangeTimeUnchangedWhileHandleOpen`) is real and build-safe-mutation-verified:
  removing the post-rename restore turns it red. But that is develop's existing behaviour, not
  something this PR adds.
- Copilot's backwards-Ctime point was valid and is fixed (the write now only advances, mirroring the
  atime flush).
- #2196 — `restoreFrozenTimestamps` not preserving a frozen ChangeTime — reproduced on unmodified
  develop and is unaffected by any of this.

## What I would do next, not in this PR

Establish whether the defect exists before designing for it again: a fixture that can close a
regular file **without** deferred writes, or a live server rename-close-reopen with the ChangeTime
read at each step. If ChangeTime already settles on close today, #2182 should be closed as not
reproducible and this PR dropped entirely.

Third iteration on this issue, and the pattern is mine: I keep encoding what I believe the
behaviour is and checking that the code matches the belief. Handing it back rather than trying a
fourth shape without measurement.

marmos91:
Closing unmerged. Both tests in this PR pass against unmodified develop, so nothing here changes
behaviour either of them can observe, and the model it is built on does not match what the code does.

Measured on develop: a rename advances Ctime immediately, while the handle is still open; CLOSE with
no rename moves nothing. The deferral this PR adds is settling a debt that was already paid at rename
time, and the pre rename value the open handle half asserts is delivered by a restore that only
sometimes runs.

Root cause is #2205: `saveTimestamps` discards a `PermissionDenied` from its restoring
`SetFileAttributes`, so which ChangeTime a client reads turns on file mode. #2182 closed as not
reproducible. #2196 stands on its own.

Fourth shape on this path would have been another guess. Measurement first this time.

PR #2210, opened 2026-08-27
fix(nfs4): reject the byte ranges RFC 7530 calls invalid in LOCK

chessman:
# Description

## Type of change

## Changes Made

Validate LOCK, LOCKT, LOCKU ranges.

Makes 6 pynfs tests green.

## Testing

## Checklist

### DIFF
--- a/internal/adapter/nfs/v4/handlers/lock.go
+++ b/internal/adapter/nfs/v4/handlers/lock.go
@@ -3,6 +3,7 @@ package handlers
 import (
 	"bytes"
 	"io"
+	"math"
 
 	"github.com/marmos91/dittofs/internal/adapter/nfs/v4/pseudofs"
 	"github.com/marmos91/dittofs/internal/adapter/nfs/v4/state"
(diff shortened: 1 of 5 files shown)


marmos91:
Nice catch on the range rule, and the arithmetic in `lockRangeValid` is right. I checked
the boundaries (offset <= MaxUint64 minus length, with all ones exempt) and they hold.
`LOCKT` is in the right place too, since it carries no `seqid`.

The problem is `LOCK` and `LOCKU`. Both return `NFS4ERR_INVAL` before the locker4 and its
`seqid` are parsed, so the failure never reaches `consumeSeqidOnError` and the owner
sequence never moves, while the client's does. `state/openowner.go:96` already spells
this out and names `NFS4ERR_INVAL` specifically: the server falls one behind forever and
every later op for that owner comes back `NFS4ERR_BAD_SEQID`.

Reproduced both on your branch (v4.0 only, v4.1 sets `SkipOwnerSeqid`):

```
    LOCK ok(lock_seqid 1), LOCK bad range(2) = INVAL, LOCK good(3) = 10026 BAD_SEQID
    LOCK bad range(open_seqid N+1) = INVAL, LOCK good(open_seqid N+2) = 10026 BAD_SEQID
```

The second is the nastier one: a single malformed `LOCK` also wedges `OPEN` and `CLOSE` for
that open owner.

Could you move the `LOCK` and `LOCKU` checks down into `state.LockNew`, `LockExisting` and
`Unlock`, next to `validateOpenModeForLock` (`manager.go:2113` and `2249`)? Same class of
validation, and it sits deliberately after the defer `consumeSeqidOnError` lines, so the
sequence gets handled for you. Leave the `LOCKT` check where it is.

Unrelated thing I hit while testing, already true on develop so not yours to fix here:
nothing translates the `NFSv4` all ones length into pkg/metadata/lock, where zero is the
to EOF `convention. rangeLast(100, MaxUint64)` wraps to 98, so a `WRITE` lock at offset 100
through EOF followed by a `LOCKT` at 200 for 100 bytes returns `NFS4_OK` instead of `DENIED`.
I will open an issue for it. Might be worth softening the "however long the file grows"
comment until that lands.

Everything else looks good. Ordering after `RequireCurrentFH` is correct, the tests fail
properly when I neuter the zero length branch, vet and gofmt clean.

chessman:
Right, moved the validation.

marmos91:
Move verified: putting the range check back at the top of `LockNew` reproduces your wedge (`LOCK after invalid range status = 10026`), and where you've put it both `LOCK` and `LOCKU` consume the owner seqid and the next op succeeds. `LOCKT` stayed in the right place.

One fix before merge: this branch still carries `9a5c5bf71`, the pre-revision version of #2211's ACCESS commit, so `access.go` and `access_test.go` conflict between the two PRs. Could you rebase onto develop so this carries just the lock work?

Two nits: `lock_test.go` still says the LOCKT check "sits in the handler" — it's in `state.TestLock` now. And the all-ones wrap is now #2216 (reproduced: `rangeLast(100, MaxUint64) = 98`), so the "through end-of-file" comment at `lockowner.go:179` overpromises until that lands.

chessman:
Rebased + fixed nits.

PR #2219, opened 2026-09-01
fix(nfs4): deliver delegation recalls — AUTH_SYS credential and the client's callback_ident

### LINKED ISSUE #2218: NFSv4 callback RPCs use AUTH_NULL, so no delegation can ever be recalled from a Linux client
## Summary

DittoFS sends **every** NFSv4 backchannel callback RPC with `AUTH_NULL` credentials. The Linux
NFS client's callback service accepts `AUTH_NULL` for the `CB_NULL` procedure **only**, and denies
every other procedure with `MSG_DENIED` / `AUTH_ERROR` / `AUTH_BADCRED`.

The consequence is self-concealing: the `CB_NULL` probe that sets `CBPathUp = true` — and thus
enables delegation granting for that client — is the one and only callback the client will ever
accept. `CBPathUp` therefore certifies a backchannel that cannot carry a single recall. **No NFSv4
delegation granted to a Linux client can ever be recalled.**

## Observable failure

`TestNLMAxisInterop` (E2E) fails two of its four axes on `develop`:

```
NLM_vs_SMB     conflict=BLOCKED  afterRelease=ACQUIRED
NLM_vs_NFSv4   conflict=BLOCKED  afterRelease=ACQUIRED
SMB_vs_NLM     conflict=BLOCKED  afterRelease=BLOCKED   <-- fail
NFSv4_vs_NLM   conflict=BLOCKED  afterRelease=BLOCKED   <-- fail
NLM_AXIS_RESULT=FAIL
```

Run 33512142367. Reproduced on a Scaleway VM at `ff14b24cb`.

## The failure is not in the release path

The verdict block reads as "release over SMB/NFSv4 does not free a waiting NLM lock", but that is
not what happens. Extra probes added to the driver, run on the same rig, isolate it:

```
PROBE idle-lock.dat-NLM:            BLOCKED    <-- nothing held by anyone
PROBE idle-lock.dat-NLM-otherrange: BLOCKED    <-- offset 5000, a range no case ever touched
PROBE idle-lock.dat-SMB:            ACQUIRED
PROBE idle-lock.dat-V4:             ACQUIRED
PROBE virgin-file-NLM:              ACQUIRED
SMB_vs_NLM_FRESH   conflict=BLOCKED afterRelease=ACQUIRED   <-- passes on a fresh file
NFSv4_vs_NLM_FRESH conflict=BLOCKED afterRelease=ACQUIRED   <-- passes on a fresh file
```

So: with every lock released, NLM cannot lock `lock.dat` at **any** byte range, while SMB and NFSv4
lock the same file freely — and both cases pass on a file the earlier cases never touched. That is a
lingering **whole-file** row that conflicts with the NLM owner but not with the NFSv4 owner. Per
`UnifiedLock.ConflictsWith`, a delegation is exempted from conflict only for its own `ClientID`,
which is exactly that signature. Both failing axes have **one** cause, and it is residual state on
`lock.dat`, not a release that failed to propagate.

## Server-side evidence

Server log (DEBUG) from the reproducing run:

```
13:59:55 [DEBUG] CB_NULL succeeded, delegations enabled for client client_id=7680566251659198465
14:01:18 [INFO]  Delegation granted client_id=7680566251659198465 deleg_type=2 stateid_seqid=1
14:01:18 [DEBUG] NFSBreakHandler: dispatching CB_RECALL delegationID=fb5e4dea-... clientID=7680566251659198465
14:01:18 [WARN]  CB_RECALL failed client_id=7680566251659198465 error=callback reply: RPC call denied: reply_stat=1
14:01:22 [DEBUG] NFSv4 DELEGRETURN stateid_seqid=1     <-- 4s later, unprompted; long past the test window
```

Packet capture of the same run (`tcpdump` on the server-side veth):

```
Frame 353  10.50.0.1 -> 10.50.0.2  NFS CB  Type:Call  CB_COMPOUND
    Credentials  Flavor: AUTH_NULL (0)
    Verifier     Flavor: AUTH_NULL (0)
Frame 355  10.50.0.2 -> 10.50.0.1  NFS CB  Type:Reply CB_COMPOUND
    Reply State:  denied (1)
    Reject State: AUTH_ERROR (1)
    Auth State:   bad credential (seal broken) (1)
```

This is the only denied RPC reply in the entire capture.

Linux `fs/nfs/callback.c`, `nfs_callback_authenticate()`:

```c
rqstp->rq_auth_stat = rpc_autherr_badcred;
switch (rqstp->rq_authop->flavour) {
case RPC_AUTH_NULL:
        if (rqstp->rq_proc != CB_NULL)
                return SVC_DENIED;
        break;
case RPC_AUTH_UNIX:
        break;
```

The early return happens before the status is reset, so the client answers `AUTH_BADCRED` — exactly
what frame 355 carries.

## Mechanism, end to end

1. The NFSv4 client opens `lock.dat` for write and is granted a WRITE delegation — a whole-file row
   in `unifiedLocks`.
2. An NLM lock arrives. `BreakLeasesForByteRangeLock` marks the delegation `Breaking` and dispatches
   `CB_RECALL`, which the client **denies**.
3. `WaitForByteRangeLockBreak` deliberately does not wait on delegation breaks (deadlock safety), so
   `AddUnifiedLock` immediately sees the still-present whole-file delegation row, owned by a
   different `ClientID` → `NLM4_DENIED`.
4. `breakDelegations` skips any row already marked `Breaking`, so no later NLM attempt re-dispatches
   the recall. The row survives until `OpLockBreakScanner`'s delegation timeout — orders of magnitude
   longer than the test's window, and in production a long stall on a file that is in fact free.

## Source

`internal/adapter/nfs/v4/state/callback_common.go`, `BuildCBRPCCallMessage` — the single seam all
three callback senders route through (`callback.go:232` CB_COMPOUND v4.0, `callback.go:275` CB_NULL,
`backchannel.go:297` CB_COMPOUND v4.1):

```go
// Auth credentials: AUTH_NULL (flavor=0, length=0)
_ = xdr.WriteUint32(&buf, uint32(rpc.AuthNull))
_ = xdr.WriteUint32(&buf, 0)
```

## Long-standing, not a regression

The `AUTH_NULL` credential has been in `BuildCBRPCCallMessage` since `428711632` (2026-02-25), when
the callback client was introduced. `git log -S` over the credential lines returns that commit and
no other; the file has had exactly one unrelated touch since (`5f2abcbb8`, 2026-06-13). No lock-path
change caused this. It has never worked against a Linux client.

It stayed invisible because the only thing that ever exercised the backchannel was the `CB_NULL`
probe — the single procedure the credential is valid for. The health check and the payload used
different credentials, so the check proved nothing about the payload.

## Scope beyond the test

Any NFSv4 delegation recall against a Linux client: a conflicting SMB open, a conflicting NFSv4 open
from another client, a cross-protocol byte-range lock, `CB_NOTIFY` directory delegations, and the
v4.1 backchannel. All of them dispatch through the same builder and all of them are denied.

marmos91:
Closes #2218

`TestNLMAxisInterop` fails two of its four axes on `develop`. The verdict block reads as a release
bug, and it is not one.

```
NLM_vs_SMB     conflict=BLOCKED  afterRelease=ACQUIRED
NLM_vs_NFSv4   conflict=BLOCKED  afterRelease=ACQUIRED
SMB_vs_NLM     conflict=BLOCKED  afterRelease=BLOCKED   <-- fail
NFSv4_vs_NLM   conflict=BLOCKED  afterRelease=BLOCKED   <-- fail
NLM_AXIS_RESULT=FAIL
```

## What is actually wrong

**No NFSv4 delegation could ever be recalled from a Linux client.** Every recall was rejected at the
RPC layer, so a granted whole-file delegation stayed on the file and denied foreign byte-range locks
until a scanner timeout swept it. Both failing axes are that one lingering row, not two release bugs.

Two independent defects gate the same path, which is why fixing either alone changes nothing:

| | AUTH_NULL cred | AUTH_SYS cred |
| --- | --- | --- |
| **callback_ident 0** | denied (flavor) | denied (ident) |
| **callback_ident echoed** | denied (flavor) | **delivered** |

**1. Callbacks were sent with AUTH_NULL.** Linux's callback service accepts AUTH_NULL for the
`CB_NULL` procedure only and answers `AUTH_BADCRED` to every other one
(`fs/nfs/callback.c`, `nfs_callback_authenticate`).

**2. `callback_ident` was decoded and thrown away, then hardcoded to 0.** For NFSv4.0 the client
resolves an incoming `CB_COMPOUND` to one of its mounts by that ident and reports a miss as
`badcred` — a misleading error, since the credential is fine and the *identifier* is wrong
(`fs/nfs/callback_xdr.c`, `nfs4_callback_compound` → `out_invalidcred`; `net/sunrpc/svc.c`,
`svc_process_common` re-checks `rq_auth_stat` after the dispatch). Linux allocates idents from 1
upward, so 0 never matched. `SendCBRecall` now echoes the value from SETCLIENTID. The v4.1
backchannel keeps 0 — the kernel marks the field "ignored by v4.1 and v4.2".

`CB_NULL` slipped through both gates: AUTH_NULL is explicitly allowed for it, and it carries no
`CB_COMPOUND` header and therefore no ident. So the one probe that set `CBPathUp` — and thereby
enabled delegation granting — was the only callback the client would ever accept. **The health check
certified a backchannel that could not carry a single recall**, and the failure only ever surfaced
once something needed recalling.

## Evidence

Server log, pre-fix, from the reproducing run:

```
CB_NULL succeeded, delegations enabled for client client_id=7680566251659198465
Delegation granted client_id=7680566251659198465 deleg_type=2
NFSBreakHandler: dispatching CB_RECALL delegationID=fb5e4dea-... clientID=7680566251659198465
CB_RECALL failed client_id=7680566251659198465 error=callback reply: RPC call denied: reply_stat=1
NFSv4 DELEGRETURN stateid_seqid=1        <-- 4s later, unprompted, long past the test window
```

Packet capture of the same run — the only denied RPC reply in it:

```
Call   CB_COMPOUND   Credentials Flavor: AUTH_NULL (0)
Reply  CB_COMPOUND   Reply State: denied (1) / Reject State: AUTH_ERROR (1) / Auth State: bad credential (1)

client SETCLIENTID   callback_ident: 0x00000001
our    CB_COMPOUND   callback_ident: 0x00000000
```

Extra probes added to the driver isolate the residual state and rule out the release path:

```
PROBE idle-lock.dat-NLM:            BLOCKED    <-- nothing held by anyone
PROBE idle-lock.dat-NLM-otherrange: BLOCKED    <-- offset 5000, a range no case ever touched
PROBE idle-lock.dat-SMB:            ACQUIRED
PROBE idle-lock.dat-V4:             ACQUIRED
PROBE virgin-file-NLM:              ACQUIRED
SMB_vs_NLM_FRESH   conflict=BLOCKED afterRelease=ACQUIRED   <-- passes on a fresh file
NFSv4_vs_NLM_FRESH conflict=BLOCKED afterRelease=ACQUIRED   <-- passes on a fresh file
```

With every lock released, NLM cannot lock the file at any range while SMB and NFSv4 lock it freely,
and both cases pass on a file the earlier cases never touched. That is a whole-file row conflicting
with the NLM owner but not the NFSv4 one — the delegation exemption in `UnifiedLock.ConflictsWith`
is keyed on `ClientID`, which is exactly that signature.

## Verified on hardware, two builds, one rig

The driver needs root, veth and rpcbind, so this was run on a VM with a real kernel NFSv3/NLM client
in a network namespace. Pre-fix build: the four-line FAIL above. Post-fix build, unmodified driver:

```
NLM_vs_SMB     conflict=BLOCKED  afterRelease=ACQUIRED
NLM_vs_NFSv4   conflict=BLOCKED  afterRelease=ACQUIRED
SMB_vs_NLM     conflict=BLOCKED  afterRelease=ACQUIRED
NFSv4_vs_NLM   conflict=BLOCKED  afterRelease=ACQUIRED
NLM_AXIS_RESULT=PASS
```

An intermediate build with only the credential fixed still failed identically, with AUTH_SYS
verified on the wire — that is what established that both defects are required.

## Long-standing, not a regression

Both defects date to `428711632` (2026-02-25), when the callback client was introduced. `git log -S`
over the credential lines and over `encodeCBCompound(0, …)` returns that commit and nothing else. No
recent lock-path change is involved; delegation recall has never worked against a Linux client. It
stayed hidden because the only thing exercising the backchannel was the probe that both gates let
through.

## Scope beyond this test

Every NFSv4 delegation recall against a Linux client: a conflicting SMB open, a conflicting NFSv4
open from another client, a cross-protocol byte-range lock, and `CB_NOTIFY` directory delegations.
The credential fix also applies to the v4.1 backchannel, which shares the message builder.

## Also in here

Two OPEN log lines reported the *attempted* delegation type even when the lock manager refused the
grant (`deleg == nil`, which encodes `OPEN_DELEGATE_NONE` on the wire). They read as "granted a write
delegation" on an OPEN that granted none, which cost real time during this investigation. They now
report what is encoded. Wire behaviour is unchanged.

## Tests

Three regression tests, each run against a deliberately broken build and confirmed to compile:

- `TestBuildCBRPCCallMessage_CredIsParseableAuthSys` — flavor is AUTH_SYS and the body parses as
  `authsys_parms` with every field present. Killed by reverting the credential and by dropping the
  trailing gids word.
- `TestBuildCBRPCCallMessage_ProbeAndPayloadShareCredential` — pins the invariant the probe depends
  on: `CB_NULL` must carry the same credential as the callbacks it vouches for. Killed by a mutant
  that gives the probe AUTH_NULL and the payload AUTH_SYS, i.e. the exact latent shape.
- `TestSendCBRecall_EchoesCallbackIdent` and `TestSetClientID_StoresCallbackIdent` — both halves of
  the ident wiring. Killed by hardcoding the ident to 0 at each end.

A mutant that dropped the ident in SETCLIENTID initially failed to compile rather than fail an
assertion; it was rewritten into a compiling form, which then survived until
`TestSetClientID_StoresCallbackIdent` was added.

Two backchannel tests parsed the outgoing callback at a hardcoded 40-byte RPC header. They now skip
the credential and verifier by their declared length, so they no longer depend on the flavor in use.

### DIFF
--- a/internal/adapter/nfs/v4/handlers/open.go
+++ b/internal/adapter/nfs/v4/handlers/open.go
@@ -516,7 +516,7 @@ func (h *Handler) handleOpenClaimNull(
 		"createMode", createMode,
 		"stateid_seqid", openResult.Stateid.Seqid,
 		"rflags", openResult.RFlags,
-		"delegation", delegType,
+		"delegation", encodedDelegType(deleg),
 		"client", ctx.ClientAddr)
 
 	// Encode OPEN4resok
@@ -605,7 +605,7 @@ func (h *Handler) handleOpenClaimFH(
 	logger.Debug("NFSv4 OPEN CLAIM_FH successful",
 		"stateid_seqid", openResult.Stateid.Seqid,
 		"rflags", openResult.RFlags,
-		"delegation", delegType,
+		"delegation", encodedDelegType(deleg),
 		"client", ctx.ClientAddr)
 
 	// CLAIM_FH does not create or change a directory, so change_info is empty.
@@ -1009,4 +1009,14 @@ func effectiveUIDGID(authCtx *metadata.AuthContext) (uint32, uint32) {
 		}
 	}
 	return uid, gid
+}
+
+// encodedDelegType reports the delegation type that OPEN actually puts on the
+// wire. A refused grant leaves deleg nil, which encodes as OPEN_DELEGATE_NONE —
+// distinct from the type the open attempted.
+func encodedDelegType(deleg *state.DelegationState) uint32 {
+	if deleg == nil {
+		return types.OPEN_DELEGATE_NONE
+	}
+	return deleg.DelegType
 }
--- a/internal/adapter/nfs/v4/handlers/ops_test.go
+++ b/internal/adapter/nfs/v4/handlers/ops_test.go
@@ -1388,3 +1388,38 @@ func TestReadDir_NoCurrentFH(t *testing.T) {
 			decoded.Status, types.NFS4ERR_NOFILEHANDLE)
 	}
 }
+
+// TestSetClientID_StoresCallbackIdent pins that SETCLIENTID keeps the client's
+// callback_ident. An NFSv4.0 client matches an incoming CB_COMPOUND to one of
+// its mounts by this value, so discarding it here leaves every later callback
+// carrying an identifier the client rejects, with nothing on the acquire path
+// to reveal it.
+func TestSetClientID_StoresCallbackIdent(t *testing.T) {
+	h := newTestHandlerWithShares([]string{"/export"})
+	ctx := newOpsTestContext()
+
+	data := encodeCompoundWithOps("", 0, []encodedOp{
+		encodeSetClientID("ident-test-client"),
+	})
+
+	resp, err := h.ProcessCompound(ctx, data)
+	if err != nil {
+		t.Fatalf("ProcessCompound error: %v", err)
+	}
+
+	decoded, _ := decodeCompoundResp(resp)
+	if decoded.Results[0].Status != types.NFS4_OK {
+		t.Fatalf("SETCLIENTID status = %d, want NFS4_OK", decoded.Results[0].Status)
+	}
+
+	clientID := binary.BigEndian.Uint64(decoded.Results[0].ExtraData[0:8])
+	client := h.StateManager.GetClient(clientID)
+	if client == nil {
+		t.Fatalf("no client record for id %d", clientID)
+	}
+
+	// encodeSetClientID sends callback_ident = 1.
+	if client.Callback.Ident != 1 {
+		t.Errorf("stored callback ident = %d, want 1 (the value the client sent)", client.Callback.Ident)
+	}
+}
--- a/internal/adapter/nfs/v4/handlers/setclientid.go
+++ b/internal/adapter/nfs/v4/handlers/setclientid.go
@@ -65,7 +65,7 @@ func (h *Handler) handleSetClientID(ctx *types.CompoundContext, reader io.Reader
 	}
 
 	// Read callback_ident (uint32)
-	_, err = xdr.DecodeUint32(reader)
+	callbackIdent, err := xdr.DecodeUint32(reader)
 	if err != nil {
 		return &types.CompoundResult{
 			Status: types.NFS4ERR_BADXDR,
@@ -79,6 +79,7 @@ func (h *Handler) handleSetClientID(ctx *types.CompoundContext, reader io.Reader
 		Program: cbProgram,
 		NetID:   cbNetID,
 		Addr:    cbAddr,
+		Ident:   callbackIdent,
 	}
 
 	// Delegate to StateManager for the five-case algorithm
--- a/internal/adapter/nfs/v4/state/backchannel_test.go
+++ b/internal/adapter/nfs/v4/state/backchannel_test.go
@@ -392,17 +392,7 @@ func TestBackchannelSender_SequenceIDIncrement(t *testing.T) {
 		}
 		xid := binary.BigEndian.Uint32(body[0:4])
 
-		// Parse past RPC header (10 uint32s = 40 bytes) to get to CB_COMPOUND args
-		// XID(4) + msg_type(4) + rpc_vers(4) + prog(4) + vers(4) + proc(4) +
-		// cred_flavor(4) + cred_len(4) + verf_flavor(4) + verf_len(4) = 40 bytes
-		// Then CB_COMPOUND: tag_len(4) + minorversion(4) + callback_ident(4) +
-		// op_count(4) + first_op_code(4) = 20 bytes
-		// CB_SEQUENCE args start at offset 60
-		// CB_SEQUENCE args: sessionID(16 bytes) + seqID(4 bytes)
-		if len(body) < 60+16+4 {
-			t.Fatalf("body too short: %d bytes", len(body))
-		}
-		seqID := binary.BigEndian.Uint32(body[60+16 : 60+16+4])
+		seqID := cbSequenceSeqID(t, body)
 
 		replyBody := buildMockCBCompoundReplyBody(xid)
 		pending.Deliver(xid, replyBody)
@@ -472,11 +462,11 @@ func TestBackchannelSender_SeqIDAndXIDAreIndependent(t *testing.T) {
 		if _, err := io.ReadFull(clientConn, body); err != nil {
 			t.Fatalf("read body: %v", err)
 		}
-		if len(body) < 60+16+4 {
+		if len(body) < 4 {
 			t.Fatalf("body too short: %d bytes", len(body))
 		}
 		xid = binary.BigEndian.Uint32(body[0:4])
-		seqID = binary.BigEndian.Uint32(body[60+16 : 60+16+4])
+		seqID = cbSequenceSeqID(t, body)
 
 		pending.Deliver(xid, buildMockCBCompoundReplyBody(xid))
 		return xid, seqID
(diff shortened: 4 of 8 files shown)


marmos91:
### Post-fix server-side evidence (positive, not just the verdict)

Same rig, post-fix build. The recall now lands and the delegation comes back inside the same second,
where pre-fix it was denied and the client only returned it ~4s later, unprompted:

```
CB_NULL succeeded, delegations enabled for client client_id=7680574184463794177
NFSBreakHandler: dispatching CB_RECALL delegationID=b026b699-... clientID=7680574184463794177
CB_RECALL sent successfully addr=10.50.0.2:45073 program=1073741824
CB_RECALL sent successfully client_id=7680574184463794177 deleg_type=2
Delegation returned client_id=7680574184463794177 deleg_type=2 kind=file
delegation denied by lock manager ... error=delegation denied: file recently had caching broken
```

That last line is the `isRecentlyRecalled` guard doing its job: having just been recalled, the file
does not immediately hand out another delegation, so the NLM lock the recall was made for actually
gets its turn.

The diagnostic probes that isolated the defect are all clear now — including the two that were stuck
regardless of byte range:

```
PROBE idle-lock.dat-NLM:            ACQUIRED   (was BLOCKED)
PROBE idle-lock.dat-NLM-otherrange: ACQUIRED   (was BLOCKED)
PROBE idle-lock.dat-SMB:            ACQUIRED
PROBE idle-lock.dat-V4:             ACQUIRED
PROBE virgin-file-NLM:              ACQUIRED
```

### Not flaky

Four consecutive full runs of the unmodified driver on the post-fix build, all four axes green in
every one:

```
run 1  NLM_AXIS_RESULT=PASS
run 2  NLM_AXIS_RESULT=PASS   (diag variant, same four axes)
run 3  NLM_AXIS_RESULT=PASS
run 4  NLM_AXIS_RESULT=PASS
```

PR #2235, opened 2026-09-01
refactor(metadata): one body per file read instead of four

marmos91:
Part of #1828.

Six handle-addressed reads — `GetFile`, `GetChild`, `GetParent`, `GetLinkCount`,
`ListChildren`, `GetFileByPayloadID` — existed **four times each**: a store copy
and a transaction copy, in sqlite and in postgres. They are pure reads, so the
only thing that differs between a store copy and its transaction twin is which
executor runs the statement, which is exactly what `Core.X` already is.

They now live once, on `store/sql.Core`, promoted onto all four types. Net
−1,229 lines of production code.

### What had to become a hook

`Dialect` gains two members:

- **`MapError(err, operation, path)`** — dispatches to each package's own
  `mapDBError` / `mapPgError`. The classifiers stay split (postgres reads
  SQLSTATE via `errors.As(&pgErr)` and carries five classes sqlite has no
  analogue for); only the entry point is shared.
- **`Files() *FileQueries`** — the six statements per dialect. These differ by
  more than placeholder syntax: the path column and the block-ref aggregate are
  rewritten per dialect, and postgres matches a payload id through
  `content_id_hash = md5($1)` because a content id for a path near `PATH_MAX`
  overruns its 2704-byte btree key limit.

### Drift the four copies had already accumulated

Merging them settles three asymmetries rather than preserving them:

- the transaction `GetFile` carried a Debug log ("GetFile retrieved") the pool
  path never had, in both dialects. **Dropped** — it was a dev aid for a past
  file-type investigation, and the caller has the same fields.
- the store `GetFileByPayloadID` had no `ctx.Err()` check; its transaction twin
  did. Now both do.
- `ListChildren`'s nlink fallback was an `if`/nested-`if`; it is a `switch` now,
  same semantics.

### The coverage hole this found

A transaction-level file read that escapes to the **pool** — losing
read-your-writes and surviving a rollback — passed the entire postgres
integration suite. Nothing in the suite reads a file inside a transaction that
wrote it, so a pooled read answers every existing question correctly.

`storetest.FileReadTxOps` closes it for all four backends: a write is visible to
a read in the same transaction before commit, and a rolled-back write is visible
to nobody afterwards. Verified against the deliberately broken build — the
mutation survives without it and fails with it. (It fails by wedging on the
connection the transaction still holds, so the symptom is the package timeout,
not an assertion; that is noted in the test.)

### Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean
- `go test -tags=integration -p 1 ./pkg/metadata/...` green against a real
  postgres 16, plus badger / sqlite / memory
- the new conformance test passes on all four backends and refuses the broken one

### DIFF
--- a/pkg/metadata/store/postgres/dialect.go
+++ b/pkg/metadata/store/postgres/dialect.go
@@ -36,4 +36,60 @@ var chunkQueries = storesql.ChunkQueries{
 	EnumerateHashes:    enumerateHashesQuery,
 }
 
+// MapError translates a driver error into an ExportError. It is this
+// package's own mapPgError, reached through the Dialect so a shared body can
+// classify an error without importing the driver that produced it.
+func (dialect) MapError(err error, operation, path string) error {
+	return mapPgError(err, operation, path)
+}
+
+// Files returns the postgres file and directory read statements.
+func (dialect) Files() *storesql.FileQueries { return &fileQueries }
+
+// inodeSelectColumns is the full inode projection GetFile and
+// GetFileByPayloadID share, block-ref aggregate included, in the column order
+// sqlcodec.FileRowToFileWithNlinkAndBlocks scans.
+const inodeSelectColumns = `
+	f.id, f.share_name, ` + inodePathExpr + `,
+	f.file_type, f.mode, f.uid, f.gid, f.size,
+	f.atime, f.mtime, f.ctime, f.creation_time,
+	f.content_id, f.link_target, f.device_major, f.device_minor,
+	f.hidden, f.acl, f.eas, f.object_id,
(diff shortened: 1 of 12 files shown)


marmos91:
On the `SetParent` finding: confirmed, and it is the same four-copy problem this PR is unwinding — three of the four implementations return `ctx.Err()`, the postgres transaction returns bare `nil`. Rather than patch the outlier, the stacked follow-up collapses `SetParent` onto `store/sql.Core` (it is a no-op on both dialects at both levels, since `SetChild` maintains the parent edge), which removes the divergence by construction. Left out of this PR to avoid restarting a nearly-green CI run for a one-line write in a reads-only change.

PR #2239, opened 2026-09-01
refactor(metadata): one body for handle, capability and statistics reads

marmos91:
Part of #1828. Stacked on #2237.

Three more four-copy methods onto `store/sql.Core`: `GenerateHandle`,
`GetFilesystemCapabilities` and `GetFilesystemStatistics`. `Dialect` gains one
statement, `Statfs`, which lets both dialect-local `statfs.go` files go.

### A pointer into the store's own capabilities

The pool path returned `&s.capabilities` — a pointer to the field itself, so a
caller could edit the capabilities every other caller reads. Its transaction
twin already copied first. The shared body copies, which is the safe half of
the pair.

Also settled: the pool `GetFilesystemStatistics` had no `ctx.Err()` check; its
twin did. Now both do.

### The statfs test got better rather than deleted

`postgres/statfs_test.go` pinned a real invariant — statfs must never report the
store-wide total for one share — by asserting the SQL *text* contained
`share_name = $1`. That test cannot survive the query moving onto the dialect,
and it only ever covered postgres.

Replaced by `storetest.StatfsOps/CountsOnlyItsOwnShare`, which pins the
behaviour instead: fill a *second* share and assert the first share's numbers do
not move. It covers all four backends — sqlite, badger and memory never had this
coverage — and it tests the property rather than one spelling of it.

Verified against a deliberately broken build: with the share predicate replaced
by a tautology, sqlite goes red. On a single-share store, which is what most
tests build, an unscoped statfs looks perfectly healthy, so the second share is
what does the work.

### Verification

- `go build ./...`, `go vet ./...`, `gofmt` clean
- `go test -tags=integration -p 1 ./pkg/metadata/...` green against a real
  postgres 16, plus badger / sqlite / memory


marmos91:
Good catch on the zero-size probe files — that assertion could not have failed. `createTestFile` leaves `Size` at zero, so a store-wide `SUM(size)` over the probe rows is zero, which is exactly what a correctly scoped query returns; only the file count was doing any work.

Fixed in c2a7dfd93: both shares' files now carry sizes. Verified with a mutation that scopes the **count** but not the **sum** — `UsedBytes` now reports `share A's byte count moved when share B was filled` on its own, so each half of the assertion is independently load-bearing rather than one carrying the other.

Issue #2084, opened 2026-08-24
Invalidate marks a range cold without a durable cold-log entry; eviction then loses it


marmos91:
## Summary

`Invalidate` marks a range cold in memory without appending to `cold.log`, and the eviction path's cold-log scan deliberately skips intervals that are already cold. An interval marked cold by `Invalidate` therefore still owns a live record in a segment, but will never get a durable cold marker — and when that segment is evicted and unlinked, a restart loses the range entirely. It comes back as an absent interval, which the read path serves as zeros. Same failure shape as #1850 / #1879.

Found while adding a remote-only residency count for #1848; the count is what led to reading these two paths together. Not caused by that change.

## The two halves

`Invalidate` (`pkg/block/journal/store.go:1150`) sets the bit and returns — no `appendCold`:

```go
for k := range fi.ivs {
    iv := &fi.ivs[k]
    if iv.end() <= off || iv.fileOff >= end || iv.cold || !iv.synced {
        continue
    }
    iv.cold = true
}
```

`evictSegment` (`pkg/block/journal/reclaim.go:238`) builds its cold-log entries from intervals that are **not** already cold:

```go
if fi.ivs[k].loc.SegmentID == seg.id && !fi.ivs[k].cold {
    entries = append(entries, coldEntry{...})
}
```

That skip is right for the case it was written for: an already-cold interval owns no record in this segment, because eviction put it there and already logged it. `Invalidate` breaks that assumption — it is the one path that sets `cold` on an interval whose record is still live in a segment on disk.

So: `appendCold` writes no entry for it, `retireSegment` unlinks the segment holding its only copy, and recovery finds neither a record nor a cold marker. `applyColdLog` cannot resurrect what was never logged. The range becomes a hole.

The comment above `evictSegment` states the invariant this violates exactly — "they are the only remaining record that those ranges hold data, so a crash between marking and unlinking must not be able to lose them."

## Reachability

The only caller is `healCorruptWarmRead` (`pkg/block/engine/read_internal.go:83`), which demotes a range whose on-disk bytes failed their checksum and immediately re-hydrates it from the remote. On the happy path the hydrate lands a fresh clean record at a higher version and the cold interval is superseded, so the window closes in milliseconds.

It does not close when the hydrate fails — `ensureAndReadFromLocal` returns an error and the interval is left cold-in-memory with no durable marker, indefinitely. The obvious way to get there is a remote that is unreachable, which is also a plausible time to be reading from a degraded local tier. From that state:

[...]

PR #2245, opened 2026-09-02
fix(journal): truncate the torn cold.log tail at recovery

### LINKED ISSUE #2230: journal: torn cold.log tail is never truncated at recovery, poisoning every future append
`applyColdLog` (`recovery.go:358-386`) calls `loadCold` (`cold.go:192-212`), which returns the intact prefix of `cold.log` — but **never truncates the file on disk**.

`loadSegment` does exactly this repair for `.seg` files (`fd.Truncate(validUpTo)` + `Sync`, `recovery.go:243-250`). There is no equivalent for `cold.log`. `appendCold` reopens with `O_APPEND` (`cold.go:138`), so new entries land *after* the surviving garbage tail.

The next restart's `loadCold` walks from offset 0, parses the pre-crash intact entries, reaches the still-present garbage, and stops (`cold.go:203-207`) — **permanently discarding every entry the intervening process appended.**

`compactColdLog` (`recovery.go:395-405`), the only caller of `rewriteCold`, is gated by `coldLoaded <= 2*len(live) + coldCompactFloor(1024)`, so it never fires on realistic small or medium logs. The torn tail is never repaired.

A lost cold entry means a silent zero-fill instead of a remote fetch — the failure mode `cold.go`'s own doc exists to prevent.

## Fix

Have `loadCold` return the intact byte offset. In `applyColdLog`, truncate + fsync `cold.log` at that offset before any later append, mirroring `loadSegment`'s active-tail repair.

## Notes

- Verified: no wholesale `cold.log` rewrite outside `compactColdLog`.
- Reachable via every production cold path: `Evict` (`reclaim.go:280`), `SeedCold`/`SeedColdBatch` (`store.go:531`, `:625`, `:1290`).

Source: `.planning/2026-09-01-journal-audit-report.md` (verified finding, `gaps`/recovery).

marmos91:
`cold.log` is the only durable record that a byte range is safe on the remote
block store but holds no local bytes. Lose an entry and the range comes back
after restart as a POSIX hole: the read zero-fills and never fetches. That is
the silent-zeros family from #1850 / #2084.

## What was wrong

`loadCold` stops at the first torn entry and returns the intact prefix — but
nothing truncated the file back to that point. `appendCold` reopens the log
`O_WRONLY|O_CREATE|O_APPEND` (`cold.go:138`), so every entry written after the
crash lands *behind* the surviving garbage.

The next recovery walks from offset 0, replays the pre-crash entries, hits the
same garbage, and stops there — permanently discarding everything the
intervening process appended. One torn write silently kills every cold interval
recorded from then on, for the life of the store.

`loadSegment` already does exactly this repair for an active `.seg`
(`fd.Truncate(validUpTo)` + `Sync`, `recovery.go:243-250`). There was no
equivalent for `cold.log`.

## Premise verification

Every claim was checked against unmodified `develop` (`e2ce799a3`) before any
code was written:

- **`loadCold` never truncates, and no caller does it on its behalf.**
  Confirmed. It has exactly one non-test caller, `applyColdLog`
  (`recovery.go:359`), which only replays the returned entries.
- **`cold.log` really is `O_APPEND` on the path that runs after recovery.**
  Confirmed at `cold.go:138`, reached from `Evict` (`reclaim.go:280`),
  `SeedCold` (`store.go:531`) and `SeedColdBatch` (`store.go:625`, `:1290`).
- **Nothing else rewrites the log wholesale.** `rewriteCold` has exactly one
  caller, `compactColdLog`.
- **Would compaction incidentally heal a torn tail?** Yes — `rewriteCold` is
  temp-file + rename, so a compaction pass does rebuild the file from the live
  set. But it is gated on `r.coldLoaded > 2*len(live) + coldCompactFloor`, and
  `coldCompactFloor` is 1024. The log must carry more than ~1024 dead entries
  beyond twice the live set before the heal can fire. Until then the poisoning
  is permanent, and every restart inside that window loses another process's
  worth of cold intervals. Small and medium logs never heal at all.

The premise holds in full.

## The fix

`loadCold` now returns the byte offset just past the last intact entry, and
`applyColdLog` drops and fsyncs everything after it before anything can append
— the same shape as the active-segment repair. Recovery is the only point where
the log is known to end at an intact entry, so it is the only place this can be
done.

```go
func loadCold(dir string) ([]coldEntry, int64, error)
func truncateColdTail(dir string, validUpTo int64) error
```

Truncation failure fails the open rather than warning. Continuing would append
onto a log already known to be poisoned, which is exactly the loss this change
exists to prevent. That is deliberately unlike `compactColdLog`, where a failure
only costs redundant replay and a warn is the right call.

### This changes `Open`'s failure surface, deliberately

Worth stating plainly, because it is an availability change and not only a
correctness one. `applyColdLog` propagates `truncateColdTail`'s error, so a
`cold.log` that cannot be opened `O_WRONLY` — a read-only mount, permissions
drift, a snapshot restored with tight modes — now refuses to open a store that
previously opened and merely logged a warning.

That is the intended trade. The alternative is to open anyway and let the next
`Evict` or `SeedCold` append behind a tail already known to be garbage, which
silently destroys every cold interval written from then on and surfaces later as
zero-filled reads with nothing in the log. Refusing to open is loud, immediate
and repairable; the other branch is silent and permanent. Note the two cases it
does *not* catch: a missing log returns nil, and an intact log costs one open
and one stat, so the new failure mode is reachable only when the log exists and
the process genuinely cannot write to it.

Anyone later tempted to soften this to a warning should move the repair rather
than drop it — the invariant `appendCold` depends on ("only the tail can tear")
has to be restored before any append, and recovery is the only point where the
log is known to end at an intact entry.

## Verification

`TestColdLogTornTailIsRepairedOnOpen` seeds a cold range, appends a truncated
entry header to simulate a torn write, reopens, seeds a *second* range, reopens
again, and asserts both ranges still read as cold.

Against a build with only the `truncateColdTail` call removed (test unchanged):

```
=== RUN   TestColdLogTornTailKeepsIntactEntries
--- PASS: TestColdLogTornTailKeepsIntactEntries (0.07s)
=== RUN   TestColdLogTornTailIsRepairedOnOpen
[WARN] journal: cold log torn, keeping the intact entries offset=32 intact_entries=1 err=journal: torn record: cold entry runs past end of log
[WARN] journal: cold log torn, keeping the intact entries offset=32 intact_entries=1 err=journal: torn record: cold entry CRC mismatch
    cold_test.go:258: cold interval for "g" was lost: the read reports a hole and zero-fills instead of fetching from the remote
--- FAIL: TestColdLogTornTailIsRepairedOnOpen (0.10s)
FAIL
```

The second warning is the poisoning itself: the torn header plus the entry
appended after it re-parse as one CRC-bad entry, so the second recovery stops at
the same offset and `g` is gone.

With the fix:

```
=== RUN   TestColdLogTornTailKeepsIntactEntries
[WARN] journal: cold log torn, keeping the intact entries offset=32 intact_entries=1 err=journal: torn record: cold entry runs past end of log
[WARN] journal: dropping torn cold log tail valid_up_to=32 dropped_bytes=31
--- PASS: TestColdLogTornTailKeepsIntactEntries (0.10s)
=== RUN   TestColdLogTornTailIsRepairedOnOpen
[WARN] journal: cold log torn, keeping the intact entries offset=32 intact_entries=1 err=journal: torn record: cold entry runs past end of log
[WARN] journal: dropping torn cold log tail valid_up_to=32 dropped_bytes=31
--- PASS: TestColdLogTornTailIsRepairedOnOpen (0.10s)
PASS
```

`go test ./pkg/block/journal/ -race` and `go test ./pkg/block/...` are green
(the arm64-only `TestBLAKE3FasterThanSHA256` CPU gate in `pkg/block` fails on
this laptop and is unrelated).

## Adjacent defects seen, not fixed here

- `rewriteCold` leaves `cold.log.tmp` behind on every one of its error returns.
- `appendCold` skips the directory fsync forever after its first failure: the
  `O_CREATE` open makes the file even when `fsyncDir` fails, so the next call
  sees it already exists and never retries.

Both are separate audit findings and out of scope for this change.

Closes #2230

### DIFF
--- a/pkg/block/journal/cold.go
+++ b/pkg/block/journal/cold.go
@@ -188,17 +188,20 @@ func (s *Store) appendCold(entries []coldEntry) error {
 
 // loadCold reads every intact entry from dir's cold log. A missing log is not an
 // error (no eviction or seed has happened yet). It stops at the first torn entry
-// and returns what precedes it.
-func loadCold(dir string) ([]coldEntry, error) {
+// and returns what precedes it, along with the byte offset just past the last
+// intact entry: anything between that offset and the end of the file is garbage
+// the caller is expected to drop with truncateColdTail before appending.
+func loadCold(dir string) ([]coldEntry, int64, error) {
 	raw, err := os.ReadFile(filepath.Join(dir, coldLogName))
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return nil, nil
+			return nil, 0, nil
 		}
-		return nil, fmt.Errorf("journal: read cold log: %w", err)
+		return nil, 0, fmt.Errorf("journal: read cold log: %w", err)
 	}
 	var out []coldEntry
-	for off := 0; off < len(raw); {
+	off := 0
+	for off < len(raw) {
 		e, n, derr := decodeColdEntry(raw[off:])
 		if derr != nil {
 			logger.Warn("journal: cold log torn, keeping the intact entries",
@@ -208,7 +211,39 @@ func loadCold(dir string) ([]coldEntry, error) {
 		out = append(out, e)
 		off += n
 	}
-	return out, nil
+	return out, int64(off), nil
+}
+
+// truncateColdTail drops whatever follows the last intact entry and makes the
+// truncation durable. The log is only ever opened for append, so a torn tail
+// left in place sits in front of every later entry: the next load stops at the
+// same offset and discards everything appended in between, turning one torn
+// write into the permanent loss of every cold interval recorded after it.
+func truncateColdTail(dir string, validUpTo int64) error {
+	fd, err := os.OpenFile(filepath.Join(dir, coldLogName), os.O_WRONLY, 0o644)
+	if err != nil {
+		if errors.Is(err, os.ErrNotExist) {
+			return nil
+		}
+		return fmt.Errorf("journal: open cold log for tail repair: %w", err)
+	}
+	defer func() { _ = fd.Close() }()
+	st, err := fd.Stat()
+	if err != nil {
+		return fmt.Errorf("journal: stat cold log: %w", err)
+	}
+	if st.Size() <= validUpTo {
+		return nil
+	}
+	logger.Warn("journal: dropping torn cold log tail",
+		"valid_up_to", validUpTo, "dropped_bytes", st.Size()-validUpTo)
+	if err := fd.Truncate(validUpTo); err != nil {
+		return fmt.Errorf("journal: truncate torn cold log tail: %w", err)
+	}
+	if err := fd.Sync(); err != nil {
+		return fmt.Errorf("journal: fsync truncated cold log: %w", err)
+	}
+	return nil
 }
 
 // rewriteCold replaces the log with exactly entries, dropping the ones recovery
--- a/pkg/block/journal/cold_test.go
+++ b/pkg/block/journal/cold_test.go
@@ -198,6 +198,68 @@ func TestColdLogTornTailKeepsIntactEntries(t *testing.T) {
 	}
 }
 
+// TestColdLogTornTailIsRepairedOnOpen: the log is opened for append, so a torn
+// tail that recovery reads past but leaves on disk sits in front of everything
+// written afterwards. The next load stops at the same offset again, so every
+// cold interval recorded by the intervening process is lost — and a lost cold
+// interval reads as a hole that zero-fills instead of fetching from the remote.
+func TestColdLogTornTailIsRepairedOnOpen(t *testing.T) {
+	ctx := context.Background()
+	dir := t.TempDir()
+	cfg := Config{ShardCount: 1, SegmentSize: minSegmentSize}
+
+	s, err := Open(dir, cfg, newFakeRemote(), newFakeClock())
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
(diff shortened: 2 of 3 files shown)


marmos91:

Reviewed against the code rather than this description. **No blocking findings**, one note worth folding into the description.

Verified that `applyColdLog` propagates `truncateColdTail`'s error, so a repair failure fails the open rather than appending onto a log already known to be poisoned. That is the right call, and correctly distinguished from `compactColdLog`, where a failure only costs a redundant replay.

`truncateColdTail` itself is clean: `ErrNotExist` returns nil, `st.Size() <= validUpTo` short-circuits so an intact log costs one open plus one stat per recovery, and `Truncate` followed by `Sync` is sufficient — the size change lives in the inode and no directory entry moves, so no directory fsync is required.

The `loadCold` → `truncateColdTail` coupling (read by path, then reopen by path) is safe because recovery is exclusive, and `loadCold`'s doc comment carries the obligation forward to any future caller.

**Note worth adding to the description:** this changes `Open`'s *failure surface*, not only its correctness. A `cold.log` that cannot be opened `O_WRONLY` — a read-only mount, permissions drift, a snapshot restored with tight modes — now refuses to open a store that previously opened and merely warned. That looks like the right trade at this severity, but it is an availability change, and someone later deciding whether to soften it should find the intent written down rather than have to infer it from the error plumbing.

PR #2246, opened 2026-09-02
fix(journal): repack target never counted its records, so eviction could drop dirty bytes

### LINKED ISSUE #2228: journal: repackSegment never sets target.records — synced-gate can evict a still-dirty segment
`repackSegment` relocates a victim's live records into a fresh target via `writeDataRecord` (`reclaim.go:971-999`), then sets `target.liveBytes` (`:834`) and `target.syncedRecords` (`:835`) — but **never sets `target.records`**.

`segmentMeta.records` (`segment.go:54`, doc: "physical records appended (eviction synced-gate denominator)") is only incremented by the live append path (`segment.go:359`) and recovery replay (`recovery.go:340`). `createSegment` leaves it zero, so a repack target's `records` stays 0 forever.

`evictable()` (`reclaim.go:207-210`) gates on `syncedRecords == records`. `pickVictim` (`reclaim.go:621-672`) selects purely by dead-byte ratio and never checks whether the victim's remaining live records are synced — so a victim can legitimately hold entirely unsynced live records. When it does, `syncedCount == 0`, and `0 == 0` evaluates **true**: the target now looks fully synced and evictable while holding the only copy of dirty, never-remote-durable data.

A later `Evict`/`ensureSpace` pass then cold-marks and unlinks it (`evictSegment`, `reclaim.go:245-309`) — `evictSegment` cold-marks every interval regardless of `iv.synced`. A subsequent cold read tries to fetch bytes that were never pushed to the remote. This defeats `Evict`'s own stated contract at `reclaim.go:44-47`.

## Fix

```go
target.records.Store(int64(len(moves)))
```

beside the existing stores at `:834-835`, excluding tombstone/truncate markers to match live-path semantics.

## Notes

- Found independently by two audit lenses (`bugs` and `gaps`) — the strongest confirmation signal in the run.
- Verified by grep: only two non-test writers of `records` package-wide.
- `TestGCForcedRepackPreservesData` does not catch it — it checks data preservation post-repack, not the synced-gate denominator.

Source: `.planning/2026-09-01-journal-audit-report.md` (verified finding, `bugs`+`gaps`/reclaim).

marmos91:
Closes #2228

## What was wrong

`repackSegment` relocates a victim's live records into a fresh target and then
publishes the target's counters:

```go
target.liveBytes.Store(relocated)
target.syncedRecords.Store(syncedCount)
```

It never stored `target.records`, and `createSegment` leaves it zero. So a repack
target's eviction synced-gate — `evictable()`, which is
`sealed && !busy && syncedRecords == records` — compared a real numerator against
a permanent zero denominator. That is broken in *both* directions:

- **Data loss.** `pickVictim` selects purely on dead-byte ratio and never looks at
  whether the survivors are synced, so a victim can legitimately hold live records
  that were never pushed to the remote store. Then `syncedCount == 0`, and
  `0 == 0` reads as *fully synced*. The target becomes evictable while holding the
  only copy of dirty bytes; `evictSegment` cold-marks every interval regardless of
  `iv.synced`, and the subsequent cold read goes looking for bytes that were never
  uploaded. This defeats the contract `Evict` states in its own doc comment.
- **Disk leak.** The mirror case: a target whose survivors *were* synced gets
  `syncedRecords == N`, `records == 0`, so `N == 0` is never true and the segment
  can never be reclaimed. A later carve flip (`seg.syncedRecords.Add(1)`) only
  pushes the numerator further from a denominator that never moves.

## The fix

One store, next to the two that were already there:

```go
target.records.Store(int64(len(moves)))
```

`moves` is exactly the set of data records the `writeDataRecord` loop appends —
one call per entry, no batching or splitting. The tombstones and truncate markers
carried forward are deliberately excluded, which is what makes the value agree
with the other two writers of this counter: the live append path (`segment.go`)
and the recovery replay (`recovery.go`) both count data records only, so a restart
reconstructs the same denominator the live code holds.

The fix is deliberately *not* adding `records > 0` to `evictable()` the way
`sealableActive` has it. That would mask the data-loss direction and leave the
disk leak permanently in place; correcting the counter closes both.

Ordering is safe: both stores land before the `sh.mu` section that installs
`sh.sealed[target.id] = target` and repoints the index, and every path that can
discover the target goes through one of those two, so nothing can observe
`records == 0` alongside a non-zero `syncedRecords`.

## Evidence

The regression tests were run against a deliberately broken build (fix line
reverted, tests kept) before being trusted.

**Without the fix:**

```
=== RUN   TestGCForcedRepackPreservesData
    gc_test.go:185: fully synced target is not evictable: records=0 syncedRecords=1
--- FAIL: TestGCForcedRepackPreservesData (0.30s)
=== RUN   TestGCRepackDirtyTargetNotEvictable
    gc_test.go:213: target holding an unsynced record is evictable: records=0 syncedRecords=0
--- FAIL: TestGCRepackDirtyTargetNotEvictable (0.33s)
FAIL
```

A throwaway probe (not committed) that ran past the gate assertion to the actual
eviction confirms the loss is real rather than a counter cosmetic — 300 KiB
written with `WriteAt` (never synced), repacked, then a single `Evict` pass:

```
PROBE target: records=0 syncedRecords=0 evictable=true unsyncedBytes=512000
PROBE after Evict: segmentsEvicted=1 bytesFreed=307301 -> ReadAt keep err=<nil> cold=true
```

The same probe with the fix in place:

```
PROBE target: records=1 syncedRecords=0 evictable=false unsyncedBytes=512000
PROBE after Evict: segmentsEvicted=0 bytesFreed=0 -> ReadAt keep err=<nil> cold=false
```

**With the fix:**

```
=== RUN   TestGCForcedRepackPreservesData
--- PASS: TestGCForcedRepackPreservesData (0.34s)
=== RUN   TestGCRepackDirtyTargetNotEvictable
--- PASS: TestGCRepackDirtyTargetNotEvictable (0.27s)
PASS
```

`go test ./pkg/block/journal/... -race` and `go test ./pkg/block/...` are green.

## Coverage

`TestGCForcedRepackPreservesData` already reached a post-repack target, so the
synced-target (disk-leak) side is asserted there rather than duplicated.
`TestGCRepackDirtyTargetNotEvictable` is new and covers the data-loss side end to
end: repack, assert the gate, run a real `Evict` pass, then read the bytes back
byte-identical and warm. `seedRepackable` gained a `keepSynced` parameter so the
two tests share one setup instead of a near-verbatim copy that could drift; the
four existing call sites pass `true` and are unchanged. In both tests the
behavioural assertion precedes the raw counter assertion, so a regression reports
the symptom rather than the number.

## Adjacent defect, not fixed here

`repackSegment` never sets `target.lastAccess`, and `createSegment` does not
initialize it either, so a fresh repack target sits at the zero timestamp.
`claimColdestEvictable` treats that as the oldest possible access, so a target —
which by construction holds the *surviving* records of its victim — sorts to the
front of the approx-LRU and is evicted before genuinely colder segments. It is a
placement inversion rather than a correctness bug, it predates this change, and
it is left alone deliberately.

### DIFF
--- a/pkg/block/journal/gc_test.go
+++ b/pkg/block/journal/gc_test.go
@@ -85,10 +85,12 @@ func TestDeadBytesOnTombstone(t *testing.T) {
 	}
 }
 
-// seedRepackable builds a store with one sealed segment holding a synced "keep"
-// record and a dirty "gone" record, then deletes "gone" so the sealed segment is
-// 70% dead. It returns the keep payload for byte-identity checks.
-func seedRepackable(t *testing.T, s *Store) []byte {
+// seedRepackable builds a store with one sealed segment holding a "keep" record
+// and a dirty "gone" record, then deletes "gone" so the sealed segment is 70%
+// dead. keepSynced decides whether "keep" lands already synced to the remote,
+// which is what makes the repack target evictable or not. It returns the keep
+// payload for byte-identity checks.
+func seedRepackable(t *testing.T, s *Store, keepSynced bool) []byte {
 	t.Helper()
 	ctx := context.Background()
 	keep := make([]byte, 300<<10)
@@ -96,7 +98,13 @@ func seedRepackable(t *testing.T, s *Store) []byte {
 	rand.New(rand.NewSource(1)).Read(keep)
 	rand.New(rand.NewSource(2)).Read(gone)
 
-	if err := s.Hydrate(ctx, "keep", 0, keep, 0); err != nil { // synced=true
+	var err error
+	if keepSynced {
+		err = s.Hydrate(ctx, "keep", 0, keep, 0)
+	} else {
+		err = s.WriteAt(ctx, "keep", 0, keep)
+	}
+	if err != nil {
 		t.Fatal(err)
 	}
 	if err := s.WriteAt(ctx, "gone", 0, gone); err != nil { // synced=false
@@ -115,7 +123,7 @@ func seedRepackable(t *testing.T, s *Store) []byte {
 func TestGCForcedRepackPreservesData(t *testing.T) {
 	s := testStore(t, Config{SegmentSize: minSegmentSize, ShardCount: 1})
 	ctx := context.Background()
-	keep := seedRepackable(t, s)
+	keep := seedRepackable(t, s, true)
 
 	victim := onlySealed(t, s, "keep")
 	occupied := victim.liveBytes.Load()
@@ -189,23 +197,9 @@ func TestGCForcedRepackPreservesData(t *testing.T) {
 func TestGCRepackDirtyTargetNotEvictable(t *testing.T) {
 	s := testStore(t, Config{SegmentSize: minSegmentSize, ShardCount: 1})
 	ctx := context.Background()
+	keep := seedRepackable(t, s, false)
 
-	keep := make([]byte, 300<<10)
-	rand.New(rand.NewSource(4)).Read(keep)
-	if err := s.WriteAt(ctx, "keep", 0, keep); err != nil { // synced=false
-		t.Fatal(err)
-	}
-	if err := s.WriteAt(ctx, "gone", 0, make([]byte, 700<<10)); err != nil {
-		t.Fatal(err)
-	}
-	// Roll the active segment over so keep+gone land in a sealed segment.
-	if err := s.WriteAt(ctx, "trigger", 0, make([]byte, 200<<10)); err != nil {
-		t.Fatal(err)
-	}
-	if err := s.Delete(ctx, "gone"); err != nil { // 70% dead: an auto pass repacks
-		t.Fatal(err)
-	}
-
+	// 0.7 dead ratio >= default GCDeadRatioForce (0.5): an auto pass repacks it.
 	res, err := s.GC(ctx, GCOptions{})
 	if err != nil {
 		t.Fatal(err)
@@ -282,7 +276,7 @@ func TestGCCrashBeforeUnlinkOrphanSwept(t *testing.T) {
 		t.Fatal(err)
 	}
 	ctx := context.Background()
-	keep := seedRepackable(t, s)
+	keep := seedRepackable(t, s, true)
 
 	// Repack but stop right before reclaiming the victim: on disk the target is
 	// durable while the victim still exists (crash-before-unlink).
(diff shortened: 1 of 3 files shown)


marmos91:
Both Copilot points addressed.

**The `RestoreToVersion` finding is stale, not a real gap in this branch.** The PR body was briefly wrong: a scratchpad file was overwritten by a concurrently-running task and `gh pr create` picked up another change's description. Copilot reviewed that body. The body has since been replaced with the correct one, and Copilot's own generated overview describes this branch accurately — the diff is three files in `pkg/block/journal`, sets `target.records` in `repackSegment`, and touches nothing in `RestoreToVersion` or `commitDirtyShards`. Nothing is missing from the branch; only the description was.

**The wording point is real and is fixed in 776bf485.** The comment said "the synced-gate below" when `evictable` is defined earlier in the file. It now names `evictable` directly rather than pointing in the wrong direction.

marmos91:

Reviewed against the code rather than this description. **No blocking findings.**

The whole fix is one line, so the only way it is wrong is if `len(moves)` can overcount the records actually written. It cannot: every iteration of the relocation loop either calls `writeDataRecord` exactly once or returns an error after `cleanup()`. There is no `continue`, no skip, and no batching or splitting, so on the success path the target holds exactly `len(moves)` data records.

Marker exclusion is correct and for the stated reason — the live-append path increments the counter *after* the tombstone and truncate `continue` branches, so it counts data records only, and a restart therefore reconstructs the same denominator the live code holds.

Ordering verified: both counter stores land before the `sh.mu` section installs `sh.sealed[target.id]` and repoints the index, so no path can observe `records == 0` alongside a non-zero `syncedRecords`.

Two things worth recording for future readers:

- **The two pins fail for genuinely different reasons** — `1 == 0` (stranded space) and `0 == 0` (false-synced, data loss). One test passing for two different reasons would tell you nothing about either.
- **The probe carries a control past the gate to a real `Evict` pass**: `evictable=true` → `freed 307301 bytes` → `ReadAt cold=true`. That `cold=true` *is* the silent-zeros symptom, not a proxy for it. Given this repo's history (#2084, where a log line from the fixed build nearly got a silent-corruption bug demoted to a doc note), reproducing the symptom rather than a counter assertion is the right standard.

Declining to add `records > 0` to `evictable()` was also correct: it would have masked the data-loss direction while leaving the disk leak permanently in place.

Issue #2227, opened 2026-09-01
journal: RestoreToVersion is blind to cold.log — deletes remote-durable files and zero-fills cold ranges


marmos91:
`RestoreToVersion`'s phase-1 ceiling replay (`store.go:1087-1139`) rebuilds `vIndex` by scanning **only** on-disk segment records via `scanValidRecords`. It never folds in `cold.log`, the way `recover()` does via `applyColdLog` (`recovery.go:358`).

Cold intervals own no segment record — `appendCold` (`cold.go:129`) writes only to `cold.log`. They are created by `Evict` and by `SeedCold`/`SeedColdBatch` (`store.go:520`, `store.go:607`). So:

1. **A file entirely cold at V** gets no `vIndex` entry, falls into the "in head but not in vIndex" bucket (`store.go:1217-1223`), and is unconditionally `Delete`d — despite being durably present on the remote at V.
2. **A file with mixed warm/cold ranges at V** appears in `vIndex` from its warm parts only. The cold sub-ranges are absent from `exts` (`store.go:1174-1216`), so after the Delete + re-`WriteAt` sequence those ranges become genuine POSIX holes. Reads return zeros instead of fetching.

This is the HOLE-vs-COLD conflation that `doc.go` calls the module's cardinal sin, and it is the same reason `recover()` folds the cold log: "a cold interval owns no record" (`recovery.go:352-354`).

## Why there is no compensation

One might expect `shares.SeedColdFromManifest` (`snapshot.go:1598`) to re-seed cold ranges afterwards. It is gated `if remoteVerify` (`snapshot.go:1595`), while `RestoreToVersion` runs **only** in the `!remoteVerify` local-only branch (`snapshot.go:1341`). On the only path that calls it there is no compensation at all — and the erroneous tombstones are fsynced before anything could roll them back.

Cold intervals are reachable on a local-only share: `blockstore_config.go:370` disables eviction when `remoteStore == nil`, but `engine.go:271` re-enables it unconditionally via `SetEvictionEnabled(bs.syncer.IsRemoteHealthy())`, and `syncer.go:309-312` returns true when `healthMonitor == nil`. `Start` runs after line 370.

## Fix

[...]

PR #2248, opened 2026-09-02
fix(journal): fsync the view RestoreToVersion re-materializes

### LINKED ISSUE #2229: journal: RestoreToVersion never fsyncs its re-materialized view, contradicting its own doc
`RestoreToVersion` phase 2 (`store.go:1169-1224`) calls `s.Delete`, which fsyncs its tombstone via `groupCommit` (`segment.go:426-474`), then re-appends the V-view bytes with `s.WriteAt` (`:1211`) — which only buffers and never fsyncs (documented at `store.go:424-425`). There is no `Commit`/`groupCommit` call anywhere after.

The method's own doc (`store.go:1057-1059`) claims it "re-materializes the view durably at the log head, so a crash-reopen reconstructs V". That is false as written.

The exposure window is bounded by `DirtyExpiry` (~30s) or the next explicit `Commit`. A crash inside it leaves the **tombstone durable and the replacement data not** — the file reads empty. Because each subsequent `Delete` fsyncs its own shard, in practice only the last restored file per shard is exposed, but the window is real, and on a local-only share the journal is the only copy.

## Fix

Call `sh.groupCommit()` (or `s.Commit(ctx, id)`) for every shard touched, at the end of `RestoreToVersion`, before returning.

## Notes

- Verified: zero `groupCommit`/`Commit` references in the function; the caller (`snapshot.go:1539-1546`) only does `DrainRollups` afterwards.
- `RestoreToVersion` has **zero test call sites** across all 29 test files.
- Related: #2227 (same method, cold-log blindness).

Source: `.planning/2026-09-01-journal-audit-report.md` (verified finding, `gaps`/store-api).

marmos91:
## What was wrong

`Store.RestoreToVersion` phase 2 re-materializes each file's target-version view at
the log head in two steps: `s.Delete` buries the current head with a tombstone, then
`s.WriteAt` re-appends the V-view bytes on top of it.

Those two steps have different durability. `Delete` goes through `appendTombstone`,
which fsyncs the shard before it returns. `WriteAt` only buffers — that is its
documented contract, durability is a separate `Commit`. And there was no
`Commit`/`groupCommit` anywhere after: not in `RestoreToVersion`, and not in the
caller, which only runs `DrainRollups` (a carve pass, which never fsyncs the journal).

So the window between the tombstone and the next barrier — bounded only by
`DirtyExpiry` (~30s) or an unrelated client `Commit` — held a **durable burial with a
buffered replacement**. A crash inside it reconstructs the tombstone and not the data,
and the restored file reads empty. The method's own doc claimed it
"re-materializes that view durably at the log head, so a crash-reopen reconstructs V";
that was false as written.

`RestoreToVersion` is the local-only restore primitive, so on those shares the journal
is the only durable copy of the bytes. There is nothing to re-seed from.

## Premise check

The issue's claims were verified on unmodified `develop` before any code was written:

- The whole package has exactly four fsync call sites (`store.go:404` dirty-age loop,
  `store.go:653` `Commit`, `segment.go:470` `appendTombstone`, `segment.go:522`
  `appendTruncateMarker`), plus `sealInPlace`'s direct `fd.Sync`. None is reachable
  from `WriteAt`, and none appeared in `RestoreToVersion`'s body.
- The caller (`snapshot.go:1539`) follows the restore with `DrainRollups`, which is
  `local.Carve(Force: true)` — it rolls dirty payloads into CAS + the manifest and
  issues no journal fsync. It does not incidentally cover the window.
- `RestoreToVersion` touches whichever shards its file IDs hash to, not one, so the
  fix has to be shard-wide. The issue's "in practice only the last restored file per
  shard is exposed" is right, and it comes from a *subsequent* `Delete` incidentally
  fsyncing that shard — which is exactly the partial-coverage shape a fix must not
  inherit.

Premise confirmed.

## The fix

`commitDirtyShards` — which already walked every shard holding records above its
durable watermark — now returns the first failure instead of only logging, and
`RestoreToVersion` calls it once at the end. `syncLoop`, its only other caller, logs
the returned error as before. Sweeping every shard still holding unsynced records
rather than the last one written is the point: a naive "commit the shard of the last
file" fix reproduces the same partial coverage the defect already had. A shard already
at its durable watermark is skipped, so the superset costs nothing.

## Evidence

`RestoreToVersion` had **zero test call sites** across the package's 29 test files, so
`restore_test.go` is new. It carries a reusable fixture rather than a one-off, because
#2227 fixes a different defect in the same method and will extend it.

The fixture matters because `groupcommit_durability_test.go` already documents why the
obvious test is hollow: *"the crash-recovery tests reopen from the same tmpfiles, where
the OS page cache serves the bytes whether or not fsync ran — so they cannot tell a
real fsync from a no-op."* So `restoreFixture` hooks each shard's `segSync` seam to
record, per segment, the tail some **completed** fsync covered, and `crashReopen()`
truncates every unsealed segment back to that tail before reopening — a real power-cut
model, not a reopen. Sealed segments are left whole (`sealInPlace` fsyncs before it
sets the sealed bit, so they are durable by construction). `Close` is called first and
issues no fsync of its own. The dirty-age loop is disabled and the background repack is
put out of reach, so the only fsyncs in the test are the ones the code under test issues.

The test restores two files that hash to two different shards, so a barrier covering
only one shard still fails it.

Run against a deliberately broken build, three ways:

```
# with the fix
--- PASS: TestRestoreToVersion_ReMaterializedViewSurvivesCrash (0.46s)

# fix reverted (the defect as filed)
restore_test.go:187: post-crash restore-0: restored view lost (reads-empty=true, reads-V2=false);
    the burial tombstone was fsynced but its replacement data was not
--- FAIL: TestRestoreToVersion_ReMaterializedViewSurvivesCrash (0.52s)

# partial fix injected (commit shard 0 only) — pins the shard-wide requirement
restore_test.go:185: post-crash restore-1: restored view lost (reads-empty=true, reads-V2=false);
    the burial tombstone was fsynced but its replacement data was not
--- FAIL: TestRestoreToVersion_ReMaterializedViewSurvivesCrash (0.76s)
```

`go test ./pkg/block/journal/... -race` → `ok 87.800s`. `./pkg/block/...` and
`./pkg/controlplane/runtime/...` green. 260 repeats of the new test under three
concurrent full-package runs: clean. `go build ./...`, `go vet ./...`, `gofmt` and
`golangci-lint` clean.

### Known, deliberate gap in the test

`commitDirtyShards` skips a shard whose `dirty()` is already false, and the fixture
sets `DirtyExpiry: -1` so the dirty-age loop never runs. The test therefore never
exercises the case where the sweep meets a shard some earlier commit already covered.
That path is fine by inspection — already durable means there is nothing to fsync —
but it is argued rather than pinned, and it is left that way on purpose rather than
overlooked.

## Adjacent, not fixed

- `shard.dirty()` reports false once `syncFailed` is set, so a shard whose fsync has
  permanently failed is skipped and `RestoreToVersion` returns nil for it. That is the
  store's existing sticky-failure policy everywhere (`markSynced`, `groupCommit`,
  `syncLoop` all treat it the same), and the store is already wedged by then — left
  alone rather than given a special case on this path.
- `sealInPlace` fsyncs through `m.fd.Sync()` directly rather than the per-shard
  `segSync` seam the durability tests hook, so a broken rotation fsync is invisible to
  that seam. It is a limit of the seam, not a hole in this test: sealed segments are
  legitimately durable and the crash model leaves them whole.

Closes #2229

### DIFF
--- a/pkg/block/journal/restore_test.go
+++ b/pkg/block/journal/restore_test.go
@@ -23,12 +23,14 @@ type restoreFixture struct {
 	durable map[uint64]int64
 }
 
-// newRestoreFixture opens a store over a fresh directory with shardCount shards
-// and the dirty-age sync loop disabled, so the only fsyncs are the ones the code
-// under test issues.
+// newRestoreFixture opens a store over a fresh directory with shardCount shards.
+// The dirty-age sync loop is disabled and the background repack is put out of
+// reach (no segment's dead ratio can exceed 1), so the only fsyncs are the ones
+// the code under test issues and no pass moves records between the segments the
+// crash model tracks.
 func newRestoreFixture(t *testing.T, shardCount int) *restoreFixture {
 	t.Helper()
-	s := testStore(t, Config{ShardCount: shardCount, DirtyExpiry: -1})
+	s := testStore(t, Config{ShardCount: shardCount, DirtyExpiry: -1, GCDeadRatioForce: 2})
 	f := &restoreFixture{Store: s, t: t, durable: map[uint64]int64{}}
 	for _, sh := range s.shards {
 		inner := sh.segSync
@@ -77,7 +79,7 @@ func (f *restoreFixture) commitAll() uint64 {
 // helper's own Close is a no-op.
 func (f *restoreFixture) crashReopen() *Store {
 	f.t.Helper()
-	if err := f.Store.Close(); err != nil {
+	if err := f.Close(); err != nil {
 		f.t.Fatalf("Close: %v", err)
 	}
 	ids, err := scanSegmentIDs(f.dir)
--- a/pkg/block/journal/restore_test.go
+++ b/pkg/block/journal/restore_test.go
@@ -1,0 +1,206 @@
+package journal
+
+import (
+	"bytes"
+	"context"
+	"fmt"
+	"os"
+	"sync"
+	"testing"
+)
+
+// restoreFixture is a store whose fsyncs are tracked so a test can throw away
+// every segment byte no completed fsync covered — precisely what a power cut
+// takes. Reopening the same directory proves nothing on its own: the page cache
+// serves buffered writes back exactly like fsynced ones, so a plain reopen
+// cannot tell a durable record from one that only reached the page cache.
+type restoreFixture struct {
+	*Store
+	t   *testing.T
+	dir string
+	cfg Config
+
+	mu sync.Mutex
+	// durable maps a segment id to the tail some completed fsync covered.
+	durable map[uint64]int64
+}
+
+// newRestoreFixture opens a store over a fresh directory with shardCount shards
+// and the dirty-age sync loop disabled, so the only fsyncs are the ones the code
+// under test issues.
+func newRestoreFixture(t *testing.T, shardCount int) *restoreFixture {
+	t.Helper()
+	dir := t.TempDir()
+	cfg := Config{ShardCount: shardCount, DirtyExpiry: -1}
+	s, err := Open(dir, cfg, newFakeRemote(), SystemClock())
+	if err != nil {
+		t.Fatalf("Open: %v", err)
+	}
+	t.Cleanup(func() { _ = s.Close() })
+	f := &restoreFixture{Store: s, t: t, dir: dir, cfg: cfg, durable: map[uint64]int64{}}
+	for _, sh := range s.shards {
+		inner := sh.segSync
+		sh.segSync = func(seg *segmentMeta) error {
+			// Sample the tail before the barrier: only bytes already written when
+			// the fsync started are certainly covered by it.
+			tail := seg.tail.Load()
+			if err := inner(seg); err != nil {
+				return err
+			}
+			f.mu.Lock()
+			if tail > f.durable[seg.id] {
+				f.durable[seg.id] = tail
+			}
+			f.mu.Unlock()
+			return nil
+		}
+	}
+	return f
+}
+
+// write appends data at off, the plain buffered client write.
+func (f *restoreFixture) write(id FileID, off int64, data []byte) {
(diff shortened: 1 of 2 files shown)


marmos91:

Reviewed against the code rather than this description. **No blocking findings.**

The fixture is the strongest artefact in this batch. Hooking `segSync` to record the tail sampled *before* the barrier, then truncating each unsealed segment back to it, is the only honest way to test this defect — and the comment saying why is the crux: a plain reopen proves nothing, because the page cache serves buffered writes back exactly like fsynced ones. A durability test built on a plain reopen passes on the broken build.

Three details worth singling out:

- `commitAll` deliberately fsyncs the shards itself instead of calling the store's own dirty-shard sweep, so the test cannot be set up by the very call it pins.
- `idsOnDistinctShards` forces the two-shard span, so a barrier covering only the last shard written still fails.
- **The partial-fix mutation (commit shard 0 only) is above the usual bar.** It pins the shard-*wide* requirement rather than merely the presence of a barrier — and shard-wide is precisely what a plausible naive fix gets wrong. Three mutation runs where one was required.

**One non-blocking gap, argued rather than pinned:** `commitDirtyShards` skips shards where `!sh.dirty()`, and the fixture sets `DirtyExpiry: -1` to disable the sync loop. So the case where that loop has already committed a shard the restore touched is never exercised. It is fine by inspection (already durable means nothing to fsync), but it is reasoning rather than coverage, and worth naming in the description as a deliberate known gap so it is not later mistaken for an oversight.

Minor: the new `RestoreToVersion` doc says "fsync every shard the pass touched" while the implementation commits every dirty shard store-wide — a superset, and arguably the more robust choice. Behaviour is right; the comment should match it.

PR #2264, opened 2026-09-02
test(metadata): pin that transaction writes roll back with their transaction

marmos91:
## The gap

`FileReadTxOps` pins that transaction-level *reads* run inside the caller's
transaction. The transaction-level **writes** had no equivalent.

The two failure modes are not symmetric. An escaped read returns a
stale-but-plausible answer. An escaped **write is durable** — it survives the
rollback that was supposed to erase it, and every other test in the suite still
passes, because the rest of the suite only ever asks what is committed.

This matters specifically while transaction bodies are being consolidated onto
`store/sql.Core`. Such a body reaches its executor through a struct field, so
pointing one at the pool instead of the transaction is a one-character mistake
that compiles, races nothing, and leaves no trace except a write that will not
roll back.

## What it does

Six subtests — `SetChild`, `DeleteChild`, `SetLinkCount`, `DeleteFile`,
`UpdateShareOptions`, `SetServerConfig` — each writing through one method,
aborting, and asserting the store shows no trace afterwards.

## Verified against a broken build, not assumed

Pointing each dialect's transaction `Core` at the pool executor:

**postgres** — all six fail by assertion, each naming its method:

```
--- FAIL: TestConformance/TxWriteRollbackOps/SetChild
--- FAIL: TestConformance/TxWriteRollbackOps/DeleteChild
--- FAIL: TestConformance/TxWriteRollbackOps/SetLinkCount
--- FAIL: TestConformance/TxWriteRollbackOps/DeleteFile
--- FAIL: TestConformance/TxWriteRollbackOps/UpdateShareOptions
--- FAIL: TestConformance/TxWriteRollbackOps/SetServerConfig
        SetServerConfig survived the rollback: it ran outside the transaction
```

**sqlite** — wedges instead, because the escaped write blocks on the connection
its own transaction still holds. That is a detection too, but an illegible one,
so the doc comment says so explicitly rather than promising a named failure the
single-connection backends cannot deliver.

Unmodified, green on memory, sqlite, badger and postgres.

### DIFF
--- a/pkg/metadata/storetest/suite.go
+++ b/pkg/metadata/storetest/suite.go
@@ -227,6 +227,10 @@ func RunConformanceSuite(t *testing.T, factory StoreFactory) {
 	// run inside the caller's transaction. Every other test here reads only
 	// committed state, which a read that escaped to the pool would answer just
 	// as correctly.
+	t.Run("TxWriteRollbackOps", func(t *testing.T) {
+		runTxWriteRollbackOps(t, factory)
+	})
+
 	t.Run("FileReadTxOps", func(t *testing.T) {
 		runFileReadTxOps(t, factory)
 	})
--- a/pkg/metadata/storetest/suite.go
+++ b/pkg/metadata/storetest/suite.go
@@ -223,14 +223,18 @@ func RunConformanceSuite(t *testing.T, factory StoreFactory) {
 		runServerConfigOps(t, factory)
 	})
 
-	// FileReadTxOps pins that the transaction-level file and directory reads
-	// run inside the caller's transaction. Every other test here reads only
-	// committed state, which a read that escaped to the pool would answer just
-	// as correctly.
+	// TxWriteRollbackOps pins that the transaction-level writes land inside the
+	// caller's transaction. A write that escaped to the pool is durable: it
+	// outlives the rollback meant to erase it, and every other test here still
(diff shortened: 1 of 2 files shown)


marmos91:
Both findings were real; fixed in `91858027b`.

**1. `assert.Error` was too loose.** It accepted any error, so an unrelated failure would have read as proof the write rolled back — a guard that passes for the wrong reason. Now `metadata.IsNotFoundError`, matching `file_read_tx_ops.go`.

**2. Orphaned comment.** Inserting the new subtest between `FileReadTxOps`' comment and its `t.Run` left the comment describing the wrong subtest. Each now carries its own.

Re-verified the tightened assertion still detects what it exists for, by pointing postgres' transaction `Core` at the pool again — all six still fail, `SetChild` among them:

```
--- FAIL: TestConformance/TxWriteRollbackOps/SetChild
--- FAIL: TestConformance/TxWriteRollbackOps/DeleteChild
--- FAIL: TestConformance/TxWriteRollbackOps/SetLinkCount
--- FAIL: TestConformance/TxWriteRollbackOps/DeleteFile
--- FAIL: TestConformance/TxWriteRollbackOps/UpdateShareOptions
--- FAIL: TestConformance/TxWriteRollbackOps/SetServerConfig
```

Green unmodified on memory, sqlite, badger and postgres.

PR #2269, opened 2026-09-02
fix(smb): correct the seven wrong spec citations and empty the allowlist

### LINKED ISSUE #2185: Correct the seven spec citations the citation lint flags on develop
The citation lint added in #2179 flags seven citations that #2174 did not reach. They are
listed in `test/spec-citations/known_wrong.txt` so the check can gate new drift; this issue
tracks correcting them.

They are deliberately **not** fixed in that PR. #2174 established that a wrong citation and a
wrong implementation look identical from inside a source file — two real defects (#2171, #2181)
were found exactly because someone checked the spec instead of adjusting the comment. Editing
these comments to make a linter go quiet is how that class hides.

## 1. `MS-FSA 2.1.5.14.3` — fabricated section, 6 sites

| file | line |
| --- | --- |
| `internal/adapter/smb/handlers/change_notify.go` | 1729 |
| `internal/adapter/smb/handlers/close.go` | 547 |
| `internal/adapter/smb/handlers/close_notify_test.go` | 165, 228 |
| `internal/adapter/smb/handlers/handler.go` | 1600 |
| `internal/adapter/smb/handlers/set_info.go` | 1399 |

MS-FSA v20250909 §2.1.5.14 is *Server Requests a Query of Security Information* and has exactly
one child, `.1` (*Algorithm for Copying Audit or Label ACEs Into a Buffer*). There is no `.3`.

This is the same family #2174 corrected 43 times (`2.1.5.14.2/.3/.4/.5/.10/.13/.23` → per-class
sections under §2.1.5.15); these six survived it. Every one of them makes the same behavioural
claim — *setting the delete disposition on a directory completes outstanding CHANGE_NOTIFY
requests on it with `STATUS_DELETE_PENDING`*. That claim needs a real authority before the
number is replaced: §2.1.5.15.6 (`FileDispositionInformation`) is the section that sets the
disposition, but the notify-completion behaviour may belong to MS-SMB2 §3.3.5.19 instead, or to
neither — in which case the comments are asserting something no spec requires and the right fix
is to say so.

## 2. `MS-FSCC 2.4.34` cited for `FILE_RENAME_INFORMATION`, 1 site

`internal/adapter/smb/handlers/close_doc_dir_not_empty_test.go:43` describes the
`FILE_RENAME_INFORMATION` buffer and cites §2.4.34, which is `FileNetworkOpenInformation`.
`FileRenameInformation` is §2.4.42 (§2.4.42.2 for the SMB2 form). This one is mechanical —
the wrong-number class from #2174's `requests.go` block, one site further on.

## Done when

Each entry is either corrected with a verified section, or replaced by a comment that does not
cite a section it cannot support, and the corresponding line is removed from
`test/spec-citations/known_wrong.txt` (the check fails on a stale entry, so it cannot be
forgotten).

marmos91:
Closes #2185. Stacked on #2267 — that one makes the check usable locally; without
it a local run reports 685 findings from nested working copies and buries these seven.

## The corrections

**MS-FSA 2.1.5.14.3 → 2.1.5.11** (`change_notify.go`, `close.go`, `close_notify_test.go`,
`handler.go`, `set_info.go`)

2.1.5.14 is *Server Requests a Query of Security Information* and has no subsection 3, so
the cited section never existed. All five comments state the same rule, and it is
2.1.5.11 *Server Requests Change Notifications for a Directory*, step 4:

> 4. If **Open.ChangeNotifyDirectoryMarkedDeleted** is TRUE:
> 4.1. Remove **ChangeNotifyEntry** from **Open.File.Volume.ChangeNotifyList**.
> 4.2. Complete the ChangeNotify operation with status STATUS_DELETE_PENDING.

The spec's own change log corroborates it — 2.1.5.11 carries "Updated Change Notify
processing when a directory is marked for deletion", classed Major.

**MS-FSCC 2.4.34 → 2.4.42.2** (`close_doc_dir_not_empty_test.go`)

2.4.34 is `FileNetworkOpenInformation`. The comment describes `FILE_RENAME_INFORMATION`,
which is 2.4.42 — and the encoder settles which variant: a 20-byte fixed header
(1 ReplaceIfExists, 7 Reserved, 8 RootDirectory, 4 FileNameLength at offset 16) is the
SMB2 layout, so **2.4.42.2**, not the SMB1 form whose RootDirectory is 4 bytes.

**MS-FSCC 2.3.32 → 2.3.51 + 2.3.52** (`ioctl_sparse.go`)

2.3.32 is `FSCTL_GET_RETRIEVAL_POINTERS Reply` — a different FSCTL. The sentence spans
both halves of `FSCTL_QUERY_ALLOCATED_RANGES`, so it now cites the request window
(2.3.51) and the reported ranges (2.3.52) at the clauses that mean them.

## The allowlist is now empty

`known_wrong.txt` keeps its header — it is embedded, and the file is where a future
deferral goes — but holds no entries. Its comment says why an empty list is the point.

## Verified

```
before: spec-citations: 7 problem(s) across 2362 Go files
after:  spec-citations: 2362 Go files clean (0 known-wrong citation(s) allowed)
```

The tool flagged all seven allowlist entries as stale on its own once the citations were
fixed, which is how they were found to be complete rather than by my counting them.

And the check can still refuse — putting one bad citation back:

```
internal/adapter/smb/handlers/close.go:547: [MS-FSA] 2.1.5.14.3 does not exist (MS-FSA v20250909)
spec-citations: 1 problem(s) across 2362 Go files
```

An empty allowlist that cannot fail would be worse than the seven entries it replaced.

`go test ./internal/adapter/smb/handlers/` green; comments only, no behaviour change.

### DIFF
--- a/internal/adapter/smb/handlers/change_notify.go
+++ b/internal/adapter/smb/handlers/change_notify.go
@@ -1726,7 +1726,7 @@ func (r *NotifyRegistry) deliverChanges(notify *PendingNotify, changes []FileNot
 // that marks a directory for deletion is usually not the handle that is
 // watching it.
 //
-// Per [MS-FSA] 2.1.5.14.3, setting the delete disposition on a directory sweeps
+// Per [MS-FSA] 2.1.5.11, setting the delete disposition on a directory sweeps
 // every ChangeNotifyEntry whose OpenedDirectory.File is the same file, removes
 // it, and completes it with STATUS_DELETE_PENDING. The trigger is the mark, not
 // the unlink — the directory is still on disk until the last handle closes, and
--- a/internal/adapter/smb/handlers/change_notify.go
+++ b/internal/adapter/smb/handlers/change_notify.go
@@ -1726,7 +1726,7 @@ func (r *NotifyRegistry) deliverChanges(notify *PendingNotify, changes []FileNot
 // that marks a directory for deletion is usually not the handle that is
 // watching it.
 //
-// Per [MS-FSA] 2.1.5.11, setting the delete disposition on a directory sweeps
+// Per [MS-FSA] 2.1.5.15.3, setting the delete disposition on a directory sweeps
 // every ChangeNotifyEntry whose OpenedDirectory.File is the same file, removes
 // it, and completes it with STATUS_DELETE_PENDING. The trigger is the mark, not
 // the unlink — the directory is still on disk until the last handle closes, and
--- a/internal/adapter/smb/handlers/close.go
+++ b/internal/adapter/smb/handlers/close.go
@@ -544,7 +544,7 @@ func (h *Handler) Close(ctx *SMBHandlerContext, req *CloseRequest) (*CloseRespon
 				"messageID", notify.MessageID)
 		}
 
-		// Per [MS-FSA] 2.1.5.11, a directory marked for deletion completes
+		// Per [MS-FSA] 2.1.5.15.3 step 3.2.3.2, a directory marked for deletion completes
 		// every pending CHANGE_NOTIFY on it with STATUS_DELETE_PENDING. A
 		// handle carrying FILE_DELETE_ON_CLOSE from CREATE commits that mark
 		// in the election above rather than at CREATE, so this is the point
--- a/internal/adapter/smb/handlers/close.go
+++ b/internal/adapter/smb/handlers/close.go
@@ -544,7 +544,7 @@ func (h *Handler) Close(ctx *SMBHandlerContext, req *CloseRequest) (*CloseRespon
 				"messageID", notify.MessageID)
 		}
 
-		// Per [MS-FSA] 2.1.5.14.3, a directory marked for deletion completes
+		// Per [MS-FSA] 2.1.5.11, a directory marked for deletion completes
 		// every pending CHANGE_NOTIFY on it with STATUS_DELETE_PENDING. A
 		// handle carrying FILE_DELETE_ON_CLOSE from CREATE commits that mark
 		// in the election above rather than at CREATE, so this is the point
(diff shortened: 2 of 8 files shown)


marmos91:
## Correction: the sweep is 2.1.5.15.3, not 2.1.5.11 (`aad9341a3`)

An objection from a peer session, grounded in open issue #2199, showed the first version of this PR was wrong. Recording it here because the reasoning matters more than the diff.

**What was wrong.** I repointed the six MS-FSA sites to 2.1.5.11 because its step 4 completes a ChangeNotify with `STATUS_DELETE_PENDING`, matching the words in the comments. I never checked *which moment* it fires at. It fires at **registration** — a notify arriving on a directory already marked deleted. #2199 records that half as deliberately unimplemented. Citing it at the sweep would have put a conformance claim at six sites for behaviour an open issue says we do not have.

**The correct section is [MS-FSA] 2.1.5.15.3 `FileDispositionInformation`**, step 3.2.3:

```
3.2.3.1. Set Open.ChangeNotifyDirectoryMarkedDeleted to TRUE.
```

That is the mark-time sweep verbatim. **2.1.5.15.4** states the identical sweep at step 4.3.3.2 for `FileDispositionInformationEx`, which `set_info.go:1292` serves in the same `case`, so that site cites both.

`2.1.5.14.3` → `2.1.5.15.3` is one digit in the middle component. It was a typo all along, which is why the prose around it was accurate and read as trustworthy — and why a citation checker that only asks "does this section exist" was the right tool to catch it and the wrong one to fix it.

**The near-miss worth recording.** The tree already cites 2.1.5.11 correctly in four places — `stub_handlers.go:569`/`:584`, `change_notify.go:259`, `change_notify_test.go:3680` — for the `ChangeNotifyEntry` and its completion filter, which is genuinely what that section specifies. My version would have left one section standing for two different rules with one of them wrong. A citation to a section that does not exist announces itself; a plausible citation to a real section that says something else does not, and this linter cannot tell them apart. Those four are untouched.

#2199 is unaffected and stays open: 2.1.5.11 is now cited by nothing claiming to implement it.

Verified: `spec-citations: 2362 Go files clean (0 known-wrong citation(s) allowed)`, `go test ./internal/adapter/smb/handlers/` green.

PR #2272, opened 2026-09-02
fix(smbtorture): anchor the NT_STATUS_NO_MEMORY flake excuse to a connection diagnostic

### LINKED ISSUE #2198: parse-results.sh grades a real failure as a skip when its diagnostic mentions NT_STATUS_NO_MEMORY

marmos91:
## What was wrong

`parse-results.sh` buffers every `failure:`/`error:` block and rewrites it to
`skip:` when the block looks like an infrastructure flake. One of the two
triggers was the bare string `NT_STATUS_NO_MEMORY`, anywhere in the block, with
no requirement that the block be a connection diagnostic at all.

That status is not evidence of anything on its own. smbtorture's **own** iconv
returns `EILSEQ` when it converts an unpaired UTF-16 surrogate, and Samba maps
that to `NT_STATUS_NO_MEMORY`. So `smb2.charset.Testing partial surrogate` —
a real protocol assertion at `charset.c:174`, measured failing 15/15 draws and
never once observed passing — was graded a **skip** on every run.

The harm is the inverse of what a known-failure list is for. The failure was
subtracted from the count, the three sibling `charset.Testing` subcases passed,
and the row therefore surfaced under *"Known failures that now PASS — candidates
to drop from KNOWN_FAILURES.md"*. That list is what #2189 was working from; one
of its eight "now-passing" rows was manufactured by this filter.

The filter was also silent: nothing in the summary said a failure had been
reclassified, so the only trace it left was the row quietly appearing on a
removal list.

## Reproduction

Eight lines, no server, no Docker:

```
test: charset.Testing composite character (a umlaut)
success: charset.Testing composite character (a umlaut)
test: charset.Testing partial surrogate
failure: charset.Testing partial surrogate [
../../source4/torture/smb2/charset.c:174: status was NT_STATUS_NO_MEMORY, expected NT_STATUS_OK: Failed to create partial surrogate 1
]
test: charset.Testing wide-a
success: charset.Testing wide-a
```

Before — a `failure:` line in, `Failed: 0` out, and the row offered for deletion:

```
  Total:     3
  Passed:    2
  Failed:    0
### Known failures that now PASS — candidates to drop from KNOWN_FAILURES.md (1)
- smb2.charset.Testing
```

After:

```
  Total:     3
  Passed:    2
  Failed:    1
### Known failures still failing (1)
- smb2.charset.Testing
```

## The fix

`NT_STATUS_NO_MEMORY` no longer earns the excuse on its own. It excuses a block
only when the same block **also** carries a connection-setup diagnostic
(`CONN_SETUP_PATTERN`). `CONN_FAIL_PATTERN` — the canonical
`Establishing SMB2 connection failed` text — still excuses a block by itself,
unchanged. Two sticky per-block flags let the two markers arrive on any line of
the detail block, in any order, which is the ordering robustness the buffered
scan was built for in the first place.

The pattern is **not** deleted. It was added for a real client-side talloc
cascade, and a filter that has not fired lately is not a filter that is not
needed — it is a filter whose trigger is too broad.

Every reclassification is now named in the summary:

```
| Reclassified as flakes | 1 |

### Failures reclassified as infrastructure flakes — graded as skips, not failures (1)
- smb2.rename.rename1 — connection setup
```

The ChannelSequence excuse reports itself the same way, so no rewrite is silent
any more.

## Tests

`parse-results_test.sh` had no coverage of the `NO_MEMORY` path at all. Four
cases added, and **all eight of their assertions were confirmed RED against
unmodified `develop`** before any fix was written:

- a real assertion mentioning `NT_STATUS_NO_MEMORY` is graded a failure, counted
  against the blacklist, and is not a removal candidate;
- the same assertion on an unlisted test reds the job;
- `NT_STATUS_NO_MEMORY` alongside a connection-setup diagnostic is still excused,
  and says so;
- the canonical connect diagnostic is still excused, and is named with the
  `smb2.` prefix the rest of the summary uses.

Every pre-existing case still passes.

## Effect on KNOWN_FAILURES.md

`smb2.charset.Testing` changes grade from *skip / now-passing candidate* to
*known failure, still failing*. It is already on the blacklist, so the verdict
and exit code are unaffected and CI stays green — the row simply stops
advertising itself for deletion. No row is added or removed.

The row's prose caveat, which told readers not to trust the "now passing" report
and pointed at this issue, is updated since it no longer describes the harness.

Across the ten CI job logs examined for #2189, `NT_STATUS_NO_MEMORY` appeared
only in this one assertion, so no other row's grade moves.

Closes #2198


marmos91:
Drive-by from #2269, which is correcting this same typo in the Go comments — **not** editing this file, because your PR already touches it and I would rather not collide.

`KNOWN_FAILURES.md:472` carries a wrong spec citation:

> `CompleteWatchersForDeletePending` sweeps every watcher on the directory's path and share when the delete disposition is committed, and answers each `STATUS_DELETE_PENDING`, which is what **[MS-FSA] 2.1.5.14.3** requires.

**2.1.5.14.3 does not exist** — 2.1.5.14 is "Server Requests a Query of Security Information" and has no subsection 3. The prose is accurate; only the number is wrong.

The sweep it describes is **[MS-FSA] 2.1.5.15.3 `FileDispositionInformation`** step 3.2.3.2:

```
```

`2.1.5.14.3` → `2.1.5.15.3`: one digit in the middle component, which is why it read as plausible everywhere it appeared.

Worth taking in this PR since you are editing the file anyway. Two notes if you do:

- The citation lint only scans `.go`, so nothing will ever flag this line — it will sit here until a human reads it.
- Do **not** reach for `2.1.5.11`. That section is the *other* half of the rule (a CHANGE_NOTIFY arriving on an already-marked directory), which #2199 records as deliberately unimplemented. I nearly landed exactly that mistake in #2269 and a peer caught it.

Same typo also in #2199's body; flagged there separately. Entirely your call — I am not asking for a change to your PR's scope.

Issue #2281, opened 2026-09-02
SMB FILE_WRITE_ATTRIBUTES does not bridge to the metadata layer's POSIX ownership check, so a non-owner can never set a timestamp


marmos91:
Surfaced underneath #2205 and #2280 while working PR #2279. Filed separately because it is the shared cause of both, and fixing it may be the right way to fix #2280.

## The mismatch

SMB grants the right to set timestamps via **`FILE_WRITE_ATTRIBUTES`** on the handle. The metadata layer requires **POSIX ownership** for an explicit timestamp write:

`pkg/metadata/file_modify.go` — `!isOwner && !isRoot` → `ErrPermissionDenied`, evaluated *before* `checkWritePermission` is ever consulted.

`WriteAuthorizedByHandle` exists and is set from the handle's `GrantedAccess` by four of `restoreFrozenTimestamps`' five callers (WRITE, COPYCHUNK, and both SET_INFO sites), but it does **not** reach the explicit-timestamp gate — that gate is a flat ownership check sitting above it. So the flag makes no difference on this path.

## Consequence

A non-owner holding a handle opened with `FILE_WRITE_ATTRIBUTES` cannot set a timestamp over SMB at all, even though the protocol says that access right is precisely what authorizes it. Windows clients that rely on `FILE_WRITE_ATTRIBUTES` rather than ownership will see the write silently fail or be denied.

## Why it matters beyond conformance

This is the root under two other issues:

- **#2205** (fixed) — the discarded restore error made observable behaviour depend on whether this gate happened to pass.
- **#2280** (open) — the parent-directory frozen-timestamp restore fails for non-owning callers for exactly this reason. Bridging `FILE_WRITE_ATTRIBUTES` properly may be a cleaner fix there than threading `buildOpenerAuthContext` through five call sites.

## Not yet decided

Whether the right answer is to make `WriteAuthorizedByHandle` (or an equivalent) satisfy the explicit-timestamp gate, or to keep the POSIX gate authoritative and have the SMB layer refuse the open. That is a real design question about how far the SMB access model should be allowed to override POSIX semantics, and it should be settled deliberately rather than by whichever patch lands first.

marmos91:
Fixed in #2313.

Issue #2280, opened 2026-09-02
restoreParentDirFrozenTimestamps loses a directory's frozen timestamps when the child operation comes from another user


marmos91:
Measured while fixing #2205 (PR #2279), deliberately left out of that PR's scope.

## The defect

`restoreParentDirFrozenTimestamps` is broken by the same ownership gate that caused #2205, but unlike the CLOSE path in #2196 it **is reachable through the protocol**, because freeze and restore are driven by *different* principals:

- The freeze belongs to whoever holds the open **directory** handle and set the `-1` sentinel on it.
- The restore is driven by another user's child **create / rename / delete**, and runs with *that* user's identity.

So the two are not gated identically — which is exactly the property that made #2196 unreachable and makes this one reachable.

`pkg/metadata/file_modify.go` requires ownership for an explicit timestamp write (`!isOwner && !isRoot` → `ErrPermissionDenied`, decided before `checkWritePermission` is consulted), so the restore silently fails for a non-owning caller.

## Measured

The directory's frozen Mtime and Ctime are **lost** when the child rename comes from a caller who does not own the directory, and **survive** when the caller may write them. Same operation, two outcomes, selected by an unchecked permission check — the same shape as #2205, but on the parent-directory path rather than the file path.

## Why it was not fixed in #2279

The fix needs the restore to run under the frozen handle's own principal (`buildOpenerAuthContext`) rather than the caller's, which means threading an `*SMBHandlerContext` through five call sites. That is a larger change than #2205 warranted and would have made that PR's diff hard to review against its own claim.

## Related

- #2205 — same ownership gate, file path, fixed by removing the restore entirely.
- #2196 — same mechanism, CLOSE path, **not** reachable because freeze and restore share a principal there.
- Depends on the same underlying gap as the companion issue on `FILE_WRITE_ATTRIBUTES` vs POSIX ownership.

marmos91:
Fixed in #2319.

Issue #2205, opened 2026-08-26
smb: rename's timestamp restore discards a PermissionDenied, so ChangeTime on an open handle turns on file mode


marmos91:
`saveTimestamps` in `internal/adapter/smb/handlers/set_info.go:1770` restores a file's Mtime and
Ctime after a rename by calling `SetFileAttributes` and discarding the result:

```go
return func() {
    _, _ = metaSvc.SetFileAttributes(authCtx, handle, &metadata.SetAttrs{
        Mtime: &mtime,
        Ctime: &ctime,
    })
}
```

When that write is denied, the rename's ChangeTime update stands and the client reading its own
still open handle sees the post rename value. When it succeeds, the client sees the pre rename value.
Nothing distinguishes the two cases at runtime, and nothing is logged.

## Measured on unmodified develop

Two handler fixtures, same rename, opposite outcomes. I instrumented the discarded error:

| fixture | restore result | Ctime while the renaming handle is open |
| --- | --- | --- |
| `setupStreamsDisabledShare` (export mode `0o777`) | `nil` | pre rename, preserved |
| `setupReparseShare` (no mode grant) | `PermissionDenied: operation not permitted` | post rename, advanced immediately |

A probe with no rename confirms CLOSE itself never moves Ctime, so the advance is the rename's own
write surviving a failed restore, not a flush.

## Why this is reachable outside the fixtures

A rename is authorized by the parent directory. The restore writes the file. A file whose mode
denies the caller a write, sitting in a directory the caller can write, therefore renames
successfully and then keeps the advanced ChangeTime, while the same rename on a permissive file
does not. Two clients performing the same operation can read different ChangeTimes.

MS-FSA 2.1.5.15.12 requires the rename to update LastChangeTime, and product note `<187>` says NTFS
and ReFS defer that update until the handle closes. Whichever of those we implement, it cannot be
selected by whether an unchecked permission check passed.

## Scope

At minimum the error stops being discarded: log it, and decide deliberately whether a failed restore
fails the rename or is tolerated. Choosing the observable semantics is the larger half and wants its
own decision, not a silent default. See #2196 for the frozen ChangeTime case on the same path, and
#2182, which this supersedes.

marmos91:
Fixed in #2279 (a39ec44aa).

[...]

PR #2284, opened 2026-09-02
fix(smb): keep the ChangeTime a client observed across rename

marmos91:
`smb2.rename.simple_modtime` has failed on every metadata backend since #2279
landed, and it is not in `KNOWN_FAILURES.md`, so it fails the job. It is
currently the only new failure on #2282 and #2283 as well, which inherit it
from develop.

## What the test asserts

`simple_modtime` creates `file1.txt`, sleeps 5 seconds, renames it through the
*still-open* handle, then queries that handle:

```c
torture_assert_nttime_equal(torture, c1.out.change_time,
                            gi.all_info.out.change_time, "Bad timestamp\n");
```

`torture_assert_nttime_equal` stringifies its `got` argument into the message,
so

```
rename.c:1527: c1.out.change_time was 14:32:00 UTC, expected 14:32:05 UTC
```

reads as: the CREATE reply's ChangeTime was 14:32:00, the post-rename query
returned 14:32:05, and the test requires them equal. The 5-second gap is the
test's own `sleep(5)`. ChangeTime is not failing to advance — it **is**
advancing, and the test requires it not to while a handle is outstanding.

## Root cause

#2279 deleted `saveTimestamps`, which had put the pre-rename Ctime back after
`Service.Move` stamps it. That was the only thing keeping this test green.

The `restoreFrozenTimestamps` calls #2279 added to both rename branches are not
implicated: `buildFrozenAttrs` returns nil unless a `-1` sentinel actually froze
a field on the handle, and `simple_modtime` never sends one, so those calls are
inert here.

## Why #2279's unit test disagreed with conformance

`TestSetInfo_Rename_ChangeTimeAdvancesForEveryCaller` and smbtorture measure the
same field on the same path. They disagree about the correct answer, not about
the mechanism: the unit test was written from #2279's own premise and asserted
`after.Ctime.After(past)` — precisely the behaviour that breaks the conformance
run. It has been inverted to assert preservation, so the two layers can no
longer disagree. It fails on unmodified develop for both callers.

## The authority #2279 missed

MS-FSA 2.1.5.15.12's normative text says a rename updates LastChangeTime, which
is what #2279 acted on. That section's own product behaviour note says when it
becomes visible:

> **<187> Section 2.1.5.15.12**: The file system only updates LastChangeTime if
> no user has explicitly set LastChangeTime. **The NTFS and ReFS file systems
> defer setting LastChangeTime until the handle is closed.**

Both halves are needed. The normative line alone reads as an unconditional
update, and that reading is what turned the conformance run red. Both the code
comment and the test now cite the note alongside the conformance case.

## The fix

#2279 fixed a real defect (#2205): the write-back went through the
ownership-gated `SetFileAttributes` as the caller, so an owner's rename put the
old ChangeTime back and a non-owner's did not — one operation, two observable
ChangeTimes, selected by an unchecked permission check. #2279 resolved that
divergence by letting the advance stand for everyone. Note <187> and
conformance both say the other direction is the correct one, so the preserve is
restored and the divergence is closed by writing it back as
`metadata.NewSystemAuthContext`, which lands for every caller. Only Ctime is
snapshotted — `Move` leaves the renamed inode's Mtime alone, as #2279
established.

**This narrows a pre-existing clobber surface rather than introducing one.**
The write-back after `Move` is what `saveTimestamps` already did before #2279
removed it, and it did so for Mtime **and** Ctime. Restoring one field is
strictly less exposed than what shipped for months. A reviewer seeing "writes
the store after `Move`" will reasonably flag the concurrency hazard — it is
real and tracked in #2285; see the follow-up section.

This is not #2280 or #2281. Both concern the ownership gate on
`restoreFrozenTimestamps` / `restoreParentDirFrozenTimestamps`; neither is on
this path, and neither is touched here.

## Verification

Local `smb2.rename`, memory profile, same pinned samba-toolbox image as CI:

| | Total | Passed | New failures |
|---|---|---|---|
| develop `a39ec44a` | 13 | 12 | 1 — `smb2.simple_modtime` |
| this branch | 13 | 13 | 0 |

The baseline run reproduced the CI assertion character for character, so the rig
sees the bug rather than passing vacuously. `smb2.setinfo` is green on this
branch. `go test ./internal/adapter/smb/... ./pkg/metadata/...` passes.

Refs #2279, #2205. Unblocks #2282 and #2283.

## Review follow-up

**Issue numbers removed from source comments.** Both sites now state the
rationale in terms of the code, per CLAUDE.md. The references live here instead.

**The concurrency finding is real and is deliberately deferred, marked
`ponytail:` and tracked in #2285.** The rule is handle-scoped; this implements
it file-scoped by rewriting the stored timestamp, so a Ctime advanced between
the read and the write-back is walked backwards for every observer. Reachable:
`RequestOrder` orders only response emission — "Handlers keep running
concurrently" — so a pipelining client can have a WRITE in flight on the same
file, and a second handle or NFS can do it too.

Kept anyway, on three grounds:

1. It is the pre-existing behaviour, narrowed. `saveTimestamps` did the same
   read-then-write-back for Mtime **and** Ctime; this restores one field, so the
   clobber surface is strictly smaller than what shipped before #2279.
2. The overlay's read seam is small — `applyFrozenTimestamps` has two call
   sites — but its *invalidation* is not. The overlay must stop applying at the
   next genuine Ctime advance, including one through a different handle, which
   an `OpenFile` field cannot observe. Left applying too long it silently
   reports a stale ChangeTime for the life of the handle: worse than the race,
   and harder to notice. It also does not cover directory enumeration, which
   reports a child's ChangeTime with no `OpenFile` at all.
3. Develop is red on all five metadata backends and #2282/#2283 are blocked
   behind it.

### DIFF
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -1768,17 +1768,17 @@ func applyFrozenTimestamps(openFile *OpenFile, file *metadata.File) {
 
 // snapshotChangeTime reads a file's current ChangeTime and returns a function
 // that writes it back. Service.Move stamps the renamed inode's Ctime, but a
-// client that renames through an outstanding handle must keep observing the
-// ChangeTime it was handed in the CREATE reply — smbtorture
-// smb2.rename.simple_modtime compares exactly those two values.
+// client that renames through a handle it still holds open must keep observing
+// the ChangeTime that handle was handed in its CREATE reply, so the stamp is
+// put back.
 //
 // The write-back runs as the system identity on purpose. An explicit timestamp
 // write is ownership-gated in the metadata layer while the rename itself is
 // authorized on the parent directory, so writing the stamp back as the caller
-// lands for an owner and is silently refused for everyone else: one rename,
-// two observable ChangeTimes, selected by a permission check the caller never
-// asked for (#2205). As the system identity it lands for every caller, so the
-// ChangeTime a client observes no longer depends on who owns the file. A
+// would land for an owner and be silently refused for everyone else: one
+// rename, two observable ChangeTimes, chosen by a permission check the caller
+// never asked for. As the system identity it lands for every caller, so the
+// ChangeTime a client observes does not depend on who owns the file. A
 // read-only share cannot reach here — Move would already have refused.
 //
 // Only Ctime is snapshotted: Move leaves the renamed inode's Mtime alone, and
@@ -1787,9 +1787,18 @@ func applyFrozenTimestamps(openFile *OpenFile, file *metadata.File) {
 //
 // Returns a no-op if the read fails.
 //
-// ponytail: read-then-write-back costs an extra store round-trip per rename;
-// teach Service.Move to skip the stamp for its SMB caller only if rename
-// throughput ever shows up in a profile.
+// ponytail: the rule being implemented is handle-scoped — what a handle that
+// was already open keeps observing — but this writes the stored timestamp, so
+// it is file-scoped, and a Ctime advanced between the read and the write-back
+// (a WRITE, truncate or SET_INFO on another handle, or over NFS) is walked
+// backwards for every observer. The two coincide whenever nothing else is
+// touching the file. Upgrade to a per-OpenFile overlay consulted by
+// QUERY_INFO — the seam applyFrozenTimestamps already uses — once that overlay
+// can also say when to stop applying: it has to yield to the next real Ctime
+// advance, including one made through a different handle, which an OpenFile
+// field cannot see on its own. A directory enumeration reports a child's
+// ChangeTime with no OpenFile at all, so an overlay does not cover that path
+// either.
 func (h *Handler) snapshotChangeTime(authCtx *metadata.AuthContext, handle metadata.FileHandle) func() {
 	metaSvc := h.Registry.GetMetadataService()
 	file, err := metaSvc.GetFile(authCtx.Context, handle)
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -762,10 +762,12 @@ func (h *Handler) setFileInfoFromStore(
 			oldFileName := oldName.FileName
 			oldParentPath := GetParentPath(oldName.Path)
 
-			// Perform the rename. Move stamps LastChangeTime and leaves
-			// LastModificationTime untouched, which is what MS-FSA
-			// 2.1.5.15.12 asks for; neither is written back afterwards.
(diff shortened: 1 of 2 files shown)


marmos91:
Actioned the review.

Issue numbers are out of both source comments; the rationale now stands on its own and the references live in the PR body.

The concurrency finding is real, not theoretical — `RequestOrder` orders only response emission ("Handlers keep running concurrently"), so a pipelining client can have a WRITE in flight on the same file, and a second handle or NFS can do it too. Kept the file-scoped fix and marked it `ponytail:`, tracked in #2285.

Reasoning: it is the pre-existing behaviour narrowed rather than a new hazard (`saveTimestamps` clobbered Mtime *and* Ctime; this restores one field). The overlay's read seam is small — `applyFrozenTimestamps` has two call sites — but its invalidation is not: the overlay has to stop applying at the next genuine Ctime advance, including one through a different handle, which an `OpenFile` field cannot observe. Left applying too long it silently reports a stale ChangeTime for the life of the handle, which is worse than the race it replaces, and it does not cover directory enumeration at all.

`smb2.rename` re-run locally after the edits: 13/13, 0 new failures.

marmos91:
@copilot review

marmos91:
Restored the conformance case name in the test's failure message, and while verifying the citation found the authority #2279 actually missed.

MS-FSA 2.1.5.15.12's normative text says a rename updates LastChangeTime — that is what #2279 acted on. The same section's product behaviour note says when it becomes visible:

> **<187> Section 2.1.5.15.12**: The file system only updates LastChangeTime if no user has explicitly set LastChangeTime. **The NTFS and ReFS file systems defer setting LastChangeTime until the handle is closed.**

So the spec reading in #2279 was half of one. Both the helper comment and the test now cite the note alongside `smb2.rename.simple_modtime`, so the next reader learns the requirement and not just which test went red. No issue numbers in source.

`smb2.rename` re-run after the change: 13/13, 0 new failures.

marmos91:
Second-pass Copilot comments actioned — both were correct.

The comments claimed the code preserves "the ChangeTime the handle was handed in its CREATE reply". It actually snapshots the **pre-rename** ChangeTime; those coincide only when nothing advanced the file's Ctime in between. Reworded both sites to say pre-rename, and pointed at the `ponytail:` note, which already records that exact assumption — the file-scoped approximation tracked in #2285. Precision matters here specifically, since the whole point of the note <187> citation is which value is preserved.

Comments-only diff on top of a fully green run (`smbtorture / memory` and `smbtorture / badger-fs` both passed at 6c5665d0e; local `smb2.rename` 13/13). Ready for you to land.

PR #2292, opened 2026-09-02
perf(metadata): let a directory listing skip the attributes nobody reads

marmos91:
`ListChildren` filled `DirEntry.Attr` for every entry unconditionally. On badger that is a `Get`, a `decodeFile` and a link-count read **per child** — so a listing is 3N+1 operations, not the 2N+1 the shape suggests. Six of the eight callers throw the result away.

| caller | what it reads |
| --- | --- |
| `Service.LookupCaseInsensitive` | names — scans the **whole** directory on an exact-match miss |
| `file_modify.go:727`, `directory.go:209` | one entry, to test emptiness |
| `xattr.go` ×2 | names |
| `engine/audit_state.go` | names — then re-reads each file with `GetFile`, deliberately |
| `directory.go:95` (READDIRPLUS) | **attrs** |
| `smb/query_info.go` (ADS enumeration) | **attrs** — `entry.Attr.Size` |

The hot one is `LookupCaseInsensitive`, reached from SMB CREATE. Creating a new file in a large directory misses the exact-name index, scans every child, and fetches attributes for all of them to compare names.

## What changed

`ListChildren` takes a `ChildAttrs` argument. `WithAttrs` is the old behaviour and stays on the two callers that read attributes. `NamesOnly` leaves `Attr` nil and lets each backend skip the work that fills it:

- **badger** — drops the per-child `Get` + decode + link-count read
- **memory** — drops three deep copies (`Blocks`, `ACL`, `EAs`) per entry
- **sqlite / postgres** — a second query with the inode `LEFT JOIN` and seventeen columns removed, so no ACL/EA JSON decode either

An argument rather than a second method, so the compiler names every call site and the two forms cannot drift.

## Measured — badger, local NVMe, medians of five paired runs

| children | WithAttrs | NamesOnly | time | allocs |
|---:|---:|---:|---:|---:|
| 100 | 0.23 ms | 0.09 ms | 2.6× | 3,221 → 1,022 (3.2×) |
| 1,000 | 2.53 ms | 0.71 ms | 3.6× | 27,524 → 5,525 (5.0×) |
| 10,000 | 29.13 ms | 7.08 ms | **4.1×** | 270,539 → 50,534 (**5.4×**) |

Each tightening of the method moved this number down, so the history is worth
stating: a hand-rolled probe that skipped `DirEntry` construction and handle
encoding said 6.1×; the first real-API run said 4.8× but shared one store
between the two modes, letting whichever ran second read caches the first had
warmed; isolated stores and medians over five paired runs say **4.1×** at
10,000 children. Wall-clock on this machine varies 28–32 ms for the identical
`WithAttrs` case, so the allocation counts are the steadier measure — those are
stable to within ten allocations across runs.

Only badger is measured. The SQL saving is structural — nineteen columns and a
join become two columns — and is **not** quantified here.

## A bug this found

`findStreamChild`'s fallback scan returned `entries[i].Attr` from the listing. Its exact-name branch two lines above already resolves attributes with a single `GetFile`. The scan now does the same on the entry that matched, instead of paying for every child of the directory to use one. Caught by `XattrOps/StreamBackedGet` failing on sqlite, not by reading.

## Tests

`DirOps/NamesOnlyMatchesWithAttrs` pins the substitution guarantee: same entries, same order, same ids and handles, same cursor — whole and across a page boundary, since the SQL backends answer the two modes with different queries and a page boundary is where two queries meant to agree stop agreeing.

Mutation-verified rather than assumed:
- badger ignoring the flag → `entry 0 (file0.txt): NamesOnly populated Attr`
- sqlite's names-only query ordered by `child_id` → `entry 0: name = "file2.txt" under NamesOnly, "file0.txt" under WithAttrs`

Green: full unit suite; badger, memory, sqlite conformance; **postgres integration against a real postgres 16**, with the new case confirmed to run there.

## Not in scope

The scan stays O(N) — `LookupCaseInsensitive` still walks the directory, it just walks it ~5× cheaper. Removing the walk needs a case-folded name index, which is separate and not blocked by this.

One behaviour note: under `NamesOnly` a badger transaction's read set no longer includes the children's inode keys, so a concurrent write to a child's attributes no longer conflicts with a listing that never looked at them. That is the intended narrowing — structural changes to the directory are still in the read set — but it is a real change in when SSI conflicts fire.

### DIFF
--- a/internal/adapter/smb/handlers/query_info.go
+++ b/internal/adapter/smb/handlers/query_info.go
@@ -1021,7 +1021,7 @@ func (h *Handler) buildFileStreamInformation(authCtx *metadata.AuthContext, file
 			prefix := baseName + ":"
 			cursor := ""
 			for {
-				entries, nextCursor, listErr := store.ListChildren(ctx, name.ParentHandle, cursor, 1000)
+				entries, nextCursor, listErr := store.ListChildren(ctx, name.ParentHandle, cursor, 1000, metadata.WithAttrs)
 				if listErr != nil {
 					break
 				}
--- a/pkg/block/engine/audit_state.go
+++ b/pkg/block/engine/audit_state.go
@@ -213,7 +213,7 @@ func walkAuditShareFiles(
 ) error {
 	cursor := ""
 	for {
-		entries, next, err := store.ListChildren(ctx, dirHandle, cursor, 0)
+		entries, next, err := store.ListChildren(ctx, dirHandle, cursor, 0, metadata.NamesOnly)
 		if err != nil {
 			return fmt.Errorf("list children: %w", err)
 		}
--- a/pkg/controlplane/runtime/snapshot_consistency_test.go
+++ b/pkg/controlplane/runtime/snapshot_consistency_test.go
@@ -208,7 +208,7 @@ func (f *realBackupFixture) assertNoTornFiles(t *testing.T, store metadata.Store
 		t.Fatalf("snapshot %d: root handle missing from restored dump: %v", n, err)
 	}
 
-	entries, _, err := store.ListChildren(ctx, root, "", 0)
+	entries, _, err := store.ListChildren(ctx, root, "", 0, metadata.WithAttrs)
 	if err != nil {
 		t.Fatalf("snapshot %d: list children: %v", n, err)
 	}
--- a/pkg/metadata/directory.go
+++ b/pkg/metadata/directory.go
@@ -92,7 +92,7 @@ func (s *Service) ReadDirectory(ctx *AuthContext, dirHandle FileHandle, cookie u
 	token := s.cookies.GetToken(cookie)
 
 	// Call store's CRUD ListChildren method
-	entries, nextToken, err := store.ListChildren(ctx.Context, dirHandle, token, limit)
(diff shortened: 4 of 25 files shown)


marmos91:
Both review points taken, and the first one changed the headline number.

**Shared benchmark store.** Correct, and it mattered more than a nit: each sub-benchmark now builds its own store. Re-measured as medians over five paired runs, the gain at 10,000 children is **4.1×**, not the 4.8× the shared-store run reported. Wall-clock varies 28–32 ms for the identical `WithAttrs` case on this machine, so the allocation counts (270,539 → 50,534, stable to ~10 allocs across runs) are the steadier evidence. PR body updated with the full methodology history — the probe said 6.1×, the shared-store run said 4.8×, isolated medians say 4.1×.

**Overstated comment.** Also correct: paging each mode against itself would still pass if the two agreed internally and disagreed only with each other's cursors, which is exactly what a caller hits when it switches modes mid-listing. The test now resumes each mode from the other's cursor in both directions. Verified it fires: with sqlite's names-only cursor comparison mutated `>` → `>=`, it reports `resuming NamesOnly from a WithAttrs cursor yielded [... file1.txt file1.txt ...]`.

`ChildAttrs` also gets a `String`, so that message names the mode instead of printing the bool underneath it.

PR #2295, opened 2026-09-03
test(smb): make CLOSE observable on the handler fixtures

### LINKED ISSUE #2201: No SMB handler fixture can observe CLOSE-time timestamp behaviour on a regular file
Split out of the #2182 work because it is a scaffolding problem, not a protocol one, and it holds
regardless of whether the ctime defect in #2182 turns out to exist.

## Neither SMB handler fixture can measure CLOSE-time behaviour on a regular file

| fixture | closing a regular file | effect on timestamps |
| --- | --- | --- |
| `setupStreamsDisabledShare` | fails — `STATUS_INTERNAL_ERROR`, *"block store not available for handle"* | CLOSE never reaches its flush or the frozen restore |
| `setupReparseShare` (block-store backed) | succeeds | the deferred-write flush sets Ctime unconditionally, masking whatever the path under test did |

So a test that wants to assert what CLOSE does to a timestamp either cannot get there, or cannot
distinguish its subject from the flush.

## Why this is worth fixing rather than working around

It produced two failures in one PR that were both invisible until deliberately hunted:

1. **A test that passed while CLOSE was failing.** The first close-settlement test asserted only the
   Go error, not `resp.Status`. CLOSE was returning `STATUS_INTERNAL_ERROR` on every run and the
   test was green. Asserting the status is the fix for that one, and it is worth grepping for other
   `h.Close(...)` calls in tests that check `err` but not `Status` — `close_doc_dir_not_empty_test.go`
   closes a child file without checking status, which is fine for its purpose but is the same shape.

2. **A test that passed on unmodified develop.** Once moved to the block-store fixture, the test
   passed with the change, without the change, and with three separate build-safe mutants disabling
   the mechanism it was supposedly measuring. The flush was producing the result in every case.

The second is the general signal: **if a new test passes against the unmodified baseline, the
fixture is under test, not the code.** Worth running that check routinely rather than only when
something looks wrong — it is cheaper than mutation testing and catches a class mutation testing
does not (a mutant only tells you the test is not vacuous *for the code you mutated*).

## Suggested shape

A fixture that wires a local block store — so CLOSE completes — but leaves the file with **no**
pending writes, so CLOSE takes the non-flushing path. That is the configuration in which CLOSE-time
timestamp behaviour is observable at all, and there is currently no way to construct it.

Cheapest route is probably a variant of `setupReparseShare` that skips whatever marks the file dirty,
rather than a third fixture from scratch — worth checking what actually sets the pending-write state
on a freshly created file first, since a file that has never been written arguably should not have
one.

## Blocks

- #2182 — its premise cannot be tested without this.
- #2196 — reproduced, but the reproduction runs on the masking fixture, so the exact failure mode
  (restore not running vs. running and being overwritten) is not yet pinned down.

marmos91:
Closes #2201.

## Premises, measured before fixing

The issue makes two claims. One holds, one does not.

**Holds — `setupStreamsDisabledShare` cannot CLOSE a regular file.** The fixture wired no block store, so any handle carrying a payload fails CLOSEs block-store flush with `STATUS_INTERNAL_ERROR` and never reaches step 4 (metadata flush, LastAccessTime settle, frozen-timestamp restore). A file created directly in the metadata store closes fine, which is why this was not obvious; it needs the CREATE handler to allocate a payload to reproduce.

This was live in develop: `TestSetInfo_DeleteDisposition_NonEmptyDirectoryRefused` closes `parent\child` and checks only the Go error. Adding the status assertion on unmodified develop:

```
close_doc_dir_not_empty_test.go:228: close child: err=<nil> status=STATUS_INTERNAL_ERROR
```

Green on every run, with its CLOSE failing on every run.

**Does not hold — `setupReparseShare` does not mask the restore.** The issue expects the deferred-write flush to stamp Ctime unconditionally, and proposes building a third fixture that wires a block store but leaves no pending write. That fixture already exists: `setupReparseShare`s file is created and never written, so `FlushPendingWriteForFile` finds a cache-only state (`hasWriteData == false`) and returns before touching a timestamp. CLOSE-time behaviour is observable there today.

## What the mutant showed

Disabling `h.restoreFrozenTimestamps(authCtx, openFile)` in `close.go`:

| test | restore enabled | restore disabled |
| --- | --- | --- |
| `TestClose_FrozenTimestampsSurviveRename` (pre-existing) | PASS | **PASS** |
| `TestClose_FrozenTimestampsSurviveOutOfBandAdvance` (new) | PASS | FAIL |

So the issues suspicion about the existing test is confirmed, just for a different reason than the fixture: the rename preserves ChangeTime on its own since #2279 and advances none of the other three, so the restore never has anything to do. The test pins a true end-to-end outcome, it just does not cover the mechanism its name implies. Its doc comment now says so and points at the new test rather than deleting the coverage.

## Changes

- `create_streams_disabled_test.go` — cpstore + local memory block store on the share, so CLOSE completes.
- `close_doc_dir_not_empty_test.go`, `close_notify_test.go`, `ioctl_copychunk_sparse_integration_test.go` — assert `resp.GetStatus()` at the four close sites that checked only `err`.
- `close_frozen_timestamps_test.go` — `TestClose_FrozenTimestampsSurviveOutOfBandAdvance`: freeze via SET_INFO, advance Mtime/Ctime out of band as a second handle or an NFS client would, CLOSE, assert the frozen values won. Mutation-verified above.

Test-only. `go test -race ./internal/adapter/smb/handlers/` green; whole `internal/adapter/smb/...` green.

Removing the block store again reproduces the original failure through the new assertions, so the guard is verified to fire rather than merely being silent.

### DIFF
--- a/internal/adapter/smb/handlers/close_doc_dir_not_empty_test.go
+++ b/internal/adapter/smb/handlers/close_doc_dir_not_empty_test.go
@@ -170,8 +170,8 @@ func TestClose_DeleteOnClose_NonEmptyDirectorySucceeds(t *testing.T) {
 
 	// subdir-name, closed straight away.
 	fileID := dirTestCreate(t, h, ctx, "subdir-name", types.FileOpenIf, types.FileDirectoryFile)
-	if _, err := h.Close(ctx, &CloseRequest{FileID: fileID}); err != nil {
-		t.Fatalf("close subdir-name: %v", err)
+	if cr, err := h.Close(ctx, &CloseRequest{FileID: fileID}); err != nil || cr.GetStatus() != types.StatusSuccess {
+		t.Fatalf("close subdir-name: err=%v status=%v", err, cr.GetStatus())
 	}
 
 	// subname1 (a directory) renamed within subdir-name; handle left open.
@@ -224,8 +224,8 @@ func TestSetInfo_DeleteDisposition_NonEmptyDirectoryRefused(t *testing.T) {
 
 	parentID := dirTestCreate(t, h, ctx, "parent", types.FileOpenIf, types.FileDirectoryFile)
 	childID := dirTestCreate(t, h, ctx, "parent\\child", types.FileOpenIf, types.FileNonDirectoryFile)
-	if _, err := h.Close(ctx, &CloseRequest{FileID: childID}); err != nil {
-		t.Fatalf("close child: %v", err)
+	if cr, err := h.Close(ctx, &CloseRequest{FileID: childID}); err != nil || cr.GetStatus() != types.StatusSuccess {
+		t.Fatalf("close child: err=%v status=%v", err, cr.GetStatus())
 	}
 
 	resp, err := h.SetInfo(ctx, &SetInfoRequest{
--- a/internal/adapter/smb/handlers/close_doc_dir_not_empty_test.go
+++ b/internal/adapter/smb/handlers/close_doc_dir_not_empty_test.go
@@ -170,8 +170,12 @@ func TestClose_DeleteOnClose_NonEmptyDirectorySucceeds(t *testing.T) {
 
 	// subdir-name, closed straight away.
 	fileID := dirTestCreate(t, h, ctx, "subdir-name", types.FileOpenIf, types.FileDirectoryFile)
-	if cr, err := h.Close(ctx, &CloseRequest{FileID: fileID}); err != nil || cr.GetStatus() != types.StatusSuccess {
-		t.Fatalf("close subdir-name: err=%v status=%v", err, cr.GetStatus())
+	cr, err := h.Close(ctx, &CloseRequest{FileID: fileID})
+	if err != nil {
+		t.Fatalf("close subdir-name: %v", err)
+	}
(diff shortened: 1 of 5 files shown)


marmos91:
@copilot review

marmos91:
Addressed in 1782235.

**Split err/status checks — taken.** The nil dereference is not reachable against the current implementation: `h.Close` has three return statements and all three carry a non-nil `*CloseResponse`. But `GetStatus()` has a value receiver, so a nil response would panic rather than report, and a test that hides the real signal is exactly the defect this change set exists to remove. Split at all four sites, matching the form already used in `close_frozen_timestamps_test.go`.

Re-verified the guard still fires after the split — removing the fixture block store again:

```
close_doc_dir_not_empty_test.go:236: close child status = STATUS_INTERNAL_ERROR, want STATUS_SUCCESS
```

**Hard-coded future date — not taken.** `2030-01-01` is written into the store as a timestamp value and read back for an equality check against the frozen `2021-02-03`. It is never compared against the wall clock, so it does not become stale when that date passes; it only needs to differ from the frozen value. A `time.Now()`-relative date would make the test depend on clock behaviour it is not trying to measure.

Issue #2297, opened 2026-09-03
The operator CodeQL coverage guard attests source-archive membership, not analysis — it reported 22 of 22 while extraction had failed


marmos91:
Companion to #2296, which is the coverage gap this guard was added (#2180) to catch and did not.

## What the guard measures

`.github/workflows/codeql.yml`, "Verify the operator module reached the database", reads the CodeQL **source archive**:

```bash
archive=$(find "${RUNNER_TEMP}/codeql-db" -name src.zip -print -quit)
...
unzip -Z1 "${archive}" | grep -oE "k8s/dittofs-operator/.*\.go$" | sort -u >"${got}"
```

`src.zip` is the archived source text. A file lands there because the extractor saw it, not because it was successfully compiled and analysed. So the guard cannot distinguish "the operator was analysed" from "the operator failed to extract but its sources were archived anyway".

## Measured

Three runs of `Analyze (go)`, all on bases containing `fe1a155c7`, all with the **identical** extraction failure in the autobuild log (`go.mod requires go >= 1.26.0 ... Extraction failed for k8s/dittofs-operator`):

| run | extractor invocation | guard reported | job |
| --- | --- | --- | --- |
| 33726453595 (PR #2278) | `go-extractor [--source-root <repo root> ./...]` | `22 of 22` | pass |
| 33725348432 (PR #2293) | `go-extractor [--source-root <repo root> ./...]` | `22 of 22` | pass |
| 33726953310 (PR #2295) | `go-extractor [./...]` | `0 of 22` | **fail** |

Same failure in all three. The guard passed on two of them.

What actually varied is whether the extractor was invoked with `--source-root`. With it, archived paths are repo-relative and match the `k8s/dittofs-operator/...` pattern the guard greps for; without it they are module-relative and match nothing. So the guards verdict turns on an incidental extractor flag, not on coverage.

## Why this matters more than the underlying gap

#2296 is one bad pin and one line to fix. This is the check that was supposed to make that class of gap impossible, and it certified the gap as covered on two consecutive PRs. Its one true-negative was an accident.

It is also self-concealing in the direction that matters: it fails **closed** only when paths happen not to match, and passes while reporting a full `22 of 22` when the module was not analysed at all. A reader checking whether coverage exists sees the strongest possible green.

## Suggested fix

[...]

PR #2298, opened 2026-09-03
ci(codeql): take the analyser toolchain from the operator go.mod

### LINKED ISSUE #2296: CodeQL has not analysed the operator module since the go directive went to 1.26.0
Surfaced by an unrelated PR (#2295) going red on `Analyze (go)` with a test-only diff.

## The defect

`.github/workflows/codeql.yml` pins the analyser toolchain:

```yaml
      - name: Set up Go
        with:
          go-version: "1.25.x"
```

`fe1a155c7` ("bump the operator-dependencies group in /k8s/dittofs-operator") moved `k8s/dittofs-operator/go.mod` from `go 1.25.1` to `go 1.26.0`. The autobuild runs with `GOTOOLCHAIN=local`, so it cannot fetch a newer toolchain:

```
go: go.mod requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)
2026/09/03 07:14:04 Extraction failed for k8s/dittofs-operator: exit status 1
2026/09/03 07:14:04 Warning: extraction failed for 1 project(s): k8s/dittofs-operator
```

The operator module has not been extracted, and therefore not analysed, on any run since that commit. 22 source files.

`fe1a155c7`s own CodeQL run was **cancelled**, so the check never reported on the commit that introduced it — the same shape as #2104, where the analysis that would have caught a gap never ran on the change that opened it.

## Why the pin cannot simply be raised in place

The pin is deliberate and documented:

> Pin the toolchain rather than letting autobuild resolve one: a Go version drift between here and the other workflows changes which packages resolve, and an unresolved package is analyzed as empty rather than reported as an error.

That rationale is sound but assumes one toolchain can serve the analysis. It cannot: this workflow builds **two** modules with different floors — root `go 1.25.0`, operator `go 1.26.0`. Any single hardcoded version drifts from one of them on the next bump.

`operator-tests.yml` and `release.yml` already avoid this with `go-version-file: k8s/dittofs-operator/go.mod`. `codeql.yml` is the only workflow that hardcodes a version while building the operator.

## Suggested fix

`go-version-file: k8s/dittofs-operator/go.mod` — the higher of the two floors, which builds the root module fine, and self-tracks future bumps so this cannot silently recur. `GOTOOLCHAIN=auto` would also work but gives up the pin the comment argues for.

## Acceptance

- `Analyze (go)` reports 22 of 22 operator files **and** no `Extraction failed for k8s/dittofs-operator` in the autobuild log.
- Verified by the companion issue on the coverage guard, which currently cannot tell those two states apart.

marmos91:
Closes #2296. Found while an unrelated test-only PR (#2295) went red on `Analyze (go)`.

## The defect

`codeql.yml` builds **two** modules — root (`go 1.25.0`) and `k8s/dittofs-operator` (`go 1.26.0` since `fe1a155c7`) — and pinned a single `go-version: "1.25.x"`. Autobuild runs `GOTOOLCHAIN=local`, so it cannot fetch a newer toolchain:

```
go: go.mod requires go >= 1.26.0 (running go 1.25.14; GOTOOLCHAIN=local)
Extraction failed for k8s/dittofs-operator: exit status 1
Warning: extraction failed for 1 project(s): k8s/dittofs-operator
```

Below a modules floor the analysis is not degraded, the module is **dropped**. 22 operator source files have gone unanalysed on every run since that bump. `fe1a155c7`s own CodeQL run was cancelled, so the check never reported on the commit that introduced it — same shape as #2104.

## The fix

`go-version-file: k8s/dittofs-operator/go.mod`, matching what `operator-tests.yml` and `release.yml` already do.

The existing comment argues for pinning rather than letting autobuild resolve, and that still holds — this keeps a pin, it just sources it. It has to be the operators floor: it is the higher of the two and builds the root module fine, whereas no hardcoded literal can serve both modules across a bump of either. Sourcing it from the file means the next bump of either module cannot silently reopen this.

Go 1.26.0 is known to resolve on these runners — `operator-tests.yml` reads the same file and is green on `82128705b`.

## Verification

The honest limit: this cannot be fully verified before merge, because the coverage guard that would confirm it is itself unreliable — see #2297. It currently reports `22 of 22` **while extraction is failing**, because it checks `src.zip` membership rather than extraction success, and its verdict turns on whether the extractor happened to be passed `--source-root`. So a green guard on this PR is necessary but not sufficient.

What to check on this PRs `Analyze (go)` run:

- [ ] guard reports `22 of 22`
- [ ] autobuild log contains **no** `Extraction failed for k8s/dittofs-operator`

The second is the one that matters, and is the condition #2297 will make the guard actually enforce.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -40,11 +40,19 @@ jobs:
       # version drift between here and the other workflows changes which
       # packages resolve, and an unresolved package is analyzed as empty
       # rather than reported as an error.
+      #
+      # Read the version from the operator's go.mod, the higher of the two
+      # floors in the repo — it builds the root module too, and no hardcoded
+      # version can serve both modules across a bump of either. Autobuild runs
+      # with GOTOOLCHAIN=local, so a toolchain below a module's floor does not
+      # downgrade the analysis, it drops the module: the operator failed to
+      # extract on every run between its go directive reaching 1.26.0 and this
+      # change. Matches operator-tests.yml and release.yml.
       - name: Set up Go
         if: matrix.language == 'go'
         uses: actions/setup-go@v7
         with:
-          go-version: "1.25.x"
+          go-version-file: k8s/dittofs-operator/go.mod
           cache: true
 
       - name: Initialize CodeQL


marmos91:
@copilot review

Issue #2270, opened 2026-09-02
Share USED bytes never decrease on delete — a quota'd share fills permanently


marmos91:
Found during the v0.31.0 release smoke test on Scaleway (NFSv4.1 + SCW S3).

## Observed

`USED` is monotonic. It rises with writes and never falls on delete, even when the share is completely empty.

Fixed build:

```
USED baseline:      1.17GiB
write 50 MB      -> 1.22GiB
delete that file -> 1.22GiB    files in share = 0
```

Same probe on the previous release binary, from an empty share:

```
USED baseline:      0
write 50 MB      -> 50.00MiB
delete that file -> 50.00MiB   files in share = 0
```

So this is **pre-existing, not a regression** — both builds behave identically. It is filed separately from #2265 and #2268 for that reason.

At the point of the first measurement above the share held **zero files and zero S3 objects** (block GC had reclaimed all 1.2 GB correctly, bucket verified empty), yet it still reported 1.17 GiB used. The number survives a server restart, so it is persisted rather than a stale in-memory counter or a display artifact.

## Impact

The smoke share was `unlimited`, so nothing failed. On a share created with `--quota-bytes`, the quota is consumed permanently: churning data through a share monotonically exhausts it regardless of how much is actually stored, and the only recovery is recreating the share. An operator deleting files to get under quota sees no effect.

It also makes `dfsctl share list` actively misleading for capacity planning — the column reads as current usage, not as a lifetime write total.

## Not the same as #2268

#2268 is about **local journal bytes** on disk not being reclaimed while a share is idle. This is about the **accounting figure** the control plane reports and enforces quota against. They were observed together but are independent: in the measurement above the local journal held 734 MB while `USED` reported 1.17 GiB and the true size was 0.

## Reproduction

Create a remote-backed share, note `dfsctl share list` USED, write a file, delete it, wait for the metadata to settle, and read USED again. No large dataset or cap pressure is needed — 50 MB reproduces it.

marmos91:
Fixed in #2288 (`9f35691f4`).

**The counter was never missing a decrement — the inode is never deleted.** `Service.RemoveFile` deliberately keeps the inode, dropping the directory entry and setting `nlink = 0` so `fstat(2)` on a descriptor that outlived the unlink still works, and nothing reaps that row afterwards: the only production callers of `Store.DeleteFile` are `RemoveDirectory` and rename-over-a-directory-destination, neither of which is a regular file's unlink path.

[...]

PR #2299, opened 2026-09-03
refactor(metadata): share what putFile marshals, keep what the dialects decide

marmos91:
Stage S6 of #1828. Supersedes #2294, which GitHub closed when #2293 merged and its base branch was deleted; this is the same commit rebased onto develop.

Both `putFile` bodies ran to ~240 lines and about 200 of them were the same work. Normalising placeholders and error helpers, the two functions differed in ~35 lines out of 162.

**Shared now:**
- `sql.InodeValues` turns a `metadata.File` into the twenty column values, in the order both the UPDATE and the INSERT name them. One builder for both statements means a column added to the table cannot be wired into one and forgotten in the other — previously each dialect had two hand-written copies of the same list, so four places to keep in step.
- `sql.ApplyPutQuota` holds the usage-delta arithmetic: the plain size change, the type change from a row that contributed nothing, and the chown that moves both the bytes and the inode from the old owner to the new.

**Per-dialect still, because the difference is real:**

Postgres is multi-writer, so it cannot read the pre-update row in a statement of its own — another transaction could change it before the UPDATE ran. It locks the row in a leading CTE and returns the old values from the UPDATE in the same round-trip. SQLite is single-writer and the transaction holds an exclusive write lock for its duration, so a SELECT followed by an UPDATE has no interleaving window, and it reads the row separately.

Writing those two the same way would either hand sqlite a CTE it does not need or hand postgres a race it cannot afford. The comments in each now say which property they are relying on, rather than describing the mechanics.

Each `putFile` is 139 lines, down from 239 and 255.

## Verification

The risk in this change is the quota arithmetic — it decides share usage counters, and a silent error there is the family of #2270. It is genuinely covered, which I confirmed rather than assumed: changing `ApplyPutQuota`'s chown branch from `-int64(oldSize)` to `0` fails `StoreSurface/GetQuotaUsageChown`.

- Full unit suite; `GetQuotaUsage`, `GetQuotaUsageChown` and `GetQuotaUsagePerShare` all run and pass
- **postgres integration against a real postgres 16** — the CTE path is not exercised by any unit test
- `golangci-lint`: 0 issues

### DIFF
--- a/pkg/metadata/store/postgres/transaction.go
+++ b/pkg/metadata/store/postgres/transaction.go
@@ -2,7 +2,6 @@ package postgres
 
 import (
 	"context"
-	"database/sql"
 	"encoding/json"
 	"errors"
 	"log/slog"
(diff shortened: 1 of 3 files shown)


marmos91:
Two of the four taken, one rejected on evidence.

**`inodes_object_id_idx`** — correct, fixed. Migration `000032_rename_files_to_inodes` renames it, and sqlite's initial schema creates it under that name outright, so `files_object_id_idx` had been stale in that comment since the rename.

**"share owner"** — correct, fixed in both dialects. The insert charges `file.UID`/`file.GID`, so it now says "its owning identity". The wording predates this PR but the line is mine to get right now that I moved it.

**The `database/sql` import** — this one is wrong. Inside `package sql`, an unaliased `import "database/sql"` binds `sql` in *file* scope to the standard library, and the package's own identifiers are referenced unqualified, so `sql.NullInt64` resolves exactly as intended. It compiles, and the whole suite plus CI on this head is green, which settles it independently of the language argument. It is also the existing convention in this package: `chunks.go`, `scan.go` and `files.go` all import `database/sql` unaliased. Aliasing only `put_file.go` would make it the odd one out.

PR #2301, opened 2026-09-03
fix(test/e2e): build against the testcontainers v0.44 port API

### LINKED ISSUE #2300: E2E has not run since the go-dependency bump: test/e2e/framework fails to build against testcontainers v0.44
Found while babysitting develop. `E2E Tests` is red on develop and has been for six commits, and the failure is a **build** failure — so no E2E test has executed in that window, on any merge.

## The break

`0cc2e2804` ("chore(deps): bump the go-dependencies group with 32 updates") moved `testcontainers-go` 0.40.0 → 0.44.0, which changed the mapped-port API:

- `Container.MappedPort(ctx, string)` now takes a plain `string`, not a `nat.Port`.
- It returns `github.com/moby/moby/api/types/network.Port`, which has **no** `Int()`. The replacement is `Num() uint16` (`Port() string` / `String()` also exist).

```
test/e2e/framework/containers.go:512:19: port.Int undefined (type "github.com/moby/moby/api/types/network".Port has no field or method Int)
test/e2e/framework/kerberos.go:76:47: cannot use nat.Port("88/tcp") (constant "88/tcp" of string type nat.Port) as string value in argument to container.MappedPort
test/e2e/framework/kerberos.go:91:25: mappedPort.Int undefined (...)

FAIL github.com/marmos91/dittofs/test/e2e           [build failed]
FAIL github.com/marmos91/dittofs/test/e2e/framework [build failed]
FAIL github.com/marmos91/dittofs/test/e2e/helpers   [build failed]
```

The files are behind `//go:build e2e`, so a plain `go build ./...` / `go vet ./...` / `go test ./...` cannot see any of it. Only the E2E workflow (`go test -tags=e2e`) compiles these packages, and Dependabot bump PRs do not run it.

## When

| commit | E2E |
| --- | --- |
| `9f35691f4` | success — last green |
| `ecdb4ed10` | not run |
| `eccc484d4` | cancelled |
| `fe1a155c7` | cancelled |
| `0cc2e2804` | **failure** — first |
| `82128705b` … `0bd9414e0` | failure |

Six commits merged into develop since, none with E2E coverage.

## Why this got missed

The check was visibly red, not silently absent — this is not #2104s failure mode. What hid it is that a **build** failure and a **test** failure look identical at the check level, and three of the runs immediately before the break were `cancelled` or absent, so the transition from green to red did not read as a step change in the run list.

Worth noting `#2212` made a failing E2E run conclude `failure` rather than `cancelled`; without it this would still be showing as cancelled and be even easier to skim past.

## Suggested guard (not in the fix PR)

A build-only job that runs `go test -tags=e2e -c -o /dev/null ./test/e2e/...` on every PR would have caught this in the Dependabot PR itself, in seconds, with no docker/sudo. Every build-tagged tree in the repo has the same blind spot — `windows`, `darwin`, and `e2e` files are invisible to the default test matrix. Filing separately if wanted.

marmos91:
Closes #2300. Found while babysitting develop, which has been red on `E2E Tests` for six commits.

## The defect

`0cc2e2804` bumped `testcontainers-go` 0.40.0 → 0.44.0, changing the mapped-port API:

- `MappedPort(ctx, string)` takes a plain `string`, no longer a `nat.Port`
- it returns `moby/api/types/network.Port`, which has `Num() uint16` where `nat.Port` had `Int()`

Three call sites still used the old shape, so **three packages failed to build** and no E2E test has executed since:

```
FAIL github.com/marmos91/dittofs/test/e2e           [build failed]
FAIL github.com/marmos91/dittofs/test/e2e/framework [build failed]
FAIL github.com/marmos91/dittofs/test/e2e/helpers   [build failed]
```

## The change

Three lines, plus the now-unused `nat` import:

- `containers.go:512` — `port.Int()` → `int(port.Num())`
- `kerberos.go:76` — `MappedPort(ctx, nat.Port("88/tcp"))` → `MappedPort(ctx, "88/tcp")`
- `kerberos.go:91` — `mappedPort.Int()` → `int(mappedPort.Num())`

`Num()` returns `uint16`; both destination fields are `int` and are consumed as `int` throughout (`KDCHelper.kdcPort` into `krb5.conf`, `PostgresHelper.Port` into a DSN), so the conversion happens at the boundary rather than widening the field types and rippling through every caller.

I checked all four `MappedPort` call sites, not just the ones the compiler stopped at — `containers.go:114` and `:377` already pass strings and use the result without `Int()`, so they were already correct.

## Verification

The compiler is the test here, and it could not run before this change. All four e2e packages now build their **test binaries** (`go build` alone would not have caught this — it skips `_test.go`):

```
ok: github.com/marmos91/dittofs/test/e2e
ok: github.com/marmos91/dittofs/test/e2e/framework
ok: github.com/marmos91/dittofs/test/e2e/helpers
ok: github.com/marmos91/dittofs/test/e2e/snapshot
```

`go vet -tags=e2e ./test/e2e/...` clean, `go build ./...` unaffected. The pre-fix tree reproduces the three errors, so the change is load-bearing rather than cosmetic.

What this PR does **not** establish is that the E2E suite passes — it has not run in six commits, so this run is the first real signal since `9f35691f4`. If it comes back red on an actual test, that is a separate finding, not a regression from this diff.

## Follow-up, deliberately not here

A `go test -tags=e2e -c -o /dev/null ./test/e2e/...` job would have caught this inside the Dependabot PR in seconds with no docker or sudo. Build-tagged trees (`e2e`, `windows`, `darwin`) are invisible to the default matrix — noted in #2300.

### DIFF
--- a/test/e2e/framework/containers.go
+++ b/test/e2e/framework/containers.go
@@ -509,7 +509,7 @@ func NewPostgresHelper(t *testing.T) *PostgresHelper {
 		T:         t,
 		Container: container,
 		Host:      host,
-		Port:      port.Int(),
+		Port:      int(port.Num()),
 		Database:  "dittofs_e2e",
 		User:      "dittofs_e2e",
 		Password:  "dittofs_e2e",
--- a/test/e2e/framework/kerberos.go
+++ b/test/e2e/framework/kerberos.go
@@ -12,7 +12,6 @@ import (
 	"testing"
 	"time"
 
-	"github.com/docker/go-connections/nat"
 	"github.com/stretchr/testify/require"
 	"github.com/testcontainers/testcontainers-go"
 	"github.com/testcontainers/testcontainers-go/wait"
@@ -73,7 +72,7 @@ func NewKDCHelper(t *testing.T, cfg KDCConfig) *KDCHelper {
 	require.NoError(t, err, "failed to start KDC container")
 
 	// Get the mapped port
-	mappedPort, err := container.MappedPort(ctx, nat.Port("88/tcp"))
+	mappedPort, err := container.MappedPort(ctx, "88/tcp")
 	require.NoError(t, err, "failed to get KDC port")
 
 	host, err := container.Host(ctx)
@@ -88,7 +87,7 @@ func NewKDCHelper(t *testing.T, cfg KDCConfig) *KDCHelper {
 		keytabDir: keytabDir,
 		ccacheDir: ccacheDir,
 		kdcHost:   host,
-		kdcPort:   mappedPort.Int(),
+		kdcPort:   int(mappedPort.Num()),
 	}
 
 	// Generate krb5.conf for this KDC instance


marmos91:
@copilot review

PR #2302, opened 2026-09-03
fix(block): hold the retention pin in the local store so Start cannot lift it

### LINKED ISSUE #2257: block: RetentionPin is silently defeated at every startup — Start overwrites the constructor's eviction gate
`blockstore_config.go:370` builds the local store with eviction correctly gated on the retention policy:

```go
// Eviction requires a remote store (so evicted blocks can be re-fetched) and
// must not be pin mode (pin keeps blocks stored locally indefinitely).
localStore.SetEvictionEnabled(remoteStore != nil && config.RetentionPolicy != block.RetentionPin)
```

`bs.Start` then runs `engine.go:271`:

```go
bs.local.SetEvictionEnabled(bs.syncer.IsRemoteHealthy())
```

which ignores the retention policy entirely and overwrites the decision. `Start` runs after the constructor on every boot, so **a share configured `RetentionPin` has eviction enabled from startup onward**, and blocks the policy promises to keep locally are evicted as soon as the local tier comes under pressure.

Two more writers have the same defect:

- `engine.go:250`, the health-transition callback — `bs.local.SetEvictionEnabled(healthy)`, again with no retention term.
- the attach-a-remote path in `syncer.go:1051`.

The only writer that re-applies the pin is an explicit retention-policy *update* (`share_config.go:41`). So the pin holds only between a policy edit and the next restart, and a share that is never edited never has it.

## Impact

The retention policy is a durability/locality promise to the operator, and it is silently not kept. Nothing logs the override — the constructor's decision is simply replaced. An operator who sets `RetentionPin` and restarts gets eviction with no diagnostic.

This is also what makes #2227 reachable on a local-only share: it is the same overwrite that re-enables eviction where `:370` had disabled it for a nil remote, which is how cold intervals come to exist on a share that has no remote to fetch them back from.

## Fix

The retention policy is a property of the share's configuration, not of remote health, so the two must be combined rather than one overwriting the other. Every writer of `SetEvictionEnabled` needs the retention term — or, better, the local store should hold the pin itself and AND it in, so a caller cannot forget. The latter matches the shape the repo has just been bitten by twice elsewhere (#2228, #2251): a value a construction path sets correctly, which a later unconditional write clobbers, is a pattern worth removing structurally rather than patching per call site.

## Notes

- Verified against `origin/develop` at `3ef4d7d74`.
- Found while fixing #2227; not fixed there, since it is a distinct defect in a different package with its own blast radius.
- No production exposure (customers are evaluating only), consistent with D4 of the block-dataflow master plan.

marmos91:
Closes #2257.

## The defect

`blockstore_config.go` builds the local store with eviction gated on the retention policy, then `bs.Start` overwrites that decision with `bs.local.SetEvictionEnabled(bs.syncer.IsRemoteHealthy())` — which knows nothing about retention. `Start` runs on every boot, so **a share configured `RetentionPin` has eviction enabled from startup onward**. Two more writers had the same defect: the health-transition callback and the attach-a-remote path.

## Premise verification

Against unmodified `origin/develop`, a pinned share freed **all 16,778,528** of its local bytes after `Start` — byte-for-byte identical to an unpinned control run alongside it. `retention: pin` has never pinned anything on a healthy remote. The control matters: without it, "freed 0" would only prove the probe cannot evict at all.

## The fix

Retention stops being a value any health writer can clobber. The journal store holds `evictionSuspended` (health) and `evictionPinned` (policy) as independent atomics, and `evictionHeld()` reads them together. The three writers keep saying what they know about health; none can lift the pin.

`FSStore` inherits the new method through its embedded `*journal.Store`; only `MemoryStore` needed a no-op. `UpdateShare` collapses from a six-line if/else-if to one line.

Both review passes independently caught a race in the first attempt — I had cached the gate in a derived field refreshed by a helper, an unsynchronised read-modify-write over two atomics that reintroduced the same class of bug one layer down, on a health-flap-meets-admin-toggle interleaving. Computing on read is both correct and the smaller diff.

## A second latent bug, fixed incidentally

The old `UpdateShare` forced eviction back on when unpinning a remote-backed share **even if the remote was unhealthy**.

## Tests

`TestRetentionPinSurvivesStart` (real journal store, real memory remote, no fake sink, with an unpinned control) and `TestEvictionPinOutranksEnable`. Both verified against a deliberately broken build — with the pin dropped from `evictionHeld`, they fail with `pinned share freed 16778528 local bytes` and `pinned but evicted 1 segments`.

## One consequence to flag

The existing startup warning about pinned + bounded shares hitting `ErrDiskFull` described a state that could not previously occur, because the pin was lifted seconds after being set. It can occur now — which is the documented intent. `ensureSpace` degrades in a bounded way rather than hanging.

### DIFF
--- a/pkg/block/engine/api_accessors_test.go
+++ b/pkg/block/engine/api_accessors_test.go
@@ -73,14 +73,6 @@ func TestStore_Accessors(t *testing.T) {
 	_ = bs.LocalStats()
 }
 
-// TestStore_EvictionSetter exercises the delegating setter (it must not panic).
-func TestStore_EvictionSetter(t *testing.T) {
-	bs := newTestEngine(t, 64*1024*1024, 0)
-
-	bs.SetEvictionEnabled(true)
-	bs.SetEvictionEnabled(false)
-}
-
 // TestStore_HealthCheck covers both the legacy error probe and the
 // structured Healthcheck for a healthy local-only engine.
 func TestStore_HealthCheck(t *testing.T) {
--- a/pkg/block/engine/health.go
+++ b/pkg/block/engine/health.go
@@ -86,3 +86,10 @@ func (bs *Store) HasRemoteStore() bool {
 func (bs *Store) SetEvictionEnabled(enabled bool) {
 	bs.local.SetEvictionEnabled(enabled)
 }
+
+// SetEvictionPinned pins the local store's bytes in place for a pin-retention
+// share. It survives the health-driven SetEvictionEnabled calls the syncer and
+// Start make, so it is the only correct way to express a retention pin.
+func (bs *Store) SetEvictionPinned(pinned bool) {
+	bs.local.SetEvictionPinned(pinned)
+}
--- a/pkg/block/engine/health.go
+++ b/pkg/block/engine/health.go
@@ -81,12 +81,6 @@ func (bs *Store) HasRemoteStore() bool {
 	return bs.remote != nil
 }
 
-// SetEvictionEnabled controls whether the local store can evict blocks to free disk space.
-// Delegates to the local store's SetEvictionEnabled method.
-func (bs *Store) SetEvictionEnabled(enabled bool) {
-	bs.local.SetEvictionEnabled(enabled)
-}
-
 // SetEvictionPinned pins the local store's bytes in place for a pin-retention
 // share. It survives the health-driven SetEvictionEnabled calls the syncer and
 // Start make, so it is the only correct way to express a retention pin.
--- a/pkg/block/engine/retention_pin_start_test.go
+++ b/pkg/block/engine/retention_pin_start_test.go
@@ -11,10 +11,11 @@ import (
 	metadatamemory "github.com/marmos91/dittofs/pkg/metadata/store/memory"
 )
 
-// premiseDrainFreed builds a remote-backed engine, optionally disables eviction
-// the way the share constructor does for a pinned share, fills it and reports
-// how many local bytes an explicit force-evict reclaims.
-func premiseDrainFreed(t *testing.T, pin bool) int64 {
+// drainAfterFill builds a remote-backed engine the way the share constructor
+// does — pinning the local store before Start when the share's retention policy
+// is pin — fills it with a file that is rolled up and uploaded, then force-evicts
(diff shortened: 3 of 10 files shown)


marmos91:
Rebased onto current develop and folded in the review findings; requesting a fresh look.

marmos91:
On the eviction-on-a-local-only-share point: the concern is real but the suggested predicate is not, and the fix does not belong in this PR.

**`bs.HasRemoteStore()` is the wrong gate.** It reads `bs.remote != nil`, which is not the same thing as "this share can fetch bytes back" — the fetch path goes through the syncer's `remoteBlockStore`, wired separately by `SetRemoteBlockStore`. I tried the suggested line and it broke this PR's own control subtest: `TestRetentionPinSurvivesStart/unpinned_evicts` builds a share with a healthy remote on the syncer but no `BlockStoreConfig.Remote`, so the gate closed and the unpinned control freed 0 bytes. A gate that answers "no remote" for a share that demonstrably has one would suppress eviction on real shares.

**On the reachability.** Two independent reviews traced it: nothing is evictable on a share that never synced, because eviction requires `syncedRecords == records` (`reclaim.go:115`, `:211`) and the synced flag only advances through the carve path, which needs a remote. Verified byte-identical on `origin/develop`, so this PR neither introduces nor widens it — the `remoteStore != nil` term it removed was already dead, clobbered by the same `Start` call on every boot.

There **is** one narrow chain that reaches it, and it deserves its own issue rather than a line wedged in here: synced flags are persisted per record and restored on replay (`recovery.go:342`), and `RemoveShare` deletes only `snapshots/`, never the journal dir (`lifecycle.go:494-542`). A share recreated local-only over a surviving journal replays synced records, and both nil-remote fetch branches then return nil error without hydrating (`fetch.go:382-385`, `:484-486`) — silent zeros. Filing that separately: it needs the right predicate, its own regression test, and a decision about whether `RemoveShare` should be taking the journal dir with it, none of which is a retention-pin change.

The comment that made this claim was wrong and is fixed — it asserted `Start` leaves eviction disabled for a local-only share, when `IsRemoteHealthy()` returns true for a nil monitor. The grammar nit is gone with it.

PR #2305, opened 2026-09-03
chore(block): remove the cas→blocks migration

marmos91:
First PR of the block-ingestion improvement program: pure subtraction, no behaviour change beyond one deliberate refusal.

**1,596 lines removed**, 663 of them non-test.

## What goes

The one-shot cas→blocks migration, which imported pre-block standalone objects and repacked them. There is no store left for it to run against (master plan D4: no production stores), so it goes along with the `LegacyCASStore` interface, its four backend implementations, and the GC pass that purged the `cas/` namespace behind it.

## One deliberate behaviour change

A synced locator carrying no `BlockID` was written by a release predating the packed-block format. The reader for it is gone, so it now **fails closed** with `ErrChunkNotFound`. Removing the fallback without this would have fallen through to `block.FormatBlockKey("")` — a bogus object key, i.e. the serve-wrong-bytes class. `TestDispatchRemoteFetch_StandaloneLocatorServedViaFallback` is rewritten as `..._PreBlockFormatLocatorIsRefused` so the coverage is kept rather than dropped.

## Three files were named for something they do not contain

`legacy_cas_migration.go` in each backend in fact held the **entire hash-keyed CAS surface** (`Put`/`Get`/`GetRange`/`Has`/`Head`/`Delete`/`Walk`/`ReadBlockVerified`), and `legacy_cas_verifier.go` held a general-purpose BLAKE3 verifying reader. Deleting them by name would have removed live code. They are renamed to say what they hold: `cas.go`, `verifier.go`, `cas_key.go`.

Whether the hash-keyed `cas/` surface is itself still reachable in production is a separate question and is **not** answered here.

## Consequence worth stating

Nothing purges the `cas/` namespace any more, so a bucket still holding those objects keeps them — unreadable and unreclaimable. Under D4 no such store exists, and it is one `aws s3 rm --recursive s3://<bucket>/cas/` for an operator.

## Verification

- `go build ./...`, `go vet ./...` clean repo-wide
- `go test ./...` green; `go test -race ./pkg/block/...` green
- `GOOS=windows` and `GOOS=darwin` builds clean — build-tagged trees are invisible to `go build ./...`, which is how #2300 hid a broken `test/e2e/framework` for six commits
- E2E suite dispatched on this branch (it never runs on PRs)

Also removed: a GC test asserting "no `cas/` objects were produced, proving `mirrorChunk` never ran" — `mirrorChunk` does not exist in non-test code at all, so that assertion was already vacuous.

## Not in this PR

The pre-journal `local/fs` legacy layout. Removing it cascades into the cold-seed path (`legacy_verify.go` also holds `seedColdIfNeeded` and friends), which needs its own tracing. Separate PR.


marmos91:
Rebased onto current develop and folded in the review findings; requesting a fresh look.

Issue #2268, opened 2026-09-02
Deleting every file in a share leaves its local journal bytes on disk indefinitely


marmos91:
Found during the v0.31.0 release smoke test on Scaleway (NFSv4.1 + SCW S3). Distinct from #2265, though it shares a root cause.

## Observed

After deleting all 12 x 100 MB files from a remote-backed share:

```
files in share:      0
S3 objects:          0        (block GC reclaimed all 1.2 GB correctly)
local journal bytes: 734,122,443
```

The 734 MB survives a server restart. Nothing on an idle share ever reclaims it.

## Why

Tombstones mark the records dead, but local reclamation only ever scans the **sealed** set:

- `pickVictim` (`pkg/block/journal/reclaim.go`) returns `nil` when `len(sh.sealed) == 0`, so the 30s background repack is a no-op.
- `claimColdestEvictable` likewise iterates `sh.sealed` only.

Sealing happens in exactly one place — an append that would overflow `SegmentSize`. A share whose files never push a shard past 256 MiB therefore has an empty sealed set, and every dead byte sits in an active segment nothing looks at. Deleting the files removes the last writer, so no future append will ever seal them either.

## Scope

Bounded, unlike #2265: at most one active segment per shard, so 16 x 256 MiB = 4 GiB worst case with the defaults. It is not a wedge — #2266 lets the write-path gate force-seal synced actives, so a subsequent writer on the same share reclaims this under cap pressure. The leak only persists while the share stays idle.

That makes it a disk-usage surprise rather than a correctness bug: an operator who deletes everything to free space sees nothing returned until they write again.

## Note on what is deliberate

Repacking a *live* active segment would fight the append path, so `pickVictim`'s sealed-only scan is a reasonable design choice, not an oversight. The gap is that nothing force-seals on the **delete** path the way #2266 now does on the write path. A tombstone that empties a segment is the natural trigger — `reclaimEmptied` already handles the narrower case where a cold read hydrated the removed file's bytes back into the active segment.

## Reproduction

`test/crash`-style rig used here: create a remote-backed share, write 12 x 100 MB over NFSv4.1, delete all of it, run `dfsctl store block gc <share>` with a grace period below the object age, then compare `du -sb <share>/journal` against the now-empty bucket.

marmos91:
**Correcting my own root cause in the issue body — it is wrong.**

[...]

PR #2309, opened 2026-09-03
fix(metadata): rename over an existing file never freed the clobbered file's bytes

### LINKED ISSUE #2287: Rename over an existing destination leaks the clobbered file's payload (NFSv3 and v4)
Surfaced while fixing #2268 (PR #2286). Not fixed there — it is a different path with a different owner, and it needs a metadata-layer change rather than an adapter one.

## The gap

A POSIX rename onto an existing destination silently unlinks that destination. When it does, the clobbered file's payload is never handed to the block store, so its local bytes are never reclaimed:

- `internal/adapter/nfs/v4/handlers/rename.go:152`
- `internal/adapter/nfs/v3/handlers/rename.go:258`

This is the same class as #2268 — a caller that drops a `PayloadID` the metadata layer expects it to act on — but it is **not** fixable the same way. `metadata.Service.Move` drops the victim's last link inside its own transaction and never surfaces the victim's `PayloadID`, so neither adapter can call `blockStore.Delete` even if it wanted to. The fix has to start by having `Move` return what it removed.

## Why it is not already covered

- **Trash-enabled shares are fine**: the victim goes through `recycleNode` rather than being unlinked outright.
- **SMB is fine**: it pre-deletes the destination, so the payload is released on that path.
- So the exposure is NFSv3 and NFSv4 rename-over-existing on a share without trash enabled.

## Impact

Local cache bytes accumulate with no way to reclaim them short of the write-path pressure gate (#2266) evicting them once writes resume. Same shape as #2268: bounded and self-clearing under write pressure, permanent on an idle share. Workloads that rename over a destination repeatedly — atomic write-to-temp-then-rename, which is extremely common — churn this steadily.

Unlike #2268 the remote side is likely fine, since block GC derives orphans from (synced − live) in the metadata store and the metadata unlink does happen. Worth confirming rather than assuming.

## Suggested shape

Have `metadata.Service.Move` return the clobbered victim's `PayloadID` the way `RemoveFile` already returns the removed file's, then have both rename handlers release it — mirroring what #2286 does for NFSv4 REMOVE. Doing it in the service rather than in each adapter is what keeps v3, v4 and any future adapter from having to remember.

A regression test should assert the observable (the victim's bytes actually leave the local tier after a clobbering rename) rather than that a particular function was called.

marmos91:
## What was wrong

POSIX rename onto an existing name silently unlinks whatever was there. `metadata.Service.Move`
did that unlink inside its own transaction and returned only WCC data, so the clobbered file's
`PayloadID` never left the metadata layer:

```go
func (s *Service) Move(...) (*RenameWcc, error)
```

No adapter could call `blockStore.Delete` for the victim even if it wanted to. The victim's
records therefore stayed indexed as **live** in the per-share local tier, where no reclamation
path — fast-path retire, eviction, repack — can treat them as dead, and the bytes survived every
restart. Only the local tier was affected: remote block GC derives orphans from (synced − live)
in the metadata store, which the metadata unlink alone already drives.

Write-to-temp-then-rename is one of the most common write patterns there is, so this churns
steadily rather than being a corner case.

This is the same class as #2268 (fixed in #2286: a caller discarding a `PayloadID` the metadata
layer handed it), but it was not fixable the same way — there was nothing to discard yet.

## The issue's premise was partly wrong: SMB is not exempt

The issue reasoned that SMB is safe because it pre-deletes the destination. It does not, in
general. In `internal/adapter/smb/handlers/set_info.go` the pre-remove is gated on a *case
mismatch*:

```go
if isOverwrite && dstMatchedName != "" && dstMatchedName != toName {
    if _, _, rmErr := metaSvc.RemoveFile(authCtx, toDir, dstMatchedName); rmErr != nil {
```

An exact-case `ReplaceIfExists` rename — the overwhelmingly common case — skips that branch
entirely and reaches `Move`'s clobber path exactly like NFS does. And the branch that *does*
fire discarded `RemoveFile`'s returned file, so it leaked the payload too, in the #2268 shape.

Both SMB sites are fixed here. Stream rename (`foo.txt:a` → `foo.txt:b`) had the same exposure
and is fixed as well.

Trash-enabled shares really are exempt, as the issue says — `recycleNode` relocates the victim
instead of destroying it, and `Move` nils its destination snapshot on that path, so nothing is
released.

## The fix

`Move` now returns the clobbered victim, mirroring the contract `RemoveFile` has carried all
along:

```go
func (s *Service) Move(...) (*File, *RenameWcc, error)
```

`nil` when the rename replaced nothing, replaced a directory (which owns no content), or trash
recycled the victim. Non-nil with an empty `PayloadID` when a remaining hard link still
references the content. The link count the transaction actually wrote is captured inside the
closure, so the ownership decision comes from the committing attempt rather than from a read
taken outside it.

**No metadata store interface change was needed.** The victim's `PayloadID` was already read
outside the transaction (`store.GetFile(ctx, dstHandle)`), and the transaction aborts with
`ErrConflict` if the destination's identity changed under it, so that snapshot is guaranteed to
name the entry the transaction unlinked. No backend, and no `storetest` contract, gains a
method.

Adapters release the bytes through one shared seam, `common.ReleaseClobberedPayload`, modelled
on the existing `common.ReclaimTruncatedBlocks` — so a future protocol adapter cannot forget it
while another gets it right. Best-effort by contract: the metadata rename has already committed
and the client already has its answer, so a block-store failure is logged rather than surfaced,
consistent with every other release path.

## Verification

Two regression tests, both asserting the observable outcome — the victim's bytes actually leave
the local tier — rather than that a particular function was called. Each pins **both**
directions: the victim's payload must be gone AND the renamed file's payload must survive, so a
fix that releases the wrong payload fails instead of passing.

Both were run against a deliberately unfixed build first (the release call stubbed out, tests
kept):

```
--- FAIL: TestRename_ReclaimsClobberedPayloadBytes (0.65s)
    io_test.go:1729: clobbered payload "export/e9d8ed0d-36df-47ec-82df-6d110dd87984"
        still holds 4194304 live bytes in the local tier after RENAME

--- FAIL: TestRename_ReclaimsClobberedPayloadBytes (0.31s)
    rename_test.go:476: clobbered payload "export/fd48cc45-18b3-419a-bfdc-996d57009687"
        still holds 1048576 live bytes in the local tier after RENAME
```

and pass with the fix. `go build ./...`, `go vet ./...`, `go vet -tags integration ./...` and
`go test ./...` are all clean.

Closes #2287

### DIFF
--- a/internal/adapter/common/rename_reclaim.go
+++ b/internal/adapter/common/rename_reclaim.go
@@ -1,0 +1,53 @@
+package common
+
+import (
+	"context"
+
+	"github.com/marmos91/dittofs/pkg/metadata"
+)
+
+// ReleaseClobberedPayload frees the content of the file a rename silently
+// unlinked by renaming over it.
+//
+// POSIX rename onto an existing name removes whatever was there.
+// metadata.Service.Move drops that victim's last link but, like RemoveFile,
+// never its bytes — it returns the victim so the caller can coordinate content
+// deletion. Without this call the victim's records stay indexed as live in the
+// per-share local tier, so no reclamation path (fast-path retire, eviction,
+// repack) can treat them as dead and the bytes survive every restart. Only the
+// local tier depends on it: the remote block GC derives orphans from
+// (synced − live) in the metadata store, which the unlink alone already drives.
+//
+// This is the single seam every protocol adapter drives after a clobbering
+// rename, so one protocol cannot forget it while another gets it right.
+//
+// No-ops when there is nothing to release: no victim (the rename replaced
+// nothing, replaced a directory, or trash recycled the victim), or an empty
+// PayloadID, which is the contract's signal that the content must survive
+// because another hard link still references it.
+//
+// Best-effort by contract — the rename has already committed and the client
+// already has its answer, so callers log a non-nil error rather than failing
+// the operation. `handle` only has to belong to the victim's share; handles are
+// opaque and encode share identity, so the destination directory resolves the
+// same per-share block store the victim's content lives in.
+func ReleaseClobberedPayload(
+	ctx context.Context,
+	reg BlockStoreRegistry,
+	handle metadata.FileHandle,
+	clobbered *metadata.File,
+) error {
+	if clobbered == nil || clobbered.PayloadID == "" {
+		return nil
+	}
+
+	blockStore, err := ResolveForWrite(ctx, reg, handle)
+	if err != nil {
+		return err
+	}
+
+	// Pass nil blocks: the block store resolves this payload's manifest itself
+	// and reaps each row's refcount so the chunks become GC-eligible, in
+	// addition to purging the append log and the tracked size.
+	return blockStore.Delete(ctx, string(clobbered.PayloadID), nil)
+}
--- a/internal/adapter/common/rename_reclaim.go
+++ b/internal/adapter/common/rename_reclaim.go
@@ -7,30 +7,25 @@ import (
 )
 
 // ReleaseClobberedPayload frees the content of the file a rename silently
-// unlinked by renaming over it.
+// unlinked by renaming over it. metadata.Service.Move drops that victim's last
+// link but, like RemoveFile, never its bytes — it returns the victim so the
+// caller can coordinate content deletion. Without this call the victim's
+// records stay indexed as live in the per-share local tier, so no reclamation
+// path (fast-path retire, eviction, repack) can treat them as dead and the
+// bytes survive every restart. The remote tier needs nothing: its GC derives
+// orphans from (synced − live) in the metadata store, which the unlink already
+// drives.
 //
-// POSIX rename onto an existing name removes whatever was there.
-// metadata.Service.Move drops that victim's last link but, like RemoveFile,
-// never its bytes — it returns the victim so the caller can coordinate content
-// deletion. Without this call the victim's records stay indexed as live in the
-// per-share local tier, so no reclamation path (fast-path retire, eviction,
-// repack) can treat them as dead and the bytes survive every restart. Only the
-// local tier depends on it: the remote block GC derives orphans from
-// (synced − live) in the metadata store, which the unlink alone already drives.
-//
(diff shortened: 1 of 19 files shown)


marmos91:
Sibling leak found while investigating this one, filed separately as #2310: the SMB `FileLinkInformation` + `ReplaceIfExists` pre-remove in the same file discards `RemoveFile`'s returned `PayloadID` the same way. Left out of this PR to keep it scoped to rename.

marmos91:
Three review passes plus Copilot. Copilot recommended approval with no comments. Two accuracy findings were real and are fixed in `1fb0cd6` and `301b48c` (both comment-only):

1. The `Move` doc grouped the trash-recycle case with the hard-link case as "non-nil File with an empty PayloadID". Wrong — the trash branch nils `dstFile` before the transaction (`file_modify.go:778`), so a recycled victim comes back **nil**, and a future caller branching on `clobbered.PayloadID == ""` would have missed it entirely. Reworded.
2. `ReleaseClobberedPayload`'s doc called itself "the single seam all protocol adapters drive". Also wrong: SMB goes through `purgeBlockStorePayload`, and NFSv3/v4 REMOVE each inline the same delete. Claim dropped rather than made true.

Two suggestions I did **not** take, and why:

- **Rename the helper to `ReleasePayload` so the REMOVE sites can adopt it later.** Speculative — nothing adopts it today, and a rename is cheap whenever something does.
- **Drop the defensive copy of the victim (`victim := *dstFile`) and return `dstFile` directly.** Correct that it is redundant today, since nothing reads `dstFile` after the return. Kept anyway: handing a caller a pointer into `Move`'s own snapshot is the kind of aliasing that only bites once someone adds a read below it, and the copy costs one allocation on a path that only runs when a rename actually clobbers something.

One residual gap, pre-existing and deliberately not fixed here: `RemoveFile` pops buffered pending-write state under the flush lock when it drops the last link, and `Move` does not. A WRITE racing the release could recreate local-tier records for an nlink=0 inode — but REMOVE has the identical window, so it is not this change's to close.

PR #2318, opened 2026-09-03
fix(smb): hardlink replace never freed the replaced file's bytes

### LINKED ISSUE #2310: SMB hardlink ReplaceIfExists leaks the replaced file's payload
Found while fixing #2287 (rename-over-existing leaked the clobbered file's payload). Same class as #2268 / #2287, different operation, so it is not covered by either fix.

## The gap

SMB `SET_INFO` with `FileLinkInformation` and `ReplaceIfExists=TRUE` deletes the existing destination before creating the hard link, and discards what `RemoveFile` hands back (`internal/adapter/smb/handlers/set_info.go`):

```go
if linkInfo.ReplaceIfExists {
    if existing, matchedName, lookupErr := metaSvc.LookupCaseInsensitive(authCtx, dstDir, linkName); lookupErr == nil && existing != nil {
        if _, _, rmErr := metaSvc.RemoveFile(authCtx, dstDir, matchedName); rmErr != nil {
```

`RemoveFile`'s contract is explicit that it drops the name and the inode but never the bytes — the returned `File` carries the `PayloadID` so the caller can coordinate content deletion, and it is empty exactly when the content must survive (a remaining hard link, or a recycle to trash). Dropping it on the floor leaves the removed file's records indexed as **live** in the per-share local tier, where no reclamation path (fast-path retire, eviction, repack) can treat them as dead, so the bytes survive every restart.

Only the local tier is affected: remote block GC derives orphans from (synced − live) in the metadata store, which the metadata removal alone already drives.

## Fix shape

The same one #2286 used for NFSv4 REMOVE and #2287 used for the rename sites: capture the returned file and hand its `PayloadID` to `h.purgeBlockStorePayload`, guarded on a non-empty `PayloadID`, best-effort — the metadata removal has already committed, so a block-store failure is logged rather than surfaced.

The neighbouring rename-overwrite pre-remove in the same file had exactly this bug and is fixed in #2309; this hardlink site was left out of that PR to keep it scoped to rename.

## Exposure

Only SMB, only `FileLinkInformation` with `ReplaceIfExists=TRUE` onto a name that already exists, and only on a share without trash enabled (a trash-enabled share recycles the victim instead, so nothing is destroyed). Most clients pass `ReplaceIfExists=FALSE`, so the churn is much lower than #2287's — but it is unbounded and does not self-clear on an idle share.

A regression test should assert the observable outcome (the replaced file's bytes actually leave the local tier), not that a particular function was called.

marmos91:
## What was wrong

SMB `SET_INFO` with `FileLinkInformation` and `ReplaceIfExists=TRUE` deletes the existing
destination before creating the hard link. It threw away what `RemoveFile` handed back:

```go
if _, _, rmErr := metaSvc.RemoveFile(authCtx, dstDir, matchedName); rmErr != nil {
```

`RemoveFile`'s contract is explicit that it drops the name and the inode but never the bytes —
the returned `File` carries the `PayloadID` so the caller can coordinate content deletion, and
it is empty exactly when the content must survive (a remaining hard link, or a recycle into the
trash bin). Discarding it meant `blockStore.Delete` was never called, so the replaced file's
records stayed indexed as **live** in the per-share local tier, where no reclamation path —
fast-path retire, eviction, repack — can treat them as dead. The bytes survived every restart.

Only the local tier was affected: remote block GC derives orphans from (synced − live) in the
metadata store, which the metadata removal alone already drives.

Same class as #2268 (fixed in #2286) and #2287 (fixed in #2309) — a caller discarding a
`PayloadID` the metadata layer handed it. Found while fixing #2287; kept out of that PR because
it is a different operation, and the neighbouring rename-overwrite pre-remove in this same file
is already fixed there.

## The fix

Capture the removed file and hand its `PayloadID` to the pre-existing `purgeBlockStorePayload`
helper, which already no-ops on an empty `PayloadID` and on an empty handle. Best-effort by
contract — the metadata removal has already committed, so a block-store failure is logged rather
than surfaced, consistent with every other release path.

One detail worth flagging for review: the purge is passed `authCtx.Context`, **not**
`ctx.Context`. `handleFileLinkInformation` takes a `ctx *SMBHandlerContext` that it dereferences
nowhere else, and the existing tests call `setFileInfoFromStore` with a nil `ctx` — so reaching
for it here would have been a nil-pointer panic on the very path the fix adds.

## The release alone WOULD have caused data loss — and the guard that stops it

Review caught this before merge, and I reproduced it on the real handler: releasing the payload
is unsafe when the destination name is the file being linked. A client that links a file onto
the name it already has, with `ReplaceIfExists=TRUE` and no other hard link, gets:

1. `LookupCaseInsensitive` finds the file's own directory entry.
2. `RemoveFile` drops its **last** link, so it correctly returns a **non-empty** `PayloadID`.
3. The new release destroys those bytes.
4. `CreateHardLink` re-attaches the name to the now-empty inode.

The client sees `STATUS_SUCCESS`, the file is still listed, and its contents are gone.
Reproduced against the handler before fixing:

```
PROBE status=STATUS_SUCCESS
PROBE payload "hl/70bd5423-..." exists after self-link-replace: false
PROBE entry a.txt still resolves: err=<nil>
DATA LOSS CONFIRMED: a.txt still exists but its payload was deleted
```

Note this is a regression the *release* introduces — before it, discarding the `PayloadID` made
the self-link case accidentally harmless. So the fix is not just "call the purge": it is
"call the purge only when the destination is a different inode".

The guard compares the destination's encoded handle to the open file's and short-circuits to
success, which is the requested end state — the same shortcut `Move` already takes for a rename
onto its own name. `TestSetInfo_HardlinkReplace_SelfLinkKeepsContent` pins it, and was verified
to fail (`payload ... was destroyed by linking a.txt onto its own name`) with the guard removed.

## A second, smaller safety fix: the guard used to fail open

The identity check reads the destination's handle, and the first version ignored the error:

```go
existingHandle, encErr := metadata.EncodeFileHandle(existing)
if encErr == nil && bytes.Equal(existingHandle, openFile.MetadataHandle) {
```

An encode failure fell straight through into the destructive remove-and-release. Unreachable
today — `EncodeFileHandle` only errors on an over-long handle, and `openFile.MetadataHandle`
already encoded for the same share — but the direction is wrong on this particular branch: when
identity is unprovable, the path that destroys the caller's content is precisely the one that
must not be taken by default. It now refuses with `STATUS_INVALID_PARAMETER`.

Neither of these two safety changes is implied by #2310's text, which is only about a leak. They
exist because making the leak fix correct required them.

## Why the remaining paths cannot lose data

The guard is the `PayloadID` itself, and the metadata layer is what decides it:

- **Remaining hard link** — `RemoveFile` reads the link count *inside* its transaction and, when
  more links remain, returns `PayloadID: ""`. Nothing is released.
- **Trash-enabled share** — the victim goes through `recycleNode`, which returns a copy with the
  `PayloadID` cleared. Nothing is released.
- **Last link** — the only case with a non-empty `PayloadID`, and the only case where the bytes
  are genuinely unreferenced.

- **Self-link** — guarded above, before `RemoveFile` is ever called.

The regression tests pin the surviving direction explicitly, and the existing
`TestPurgeBlockStorePayload_PreservesHardLinkedContent` already pins that a still-hard-linked
payload is untouched by this helper.

## Verification

New `TestSetInfo_HardlinkReplace_ReclaimsReplacedPayload` drives the real handler through
`setFileInfoFromStore(..., types.FileLinkInformation, ...)` against a real runtime, share and
block store. It asserts observable block-store state rather than that a function was called, and
pins both directions — the replaced payload must be gone AND the linked file's payload must
survive, so releasing the wrong payload fails here instead of passing. The two files are given
different sizes so a mixed-up payload id cannot slip through.

Both tests were run against deliberately broken builds first, each against the specific defect
it guards. With the purge stubbed out:

```
--- FAIL: TestSetInfo_HardlinkReplace_ReclaimsReplacedPayload (0.01s)
    set_info_hardlink_replace_test.go:161: replaced payload "hl/e9efae38-f1ba-4ef4-8c5e-65214101b110"
        still holds 4096 live bytes after the hardlink replace
```

and with the self-link guard removed:

```
--- FAIL: TestSetInfo_HardlinkReplace_SelfLinkKeepsContent (0.01s)
    set_info_hardlink_replace_test.go:261: payload "hl/adf859b2-..." was destroyed by
        linking a.txt onto its own name
```

Both pass with the fix. `go build ./...`, `go vet ./...` and `go test ./...` are clean.

## Not covered here

An independent sweep of the remaining SMB `RemoveFile` call sites found three more that discard
the returned `PayloadID`; they are recorded in a comment on #2310 rather than fixed here, since
none of them overwrite an existing name:

- `close.go` cascade ADS delete on CLOSE
- `durable_scavenger.go` delete-on-close
- `create_post_break.go` create-rollback (usually no bytes written yet)

Closes #2310

### DIFF
--- a/internal/adapter/smb/handlers/ioctl_copychunk_sparse_integration_test.go
+++ b/internal/adapter/smb/handlers/ioctl_copychunk_sparse_integration_test.go
@@ -477,52 +477,8 @@ func TestCopyChunk_SparseDest_SurvivesPriorPayloadReuse(t *testing.T) {
 // PayloadID precisely when content must survive (nlink>1 or recycle-to-trash);
 // the purge must honour that signal rather than the open handle's PayloadID.
 func TestPurgeBlockStorePayload_PreservesHardLinkedContent(t *testing.T) {
-	ctx := context.Background()
-
-	cps, err := cpstore.New(&cpstore.Config{
-		Type:   cpstore.DatabaseTypeSQLite,
-		SQLite: cpstore.SQLiteConfig{Path: ":memory:"},
-	})
-	if err != nil {
-		t.Fatalf("cpstore.New: %v", err)
-	}
-	rt := runtime.New(cps)
-
-	if _, err := cps.CreateMetadataStore(ctx, &models.MetadataStoreConfig{Name: "hlmeta", Type: "memory"}); err != nil {
-		t.Fatalf("CreateMetadataStore: %v", err)
-	}
-	metaStore := metamemory.NewMemoryMetadataStoreWithDefaults()
-	if err := rt.RegisterMetadataStore("hlmeta", metaStore); err != nil {
-		t.Fatalf("RegisterMetadataStore: %v", err)
-	}
-	localBSID, err := cps.CreateBlockStore(ctx, &models.BlockStoreConfig{
-		Name: "hlbs", Kind: models.BlockStoreKindLocal, Type: "memory",
-	})
-	if err != nil {
-		t.Fatalf("CreateBlockStore: %v", err)
-	}
-
-	const shareName = "/hl"
-	if err := rt.AddShare(ctx, &runtime.ShareConfig{
-		Name:              shareName,
-		MetadataStore:     "hlmeta",
-		Enabled:           true,
-		LocalBlockStoreID: localBSID,
-		RootAttr:          &metadata.FileAttr{Type: metadata.FileTypeDirectory, Mode: 0o777},
-	}); err != nil {
-		t.Fatalf("AddShare: %v", err)
-	}
-
-	rootHandle, err := rt.GetRootHandle(shareName)
-	if err != nil {
-		t.Fatalf("GetRootHandle: %v", err)
-	}
-
-	uid, gid := uint32(0), uint32(0)
-	authCtx := &metadata.AuthContext{
-		Context:  ctx,
-		Identity: &metadata.Identity{UID: &uid, GID: &gid},
-	}
+	rt, rootHandle, authCtx := newHardlinkTestShare(t)
+	ctx := authCtx.Context
 	metaSvc := rt.GetMetadataService()
 
 	// Create "a", write a non-zero pattern, then hard-link it as "b".
@@ -573,11 +529,7 @@ func TestPurgeBlockStorePayload_PreservesHardLinkedContent(t *testing.T) {
 	if removed.PayloadID != "" {
 		t.Fatalf("RemoveFile of a hard-linked file should return empty PayloadID, got %q", removed.PayloadID)
 	}
-	var removedPayloadID metadata.PayloadID
-	if removed != nil {
-		removedPayloadID = removed.PayloadID
-	}
-	h.purgeBlockStorePayload(ctx, handle, removedPayloadID, "a", "TEST")
+	h.purgeBlockStorePayload(ctx, handle, removed.PayloadID, "a", "TEST")
 
 	// The surviving "b" link must still read the full pattern.
 	dst := make([]byte, 4096)
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -2571,7 +2571,7 @@ func (h *Handler) handleFileLinkInformation(
 			// Without it the replaced file's records stay indexed as live in
 			// the local tier, where no reclamation path can reach them.
 			if removed != nil {
-				h.purgeBlockStorePayload(ctx.Context, dstDir, removed.PayloadID, matchedName, "SET_INFO hardlink replace")
+				h.purgeBlockStorePayload(authCtx.Context, dstDir, removed.PayloadID, matchedName, "SET_INFO hardlink replace")
 			}
 		}
 	}
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -2559,6 +2559,17 @@ func (h *Handler) handleFileLinkInformation(
 	metaSvc := h.Registry.GetMetadataService()
 	if linkInfo.ReplaceIfExists {
 		if existing, matchedName, lookupErr := metaSvc.LookupCaseInsensitive(authCtx, dstDir, linkName); lookupErr == nil && existing != nil {
+			// A destination that already names the file being linked is the
+			// requested end state, so there is nothing to do. Removing it
+			// would drop the inode's last link and free its content, and the
+			// CreateHardLink below would then resurrect the name over bytes
+			// that no longer exist. Move takes the same shortcut for a rename
+			// onto its own name.
+			existingHandle, encErr := metadata.EncodeFileHandle(existing)
+			if encErr == nil && bytes.Equal(existingHandle, openFile.MetadataHandle) {
+				return setInfoStatus(types.StatusSuccess), nil
+			}
+
 			removed, _, rmErr := metaSvc.RemoveFile(authCtx, dstDir, matchedName)
 			if rmErr != nil {
 				logger.Debug("SET_INFO: hardlink replace failed to remove existing",
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -2559,10 +2559,19 @@ func (h *Handler) handleFileLinkInformation(
 	metaSvc := h.Registry.GetMetadataService()
 	if linkInfo.ReplaceIfExists {
 		if existing, matchedName, lookupErr := metaSvc.LookupCaseInsensitive(authCtx, dstDir, linkName); lookupErr == nil && existing != nil {
-			if _, _, rmErr := metaSvc.RemoveFile(authCtx, dstDir, matchedName); rmErr != nil {
+			removed, _, rmErr := metaSvc.RemoveFile(authCtx, dstDir, matchedName)
+			if rmErr != nil {
 				logger.Debug("SET_INFO: hardlink replace failed to remove existing",
 					"name", matchedName, "error", rmErr)
 				return setInfoStatus(common.MapToSMB(rmErr)), nil
+			}
+			// RemoveFile drops the name and the inode but never the bytes; the
+			// returned PayloadID is empty exactly when the content must
+			// survive, so releasing it here is hard-link- and trash-safe.
+			// Without it the replaced file's records stay indexed as live in
+			// the local tier, where no reclamation path can reach them.
+			if removed != nil {
+				h.purgeBlockStorePayload(ctx.Context, dstDir, removed.PayloadID, matchedName, "SET_INFO hardlink replace")
 			}
 		}
 	}
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -2566,7 +2566,15 @@ func (h *Handler) handleFileLinkInformation(
 			// that no longer exist. Move takes the same shortcut for a rename
 			// onto its own name.
 			existingHandle, encErr := metadata.EncodeFileHandle(existing)
-			if encErr == nil && bytes.Equal(existingHandle, openFile.MetadataHandle) {
+			if encErr != nil {
+				// Identity is unprovable, so the removal below cannot be shown
+				// to be safe. Refuse rather than fall through into it: the
+				// wrong branch here destroys the caller's content.
+				logger.Debug("SET_INFO: hardlink replace cannot identify existing destination",
+					"name", matchedName, "error", encErr)
+				return setInfoStatus(types.StatusInvalidParameter), nil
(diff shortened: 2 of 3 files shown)


marmos91:
Both Copilot comments addressed.

**GetSize error ignored** — fair, fixed. The failure branch now reports the error instead of printing a zero that looks like a size.

**Use `ctx.Context` when available, fall back to `authCtx.Context`** — declining, because the two are the same object. `BuildAuthContext` (`internal/adapter/smb/handlers/auth_helper.go:74`) does:

```go
Context: ctx.Context,
```

and `setFileInfoFromStore` is only ever reached from `handleSetInfo`, which builds `authCtx` from that same `ctx` immediately before the call. So on every production path `authCtx.Context == ctx.Context`, and the proposed nil-check would branch between two identical values while reintroducing the deref the fallback exists to avoid. `authCtx.Context` is strictly the safer of two equal choices.

Worth noting a second reason not to reach for the request context here even if it did differ: this purge is best-effort cleanup that runs *after* the metadata removal has already committed. Propagating a client cancellation into it would abandon the block-store delete and leak exactly the bytes this PR is about.

For consistency: the three sibling SMB purge sites added in #2309 do use `ctx.Context`. That is equivalent in production for the same reason, but it is the more fragile spelling — those call sites are the only `ctx.` dereferences in their enclosing functions, so a future unit test passing a nil `ctx` would panic there and not here. Not changing them in this PR; flagging it so the asymmetry is a known one rather than an oversight.

marmos91:
Review round 2 found a **data-loss regression** in the first version of this fix, and it is now fixed. Recording the verification, since this is the part of the PR that matters most.

**What was wrong.** Releasing the payload is unsafe when the destination name is the file being linked. `RemoveFile` drops the inode's last link and therefore correctly returns a *non-empty* `PayloadID`; the release then destroys the bytes, and `CreateHardLink` resurrects the name over nothing. Reproduced on the real handler before fixing:

```
PROBE status=STATUS_SUCCESS
PROBE payload "hl/70bd5423-..." exists after self-link-replace: false
PROBE entry a.txt still resolves: err=<nil>
DATA LOSS CONFIRMED: a.txt still exists but its payload was deleted
```

This was a regression the release *introduced* — discarding the `PayloadID` had made the self-link case accidentally harmless.

**The guard, and why it holds.** Skip the whole remove-and-release when the destination already names the open file. The identity test is `bytes.Equal` on encoded handles, which is sound because encoding is total and deterministic (`pkg/metadata/types.go:57-68`):

```go
func EncodeFileHandle(file *File) (FileHandle, error) { return EncodeShareHandle(file.ShareName, file.ID) }
func EncodeShareHandle(shareName string, id uuid.UUID) (FileHandle, error) { encoded := shareName + ":" + id.String() ... }
```

No generation counter, no field-ordering ambiguity, no mutable state — just (ShareName, ID). Two encodings of one inode are byte-identical regardless of which call site produced them, so the guard cannot silently miss. `bytes.Equal` on handles is also the established identity test in this file (`set_info.go:1110`).

Two consequences worth stating explicitly:

- **Case-insensitive spelling is covered.** `LookupCaseInsensitive` returns the matched entry's `*File`, whose ID is the inode's regardless of the spelling the client sent, so "A.TXT" onto on-disk "a.txt" hits the guard too.
- **A different name for the same inode is also covered, and correctly.** If the client links onto a second existing hard link of the same file, the requested end state already holds, so the no-op is right. The unguarded path would also have been safe there (`RemoveFile` returns an empty `PayloadID` while nlink > 1), but the no-op avoids the pointless nlink churn.

The guard skips nothing it should handle: a destination that is a *different* inode still gets removed and released, which is the whole point of the PR.

`TestSetInfo_HardlinkReplace_SelfLinkKeepsContent` pins it, and fails with the guard removed:

```
--- FAIL: TestSetInfo_HardlinkReplace_SelfLinkKeepsContent
    payload "hl/adf859b2-..." was destroyed by linking a.txt onto its own name
```

marmos91:
Third review pass (focused specifically on the hazard class the self-link defect exposed: a non-empty `PayloadID` whose bytes must nonetheless survive). Six candidate paths enumerated, **all safe in this diff**:

| Path | Verdict | What decides it |
| --- | --- | --- |
| Remaining hard link | SAFE | `RemoveFile` reads nlink inside the tx and clears `PayloadID` when links remain |
| Recycle to trash | SAFE | `recycleNode` unconditionally returns a copy with `PayloadID` cleared; failures propagate before the purge |
| Case-insensitive spelling | SAFE | `LookupCaseInsensitive` returns the matched inode's own `*File`, so "A.TXT" vs "a.txt" yields byte-equal handles and the guard fires |
| Different name, same inode | SAFE | Guard fires; skipping is correct (the requested end state already holds) and the unguarded path was safe anyway (nlink > 1 ⇒ empty PayloadID) |
| Dedup / CAS | SAFE | Refcount reap is keyed by `{payloadID}/{offset}`; a `PayloadID` is unique per inode, so sibling rows in other files keep the hash live |
| Alternate data streams | SAFE | Streams are sibling inodes with their own `PayloadID`; replacing a base releases only the base's bytes |

One finding, **pre-existing and not introduced here**, filed separately as #2320: on postgres, `RemoveFile`'s in-transaction nlink read is not actually protected. `pool.Begin` passes no `TxOptions` (so READ COMMITTED), `GetLinkCount` has no `FOR UPDATE`, and `SetLinkCount` writes an absolute value — a textbook lost update. A concurrent `CreateHardLink` on the victim inode can make `RemoveFile` return a non-empty `PayloadID` for content that is still linked, and no 40001/40P01 fires so the retry loop never engages. I verified those three structural preconditions directly; I did **not** reproduce the race, and #2320 says so.

It predates this PR and exposes NFSv3/v4 REMOVE, SMB delete-on-close and the #2309 rename sites identically — every caller that treats a non-empty `PayloadID` as permission to delete. The fix belongs in `pkg/metadata/store/postgres/dialect.go`, not in this handler.

Copilot on the final SHA `f76903b1a`: approval recommended, 0 new comments.

marmos91:
Second review pass on the final revision: **no issues in the hardlink fix**. All seven questions came back clean — guard placement (runs before any removal), `EncodeFileHandle` determinism, alias/case-insensitive firing, the recycle-to-trash path, both regression tests' ability to fail, the extracted fixture not weakening `TestPurgeBlockStorePayload_PreservesHardLinkedContent`, and no stray edits in `set_info.go` outside `handleFileLinkInformation`.

It also independently confirmed the `authCtx.Context` choice is **required rather than stylistic**, noting something I had missed: every other `ctx` dereference in `handleFileLinkInformation` is already nil-guarded (`set_info.go:2592`). So the function is written throughout on the assumption that `ctx` may be nil, and the purge call now matches that.

One process note, since it produced two dramatic-looking false alarms worth not repeating: I handed the reviewer a diff generated with two-dot `git diff origin/develop`. Develop had moved ahead in the meantime, so #2302 and #2308 — both already merged — showed up **reversed**, and the reviewer correctly read those reversals as regressions (a retention pin that stops holding, and unstamped segments sorting ahead of everything). They are not in this branch. Three-dot `git diff origin/develop...HEAD` gives this PR's real change:

```
 internal/adapter/smb/handlers/ioctl_copychunk_sparse_integration_test.go |  54 +----
 internal/adapter/smb/handlers/set_info.go                                |  21 +-
 internal/adapter/smb/handlers/set_info_hardlink_replace_test.go          | 263 +++++
 3 files changed, 286 insertions(+), 52 deletions(-)
```

Three files, all SMB handlers. The reviewer's clean bill on part 1 is the part that applies here.

marmos91:
Simplifier pass on the final revision: test-only, and it caught one thing worth fixing in the handler.

**Refactor** (`687773397`) — two helpers, `createHardlinkTestFile` and `openHardlinkTestFile`, absorb the setup both tests duplicated, so each body is now only the thing it pins. `set_info.go` and the copychunk test were left byte-identical.

**Fail-open on a data-loss path** (`1f002632`) — the guard read:

```go
existingHandle, encErr := metadata.EncodeFileHandle(existing)
if encErr == nil && bytes.Equal(existingHandle, openFile.MetadataHandle) {
```

An encode failure fell straight through into the destructive remove-and-release. It is unreachable today (`EncodeFileHandle` only errors on an over-long handle, and `openFile.MetadataHandle` already encoded for the same share), but the direction is wrong: when identity is unprovable, the branch that destroys the caller's content is the one you must not take. It now refuses with `STATUS_INVALID_PARAMETER` instead.

**Re-verified after both changes**, since a refactor can silently defang a test. Each mutation fails exactly its own test and leaves the other passing:

| Mutation | ReclaimsReplacedPayload | SelfLinkKeepsContent |
| --- | --- | --- |
| purge stubbed out | **FAIL** | PASS |
| self-link guard removed | PASS | **FAIL** |

That clean split is the useful part — it shows the two tests are pinning different things rather than both riding on one code path.

Issue #1603, opened 2026-07-08
Evaluate xfstests alongside pjdfstest for broader FS-correctness coverage


marmos91:
## Idea

A colleague ran [xfstests](https://github.com/kdave/xfstests) against DittoFS. It's worth evaluating as a standing complement to our existing POSIX-compliance testing in `test/posix/` (pjdfstest).

## Why it may be useful

- **pjdfstest** covers POSIX *syscall semantics* (chmod/chown/link/rename/… correctness).
- **xfstests** `generic/` adds broader filesystem correctness, plus crash-consistency and stress tooling (`fsstress`, `fsx`) that pjdfstest doesn't exercise.

They're complementary, not overlapping — xfstests would widen correctness coverage where pjdfstest is narrow.

## Scope

- **Out of scope** for the performance harness (that's fio-only; see the dfsbench consolidation issue).
- This is a **correctness** axis: it belongs next to `test/posix/` with its own `KNOWN_FAILURES`, setup/teardown, and CI wiring (mirroring `posix-tests.yml`).

## Task

- [ ] Evaluate which xfstests `generic/` groups are applicable over NFS/SMB mounts of DittoFS.
- [ ] Stand up a `test/xfstests/` harness (config + KNOWN_FAILURES) parallel to `test/posix/`.
- [ ] Decide whether to gate in CI or run on-demand initially.

PR #2321, opened 2026-09-03
test(nfs): run pynfs as an NFSv4 protocol conformance suite in CI

### LINKED ISSUE #2204: nfs: run pynfs as a protocol conformance suite and triage the failures
@chessman ran [kofemann/pynfs](https://github.com/kofemann/pynfs) against DittoFS
([#2146](https://github.com/marmos91/dittofs/pull/2146#issuecomment-5431263983)):

```
NFSv4.0: 354 passed, 97 skipped, 143 fail
NFSv4.1: 105 passed,   1 skipped,  92 fail
```

pynfs is the reference NFSv4 protocol suite and we have never run it.

We are not starting from nothing on the NFS side: `test/posix/` runs pjdfstest over both
v3 and v4 with `KNOWN_FAILURES.md` / `KNOWN_FAILURES_V4.md`, and `test/e2e/` mounts with a
real kernel client. What neither covers is the v4 protocol itself, which is where state,
sessions, owners, locking and wire error codes live. #2146 is a good example of that class:
a real interop bug that every POSIX test passes straight over.

The raw counts are not yet actionable. A failure can mean a real protocol bug, an operation
we deliberately do not implement, or a suite assertion no server satisfies. Separating those
is the work.

Plan, following the shape `test/posix/` and `test/smb-conformance/` already share (same table
format, same `test/common/known-failures.sh` parser):

1. Reproduce the run locally and capture per test verdicts, not just totals.
2. Baseline the same suite against Linux nfsd (and/or nfs-ganesha) in the same harness, so
   "everyone fails this" is measured rather than assumed.
3. Land the runner under `test/nfs-conformance/pynfs/` with a `KNOWN_FAILURES.md` carrying a
   reason per row, grading a run as "no new failures" the way the other two suites do.
4. File the real gaps triage surfaces as their own issues.

Step 2 is what makes the rest honest. Without it the known failures list becomes a place to
park bugs.

marmos91:
Adds the reference NFSv4 conformance suite as a gating CI check, in the same
shape as the POSIX and SMB suites.

## Why

`test/posix/` (pjdfstest) and `test/e2e/` prove filesystem *semantics* through a
kernel NFS client, and a kernel client only ever sends the subset of the protocol
it happens to need. Nothing covered the NFSv4 protocol itself — state, sessions,
owners, locking, wire error codes — which is where interop bugs live. #2146 is a
good example of that class: a real bug every POSIX test passes straight over.

## What is here

- `nix build .#pynfs` — pynfs packaged in the flake, `platforms.unix`, so the
  same runner works on a macOS laptop and on CI. Python 3.12 is pinned
  deliberately: pynfs unpacks XDR with the stdlib `xdrlib`, removed in 3.13.
  `stdenvNoCC`, because a pure-Python suite should not drag in a C toolchain.
- `test/nfs-conformance/pynfs/run-pynfs.sh` — provisions a server, runs the
  suite, grades it. No mount and no root: pynfs is its own NFSv4 client.
- `parse-results.sh` + `parse-results_test.sh` — grading against
  `KNOWN_FAILURES_V40.md` / `KNOWN_FAILURES_V41.md` through the shared
  `test/common/known-failures.sh` parser. Red only on a NEW failure.
- `baseline-knfsd.sh` + `baseline-knfsd.md` — the same suite against the Linux
  kernel server.
- `.github/workflows/nfs-pynfs.yml` — memory + postgres-s3 x {4.0, 4.1} on PRs,
  all four backends on develop and nightly, plus a nightly knfsd baseline.

The server comes from `test/posix/setup-posix.sh --no-mount`, which already owns
the backend profiles, rather than a second copy that would drift. That flag also
skips the root check, since nothing needs privilege without a mount.

## Measured

| Suite | DittoFS | Linux knfsd |
|---|---|---|
| NFSv4.0 | 433 passed, 150 failed, 13 skipped | **589 passed, 4 failed** |
| NFSv4.1 | 109 passed, 74 failed, 1 skipped | **182 passed, 2 failed** |

Step 2 of #2204 — baselining — is what makes the blacklist honest, and it is
unforgiving here: knfsd passes almost everything we fail. **Only 3 of our 226
failures have any suite-defect defence** (`RPLY8`, `DELEG24`, `DELEG25`).
Everything else is ours, so the tables are 219 `bug` rows and 4 `feature` rows,
grouped into 18 tracking issues: #2325-#2342.

That is a large blacklist, and deliberately not a comfortable one. What it buys
is a regression gate: the protocol cannot get quietly worse while the backlog is
worked down, and every row names the issue that should delete it.

Rows are the **union across backends**, because the two do not fail the same
set: postgres-s3 additionally fails `SATT12x`, and the courtesy-client test
alternates between `COUR3` and `COUR5` (#2327).

## Verification

- All four CI cells grade green against the committed tables.
- Negative check: deleting one row turns the run red naming exactly that test.
  A guard that has never refused anything is unverified.
- `parse-results_test.sh` runs as its own CI job (~1s), so a grading regression
  is caught without waiting for the 20-minute suite.
- `posix-tests` is green on this branch, so `--no-mount` did not regress it.

## Two things I got wrong on the way, since they shaped the design

- I first ran knfsd in Docker and read a truncated log tail plus a successful
  port check as "it works". `rpc.nfsd` was failing with errno 111 the whole
  time — the kernel will not take sockets from a container netns, and Docker
  binds the host port regardless of whether anything listens. The baseline now
  runs knfsd directly on Linux; the container image is gone.
- pynfs `-v` reprints every test in the *same* `: FAILURE` format as it runs, so
  a whole-file scan counts each failure three times. The grader reads only the
  final results block, and `parse-results_test.sh` pins that.

## Follow-ups filed

#2322 (unify the five conformance suites behind one runner and a manifest,
folding in #1603 xfstests) and #2323 (flake ergonomics).

Closes #2204


marmos91:
All three review findings were real and are fixed in 2fe5e6da6.

- **Unchecked `mktemp`** — the sharpest one. With `set -uo pipefail` and no `-e`, a failed `mktemp` left `XDG_CONFIG_HOME` empty, which Go's `os.UserConfigDir()` resolves straight back to the real config directory. The guard would have silently become the bug it exists to prevent. Now fails fast, and `cleanup` no longer runs `rm -rf ""`.
- **Globbing in `--tests`** — word splitting was intended, pathname expansion was not; a selector like `LOCK*` would have expanded against the working directory. Now `read -ra`.
- **Root-owned `/tmp` state** — genuinely introduced by `--no-mount`, since that flag is what makes an unprivileged run possible in the first place. The script now checks the three paths it must write and points at `sudo teardown-posix.sh`, instead of failing later inside `rm`.

Verified after the change: a `--tests "SEQ1 SEQ2 CSESS1"` run provisions, runs and grades clean, and `~/.config/dfsctl/config.json` is untouched.

Issue #2324, opened 2026-09-03
Metadata writes read the inode outside the transaction and write the whole row back inside it


marmos91:
## What

Several metadata writes read an inode outside a transaction, mutate the Go struct, then write the whole row back inside one. Anything another writer committed in between is overwritten by the stale snapshot — not just the field the caller meant to change, but every field on the row.

Two confirmed members:

- **`Service.Move`** — `srcFile` is read at `file_modify.go:691` via `store.GetFile`, outside the transaction; the transaction opens at `file_modify.go:834` and writes that same struct with `tx.UpdateAttrs(ctx.Context, srcFile)`.
- **`Service.SetFileAttributes`** non-size path — `file` is read at `file_modify.go:224`, and the write is `tx.UpdateAttrs(ctx.Context, file)` inside `withRelaxedTransaction`.

## Why it is not closed by per-backend isolation

badger (SSI + retry), sqlite (single connection) and memory (global mutex) make a compare/write pair atomic when both halves are inside the transaction. That does not help here, because the *read* is outside it — the stale row is already in hand before the transaction opens.

postgres is weaker still: `postgres/transaction.go:87` calls `pool.Begin` without setting an isolation level, so it runs READ COMMITTED, and `dialect.go:65-66` issues a plain `SELECT` with no `FOR UPDATE`. A read taken *inside* a postgres transaction is therefore not protected against a concurrent commit either.

## Reachable interleaving

Demonstrated on sqlite by widening Move's read-to-transaction window with a sleep and landing a `SetFileAttributes` inside it. With the pre-rename ChangeTime taken from the outside-transaction read, the stored value moved backwards:

```
advance=2026-09-03 21:51:57.579271 +0000 UTC   final=2026-09-03 21:50:27.579271 +0000 UTC
```

The window is real without the sleep; it is just narrow.

## Scope

Reading the source inode inside Move's transaction closed this for the rename ChangeTime preserve specifically. The general case is untouched: Move's `tx.UpdateAttrs(srcFile)` still writes a row snapshotted before the transaction, so a concurrent WRITE's `Size` and `Mtime` can still be lost across a rename, and `SetFileAttributes` has the same shape.

## Suggested direction

Re-read the inode inside the transaction and apply the intended field changes to *that* row, rather than writing a struct captured earlier — the same shape the ChangeTime work used. Doing it generally means auditing each caller for which fields it actually intends to change, which is why it was kept out of that PR.

[...]

PR #2343, opened 2026-09-03
fix(smb): make the rename ChangeTime preserve conditional

### LINKED ISSUE #2285: SMB rename's ChangeTime preserve is file-scoped and can walk a concurrent update backwards
## What

`snapshotChangeTime` (`internal/adapter/smb/handlers/set_info.go`) reads a
file's Ctime before `metadata.Service.Move` and writes it back after, so a
client that renames through a handle it still holds open keeps observing the
ChangeTime that handle was handed at CREATE. It is marked `ponytail:` in place.

The rule being implemented is **handle-scoped** — what an already-open handle
continues to observe. The implementation is **file-scoped**: it rewrites the
stored timestamp. The two coincide only while nothing else is touching the
file.

## The race

Between the read and the write-back, another operation can advance the same
file's Ctime. The write-back then walks it **backwards for every observer**,
not just for the renaming handle.

Reachable, not theoretical:

- `RequestOrder` (`internal/adapter/smb/request_order.go`) orders only the
  moment a response is written — "Handlers keep running concurrently". A
  pipelining client can have a WRITE in flight on the same file while its
  SET_INFO rename runs.
- Any second handle, session or connection on the same file.
- NFS on the same file: the metadata store is shared.
- WRITE genuinely advances Ctime (`pkg/metadata/io.go`), as do truncate,
  punch-hole and explicit SET_INFO.

The window is `Move`'s transaction plus one `SetFileAttributes`.

## Why it shipped this way

The same read-then-write-back was the behaviour before
`fix(smb): let rename's ChangeTime advance stand` (#2279) removed it, but it is
**not simply narrower** — it is narrower in fields and **wider in callers**, and
the second half matters more:

- **Narrower in fields.** `saveTimestamps` restored Mtime *and* Ctime, so it
  could clobber two. The current form restores Ctime only.
- **Wider in callers.** `saveTimestamps` wrote back **as the caller**, through
  the ownership-gated explicit-timestamp path. On the non-owner path that write
  was *refused*, so nothing was clobbered — the old race needed an owning
  caller. The current form writes as `NewSystemAuthContext` so that it lands for
  every caller, which is exactly what closes the divergence the original defect
  was about, but it also opens the clobber window on paths where a permission
  refusal previously closed it.

So the honest summary is: a pre-existing hazard reshaped, with one axis reduced
and one enlarged — not a strict reduction. Deferring it was still right, since
develop was red on all five metadata backends without the fix, but it should not
be triaged as "the old bug, smaller".

## Upgrade path

A per-`OpenFile` overlay consulted by QUERY_INFO, reusing the seam
`applyFrozenTimestamps` already occupies (`query_info.go`, `close.go` — two
call sites). The blocker is invalidation, not the read seam: the overlay has to
stop applying at the next genuine Ctime advance, **including one made through a
different handle**, which an `OpenFile` field cannot observe on its own. Left
applying too long it silently reports a stale ChangeTime for the life of the
handle, which is worse than the race it replaces. A generation/version stamp on
the stored attrs would give the overlay something to compare against.

Note an overlay does not cover every path either: a directory enumeration
(`query_directory.go`) reports a child's ChangeTime with no `OpenFile` in hand.

## Acceptance

- Concurrent WRITE-vs-rename on one file no longer moves the stored ChangeTime
  backwards, with a test that fails without the fix.
- `smb2.rename` stays 13/13.

marmos91:
## What was wrong

The SMB rename ChangeTime preserve read a file's ChangeTime before
`Service.Move` and wrote it back afterwards, unconditionally. The rule it
implements is handle-scoped — what a handle that is already open keeps
observing — but the implementation rewrote the *stored* timestamp. Anything
that advanced the same file's ChangeTime between the read and the write-back —
a WRITE, a truncate, a SET_INFO on another handle, or NFS on the same file —
was walked backwards for every observer.

Reproduced on unmodified `develop` (`0973be03d`) before writing anything:

```
T0=21:33:50.450252   concurrent advance=21:35:20.450252   final=21:33:50.450252
BACKWARDS WALK ON UNMODIFIED DEVELOP: stored ChangeTime 21:35:20 -> 21:33:50
```

## A root cause I refuted by running it

`Move` reads `srcFile` outside its transaction and writes that whole row back
inside it, which reads like a second, independent backwards-walker. It is not:
`Move` stamps `now`, which is always *later*, so it cannot walk ChangeTime
backwards. Only the write-back writes an older value. Trusting the code read
would have widened this PR into `Move` for nothing.

(The stale-row write is a real hazard for *other* fields — it can lose a
concurrent writer's `Size` and `Mtime`. That is a pre-existing family, filed
separately as #2324, deliberately not fixed here.)

The `Move`-side reasoning that *did* hold — that nothing assigns to `srcFile`
between its read and the transaction — is true but proves less than it looks:
the two reads are defined to differ exactly when a writer commits in that gap,
so their single-threaded coincidence says nothing about whether the defect is
closed. What closes it is the store's isolation, and the test below constructs
the concurrent writer rather than arguing it away.

## The fix

Put the value back only while the stored ChangeTime is still the one the rename
stamped. `Move` reports both timestamps on `RenameWcc`, each read back **through
the transaction that performed the rename**, and the comparison runs inside the
transaction that performs the restore.

`snapshotChangeTime` is deleted outright rather than narrowed. A caller-side
read that is usually fine is worse than no caller-side read at all.

Reading both values inside the transaction is what makes this work, and the
first two attempts at this fix were wrong in ways that only showed up when
attacked:

**The "before" must come from inside the transaction**, or it is stale by the
width of `Move`, and an advance committed in that window gets erased rather than
kept — the original defect, moved rather than fixed. Pinned by a regression test
that commits a write inside exactly that window (see Tests below).

**The "after" must come from inside the transaction too**, because the SQL
backends store timestamps as Windows FILETIME ticks and truncate on write
(`sqlcodec.go:60`, `int64(t.Nanosecond())/100`). Comparing against an in-memory
`time.Time` never matches, so the restore silently never fires — green
everywhere, and broken on sqlite and postgres in production.

## Why that second one hid, measured rather than argued

It is invisible on the machine most of this was written on:

```
darwin:  ns%100 != 0 in    0/1000 samples     <- microsecond clock, always round-trips intact
linux:   ns%100 != 0 in 1000/1000 samples     <- full nanoseconds, never round-trips intact
```

On darwin an unstored `time.Time` compares equal to its own truncated round
trip, so the no-op cannot be observed. On Linux `Equal` is false essentially
always. Presubmit smbtorture runs memory + badger-fs, both of which keep
nanosecond fidelity via `time.MarshalBinary`; sqlite/postgres smbtorture is
nightly, and the postgres presubmit job is WPTS BVT, which does not include
`smb2.rename`. A green run here would have said nothing about the backends that
break.

## Tests

Four, on a real sqlite store — the backend that truncates. Every row below is
**measured on both platforms**, not reasoned.

| Mutation to the fix | darwin | linux |
| --- | --- | --- |
| `SourceCtime` = in-memory `now` (defect B) | *none* | 3 tests |
| `SourcePreCtime` from the outside-tx read (defect A) | **1 test** | **1 test** |
| `SourcePreCtime` zero | 3 tests | 3 tests |
| `SourceCtime` zero | 3 tests | 3 tests |
| the two fields swapped | 3 tests | 3 tests |
| *(real tree)* | none | none |

The test that catches defect A hooks the store's `WithTransaction` from the test
side — no production code is touched — to commit a write in the gap between
`Move`'s read of the source inode and the transaction that rewrites it. That gap
is the only place the two reads differ, so a test that merely advances the
ChangeTime *before* calling `Move` pins nothing; an earlier version of this PR
had exactly that test, and it passed against the mutation it claimed to catch.

The hook is sqlite-bound on purpose: sqlite does not implement
`RelaxedTransactor`, so `Move`'s `withRelaxedTransaction` falls through to the
overridden method, and its single connection makes the ordering exact. badger
would take the relaxed path and re-run the closure on an SSI retry.

**Defect B is invisible on darwin.** No test in this PR fails the B mutation
there, because the darwin clock cannot produce a timestamp that fails to round
trip. CI runs Linux, where three tests catch it. `ComparesValuesTheStoreHolds`
is platform-independent, but it pins `RestoreChangeTimeIfUnchanged`'s `Equal`
semantics and never calls `Move` — it is not coverage for B.

## Scope and what this does not claim

This closes the window on badger, sqlite and memory and **narrows** it on
postgres. What closes it is the store's isolation, not the shape of `Move`:
badger retries under SSI, sqlite serialises on one connection, memory holds its
mutex across the closure. postgres begins without setting an isolation level
(`transaction.go:115`), so READ COMMITTED, and its pre-read is a plain `SELECT`
with no `FOR UPDATE` (`sql/files.go:66`) — a writer can still commit between
that read and the row lock the update takes.

**That is not a regression.** The old unconditional restore lost the same race
and produced the same result on that interleaving; this narrows the window from
[caller's snapshot .. Move commit] to [Move's pre-read .. Move's row lock].
Closing the postgres half needs a narrow conditional
`UPDATE … WHERE ctime = $expected`, which means a new store-interface method
across all backends plus the conformance suite — out of scope here, tracked in
#2324 along with the rest of the family.

Both in-transaction reads now return their error rather than discarding it. A
failed "before" with a successful "after" would leave `SourcePreCtime` zero and
still match, and the restore would write a zero ChangeTime; returning the error
removes the only path by which either field can be zero.

That error path is deliberately left untested: reverting either read to
`pre, _ := tx.GetFile(...)` would still pass all four tests, since nothing
forces those reads to fail. Pinning it needs a fault-injecting `Transaction`
wrapper stacked on the store wrapper, whose own correctness then becomes a
question, and the thing it would pin cannot produce an observable wrong result —
the error aborts the transaction, `Move` returns a nil `RenameWcc`, and both
restore call sites sit behind `Move`'s error guard, so no half-populated struct
can escape. The containment is structural, not probabilistic.

The preserve stays file-scoped rather than handle-scoped. The `ponytail:` marker
is kept and rewritten to say what remains and to name the per-`OpenFile` overlay
as the upgrade path.

## Verification

- `go build ./...`, `go vet ./...`, `go test ./...` green on darwin **and** in a
  Linux container — the second is not optional here, since one whole defect
  class is invisible on darwin.
- `smb2.rename` **13/13**, 0 failures, no cell moved.
- Every test run against a build with the fix reverted, per defect.

Closes #2285

### DIFF
--- a/internal/adapter/smb/handlers/set_info.go
+++ b/internal/adapter/smb/handlers/set_info.go
@@ -778,10 +778,10 @@ func (h *Handler) setFileInfoFromStore(
 			// holding this handle open must keep observing the ChangeTime it
 			// was handed at CREATE, so put the pre-rename value back.
 			metaSvc := h.Registry.GetMetadataService()
-			restoreChangeTime := h.snapshotChangeTime(authCtx, openFile.MetadataHandle)
 
 			var clobberedStream *metadata.File
-			clobberedStream, _, err = metaSvc.Move(authCtx, toDir, oldFileName, toDir, toName)
+			var renameWcc *metadata.RenameWcc
+			clobberedStream, renameWcc, err = metaSvc.Move(authCtx, toDir, oldFileName, toDir, toName)
 			if err != nil {
 				logger.Debug("SET_INFO: stream rename failed",
 					"from", oldFileName,
@@ -790,7 +790,7 @@ func (h *Handler) setFileInfoFromStore(
 				return setInfoStatus(common.MapToSMB(err)), nil
 			}
 
-			restoreChangeTime()
+			h.restorePreRenameChangeTime(authCtx.Context, openFile.MetadataHandle, renameWcc)
 
 			// Renaming a stream onto an existing stream name unlinks the
 			// stream that was there; free its content.
@@ -1219,10 +1219,9 @@ func (h *Handler) setFileInfoFromStore(
 		// Move stamps the renamed inode's LastChangeTime. A client holding
 		// this handle open must keep observing the ChangeTime it was handed at
 		// CREATE, so put the pre-rename value back.
-		restoreChangeTime := h.snapshotChangeTime(authCtx, openFile.MetadataHandle)
-
 		var clobbered *metadata.File
-		clobbered, _, err = metaSvc.Move(authCtx, srcParentHandle, oldFileName, toDir, toName)
+		var renameWcc *metadata.RenameWcc
+		clobbered, renameWcc, err = metaSvc.Move(authCtx, srcParentHandle, oldFileName, toDir, toName)
 		if err != nil {
 			logger.Debug("SET_INFO: rename failed",
 				"from", openFile.Name().Path,
@@ -1231,7 +1230,7 @@ func (h *Handler) setFileInfoFromStore(
 			return setInfoStatus(common.MapToSMB(err)), nil
 		}
 
-		restoreChangeTime()
+		h.restorePreRenameChangeTime(authCtx.Context, openFile.MetadataHandle, renameWcc)
 
 		// The pre-remove above only fires for a case-mismatched destination, so
 		// an exact-case ReplaceIfExists overwrite reaches Move's own clobber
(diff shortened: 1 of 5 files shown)


marmos91:
Good catch — fixed in 2eac457e2.

The atomicity claim was scoped to backends but the no-clobber claim next to it was not, which is the same overclaim one paragraph later. The comment now says the winning path spares a concurrent size or mtime advance *only where that read is serialised against the writer*, and that where it is not, a write committing between the read and the update is overwritten wholesale — the same lost-update shape the surrounding rename and attribute paths already have, tracked in #2324.

Worth noting this is not a regression: the previous unconditional restore went through `SetFileAttributes`, which has the identical read-outside/write-inside shape, so postgres lost the same race before this change.

Issue #2241, opened 2026-09-02
Transaction writes diverge across backends on "the thing is already/not there": DeleteChild and CreateShare


marmos91:
The `metadata.Transaction` / `metadata.Files` contract says:

```go
// DeleteChild removes a child entry from a directory.
// Returns ErrNotFound if name does't exist in the directory.
DeleteChild(ctx context.Context, dirHandle FileHandle, name string) error
```

Two of the four backends do that. The other two do not:

| backend | missing name |
| --- | --- |
| memory | `ErrNotFound` |
| badger | `ErrNotFound` (via `mapBadgerError` on the `txn.Get`) |
| sqlite | `nil` — `RowsAffected` deliberately unchecked |
| postgres | `nil` — same |

This predates the `store/sql` merge; the SQL pair carried the same comment verbatim in both copies before the bodies moved onto `Core`.

**The SQL side is not simply wrong.** Deleting an inode cascades to its `parent_child_map` rows through the `child_id` foreign key, so by the time the service layer calls `DeleteChild` the row may already have gone with the file. Badger and memory do not cascade, so their row is still there to find. Both behaviours are locally correct; the contract only describes one of them.

**Not observable today.** All three production call sites — `file_remove.go:217`, `directory.go:250`, and both `file_modify.go:893,899` — call `DeleteChild` only for a name they have already resolved, and none of them inspects `ErrNotFound`. The conformance suite never deletes a name that is not there, which is why the split has stood.

**What to decide:** whether the contract is idempotent (relax memory and badger, amend the doc) or strict (make the SQL pair check `RowsAffected`, and confirm the cascade cannot beat it). Idempotent looks right — a caller that wants to know whether the name existed has `GetChild` — but that is the call to make, not to assume.

Whichever way it goes, it needs a `storetest` case that deletes an absent name, since nothing currently pins either behaviour.

marmos91:
A second instance of the same class, found while merging the share writes: **`Transaction.CreateShare` on a name that already exists.**

| backend | existing name |
| --- | --- |
| memory | `ErrAlreadyExists`, unless the row is a `seeded` stub from `CreateRootDirectory` |
| badger | `ErrAlreadyExists` |
| sqlite | silently overwrites the options |
| postgres | silently overwrites the options |

[...]

PR #2344, opened 2026-09-03
refactor(metadata): delete the CreateShare surface

### LINKED ISSUE #2262: Delete the CreateShare surface: no production caller, and the SQL transaction bodies cannot create a share
## Summary

`CreateShare` exists on the `Shares` interface — and therefore on both `metadata.Store` and
`metadata.Transaction` — in eight backend bodies plus `metadata.Service.CreateShare`. Nothing in
production calls any of it. It should be deleted.

## Evidence

- `metadata.Service.CreateShare` (`pkg/metadata/service.go:955`) has no callers.
- Store-level `CreateShare` is reached only from `pkg/metadata/storetest/*`, where it is used as a
  setup helper.
- Production creates shares through `CreateRootDirectory`. `pkg/controlplane/runtime/shares/lifecycle.go:419`
  says so directly: *"The runtime seeds metadata via CreateRootDirectory (not CreateShare), which
  leaves Options zero-valued"*.
- `tx.CreateShare` has one caller in the whole tree: badger's own store-level `CreateShare`
  delegating to it (`store/badger/shares.go:83`).

## The transaction bodies do not do what the name says

sqlite and postgres implement `tx.CreateShare` as:

```go
// Update options for existing share (created by CreateRootDirectory)
query := `UPDATE shares SET options = ?1 WHERE share_name = ?2`
```

It cannot create a share. Against a name that does not exist it updates zero rows and returns nil.
Against one that does it overwrites the options — which is exactly what `tx.UpdateShareOptions`,
declared a few lines below, already does by delegating to `Core`. Memory and badger meanwhile
implement a real create that rejects duplicates with `ErrAlreadyExists`.

That gap is what #2241 recorded as a semantic divergence. It is better described as a real
implementation on two backends and a misnamed stub on the other two.

## Why it has stayed invisible

The only conformance coverage is `storetest.testDuplicateCreateShare` (`store_surface.go:269`),
which exercises the **store** path, where all four backends do agree on `ErrAlreadyExists`. No test
calls `tx.CreateShare`, so the stub has never been asked to create anything.

## Proposed change

1. Remove `CreateShare` from the `Shares` interface, and with it `Service.CreateShare` and all eight
   bodies.
2. Re-point the ~10 `storetest` call sites at `CreateRootDirectory`, which is what production uses
   and what those tests actually mean.
3. Keep `testDuplicateCreateShare`'s intent by asserting it against whatever survives, or drop it
   with the surface it pins.

## Constraint

`Store` and `Transaction` embed the same `Shares` interface (`pkg/metadata/store.go:276`), so the
transaction variant cannot be removed on its own. It is one deletion or none — which is why #2241
could not simply pick a semantic for the tx half.

Split out of #2241.

marmos91:
Closes #2262.

`CreateShare` existed on the `Shares` interface — so on both `metadata.Store` and
`metadata.Transaction` — in eight backend bodies plus `metadata.Service.CreateShare`.
Nothing in production called any of it. Production registers a share by creating its
root directory.

## The premise, re-tested rather than assumed

The issue's "no production caller" claim is correct, but it is easy to talk yourself
out of: `internal/controlplane/api/handlers/shares.go:554` and
`pkg/controlplane/store/shares.go:30` do call something named `CreateShare`. That is
the **control-plane GORM store** — `CreateShare(ctx, *models.Share) (string, error)`,
a different interface with a different signature. The metadata surface
(`CreateShare(ctx, *metadata.Share) error`) had exactly one caller,
`Service.CreateShare`, which had none of its own.

## What the two SQL transaction bodies actually did

```go
// Update options for existing share (created by CreateRootDirectory)
query := `UPDATE shares SET options = ?1 WHERE share_name = ?2`
```

That cannot create a share. Against a name that does not exist it updates zero rows
and returns nil; against one that does it overwrites the options, which is what
`UpdateShareOptions` — declared a few lines below — already does. Memory and badger
implemented a real create that rejected duplicates. #2241 recorded that as a semantic
divergence; it is better described as two implementations and two misnamed stubs.

Nothing had ever noticed because the only conformance coverage,
`storetest.testDuplicateCreateShare`, exercised the **store** path, where all four
backends did agree.

## Also deleted: `shareData.seeded`

The memory backend carried a `seeded` flag so `CreateShare` could "finish" a share
record `CreateRootDirectory` had started. Its only two readers were the two
`CreateShare` bodies, so it became write-only and went with them.

## Test call sites

Setups that called `CreateShare` before `CreateRootDirectory` drop the call. Setups
that called it alone create the root directory instead, since that is what registers
a share; where they passed options, they now set them through `UpdateShareOptions`.

Three scenarios go with the surface they pinned: `testDuplicateCreateShare`, memory's
"CreateShare finishes a seed" subtest, and `postgres/createshare_contract_test.go`.
Each was checked against still-live code first — `CreateRootDirectory` is documented
and implemented as idempotent and reconciling, so duplicate-rejection is a contract
that was deliberately abandoned, not one left unpinned.

## Verification

- `go build`, `go vet` — plain **and `-tags integration`**, since build-tagged files
  are invisible to the plain run.
- `./pkg/metadata/...`, `./pkg/block/engine/...`, `./pkg/snapshot/...`,
  `./pkg/controlplane/...` green.
- postgres integration suite against a real postgres 16: green, 34s.
- Every remaining `CreateShare` in `test/`, `internal/` and `cmd/` confirmed to be
  the control-plane REST/CLI surface.

−590/+91 before review passes.


marmos91:
Adversarial review folded in.

**Finding on the badger repair paths — accepted, and it corrects a comment I had just written.**

I had annotated badger's `len(RootHandle) == 0` branch as reachable for "records a previous version wrote". That is wrong. The only writer that ever produced a share record without a root handle was `tx.CreateShare`, and `git log -S` across all refs finds no production caller of it in any commit — the one hit in `lifecycle.go` is a comment saying the runtime deliberately does *not* use it. So no database can hold that state, and the branch is dead for every store, not just new ones.

Deleted with it:
- the pool-path option-preservation block, whose only other caller runs precisely when the share record is absent;
- the transaction-path copy, which was unreachable on its own terms — it re-probes a key whose absence is why control reached it, in one transaction that writes nothing to that key in between.

An empty root handle now fails to decode rather than silently minting a second root: fail-closed and visible.

**The behaviour those blocks claimed to protect was real and untested.** A share must keep its options when it is attached again — `CreateRootDirectory` runs a second time on every restart. Added `ShareOptionsOps/OptionsSurviveReattach` to the shared suite and verified it against two deliberately broken builds: memory overwriting the share record unconditionally, and badger's `loadExistingRoot` rewriting it with defaults. Both fail the new test; both pass without it. Green on all four backends, postgres included.

**Finding on the branch containing an unrelated hardlink fix — refuted.** It compared against the stale local `develop` ref (`29ffe1d4b`) rather than `origin/develop` (`923c40e41`, the squash-merge of #2318). `git log origin/develop..HEAD` is this PR's commits only, and the PR's own file list has 47 files, none of them `file_modify.go`.

Also noted from that pass, unrelated to this PR and worth its own issue: `TestPostgresPutFile_ObjectIDConflictMapsToErrConflict` and `_DuplicatePathAllowed` parse `DITTOFS_TEST_POSTGRES_DSN` as keyword-form only, so a URL-form DSN fails them while every other test treats the variable as a boolean gate.

PR #2348, opened 2026-09-04
refactor(metadata): one block-record implementation over both SQL dialects

marmos91:
Part of #1828 (S6c). The sqlite and postgres block-record stores were the same
five statements twice, so they collapse onto `sql.Core` the way chunks, files,
shares and locks already have: the bodies live once, and each dialect
contributes only its statement text through a new `BlockRecords()` entry.

Two methods stay behind in each backend. `CommitBlock` needs a `Transactor`
rather than an executor, and `Core` is not one. `DecrLiveChunkCount` shadows its
promoted namesake — see below.

### The retry the store path used to carry

Four of the five store-level methods were already direct execs, or reads that
never wrapped. Only `DecrLiveChunkCount` ran under `WithTransaction`, and its
comment gave atomicity as the reason. That reason is now met by the statement
itself: one `UPDATE … SET live_chunk_count = MAX(0, live_chunk_count - ?) …
RETURNING` needs no surrounding transaction to be atomic.

But the wrapper carried a second thing the comment did not mention — sqlite's
BUSY/LOCKED retry budget — and that half is load-bearing. Block reclamation
(`pkg/block/engine/gc_block.go:134`) clears the synced marker *before* it
decrements, deliberately, so a re-visit resolves the block as unsynced and
cannot double-decrement a live sibling. The price of that ordering is that a
failed decrement is never retried: the count stays inflated and the block is
never reclaimed. A transient conflict there leaks a remote object permanently,
where the same conflict on `PutBlockRecord` or `DeleteBlockRecord` merely fails
a sweep the caller repeats.

So `DecrLiveChunkCount` keeps its wrapper on both SQL stores, in the same shape
`memory` already uses. The shared single-statement body is unchanged; what the
merge removed was the read-modify-write duplication, not the retry.

For postgres the retry could not have fired regardless: the pool passes no
`TxOptions` and the DSN carries only `sslmode` and `connect_timeout`, so these
run at READ COMMITTED where a single statement cannot serialize-fail, and 40P01
needs two locks taken in opposing order, which five single-row-by-primary-key
statements cannot produce. It keeps the wrapper for symmetry, not for need.

### Connection acquisition

The old store path wrapped acquisition in its own timeout and retried it; a
direct pool exec waits on the caller's context instead. That is how every other
`Core` method has behaved since S3a, so this is alignment rather than a new
exposure.

### Tests

The suite gained `BlockRecordTxOps`, covering the transaction-level block-record
surface that nothing exercised before: read-your-writes, rollback discards, and
a committed delete. That last one is not redundant with rollback — a transaction
whose delete silently does nothing leaves exactly the state a rolled-back delete
leaves, so only a committed one can tell them apart. Verified by stubbing
memory's transaction delete to a no-op: the rollback case still passed, the new
case failed.

Integration-tagged `./pkg/metadata/...` green against a local postgres 16 and
sqlite.


marmos91:
Copilot's finding was right, and the consequence is worse than backpressure.

`DecrLiveChunkCount` was the one block-record write sqlite wrapped in `WithTransaction` — its comment gave atomicity as the reason, which the single `UPDATE … SET live_chunk_count = MAX(0, …) … RETURNING` now supplies on its own. But the wrapper also carried the busy/conflict retry budget, and that half is load-bearing.

The sole production caller (`pkg/block/engine/gc_block.go:134`) clears the synced marker **before** decrementing, deliberately, so a re-visit resolves the block as unsynced and cannot double-decrement a live sibling. The cost of that ordering is that a failed decrement is never retried: the count stays inflated and the block is never reclaimed. So a transient conflict here leaks a remote object permanently, where the same conflict on `PutBlockRecord` or `DeleteBlockRecord` just fails a sweep the caller repeats.

Restored the wrapper for that one method on both SQL stores — the same shape `memory` already uses — and left the other four on the direct path. The shared single-statement body is unchanged, so the read-modify-write duplication this PR removes stays removed.

The PR description's claim about the removed transactions still holds for postgres (READ COMMITTED, single-row-by-PK, so 40001/40P01 cannot fire); it was sqlite's BUSY/LOCKED retry that it failed to account for.

Issue #2341, opened 2026-09-03
NFSv4.0: bad, old and stale stateids are accepted instead of rejected


marmos91:
Found by the pynfs NFSv4 conformance suite added in #2321 (#2204).

**14 failing tests.** The Linux kernel NFS server passes every one of them in the same harness (see `test/nfs-conformance/pynfs/baseline-knfsd.md`), so these are ours, not assertions no server satisfies.

### NFSv4.0 — 14 tests

- `CLOSE4` (st_close.testBadStateid) — CLOSE with a bad stateid should return NFS4ERR_BAD_STATEID, instead got NFS4_OK
- `CLOSE5` (st_close.testOldStateid) — CLOSE with an old stateid should return NFS4ERR_OLD_STATEID, instead got NFS4_OK
- `CLOSE6` (st_close.testStaleStateid) — CLOSE with a stale stateid should return NFS4ERR_STALE_STATEID, instead got NFS4ERR_BAD_STATEID
- `LKT9` (st_lockt.testStaleClientid) — LOCKT with a bad clientid should return NFS4ERR_STALE_CLIENTID, instead got NFS4_OK
- `LKU6b` (st_locku.testBadLockSeqid2) — LOCKU with a bad lockseqid=1 should return NFS4ERR_BAD_SEQID, instead got NFS4_OK
- `LOCK10` (st_lock.testStaleClientid) — LOCK with a bad clientid should return NFS4ERR_BAD_STATEID or NFS4ERR_STALE_CLIENTID, instead got NF
- `LOCK9b` (st_lock.testOldOpenStateid) — LOCK with old openstateid should return NFS4ERR_OLD_STATEID, instead got NFS4_OK
- `LOCK9c` (st_lock.testOldOpenStateid2)
- `OPCF1` (st_openconfirm.testConfirmCreate) — OPEN_CONFIRM done twice in a row should return NFS4ERR_BAD_STATEID, instead got NFS4_OK
- `OPCF6` (st_openconfirm.testStaleStateid) — OPEN_CONFIRM with a stale state should return NFS4ERR_STALE_STATEID, instead got NFS4ERR_BAD_STATEID
- `OPDG2` (st_opendowngrade.testNewState1) — OPENDOWNGRADE to never opened mode should return NFS4ERR_INVAL, instead got NFS4_OK
- `OPDG3` (st_opendowngrade.testNewState2)
[...]

PR #2356, opened 2026-09-06
fix(nfs/v4): three NFSv4.0 conformance defects behind the RPLY failures

### LINKED ISSUE #2337: NFSv4.0: the duplicate request cache replays the wrong status
Found by the pynfs NFSv4 conformance suite added in #2321 (#2204).

**3 failing tests.** The Linux kernel NFS server passes every one of them in the same harness (see `test/nfs-conformance/pynfs/baseline-knfsd.md`), so these are ours, not assertions no server satisfies.

### NFSv4.0 — 3 tests

- `RPLY10` (st_replay.testCloseWait) — Call to be replayed should return NFS4_OK or NFS4ERR_EXPIRED, instead got NFS4ERR_BAD_STATEID
- `RPLY14` (st_replay.testMkdirReplay) — Replay the first time should return NFS4_OK, instead got NFS4ERR_EXIST
- `RPLY3` (st_replay.testReplayState2) — Call to be replayed should return NFS4ERR_ISDIR, instead got NFS4_OK

### Reproduce

```bash
cd test/nfs-conformance/pynfs
./run-pynfs.sh --minor-version 4.0 --tests "RPLY10 RPLY14 RPLY3" --verbose
```

These are blacklisted as `bug` in `test/nfs-conformance/pynfs/KNOWN_FAILURES_V40.md` / `_V41.md`; walk the rows out as they are fixed.

marmos91:
## The issue's premise is only one-third right

#2337 is titled "the duplicate request cache replays the wrong status", and it
groups three pynfs NFSv4.0 failures under that one cause. Two of the three never
reach the duplicate request cache at all — they fail on the *first* call, before
anything is replayed. pynfs labels its checks, and the labels in the issue body
say which:

| Test | pynfs check that failed | What that means |
| --- | --- | --- |
| RPLY3 | `Call to be replayed should return NFS4ERR_ISDIR, instead got NFS4_OK` | the original call, not a replay |
| RPLY10 | `Call to be replayed should return NFS4_OK or NFS4ERR_EXPIRED, instead got NFS4ERR_BAD_STATEID` | the original call, not a replay |
| RPLY14 | `Replay the first time should return NFS4_OK, instead got NFS4ERR_EXIST` | an actual replay defect |

`_replay()` in `nfs4.0/servertests/st_replay.py` checks the first compound under
the label "Call to be replayed", captures whatever status came back, and only
then retransmits under "Replay the first time" / "Replay the second time". So
RPLY3 and RPLY10 are two independent bugs that happen to live in a file named
`st_replay.py`, and the third one is the DRC.

The third one also isn't the DRC "replaying the wrong status". The server has a
duplicate request cache; NFSv4.0 was simply never routed through it, so there was
nothing to replay.

Three separate root causes, one commit each.

## RPLY3 — OPEN had no file-type gate

OPEN is defined over regular files only. The handler resolved the component,
checked the caller's permissions against the requested `share_access`, and
opened whatever it found. Opening a directory therefore returned `NFS4_OK` and
handed back an open stateid for it.

RFC 7530 §16.16.6:

> If the component provided to OPEN resolves to something other than a regular
> file (or a named attribute), an error will be returned to the client. If it is
> a directory, NFS4ERR_ISDIR is returned; otherwise, NFS4ERR_SYMLINK is
> returned. Note that NFS4ERR_SYMLINK is returned for both symlinks and for
> special files of other types.

The fix folds the type check into `checkOpenTarget` (was `checkOpenAccess`), the
single gate every OPEN of an already-existing object passes through, so all four
paths get it rather than only the one the test exercises: CLAIM_NULL NOCREATE,
CLAIM_NULL CREATE landing on an existing name, CLAIM_FH, and
CLAIM_DELEGATE_CUR. The function now returns an `nfsstat4` because ISDIR and
SYMLINK have no `metadata.ExportError` spelling.

**One behaviour I traded, stated plainly.** The gate refuses in the handler,
before `StateManager.OpenFile`, which is where `consumeSeqidOnError` is deferred
— so an ISDIR rejection does not consume the open-owner seqid. Per RFC 7530
§9.1.7 it arguably should: the exempt list in `failedSeqidReply` is
STALE_CLIENTID, STALE_STATEID, BAD_STATEID, BAD_SEQID, BADXDR, RESOURCE,
NOFILEHANDLE and MOVED, matching nfsd's `seqid_mutating_err()`, and ISDIR is not
on it.

I did not change that, because it is deliberate and tested.
`TestOpen_ClientIDSpansPrincipals` says so in its own comment — "checkOpenAccess
refuses before the state layer sees the request, so seqid 3 is still unused when
the read OPEN below presents it" — and then reuses seqid 3. The same is already
true of NFS4ERR_ACCESS and predates this PR; my ISDIR and SYMLINK returns follow
that convention rather than break it.

So for the specific input "OPEN a directory" this PR does trade behaviour: it
used to wrongly succeed and consume the seqid, and now correctly returns ISDIR
and does not. If the pinned convention turns out to be the wrong one, this adds a
case to it. That question — whether a permission or type refusal has "reached the
owner's sequence" — is #2359, and it wants checking against knfsd and
nfs-ganesha rather than against the RFC alone.

## RPLY10 — lease cancellation left no trace

When a v4.0 lease lapsed, `onLeaseExpired` dropped the client's open, lock and
delegation state and removed the client record. Nothing recorded that the state
had ever existed, so a client returning after the partition and using one of
those stateids was answered `NFS4ERR_BAD_STATEID` — the code for a stateid the
server never issued.

RFC 7530 §9.6.3.2:

> When a lease is canceled, all locking state associated with it is freed, and
> the use of any of the associated stateids will result in NFS4ERR_EXPIRED being
> returned.

The two statuses send a client into the same recovery, so this is not a
correctness hole so much as a lie about *why*. It is still what the RFC
specifies, and it is what a client's error reporting shows the user.

The expiry callback now records the stateids it is about to free, and every miss
path a returning client can reach consults that set before falling back:
validation of open, lock and delegation stateids, plus CLOSE, OPEN_CONFIRM and
OPEN_DOWNGRADE. Deliberately **not** touched: the stale-vs-bad epoch
classification, which is #2341's surface — this change adds an `EXPIRED` branch
where a miss already existed and changes no existing classification.

The set is capped and dropped wholesale on overflow rather than aged out per
entry; it carries a `ponytail:` comment naming that ceiling. Losing an entry only
degrades the answer back to the `NFS4ERR_BAD_STATEID` the server gave before.

## RPLY14 — NFSv4.0 never reached a duplicate request cache

This is the real one, and the finding is that the cache was not being consulted
rather than that it replayed the wrong status.

NFSv4.0 has no SEQUENCE. Sequenced operations — OPEN, CLOSE, LOCK, LOCKU,
OPEN_CONFIRM, OPEN_DOWNGRADE, RELEASE_LOCKOWNER — are covered by the open- and
lock-owner seqid replay caches in the state manager, which is why RPLY1, RPLY2,
RPLY9 and friends already pass. CREATE, REMOVE, RENAME and LINK carry no seqid.
A retransmission of one of those after a lost reply is indistinguishable from a
new request, so it was executed a second time and the client was told
`NFS4ERR_EXIST` for a directory create that had in fact succeeded. RFC 7530
§16.16.6 calls the missing machinery "the server duplicate request cache
mechanism".

**The cache already existed.** `pkg/adapter/nfs/drc.go` has been in the tree
since June (`3e37a2add`): sharded, keyed on (source address, XID, CRC-32 of the request body),
with in-flight duplicate detection, a TTL and cap eviction, modelled on knfsd's
`fs/nfsd/nfscache.c`, and covered by nine tests in `drc_test.go`. The adapter
already gates every non-idempotent NFSv3 procedure through
`drc.lookup` / `drc.record` / `drc.abort`. NFSv4 simply never reached it.

So the defect is not "there is no duplicate request cache" but "v4.0 does not
route through the one we have", which is also what shrinks the risk here: a
reimplementation carrying its own eviction policy would have to earn trust on its
own, whereas a branch into infrastructure the v3 path has exercised for three months
inherits it. An earlier version of this commit did write that second cache, in
the v4 handlers, before I found this one; it is gone.

The wiring is: `handleNFSv4Procedure` consults the cache for minorversion-0
COMPOUNDs only, and `dispatchV40` sets `compCtx.CacheReply` when the COMPOUND
actually ran one of those four operations, so the adapter records the reply or
releases the reserved slot accordingly. v4.1 and v4.2 are gated out — they take
exactly-once semantics from the session slot table, and a second cache answering
their retransmissions would be a second source of truth.

Deciding cacheability *after* execution (from the result opcodes) rather than
before is what keeps READ replies out of the cache without having to decode
every operation's arguments up front to find out what the COMPOUND contains.

A side effect worth naming: v4.0 now also drops a duplicate that arrives while
the original is still executing, instead of running it twice concurrently. That
is the behaviour pynfs itself documents for the Linux server — `_replay()` sleeps
0.3s between retransmissions with the comment "we happen to know the current
Linux server implementation will drop a replay if it comes too quickly".

### Two things a reviewer will want to check

**Aliasing.** The cache hands the same backing array to every replay.
`drc.record` copies the reply before storing it, and the only consumer,
`rpc.MakeSuccessReply`, marshals into a freshly allocated buffer and patches the
record-marking fragment header inside *that* buffer — the payload it was handed
is never written through. Verified by reading
`internal/adapter/nfs/rpc/parser.go`, not assumed. This is the same property the
v3 path has depended on since the cache was written, which is why it holds rather
than merely happening to.

**Leaked in-progress slots.** `lookup` reserves a slot, and a slot never
released is worse than it first looks: `lookup` matches an in-progress entry
*before* it considers the entry's age, and `evictIfNeeded`'s TTL sweep only runs
when a shard is at capacity. So a leaked slot does not expire — it answers every
later retransmission of that exact request with a silent drop, indefinitely.

The normal paths are covered: between the reserve and the record, the only call
is `ProcessCompound`; the GSS auth-failure early return happens *before* the
lookup; and the two early returns after it — replay and in-flight duplicate —
reserve nothing.

The path that is not covered by ordinary control flow is a panic.
`ProcessCompound` has no internal recover, and `handleRequestPanic` recovers
**per request** and leaves the connection open — so the client keeps its source
port, keeps generating the same cache key, and that request stays wedged for the
life of the connection. The release is therefore a `defer`, not a trailing
branch. `abort` deletes an entry only while it is still in-progress, so it is a
no-op once the reply has been recorded and needs no flag to guard it.

This is the same mistake as the CRC cost below, in a different currency, and it
is the transferable lesson from this PR. Neither problem exists at v3's scale,
because v3's DRC covers nine small procedures. Routing *every* v4.0 COMPOUND
through the same machinery turned two things that were fine at a narrow blast
radius into problems at a wide one — an unguarded release that used to leak
rarely, and a per-byte checksum that used to run only on tiny requests. The
lesson is not "add a size bound" or "add a defer"; it is that widening a narrow
mechanism means re-checking every cost and every failure mode it was allowed to
have while it was narrow. I found this one twice, by two different routes.

The v3 path has the same unguarded shape. It is deliberately left alone here —
that is regression surface this PR has no reason to touch — and is tracked as
#2357. What made it worth fixing on the v4 path anyway is that this PR *widens*
the exposure considerably: v3 reserves a slot only for its nine non-idempotent
procedures, whereas every v4.0 COMPOUND now reserves one. "The v3 path has the
same shape" is an argument about precedent, and precedent stops transferring once
the rate changes by that much.

#2357 also carries the better question, which this PR does not attempt: whether
`lookup` should consider an entry's age *before* matching it as in-progress. That
would bound the damage of any future leak at the TTL, instead of leaving every
call site responsible for remembering the release.

A replayed reply returns early and so skips `maybeRegisterBackchannel` and the
`noteNFSVersion` refinement. Both are correct: backchannel registration only
fires for a connection carrying a v4.1 session binding, which a v4.0 COMPOUND
cannot have, and `noteNFSVersion("4")` has already run in the dispatcher before
the refinement to "4.0".

### What review caught

Copilot flagged that routing every v4.0 COMPOUND through the cache makes each
one pay a CRC-32 over the full request body, and that `record` keeps the whole
reply while the cache bounds entry count rather than bytes. Both were right, and
the first is worse than it was reported: the v4.0 path checksums the body
**twice**, once in `lookup` to reserve the slot and once in the deferred `abort`
to release it. Measured `crc32.ChecksumIEEE`:

| Payload | ns/op | throughput |
| --- | --- | --- |
| 1 MiB | 252,830 | 4.1 GB/s |
| 64 KiB | 11,122 | |
| 8 KiB | 1,369 | |
| 256 B | 25 | |

So a 1 MiB v4.0 WRITE was paying ~506µs of CRC-32 for a compound the cache would
never keep — only CREATE, REMOVE, RENAME and LINK are recorded, and those carry a
filehandle and a name. NFSv3 never pays this because its DRC covers nine small
procedures; putting it on the whole v4.0 surface is what made payload size
matter — the same widening that made the slot leak worth guarding, arrived at
by an entirely different route.

Copilot's second point was independent and equally right: bounding entry *count*
while storing whole replies is a memory bound in name only. 4096 entries times an
unbounded reply is not a cap.

Both directions are now bounded by `maxDupReqBytes` (8 KiB), far above any
namespace-mutating compound and far below bulk data. Anything larger is left to
re-execute on a retransmission, which is what the server did before the cache
covered v4.0. Worst-case cache footprint goes from unbounded to ~32 MiB.

The third comment was a wording fix — the COMPOUND tag is a `utf8str_cs`, not an
opaque, though the wire encoding is the same. Checked against RFC 7530 §15.2.2
before changing it.

## Evidence

`go build ./...`, `go vet ./...`, `gofmt` clean, `go test ./...` green,
`go test -race ./internal/adapter/nfs/...` green.

Every new test was run against a deliberately broken build before being trusted,
and that step earned its keep. The first version of the replay test had its
controls ordered so the differing-body control ran first and overwrote the cached
entry, which masked two mutations: keying the cache without the XID and keying it
without the client address both left the test green. That version of the cache is
gone, but the lesson stood up the replacement's tests, each of which now fails
under every mutation that would break the property it claims to pin:

| Test | Mutation | Result |
| --- | --- | --- |
| `TestIsV40Compound` | accept any minorversion (v4.1 leaks into the cache) | FAIL |
| `TestDispatchV40_CacheReply` | mark every COMPOUND cacheable | FAIL |
| `TestDispatchV40_CacheReply` | mark no COMPOUND cacheable | FAIL |
| `TestOpen_NonRegularTarget` | drop the ISDIR branch | FAIL |
| `TestOpen_NonRegularTarget` | drop the SYMLINK branch | FAIL |
| `TestExpiredLease_StateidsReportExpired` | drop the recording call | FAIL |
| `TestV40Compound_ReleasesDRCSlotOnPanic` | release the slot in a trailing branch instead of a `defer` | FAIL (`lookup` = `drcInProgressDup`) |
| `TestDRCEligibleV40Compound` | drop the request size gate | FAIL |
| `TestDRCRecordableReply` | drop the reply size bound | FAIL |
| `TestDRCRecordableReply` | ignore the "ran a mutation" flag | FAIL |
| all of the above | none (control) | ok |

Both halves of the size bound are pure predicates (`drcEligibleV40Compound`,
`drcRecordableReply`) precisely so they can be tested without standing up a
runtime to provoke an 8 KiB reply. The handler keeps one call to each.

`TestExpiredLease_StateidsReportExpired` also carries a never-issued stateid as a
control, so it cannot pass by everything answering `NFS4ERR_EXPIRED`.

The cache's own contract — replay on a match, a differing checksum treated as a
new request, in-flight duplicates dropped, TTL and cap eviction, concurrent
access — is already covered by the nine tests in `pkg/adapter/nfs/drc_test.go`
and is unchanged by this PR.

## pynfs verification

The three rows are **removed** from `test/nfs-conformance/pynfs/KNOWN_FAILURES_V40.md`
in this PR, which is what makes them verify anything. The grader counts only NEW
failures, so a blacklisted test that starts passing is invisible while its row
stands — the row has to come out before the test is load-bearing. With it gone,
`pynfs v4.0 / memory` and `pynfs v4.0 / postgres-s3` go red and name the code if
any of RPLY3, RPLY10 or RPLY14 still fails, and stay green only if all three
pass. Two backends, graded by the harness rather than by a human reading a
terminal.

The backend split is not decoration: some rows in these tables are
backend-specific — the SATT12x family is postgres-only — so a memory-only run
could walk out a row that fails on postgres. CI catches that; a local run would
not have.

**RPLY8 stays.** It is categorised `suite`, and `baseline-knfsd.md` records the
Linux kernel server failing it in this same harness, along with RPLY5, RPLY6 and
RPLY7. That baseline is also what establishes the three codes here are ours:
knfsd passes all three.

Verified on the merged base (`c810d03db`), with the rows removed, on both
backends:

```
pynfs v4.0 / memory        Loaded 140 known failures
                           13 Skipped, 130 Failed, 5 Warned, 453 Passed
                           New failures: 0

pynfs v4.0 / postgres-s3   Loaded 140 known failures
                           13 Skipped, 131 Failed, 5 Warned, 452 Passed
                           New failures: 0
```

`Loaded 140` is the load-bearing number. The table held 151 rows before this
batch; #2358 removed six (LINK2, LINK9, LOOK8, RNM10, RNM11, RNM20) and #2361
removed two (LOCK20, LOCKMRG), leaving 143, and this PR removes three more. If
the rebase had restored a row or dropped an extra one, that number would not be
140 — it is what proves the run graded against the merged table rather than a
stale copy.

601 tests ran on each backend, so the suite genuinely executed rather than
silently skipping. In both failing lists RPLY8 appears once and RPLY3, RPLY10 and
RPLY14 do not appear at all. RPLY8's presence is the control: it proves the
search can find a code that is there, so the other three absences mean something.

With the rows gone, a still-failing RPLY3, RPLY10 or RPLY14 is by definition a
NEW failure. Both backends report zero. That is the gate, not the absence.

Closes #2337

### DIFF
--- a/internal/adapter/nfs/v4/handlers/compound.go
+++ b/internal/adapter/nfs/v4/handlers/compound.go
@@ -387,9 +387,9 @@ func (h *Handler) dispatchV40(compCtx *types.CompoundContext, tag []byte, numOps
 		return encodeCompoundResponse(types.NFS4ERR_RESOURCE, tag, nil)
 	}
 
-	// v4.0 has no SEQUENCE/replay cache, so an opcode that fails to decode is a
-	// fatal protocol error (hardErrOnDecodeError) rather than a cached partial
-	// reply.
+	// v4.0 has no SEQUENCE, so an opcode that fails to decode is a fatal
+	// protocol error (hardErrOnDecodeError): there is no slot to cache a partial
+	// reply in.
 	results, lastStatus, err := h.runCompoundOps(compCtx, numOps, reader, compoundLoopParams{
 		isV41:                false,
 		hardErrOnDecodeError: true,
@@ -398,7 +398,40 @@ func (h *Handler) dispatchV40(compCtx *types.CompoundContext, tag []byte, numOps
 		return nil, err
 	}
 
+	// Tell the adapter whether this reply has to survive for a retransmission.
+	// Sequenced operations carry their own replay protection in the open- and
+	// lock-owner seqid caches; these four carry none, so re-executing a
+	// retransmitted one turns a success into NFS4ERR_EXIST or NFS4ERR_NOENT.
+	compCtx.CacheReply = compoundIsNonIdempotent(results)
+
 	return encodeCompoundResponse(lastStatus, tag, results)
+}
+
+// nonIdempotentV40Ops are the NFSv4.0 operations that change the namespace and
+// have no seqid of their own, and so rely entirely on the duplicate request
+// cache to survive a retransmission (RFC 7530 Section 16.16.6 refers to it as
+// "the server duplicate request cache mechanism").
+//
+// The rest of the operation set is either seqid-protected (OPEN, CLOSE, LOCK,
+// LOCKU, OPEN_CONFIRM, OPEN_DOWNGRADE, RELEASE_LOCKOWNER) or safe to repeat, and
+// caching a READ reply would pin megabytes to protect an operation that can
+// simply run again.
+var nonIdempotentV40Ops = map[uint32]bool{
+	types.OP_CREATE: true,
+	types.OP_REMOVE: true,
+	types.OP_RENAME: true,
+	types.OP_LINK:   true,
+}
+
+// compoundIsNonIdempotent reports whether a COMPOUND ran an operation that must
+// not be executed a second time.
+func compoundIsNonIdempotent(results []types.CompoundResult) bool {
+	for i := range results {
+		if nonIdempotentV40Ops[results[i].OpCode] {
+			return true
+		}
+	}
+	return false
 }
 
 // dispatchV41 executes the v4.1 COMPOUND dispatch loop with SEQUENCE gating.
--- a/internal/adapter/nfs/v4/handlers/compound_cacheable_test.go
+++ b/internal/adapter/nfs/v4/handlers/compound_cacheable_test.go
@@ -1,0 +1,82 @@
+package handlers
+
+import (
+	"testing"
+
+	"github.com/marmos91/dittofs/internal/adapter/nfs/v4/types"
+)
+
+// TestDispatchV40_CacheReply pins which v4.0 COMPOUNDs ask the adapter to keep
+// their reply for a retransmission.
+//
+// NFSv4.0 has no SEQUENCE. Sequenced operations are covered by the open- and
+// lock-owner seqid replay caches, but CREATE, REMOVE, RENAME and LINK carry no
+// seqid: re-executing a retransmitted one reports NFS4ERR_EXIST or
+// NFS4ERR_NOENT for an operation that in fact succeeded. Everything else is
+// either seqid-protected or safe to repeat, and caching a READ reply would pin
+// megabytes to protect nothing.
+func TestDispatchV40_CacheReply(t *testing.T) {
+	cases := []struct {
+		name string
+		ops  []compoundOp
+		want bool
+	}{
+		{
+			name: "create",
+			ops:  []compoundOp{{opCode: types.OP_CREATE, data: encodeCreateDirArgs("subdir")}},
+			want: true,
+		},
+		{
+			name: "remove",
+			ops:  []compoundOp{{opCode: types.OP_REMOVE, data: encodeRemoveArgs("victim")}},
+			want: true,
+		},
+		{
+			name: "read-only",
+			ops:  []compoundOp{{opCode: types.OP_GETFH}},
+			want: false,
+		},
+		{
+			name: "no operations",
+			want: false,
+		},
+	}
+
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			fx := newRealFSTestFixture(t, "/export")
+			fx.createTestFile(t, fx.rootHandle, "victim", 0, 0o644, 0, 0)
+
+			ctx := newRealFSContext(0, 0)
+			ops := append([]compoundOp{{opCode: types.OP_PUTFH, data: encodePutFHArgs(fx.rootHandle)}}, tc.ops...)
+			data := buildCompoundArgsWithOps(nil, types.NFS4_MINOR_VERSION_0, ops)
+
+			if _, err := fx.handler.ProcessCompound(ctx, data); err != nil {
+				t.Fatalf("ProcessCompound error: %v", err)
+			}
+			if ctx.CacheReply != tc.want {
+				t.Fatalf("CacheReply = %v, want %v", ctx.CacheReply, tc.want)
+			}
+		})
+	}
+}
+
+// TestDispatchV41_NeverCachesReply checks the v4.1 path leaves CacheReply
+// alone: a v4.1 retransmission is recognised by the session slot table, and a
+// second cache answering it would be a second source of truth.
+func TestDispatchV41_NeverCachesReply(t *testing.T) {
+	fx := newRealFSTestFixture(t, "/export")
+
+	ctx := newRealFSContext(0, 0)
+	data := buildCompoundArgsWithOps(nil, types.NFS4_MINOR_VERSION_1, []compoundOp{
+		{opCode: types.OP_PUTFH, data: encodePutFHArgs(fx.rootHandle)},
+		{opCode: types.OP_CREATE, data: encodeCreateDirArgs("subdir")},
+	})
+
+	if _, err := fx.handler.ProcessCompound(ctx, data); err != nil {
+		t.Fatalf("ProcessCompound error: %v", err)
+	}
+	if ctx.CacheReply {
+		t.Fatal("v4.1 COMPOUND asked for its reply to be cached")
+	}
+}
--- a/internal/adapter/nfs/v4/handlers/open.go
+++ b/internal/adapter/nfs/v4/handlers/open.go
@@ -322,9 +322,10 @@ func (h *Handler) handleOpenClaimNull(
 		if encErr != nil {
 			return openError(types.NFS4ERR_SERVERFAULT)
 		}
-		// Check access permissions for the requested share_access
-		if accessErr := checkOpenAccess(metaSvc, authCtx, fh, shareAccess); accessErr != nil {
-			return openError(common.MapToNFS4(accessErr))
+		// Check that the target is a regular file the caller may open with the
+		// requested share_access.
+		if status := checkOpenTarget(metaSvc, authCtx, fh, shareAccess); status != types.NFS4_OK {
+			return openError(status)
 		}
 		fileHandle = fh
 
@@ -376,9 +377,10 @@ func (h *Handler) handleOpenClaimNull(
 			if encErr != nil {
 				return openError(types.NFS4ERR_SERVERFAULT)
 			}
-			// Check access permissions for the requested share_access
-			if accessErr := checkOpenAccess(metaSvc, authCtx, fh, shareAccess); accessErr != nil {
-				return openError(common.MapToNFS4(accessErr))
+			// Check that the target is a regular file the caller may open with
+			// the requested share_access.
+			if status := checkOpenTarget(metaSvc, authCtx, fh, shareAccess); status != types.NFS4_OK {
+				return openError(status)
 			}
 			fileHandle = fh
 
@@ -555,9 +557,9 @@ func (h *Handler) handleOpenClaimFH(
 		return openError(types.NFS4ERR_SERVERFAULT)
 	}
 
-	// Enforce the requested share_access against the file's permissions.
-	if accessErr := checkOpenAccess(metaSvc, authCtx, fileHandle, shareAccess); accessErr != nil {
-		return openError(common.MapToNFS4(accessErr))
+	// Enforce the file type and the requested share_access against the file.
+	if status := checkOpenTarget(metaSvc, authCtx, fileHandle, shareAccess); status != types.NFS4_OK {
+		return openError(status)
 	}
 
 	// If another client holds a conflicting delegation, recall it and ask the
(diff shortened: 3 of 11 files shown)


marmos91:
All three review comments are addressed in `ea39893f9`.

**1. CRC-32 on every v4.0 COMPOUND.** Correct, and worse than reported: the v4.0 path checksums the request body *twice* — once in `lookup` to reserve the slot, once in the deferred `abort` to release it. Benchmarked `crc32.ChecksumIEEE` rather than taking it on assertion:

| Payload | ns/op |
| --- | --- |
| 1 MiB | 252,830 |
| 64 KiB | 11,122 |
| 8 KiB | 1,369 |
| 256 B | 25 |

So a 1 MiB v4.0 WRITE was paying ~506µs for a compound the cache would never keep — only CREATE, REMOVE, RENAME and LINK are recorded. NFSv3 never pays this because its DRC covers nine small procedures; putting it on the whole v4.0 surface is what made payload size matter. Gated behind `drcEligibleV40Compound`.

**2. Unbounded reply bytes.** Also correct and independent — bounding entry *count* while storing whole replies is a cap in name only. Both directions now bound at `maxDupReqBytes` (8 KiB), far above any namespace-mutating compound. Worst-case footprint goes from unbounded to ~32 MiB. Gated behind `drcRecordableReply`.

Both are pure predicates so the bounds are testable without standing up a runtime to provoke an 8 KiB reply. `TestDRCEligibleV40Compound` and `TestDRCRecordableReply` each fail under every mutation that would break what they pin.

**3. Tag type.** Right — checked RFC 7530 §15.2.2 before changing (`utf8str_cs tag;`, `typedef utf8string utf8str_cs;`). Reworded to say it is a `utf8str_cs` whose wire encoding is that of a variable-length opaque.

PR #2365, opened 2026-09-06
refactor(metadata): one file_block_refs implementation over both SQL dialects

### LINKED ISSUE #2364: sqlite's manifest load drops start_offset, so FindByObjectID returns ChunkRefs pointing at the head of every block
`SQLiteMetadataStore.loadFileChunkRefs` selects `"offset", size, hash` and scans three columns. The `start_offset` column — where a chunk's data begins *inside* its block, which is what makes block packing addressable — is neither selected nor set, so every `block.ChunkRef` it returns carries `StartOffset == 0`.

Postgres's copy of the same function selects and sets all four. The two were written as separate near-verbatim ports of one another and drifted; nothing compares them, so nothing noticed.

### Reachability

Latent today. The only reader is `FindByObjectID`, and its only production consumer is the file-level dedup retry described in `pkg/block/engine/coordinator.go:164` — `applyFileLevelDedupHit`, which does not exist yet (grep finds it only in comments, alongside `ErrPersistFileChunksNotWired`). Every other caller is a test fake.

So no user has been served a wrong offset. But when that path is wired, a sqlite-backed share would resolve a deduped file's chunks to the head of each packed block and read the wrong bytes — silently, with no error and no checksum mismatch to catch it, since the read itself succeeds.

### Status

Fixed as a side effect of the sqlite/postgres `file_block_refs` merge in the #1828 track: once one implementation serves both dialects, the column list and the scan targets come from the same constant and cannot disagree. Filed for the record — the fix ships in that PR, not separately.

The general shape is worth noting for the rest of the merge: this is the second time a pair that looked like pure boilerplate turned out to hold a behavioural difference. Reviewing each collapse for drift is worth more than reviewing it for style.

marmos91:
Part of #1828 (S6d). Closes #2364.

The sqlite and postgres backends each carried a copy of the file-manifest
delta writer — the code that brings `file_block_refs` into agreement with a
file's `[]block.ChunkRef` by writing only the rows that changed. Every statement
in it is identical in both dialects apart from its placeholders, so the two
copies collapse into one implementation in `store/sql`.

`Dialect` gains `Placeholder(i int)`. Its doc has always rejected such a method,
and that rejection still holds for the statements it was written about: a
constant differing by `?N` versus `$N` should just be spelled twice. It does not
reach the statements here, which bind one parameter per changed offset and so
are assembled at runtime — there is no constant for a dialect to spell in
advance. The doc now says which case is which.

### What changed for postgres

It loses `pgx.Batch` and the `= ANY($2)` array form for the batched IN-list and
multi-row insert sqlite already used, both of which postgres accepts. Under 180
changed chunks that is one statement where it previously queued one per ref;
past that it pays a round-trip per batch, which the `ponytail:` marker on
`maxBoundParams` names along with the upgrade path.

One behavioural difference came with it and is fixed rather than shipped: an
offset repeated within a single manifest used to resolve last-wins on postgres,
because each ref was its own statement. Two conflicting rows in one multi-row
insert are rejected outright by postgres and quietly applied twice by sqlite, so
the delta now drops the earlier ref and both dialects agree on last-wins.

### The write path does not ride on Core

Every other shared body is a method on `Core`, promoted onto both the store and
its transaction. This one is a function taking an executor, because it is the
only shared body that spans several statements. Promoted, it would also be
callable on the pool-backed store, where the delete and upsert batches
autocommit separately — a crash between them leaves the manifest with holes,
and a hole reads back as zeros rather than as an error. Taking the executor as a
parameter keeps the caller's transaction visible at the call site and turns a
store-level call into a compile error.

### The drift it uncovered (#2364)

sqlite's manifest load selected `"offset", size, hash` and never set
`StartOffset` — where a chunk's bytes begin inside a packed block. Postgres's
copy set all four columns. `FindByObjectID` is the only reader, and its only
production consumer (`applyFileLevelDedupHit`) exists solely in comments, so the
zero was never reachable; it would have detonated silently on sqlite-backed
shares the moment file-level dedup was wired. One implementation means the
column list and the scan targets now come from the same constant and cannot
disagree.

### Tests

The suite could not have caught any of this. Nothing wrote a manifest longer
than four refs, so the point where a batching backend starts its second
statement was never reached; nothing gave a `ChunkRef` a non-zero `StartOffset`;
and `assertBlocks` compares hash, offset and size only.

- `ChunkRefOps/ManyBlocksWithStartOffset` writes 401 refs with distinct non-zero
  StartOffsets, then shrinks to one, crossing the batch boundary in both
  directions and comparing whole structs.
- `ObjectIDOps/FindByObjectID` gains non-zero StartOffsets and now compares its
  refs instead of only counting them.

Verified against three deliberately broken builds — writing `start_offset` as
zero, numbering a batch's placeholders from the global row index, and dropping
`StartOffset` from the manifest load. Each fails these tests and no others; the
last reproduces #2364 exactly.

Integration-tagged `./pkg/metadata/...` green against a local postgres 16 and
sqlite, after rebasing onto develop.


marmos91:
The build is fine — `for col := range colsPerRow` is range-over-int, legal since Go 1.22, and this module is on `go 1.25.0`. `go build ./...`, `go vet` (plain and integration-tagged) and the full integration suite against a local postgres 16 all pass on this branch.

The check that actually failed was golangci-lint, on something else: staticcheck QF1008 for `tx.Core.X` naming an embedded field that can be elided. Fixed by spelling it `tx.X, tx.D` — same value, and `golangci-lint run ./pkg/metadata/...` is now clean locally.

Issue #2366, opened 2026-09-06
journal: FileSize is an O(intervals) scan under the shard lock, and share start pays it per file


marmos91:
`journal.Store.FileSize` (`pkg/block/journal/store.go:842`) computes a file's size by
scanning every interval in its index, holding the shard lock for the whole scan:

```go
var size int64
for _, iv := range fi.ivs {
    if e := iv.end(); e > size {
        size = e
    }
}
```

Measured on the new `BenchmarkFileSize`, cost is exactly linear in interval count:

| intervals | ns/op | allocs |
|---|---|---|
| 64 | 199.6 | 0 |
| 512 | 1519 | 0 |
| 4096 | 12963 | 0 |
| 32768 | 101871 | 0 |

Eight times the intervals, eight times the time. A 32k-interval file costs ~100µs
per call, and every read on that shard blocks behind it.

### Where it is actually reached

The pier design plan asserts this sits on the hot read path, citing
`engine/readwrite.go:40,53`, and keeps `Size` out of the proposed merged `Extents`
query for that reason. **That premise is wrong.** Those two lines are inside
`engine.Store.GetSize` and `engine.Store.Exists`, and neither has a non-test
caller — nor does the `block.Reader` interface that declares them
(`pkg/block/filechunk.go:168-177`). It is a dead surface, the same shape as the
hash-keyed CAS island.

The one live caller is `findStaleSizes`
(`pkg/controlplane/runtime/shares/lifecycle.go:241`), which runs at share start
over every locally-resident file. Its own comment states the stakes:

> share start compares every locally-resident file, and on a store holding
> millions of them that is what the share's visibility waits on.

So the cost is O(files × intervals-per-file) under contended shard locks, paid
before a share becomes visible — which puts it on the same thread as the startup
latency work in #1891 and #2075, not on the read path.

### Fix shape

`fi.ivs` is offset-sorted, so the maximum end is at or near the tail; a reverse
scan answers in O(1) for the common case. It cannot simply take the last element,
because overlapping intervals are reachable (#1934: truncate-narrow plus re-carve
leaves overlapping rows), so a correct version scans backward only while an
earlier interval could still extend past the running maximum. Maintaining a
per-file `maxEnd` alongside the index is the other option and makes it O(1)
unconditionally, at the cost of keeping it correct across clip, delete and repack.

Either way the guard is the same: whatever lands must be measured against
`BenchmarkFileSize` at 32k intervals, and must agree with the current
[...]

Issue #2075, opened 2026-08-24
Share start is O(files) twice — ~4.5 min before a 10M-file share is visible


marmos91:
## Summary

Share start does O(files) work twice, on separate paths. Extrapolating measured
per-item costs, a 10M-file share spends roughly **4.5 minutes** on this before it
becomes visible. Both paths are invisible in normal testing because at ~10k files
they cost ~100 ms combined.

Surfaced while answering #1891, whose reporter could not capture a
startup-time-vs-file-count curve and assumed the cost was the `usedBytes` scan.
It is not — that is the cheaper of the two by ~10×.

## The two paths

**1. `initUsedBytesCounter` (`store/badger/store.go:626`)** — full iterate + decode
of every `f:` row at store open, before any listener binds, to seed the usage/quota
cache. Also carries the `pl:` backfill from #1886.

**2. `reconcileMetadataSizeFromJournal` (`shares/lifecycle.go:186`)** — one
`GetFileByPayloadID` per **locally-resident** journal file, called from `AddShare`
at `lifecycle.go:99` *before the share is visible*. Gated only on
`LocalBlockStoreID != "" && share.BlockStore != nil`, i.e. every normal share — it
is not one-shot-flag gated like the re-seed path in #1887. This is the loop whose
quadratic form was #1885; #1886 made each lookup an index hit but left the
per-file iteration.

## Measurements

Scaleway DEV1-L (4 vCPU / 7 GB, local SSD), badger:

| path | measured | 1M | 10M |
|---|---|---|--:|
| `initUsedBytesCounter` | 1.9 µs/file | 2.2 s | ~19 s |
| `reconcileMetadataSizeFromJournal` | 26.0 µs/lookup | 26 s | **~4.3 min** |

The scan is **decode-bound, not I/O-bound** — cold (caches dropped) 2.49 s vs warm
2.28 s at 1.15M files, so it scales with CPU and will not improve on faster disks.
The `pl:` lookup cost is flat in N (250k: 25.8 µs, 1M: 26.0 µs), so path 2 is
linear in locally-resident file count, not worse.

10M figures are linear extrapolations from 1.2M; the data supports linearity but
does not prove it at that scale.

## Why this matters more than the raw numbers suggest

Path 2 is proportional to **locally-resident** files, so it is worst exactly in a
fully warmed + pinned share — which is the configuration the offline-capable /
sovereign story (#1848) is built on. The deployment shape we would most want to
recommend is the one that pays the most startup latency.

[...]

PR #2374, opened 2026-09-06
test(block): one benchmark per journal state transition, and the recorded baseline

marmos91:
Completes design-plan §10's benchmark contract — one benchmark per state transition —
and records the baseline the pier/crane/ferry refactor will be measured against.

The write path already had eleven benchmarks. Everything a byte does *after* it is
written — going cold, coming back, being reclaimed, surviving a restart — had none, so
an extraction could lose that behaviour without a single number moving. A green suite
has already failed to catch a data-path regression here once.

## What's in it

- **Seven new transition benchmarks**: cold read, hydrate, evict, GC repack, `Open`
  recovery, truncate, delete.
- **Four residency benchmarks** — `FileSize`, `DataExtents`, `DurableExtent`,
  `ColdExtents` — as the before-receipts for the merged `Extents` query, so the
  after-numbers have something to be compared against on the same hardware.
- **Allocations on all 22.** Not one of the original eleven called `b.ReportAllocs()`.
- **Write amplification across the whole write path**, not just the random-write pair it
  started on. It is the number that decides whether the storage design is viable.
- **`BenchmarkCarveScatteredPass` no longer imports a metadata store.** A
  latency-injecting deduper shows the same packer serialisation without this package
  depending on Badger to benchmark itself.
- **The recorded baseline** under `.planning/perf/2026-09-06-block-transitions/`.

## The baseline

AMD EPYC 7543, 32C/62G, Scaleway `POP2-HC-32C-64G` — the `vmType` already in
`bench/infra/Pulumi.bench.yaml`. Three store tiers, six samples each, run sequentially.
Not a laptop, not Docker Desktop.

The two Scaleway block tiers turned out to be **the same device** for this workload —
476 vs 451 IOPS under `fdatasync=1`, geomean differing 4.78% in the *slower* direction
for the provisioned-15k volume. Provisioned IOPS raises a parallel-throughput ceiling; it
does not move the ~2.1 ms round trip a serialised fsync chain pays per barrier. tmpfs
supplies the fast end instead, which separates device cost from code cost — the split the
block-size default actually needs.

| | sbs5k | ram | |
|---|---|---|---|
| `Delete` | 14.13 ms | 14.03 µs | 1007× — almost pure fsync |
| `Truncate` | 24.62 ms | 36.67 µs | 671× — almost pure fsync |
| `ConcurrentCommit1` | 1.818 ms | 7.971 µs | 228× — on the disk's append floor |
| `ConcurrentCommit128` | 42.56 µs | 12.91 µs | 3.3× — group commit buys 43× over depth 1 |
| `Carve` | 11.71 ms | 11.70 ms | 1.0× — CPU-bound, FastCDC + BLAKE3 |
| `FileSize` | 23.93 µs | 23.91 µs | 1.0× — pure CPU |
| `DurableExtent` | 16.80 ns | 16.83 ns | 1.0× |
| `OpenRecovery` | 18.74 ms | 21.11 ms | 0.9× — allocation-bound, *slower* on RAM |

## Findings

**`FileSize` is O(intervals) under the shard lock** — filed as #2366. It answers a
narrower question than `DurableExtent` for **1424×** the time, and no disk closes that gap.

**The design plan's reason for keeping `Size` out of `Extents` was wrong**, and this PR
corrects it in place. The cited hot-read call sites are `engine.Store.GetSize` and
`engine.Store.Exists`; neither has a non-test caller, nor does the `block.Reader`
interface declaring them. A third dead island, the same shape as the hash-keyed CAS one —
worth folding into the legacy-deletion scope. `FileSize`'s one live caller is
`findStaleSizes`, which runs at share start over every locally-resident file, so the cost
lands on startup latency, not on reads.

**The stop condition is answered: the protocol layer binds first.** Against #1735's
~940 µs create wall (block COMMIT fsync ~20% of it at nj=8), `ConcurrentCommit32` at
125.9 µs is the right order for that share from an independent direction. Roughly four
fifths of a create is spent above the block engine, so the refactor cannot be justified on
create-path throughput — it has to stand on residency correctness, which is what the plan
claims for it anyway.

## Two benchmarks that were lying, and how they were caught

Both were mine, and both were caught by the machine rather than by review.

`BenchmarkEvict` reported nothing evicted in all twelve server runs. The refill assumed
eight hydrated files yield eight evictable segments; they coalesce into one active
segment, so only the first `Evict` found anything. It survived twenty iterations on a
laptop. Without the liveness guard, a fast meaningless number would have entered the
baseline as the eviction cost.

`BenchmarkHydrate`'s first guard **passed against a deliberately broken build** — counting
intervals cannot distinguish a seeded-cold file from a written one, since both hold the
same number. It counts cold intervals now, and fails on the broken build as it should.

Both guards were re-verified against deliberately broken setups.

Related: the six setup-heavy benchmarks rebuild state inside `StopTimer`, and Go sizes
`b.N` from the timed region alone — so the faster the device, the more untimed setup it
does. `Truncate` reached 28k iterations on tmpfs. They take an explicit iteration cap,
documented in the file, since that is a property of the run rather than of the source.

## Not in scope

No production code changes. Measurement only; #2366 carries the `FileSize` fix.

### DIFF
--- a/.planning/2026-09-01-pier-library-design-PLAN.md
+++ b/.planning/2026-09-01-pier-library-design-PLAN.md
@@ -896,9 +896,20 @@ Rules:
   round-trips ~2.7x and has already produced one wrong headline number in this codebase.
 - **`Extents` needs a defence.** Collapsing `FileSize`/`DataExtents`/`DurableExtent`/
   `ColdExtents` into one slice-returning query is the design's biggest performance risk:
-  `FileSize` is on the hot read path (`engine/readwrite.go:40,53`) and `ColdExtents` is
-  documented as O(live intervals) and deliberately avoids holding all shard locks at once.
-  `Size` stays separate for exactly this reason; benchmark it and keep the receipts.
+  `ColdExtents` is documented as O(live intervals) and deliberately avoids holding all
+  shard locks at once. Benchmark it and keep the receipts.
+
+  **Corrected 2026-09-06, from the receipts.** This paragraph originally kept `Size`
+  separate on the grounds that `FileSize` is on the hot read path, citing
+  `engine/readwrite.go:40,53`. Those lines are inside `engine.Store.GetSize` and
+  `engine.Store.Exists`, and **neither has a non-test caller**, nor does the
+  `block.Reader` interface declaring them — a dead surface of the same shape as the
+  hash-keyed CAS island. `FileSize`'s one live caller is `findStaleSizes`
+  (`controlplane/runtime/shares/lifecycle.go:241`), which runs at share start over every
+  locally-resident file. It *is* O(intervals) under the shard lock and measurably linear
+  (#2366), but the cost lands on startup latency, not on reads. So the hot-path argument
+  for keeping `Size` out of `Extents` does not hold; if `Size` stays separate it needs a
+  different reason. Nothing else in this section changes.
 
 ## 11. Open decisions
 
--- a/.planning/perf/2026-09-06-block-transitions/README.md
+++ b/.planning/perf/2026-09-06-block-transitions/README.md
@@ -123,6 +123,19 @@ move is about a fifth, and only at low queue depth. It has to stand on residency
 correctness, which is what the plan claims for it anyway. The one place block-engine work
 *does* pay is the code-bound set above: `FileSize`, recovery, and the carve CPU path.
 
+## Known gap in this baseline
+
+`BenchmarkColdExtentsStoreWide` was added **after** these numbers were recorded and is not
+in the tables above. It exists because `BenchmarkColdExtents` could not answer the
+question the design plan asks: `ColdExtents` walks every shard, every file in it and every
+interval of each, and a single-file store exercises none of that fan-out. Spreading the
+same work over 2048 files costs **12.6× more** (108.9 µs vs 8.6 µs on a laptop,
+shape-only) and grows with file count — so the plan's concern about that query is real,
+and the 10.31 µs in the table above understates the store-wide cost it would otherwise be
+quoted for.
+
+Record it on reference hardware with the next baseline run.
+
 ## Caveats
 
 - One machine, one run each. Six samples per point, sequential tiers, no other load.
--- a/.planning/perf/2026-09-06-block-transitions/README.md
+++ b/.planning/perf/2026-09-06-block-transitions/README.md
@@ -1,0 +1,133 @@
+# Block state-transition baseline — recorded 2026-09-06
+
+The before-numbers for the pier/crane/ferry refactor. Design-plan §10 asks for one
+benchmark per state transition, allocations on every one, write amplification on the
+write path, and a recorded baseline on real hardware. This is that baseline.
+
+Raw `go test -bench` output is in `bench-5k.txt`, `bench-15k.txt`, `bench-tmpfs.txt`;
+machine and method in `bench-env.txt`. Compare a later run with:
+
+```sh
+benchstat sbs5k=bench-5k.txt new=<later>.txt
+```
+
+## Machine
+
+AMD EPYC 7543, 32 cores, 62 GiB, Linux 6.8.0-138, Go 1.25.0, Scaleway `fr-par-1`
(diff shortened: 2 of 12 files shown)


marmos91:
## Results — recorded, and what they change

Three tiers on one EPYC 7543, six samples each, VM torn down. Full numbers in
`.planning/perf/2026-09-06-block-transitions/README.md`; the short version:

### The tiers split the transitions cleanly

| | sbs5k | ram | |
|---|---|---|---|
| `Delete` | 14.13 ms | 14.03 µs | 1007× — almost pure marker fsync |
| `Truncate` | 24.62 ms | 36.67 µs | 671× — almost pure marker fsync |
| `TinyWritesCommit` | 1.714 ms | 6.071 µs | 282× |
| `ConcurrentCommit1` | 1.818 ms | 7.971 µs | 228× — on the disk's append floor |
| `ConcurrentCommit32` | 125.9 µs | 11.75 µs | 10.7× |
| `Evict` | 2.063 ms | 194.7 µs | 10.6× |
| `GCRepack` | 4.828 ms | 1.004 ms | 4.8× |
| `ConcurrentCommit128` | 42.56 µs | 12.91 µs | 3.3× |
| `WriteAt` (64 KiB) | 99.63 µs | 35.09 µs | 2.8× |
| `Hydrate` | 55.09 µs | 35.42 µs | 1.6× |
| `Carve` (8 MiB) | 11.71 ms | 11.70 ms | 1.0× — CPU-bound |
| `FileSize` | 23.93 µs | 23.91 µs | 1.0× — pure CPU |
| `DataExtents` | 21.49 µs | 21.45 µs | 1.0× |
| `ColdExtents` | 10.31 µs | 10.34 µs | 1.0× |
| `DurableExtent` | 16.80 ns | 16.83 ns | 1.0× |
| `ReadCold` | 771.8 ns | 789.0 ns | 1.0× |
| `ReadWarm` | 4.744 µs | 5.719 µs | 0.8× |
| `RandWrite4K` | 103.7 µs | 118.4 µs | 0.9× |
| `OpenRecovery` (512 files) | 18.74 ms | 21.11 ms | 0.9× |

Below 1.0× means RAM was *slower* — the work is CPU- or allocation-bound and the tmpfs
competes with the process for memory. Those are not disk measurements at all.

### Four things this settles

**1. Group commit is the whole durability story.** Depth 1 costs 1.818 ms, depth 128
costs 42.56 µs — **43×** from amortising one barrier across the batch. This is also the
first direct explanation of why #1742 (cross-shard group commit) was correctly abandoned:
P distinct shards are P fds and therefore P mandatory barriers, so coalescing them under
one leader trades the per-shard concurrency #1740 won for a serialisation. The benchmark
that would have shown that up front now exists.

**2. `Truncate` and `Delete` are fsync, not index work.** 671× and 1007×. Anyone
optimising the index-clip half would be tuning under 0.2% of the cost.

**3. The residency queries are pure CPU, and `FileSize` is the outlier.** Identical on
every tier to three digits. `DurableExtent` answers in 16.80 ns; `FileSize` takes
23.93 µs — **1424× more for a narrower question** (#2366). No disk closes that.

**4. Recovery is allocation-bound.** `OpenRecovery` is *slower* on RAM: 36 MiB and 3426
allocations to replay 512 files. It is a startup-latency SLO and belongs on the #1891 /
#2075 share-start thread, not the I/O thread.

### The tier comparison that did not work

`sbs_5k` and `sbs_15k` are **the same device** for this workload — 476 vs 451 IOPS under
`fio --fdatasync=1 --iodepth=1`, and the benchmark geomean differs 4.78% in the *slower*
direction for the provisioned volume. Provisioned IOPS raises a parallel-throughput
ceiling; it does not move the ~2.1 ms round trip a serialised fsync chain pays per
barrier. The `sbs15k` file is kept as evidence of that, not as a data point. tmpfs
supplies the fast end instead.

### Consequence for the block-size default

The default was to be chosen from disk sensitivity. That framing does not survive: the
write path is bound by fsync *latency*, which block size does not change, and the parts
block size could plausibly move — carve, recovery, the residency queries — are CPU-bound.
Worth re-scoping before writing code rather than after.

PR #2381, opened 2026-09-07
refactor(metadata): one capability set for the SQL backends, and stop overstating timestamp precision

marmos91:
Stage S7 of #1828.

## S7 is not what the plan said it was

The plan's S7 was "switch the factory, delete the old packages". That premise is dead: `snapshot_store.go` is a genuine mechanism divergence — postgres drives raw `pgconn` COPY TO/FROM STDOUT under REPEATABLE READ, sqlite does row-by-row binary encoding — so `sqlite/` and `postgres/` stay as packages and there is no factory to switch. What follows is the part of S7 that survives that finding, plus two defects the fold surfaced.

## One capability set, and an honest resolution

The postgres and sqlite branches of `CreateMetadataStoreFromConfig` each built their own 15-field `FilesystemCapabilities` literal. Every value was identical. The comments were not:

```go
// File timestamps are stored as BIGINT unix nanoseconds (lossless),   <- postgres
// File timestamps are stored as INTEGER unix nanoseconds (lossless).  <- sqlite
TimestampResolution: time.Nanosecond,
```

Both are stale. Both stores have encoded file times as a Windows FILETIME — 100ns ticks since 1601 — since migrations `000005` (sqlite) and `000038` (postgres). Measured against the real codec before changing anything:

```
offset   1ns -> round-trip delta 0s
offset  50ns -> round-trip delta 0s
offset  99ns -> round-trip delta 0s
offset 100ns -> round-trip delta 100ns
```

**This is client-visible.** `TimestampResolution` is not internal bookkeeping — NFSv3 hands it to clients as FSINFO `time_delta` (`fsinfo.go:180`). Both SQL backends have been advertising a precision they silently discard. After this change the wire value goes from `{0, 1}` to `{0, 100}`; `durationToTimeVal` handles it correctly, and no NFSv4 or SMB path reads the field.

The value now lives as `sql.TimestampResolution`, beside the codec that decides it, rather than hand-typed in a caller that cannot see the encoding.

## dfsctl never learned about sqlite

`init.go` has accepted `--type sqlite` for as long as the branch has existed. The CLI never did — sqlite is missing from every `--type` enum, every help string, and both interactive switches. The practical effect: the store could only be created by hand-writing `--config` JSON (which bypasses the switch), and `dfsctl store metadata edit` on one failed outright with `unknown store type: sqlite`. Folded into the existing `badger` cases, since both take a single path.

## Docs

They described the pre-merge shape — independent backends, no `store/sql` — and omitted sqlite throughout. Updated `implementing-stores.md`, `architecture.md`, `configuration.md`, the `types.go` package doc, and the snapshot engine list. `cli.md` is regenerated, not hand-edited.

The interesting part is what the adversarial review made me rewrite. My first draft asserted a rule I had been following myself:

> anything running more than one statement must be a package-level function taking `(ctx, x Executor, d Dialect, ...)`, never a method on `sql.Core`

That describes the exception, not the rule. Three multi-statement bodies are `Core` methods today — `DeleteShare`, `CreateRootDirectory`, `DecrementRefCountAndReap` — and they are safe for a different reason: each is shadowed in **both** dialect packages by a store-level wrapper that runs it inside `WithTransaction`. The docs now state that actual invariant, including the part that makes it dangerous: deleting a shadow does not break the build, because the promoted `Core` method still satisfies the interface — the write just starts going to the pool.

## Verification

Every new test was run against a deliberately broken build.

`TestTimestampResolutionMatchesCodec` measures the codec's real granularity by scanning for the smallest surviving step, then compares — the expectation is derived from the codec, not restated from the constant, so a wrong value cannot agree with it by construction. Mutations, all caught:

| constant | result |
| --- | --- |
| `time.Nanosecond` (the value on develop) | FAIL |
| `200 * time.Nanosecond` | FAIL |
| `500 * time.Nanosecond` | FAIL |
| `time.Microsecond` | FAIL |

The 500ns case matters: an earlier two-probe version of this test passed against it, and the adversarial review found that. Every odd multiple of 100ns slipped through. Measuring the granularity directly closes the class rather than the instance.

`TestBuildMetadataConfig_AcceptsServerSupportedTypes` pins the CLI's accepted set to the server's. Reverting the sqlite case fails it with `server supports "sqlite" but the CLI rejects it`.

`gofmt`, `go vet`, `go run ./cmd/gendocs` (no diff) and the full `go test ./...` are green.

## Review

Three passes. The simplifier caught a real defect: inserting `sqlCapabilities()` above `configInt64` split that function from its doc comment and silently annexed it. The reviewer verified the identical-literals claim field by field against develop and traced the FSINFO arithmetic. The adversarial pass refuted the doc doctrine above and found the 500ns test gap.

## Not in this PR

`sqlite/files.go` and `sqlite/shares.go` are still 90–97% identical to their postgres counterparts — they differ only in receiver name, error sentinel, error mapper, and placeholder style, all of which `Dialect` already abstracts. Merging them needs a `Transactor` indirection because `WithTransaction` is store-only, which is a different shape of change and does not belong in a docs-and-CLI stage.

### DIFF
--- a/cmd/dfsctl/commands/store/metadata/add.go
+++ b/cmd/dfsctl/commands/store/metadata/add.go
@@ -15,7 +15,7 @@ var (
 	addName   string
 	addType   string
 	addConfig string
-	// BadgerDB specific
+	// badger and sqlite
 	addDBPath string
 )
 
@@ -27,11 +27,15 @@ var addCmd = &cobra.Command{
 Supported types:
   - memory: In-memory store (fast, ephemeral)
   - badger: BadgerDB store (persistent, embedded)
+  - sqlite: SQLite store (persistent, embedded)
   - postgres: PostgreSQL store (persistent, distributed)
 
 Type-specific options:
   badger:
     --db-path: Path to BadgerDB directory (or prompted interactively)
+
+  sqlite:
+    --db-path: Path to the SQLite database file (or prompted interactively)
 
   postgres:
     --config: JSON with connection settings, or omit for interactive prompts
@@ -46,6 +50,9 @@ Examples:
   # Add a BadgerDB store interactively
   dfsctl store metadata add --name persistent-meta --type badger
 
+  # Add a SQLite store with flags
+  dfsctl store metadata add --name persistent-meta --type sqlite --db-path /data/meta.db
+
   # Add a PostgreSQL store with JSON config
   dfsctl store metadata add --name pg-meta --type postgres --config '{"host":"localhost","dbname":"dittofs"}'
 
@@ -56,9 +63,9 @@ Examples:
 
 func init() {
 	addCmd.Flags().StringVar(&addName, "name", "", "Store name (required)")
-	addCmd.Flags().StringVar(&addType, "type", "", "Store type: memory, badger, postgres (required)")
+	addCmd.Flags().StringVar(&addType, "type", "", "Store type: memory, badger, sqlite, postgres (required)")
 	addCmd.Flags().StringVar(&addConfig, "config", "", "Store configuration as JSON (for advanced config)")
-	addCmd.Flags().StringVar(&addDBPath, "db-path", "", "Database path (required for badger)")
+	addCmd.Flags().StringVar(&addDBPath, "db-path", "", "Database path (required for badger and sqlite)")
 	_ = addCmd.MarkFlagRequired("name")
 	_ = addCmd.MarkFlagRequired("type")
 }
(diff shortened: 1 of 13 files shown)


marmos91:
Both suggestions taken in `02f49437e` — `t.TempDir()` and the explicit `+= time.Nanosecond`.

One correction on the reasoning for the first, since it matters for the next person who reads that test: the hard-coded `/tmp/...` path would **not** have failed on Windows. `buildMetadataConfig` never touches the filesystem — it puts the string in a `map[string]any{"path": path}` and returns it, so the value is inert on every platform. The Windows job does run this test (`windows-build.yml` runs `go test -short -race -v ./...`), and it would have passed.

`t.TempDir()` is still the better call — it removes the question rather than requiring a reader to go check whether the path is ever opened — so the change stands on its own merits, just not on the stated one.

PR #2386, opened 2026-09-07
fix(journal): fence a hydrate against the truncate marker's own version

### LINKED ISSUE #2247: journal: Truncate's truncVer stamp uses the pre-marker peek version, leaving a version gap where Hydrate un-truncates the file
`Truncate` stamps `sh.truncVer[id]` with the LSN it *peeks* before the truncate marker exists, never with the marker's real version. That leaves a version range that passes the staleness fence but is not covered by the clip, so a racing `Hydrate` can re-fill the truncated tail and the resulting interval survives the clip permanently.

## Where

`pkg/block/journal/store.go:968`

## What

`Truncate` (`store.go:938`) stamps the fence from a peek, under the first locked section:

```go
// store.go:968
sh.truncVer[id] = s.version.Load()   // Vpeek
```

It then releases `sh.mu`, and only afterwards does the marker get its real version:

```go
// store.go:976
truncVer, err := s.appendTruncateMarker(ctx, id, newSize)   // Vmarker, from s.nextVersion()
```

`appendTruncateMarker` assigns `Vmarker` via `s.nextVersion()` under a different lock, so `Vpeek < Vmarker` whenever any unrelated write lands in between. `sh.truncVer[id]` is **never raised to `Vmarker`** — the second locked section uses the local variable only, for the clip test:

```go
// store.go:987
if iv.version > truncVer || iv.end() <= newSize {   // local Vmarker, not sh.truncVer[id]
```

So the two guards end up fenced at different versions:

| guard | fenced at |
| --- | --- |
| `staleAfterTruncate` (`store.go:473`) | `notAfter <= sh.truncVer[id]` → **Vpeek** |
| clip loop (`store.go:987`) | `iv.version > truncVer` → **Vmarker** |

## Why it matters

A `Hydrate` whose `at` bound falls in `(Vpeek, Vmarker]` passes `staleAfterTruncate` — its bound is greater than `Vpeek`, so it does not look stale — and can land inside the clip window. `fileIndex.hydratable` treats the not-yet-clipped span as an unversioned hole and fills it. The interval it writes back is stamped with a fresh version that can exceed `Vmarker`, so it then also survives the clip's own `iv.version > truncVer` test.

The result is not a transient race that the clip cleans up behind itself: the resurrected tail is **persistent past the truncate**. A file truncated to `newSize` keeps bytes beyond `newSize`, with no error and no log line.

This is the same silent-corruption family as #1850 / #1888 / #2084, and it is adjacent to #2231 — both are holes in the same per-shard fence machinery in `shard.go` (`truncVer` at `shard.go:35-40`). #2231 adds the missing delete fence; this one closes a gap in the truncate fence that already exists.

## Fix

Raise the stamp to the marker's real version in the second locked section, after `appendTruncateMarker` returns and while `sh.mu` is already held:

```go
if truncVer > sh.truncVer[id] {
    sh.truncVer[id] = truncVer
}
```

Keep the existing `Vpeek` stamp — it is what closes the window *before* the marker exists. The point is that the fence must end up at `Vmarker`, not that it should start there. Take the max rather than assigning, so a concurrent truncate that already published a higher fence is not lowered.

## Notes

- Verified against `origin/develop` (`e2ce799a3`): `sh.truncVer[id]` is assigned at `store.go:968` and nowhere else in `Truncate`; the value returned by `appendTruncateMarker` is used only as a local in the clip loop.
- `shard.go` has `truncVer` but no `tombVer` — the delete-side equivalent tracked by #2231. Fixing both in the same locked section is cheaper than two passes over this code.
- Related, still unfiled: `Delete` (`store.go:807`) unconditionally clears `sh.truncVer[id]` even when intervals survived the tombstone race, reopening the same staleness hole for the file's surviving incarnation. #2231's fence work likely wants to address that at the same time.

Source: `.planning/2026-09-01-block-dataflow-MASTER-PLAN.md` §3a (fix now, independent of the refactor) and `.planning/2026-09-01-journal-audit-report.md` (verified finding, `bugs`/cross).

marmos91:
Closes #2247.

`Truncate` publishes its hydrate fence from a version peeked *before* the marker exists, then clips the index against the marker's version, which is higher whenever an unrelated write lands between the two locked sections. A hydrate that sampled its bound inside that window clears the fence — its bound is above the peek, so it does not read as stale — and the interval it writes back is versioned above the marker, so the clip keeps it too. The truncated tail comes back and **stays** back, with no error and no log line.

Same silent-corruption family as #1850 / #1888 / #2084, and adjacent to the delete fence #2252 added.

## Two things the issue's prescribed fix did not account for

The issue prescribed raising the fence to `truncVer` and nothing else. I applied exactly that; the two targeted repros went green and **four unrelated tests broke**.

**1. The fence is per file, so raising it also refuses the truncate's own survivors.** `DiscardLocalContent` is the live caller that breaks — server-side copy truncates the destination to zero and hydrates the copied content straight back. So the fence now also records the offset the mutation left standing, and refuses only ranges reaching past it. A truncate's survivors still arbitrate through `hydratable`; a delete leaves none, so it keeps refusing everything.

**2. The version boundary differs between truncate and delete.** A truncate admits a bound equal to its marker: that caller resolved its fetch with the marker already minted, and the survivors are there to weigh it against. A delete cannot — it scrubs the index entry, `hydratable`'s nil receiver then offers the whole range — so it fences one version higher. That is what it already did; the field is now named `minBound` and each producer sets it, rather than the difference being implied by an inclusive comparison.

Caught by running the package tree rather than the repro. `TestTruncateDoesNotFenceOffItsOwnSurvivors` is the regression guard: it passes on unmodified develop and fails only against the over-broad fix.

## A second defect in the same stamp

`Truncate` **assigned** the fence where `fenceDelete` raises it — and `fenceDelete`'s own comment says "never lower a fence a racing Truncate stamped higher". The reverse ordering was unhandled: a truncate following a delete dropped the delete's fence to its own peek and re-admitted the hydrates that delete was refusing. Both now go through `raiseHydrateFence`.

## Verification

The window needs a concurrent write to open, so a test seam lands one (`beforeTruncateMarker`, nil in production, alongside the existing `failTruncate`/`failTombstone` seams). In production any other writer to the store does the same thing.

Both defects reproduce on unmodified develop:

```
--- FAIL: TestHydrateInTheTruncateVersionWindowIsRefused
    truncated tail resurrected: FileSize = (8192,true), want (4096,true)
--- FAIL: TestTruncateDoesNotLowerADeleteFence
    Truncate lowered the fence from 1001 to 1
```

`go test ./pkg/block/...` green, `-race` on the journal package green.

### DIFF
--- a/pkg/block/journal/hydrate_delete_test.go
+++ b/pkg/block/journal/hydrate_delete_test.go
@@ -125,12 +125,12 @@ func TestDeleteFenceCountIsBounded(t *testing.T) {
 func TestDeleteFenceEvictionSparesARestampedFence(t *testing.T) {
 	sh := newShard(nil)
 	sh.fenceDelete("x", 5)
-	sh.hydrateFence["x"] = 99 // a later Truncate re-stamps the same key
+	sh.raiseHydrateFence("x", 99, 4096) // a later Truncate re-stamps the same key
 	for i := 0; i < maxDeleteFences+1; i++ {
 		sh.fenceDelete(FileID(fmt.Sprintf("f%d", i)), uint64(i+100))
 	}
-	if got := sh.hydrateFence["x"]; got != 99 {
-		t.Fatalf("eviction dropped a re-stamped fence: hydrateFence[x] = %d, want 99", got)
+	if got := sh.hydrateFence["x"]; got.minBound != 99 {
+		t.Fatalf("eviction dropped a re-stamped fence: hydrateFence[x].minBound = %d, want 99", got.minBound)
 	}
 }
 
--- a/pkg/block/journal/segment.go
+++ b/pkg/block/journal/segment.go
@@ -328,7 +328,7 @@ func (s *Store) appendRecord(ctx context.Context, id FileID, offset int64, data 
 	// and the fetched bytes are the wrong ones.
 	if synced {
 		n := int64(len(data))
-		if hydrateFenced(sh, id, notAfter) || !coversWhole(sh.index[id].hydratable(offset, n, notAfter), offset, n) {
+		if hydrateFenced(sh, id, notAfter, offset, n) || !coversWhole(sh.index[id].hydratable(offset, n, notAfter), offset, n) {
 			return nil
 		}
 	}
--- a/pkg/block/journal/shard.go
+++ b/pkg/block/journal/shard.go
@@ -32,16 +32,23 @@ type shard struct {
 	active *segmentMeta
 	sealed map[uint64]*segmentMeta
 	index  map[FileID]*fileIndex
-	// hydrateFence is, per file, the store LSN as it stood when that file's most
-	// recent truncate or delete began. A hydrate whose caller sampled its bound
-	// at or below it is refused: both mutations empty intervals away rather than
-	// recording over them, so the range they cleared leaves nothing for
-	// hydratable to weigh a stale write-back against — a delete leaves not even
-	// an index entry, and hydratable's nil receiver offers the whole range.
+	// hydrateFence is, per file, the version of that file's most recent truncate
+	// or delete together with the offset that mutation left standing. A hydrate
+	// whose caller sampled its bound at or below the version, and whose range
+	// reaches past the surviving offset, is refused: both mutations empty
+	// intervals away rather than recording over them, so the range they cleared
+	// leaves nothing for hydratable to weigh a stale write-back against — a
+	// delete leaves not even an index entry, and hydratable's nil receiver
+	// offers the whole range.
+	//
+	// The offset is what keeps a truncate from fencing off its own survivors:
+	// the prefix below newSize still holds its intervals and can be filled
+	// safely, and only the emptied tail has to be refused. A delete leaves
+	// nothing standing, so its offset is zero and it refuses every range.
 	//
 	// A delete's entry therefore has to outlive the file's index entry, and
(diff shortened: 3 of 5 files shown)


marmos91:
Copilot's finding was real and is fixed in `06f07fd7c`.

It is the same relaxation this PR's first commit fixes, on the adjacent field. `raiseHydrateFence` stopped a later truncate lowering an earlier delete's fence *version*, then overwrote that delete's `survives` with the truncate's `newSize` — widening what the fence admits by exactly as much as the version guard had just protected.

Reachable, and confirmed so rather than assumed: delete a file, write it again sparsely, truncate it. A hydrate whose bound predates the delete then clears the version test (against the truncate's marker) and the range test (against `newSize`), and fills the holes the re-created file left below `newSize` with the content the delete removed.

`survives` now only ever narrows. `TestTruncateAfterDeleteDoesNotReAdmitPreDeleteHydrates` pins it, and fails against the previous commit:

```
--- FAIL: TestTruncateAfterDeleteDoesNotReAdmitPreDeleteHydrates
    a truncate after a delete re-admitted a pre-delete hydrate: deleted content resurrected in the prefix
```

Worth naming the pattern, since it is the second time in this PR: fixing one half of a two-field invariant primed me not to look at the other half. The version guard and the range guard have to be non-relaxing for the same reason, and I only wrote that reason down for one of them.

PR #2388, opened 2026-09-07
build(nix): make the flake usable day to day

### LINKED ISSUE #2323: build(nix): expose the flake's tooling more usefully for day-to-day development
## Context

Noticed while adding the `pynfs` package in #2204. The flake works, but a few
things make it less useful than it could be for everyday development. None of
these are urgent; filing so they are not lost.

## Findings

**1. No `apps` output.** `nix run .#dfs`, `nix run .#dfsctl` and
`nix run .#pynfs` all fail — there is no `apps` attribute (`grep -c apps
flake.nix` → 0). Anyone who does not already have the repo built has to
`nix build` then reach into `./result/bin`. A handful of lines gets
`nix run github:marmos91/dittofs#dfs` working for people who have never cloned it.

**2. Helper scripts are Linux-only.** `dfs-mount`, `dfs-umount`, `dfs-posix`,
`dfs-e2e`, `dfs-postgres-start`/`-stop` are all in `linuxInputs`
(`flake.nix:375-382`), so a macOS dev shell has none of them, and the shell
banner still advertises them (`flake.nix:424-438`). `dfs-posix` genuinely needs
Linux (it reads `/proc/mounts`), but `dfs-postgres-start` is just `docker run`
and works fine on macOS. Worth splitting by what actually needs Linux rather
than by convenience.

**3. Test suites are not on the dev shell PATH cross-platform.** `pjdfstest` is
in `linuxInputs`, correctly. `pynfs` is cross-platform but is not in any shell,
so `run-pynfs.sh` shells out to `nix build` on every invocation. Because it uses
`--no-link`, the result is not a GC root, so a `nix store gc` between runs
silently triggers a full rebuild — which is how a two-second startup becomes a
ten-minute one that looks like a hang. Adding `pynfs` to `commonBuildInputs`
fixes both.

**4. `stdenvNoCC` for pure-data derivations.** `pynfs` originally used
`stdenv.mkDerivation` and pulled the entire C toolchain — on darwin, clang plus
the Apple SDK, close to a gigabyte — to run a Python script that compiles
nothing. Fixed there in #2204; worth auditing whether any other derivation has
the same shape.

**5. `checks` are thin.** `nix flake check` only verifies that `dfs` and
`dfsctl` respond to `--help` (`flake.nix:535-560`). The repo has several
self-contained shell test suites that need no server and run in about a second
each — `test/smb-conformance/smbtorture/parse-results_test.sh`,
`test/nfs-conformance/pynfs/parse-results_test.sh`. They are the things that
decide whether a conformance run grades green, so they are exactly what a
pre-push `nix flake check` should cover.

**6. `devShells.ci` duplicates `devShells.default`.** They differ only in the
banner and `darwinInputs` (`flake.nix:394` vs `:453`). `ci` could be derived
from `default` rather than maintained alongside it.

## Not proposing

Nothing here changes `version = "0.31.0"` as the single source of truth for
releases (CLAUDE.md), or the existing package/devShell names.

marmos91:
Closes #2323.

Six findings from the flake audit. Five produced a change; one legitimately found nothing.

## 1. No `apps` output

`nix run .#dfs` / `.#dfsctl` / `.#pynfs` had nothing to resolve to, so anyone
without a built checkout had to `nix build` and reach into `./result/bin`.

Adds `apps.{default,dfs,dfsctl,pynfs-4_0,pynfs-4_1}`. The pynfs attributes use an
underscore because `nix run` splits an attribute path on dots — `nix run .#pynfs-4.0`
is parsed as `pynfs-4` → `0` and fails with "does not provide attribute". A bare
`nix run .#pynfs` resolves through the new `meta.mainProgram = "pynfs-4.1"` on the
package, since pynfs ships one entrypoint per minor version and neither is "the" binary.

`apps.default` and `packages.default` both land on `dfs`, so the documented
`nix run github:marmos91/dittofs -- init` keeps working unchanged.

## 2. Helper scripts split by convenience rather than by need

`dfs-postgres-start` / `dfs-postgres-stop` only shell out to the docker CLI, which
exists on macOS, but they sat in `linuxInputs`. They move to `commonBuildInputs`.
`dfs-mount`, `dfs-umount`, `dfs-posix` and `dfs-e2e` stay Linux-only — they need the
Linux NFS client, `/proc/mounts` or pjdfstest.

**The macOS dev shell did not evaluate at all**, before this change and independently
of it: `fio` in `commonBuildInputs` pulls `libnbd`, whose `meta.platforms` is
Linux-only, so `nix develop` on aarch64-darwin failed with "not available on the
requested hostPlatform". Verified on unmodified `develop`:

```
$ nix eval --raw .#devShells.aarch64-darwin.default.drvPath
error: Package 'libnbd-1.22.5' ... is not available on the requested hostPlatform:
  hostPlatform.system = "aarch64-darwin"
```

`fio` moves to `linuxInputs`; the darwin shell now evaluates. Nothing on macOS
invoked `fio` — it could not have, the shell would not open.

The shell banner is reworked so the PostgreSQL and pynfs sections print on both
platforms instead of only in the Linux branch.

## 3. pynfs not on the dev-shell PATH

`run-pynfs.sh` shelled out to `nix build --no-link` per invocation; with no GC root,
a `nix store gc` between runs turned a two-second startup into a full rebuild that
looks like a hang. `pynfs` joins `commonBuildInputs`. It is cross-platform on purpose:
it speaks NFSv4 over TCP itself, so it needs no kernel mount and no privileges.

## 4. `stdenvNoCC` audit — nothing to change

The only `stdenv.mkDerivation` left is `pjdfstest`, which genuinely runs
`autoreconf`/`configure`/`make` and needs a compiler. `pynfs` is already
`stdenvNoCC`. Everything else in the file is `writeShellScriptBin` or `runCommand`,
both of which are `stdenvNoCC`-backed in nixpkgs already. No change.

## 5. The conformance graders are not covered by `nix flake check`

`test/smb-conformance/smbtorture/parse-results_test.sh` and
`test/nfs-conformance/pynfs/parse-results_test.sh` decide whether a conformance run
grades green, need no server or network, and run in about a second each. They now run
as `checks.conformance-graders`.

The check does not just take exit 0 for an answer. A test script that matched no cases
exits 0 having asserted nothing, which reads exactly like a pass, so the check also
requires each script to emit at least one `ok:` line. Both guards were verified against
deliberately broken inputs rather than assumed:

- grader stubbed to `exit 0` unconditionally → 32 of 43 smbtorture assertions fail, build red.
- test script replaced by one that prints `PASS:` and exits 0 without asserting →
  `test/nfs-conformance/pynfs asserted nothing`, build red.

Green on the real tree: 43 smbtorture assertions, 28 pynfs assertions.

## 6. `devShells.ci` duplicated `devShells.default`

Now `devShell.overrideAttrs { shellHook = goEnv; }` — same `buildInputs`, without the
banner and without exec-ing into zsh. `darwinInputs` is an always-empty list, so the
`ci` shell's package set is unchanged on every platform.

## Verification

- `nix flake check` green on aarch64-darwin (builds both binaries and both graders).
- `devShells.x86_64-linux.{ci,default}`, `packages.x86_64-linux.pjdfstest`,
  `checks.x86_64-linux.conformance-graders` and every app all evaluate.
- `nix run .#pynfs -- --help`, `.#pynfs-4_0 -- --help` and `.#dfsctl -- --help` all run.
- `version` and `vendorHash` untouched.

### DIFF
--- a/docs/internals/contributing.md
+++ b/docs/internals/contributing.md
@@ -42,6 +42,19 @@ The Nix flake provides:
 - golangci-lint for code quality
 - NFS utilities for E2E testing (Linux)
 - ACL libraries for POSIX compliance testing
+- pjdfstest for POSIX compliance testing (Linux)
+- pynfs for NFSv4 protocol conformance, as `pynfs-4.0` and `pynfs-4.1`
+- `dfs-postgres-start` / `dfs-postgres-stop` for a PostgreSQL test container
+
+`nix flake check` builds both binaries and runs the conformance graders, which is
+what decides whether a suite run counts as green.
+
+The binaries also run straight from the flake, without a clone:
+
+```bash
+nix run github:marmos91/dittofs#dfs -- start
+nix run github:marmos91/dittofs#dfsctl -- --help
+```
 
 ### Manual Setup (Alternative)
 
--- a/flake.nix
+++ b/flake.nix
@@ -171,7 +171,6 @@
           };
 
         # Helper script to start PostgreSQL for testing
-        # Uses sudo for docker commands to avoid docker group requirement
         dfs-postgres-start = pkgs.writeShellScriptBin "dfs-postgres-start" ''
           container_name="dittofs-postgres-test"
 
@@ -181,26 +180,31 @@
             exit 1
           fi
 
-          # Check if docker daemon is running (using sudo)
-          if ! sudo docker info &>/dev/null; then
+          # Docker Desktop and a rootless daemon answer to the invoking user;
+          # a system daemon usually does not. Probe before reaching for sudo,
+          # so the script neither demands a password it does not need nor
+          # loses, to sudo's secure_path, the docker it just found.
+          DOCKER=(docker)
+          docker info &>/dev/null || DOCKER=(sudo "$(command -v docker)")
+          if ! "''${DOCKER[@]}" info &>/dev/null; then
             echo "Error: Cannot connect to Docker daemon."
-            echo "Make sure Docker daemon is running: sudo systemctl start docker"
+            echo "Start it first (Linux: sudo systemctl start docker)."
             exit 1
           fi
 
           # Check if container already exists
-          if sudo docker ps -a --format '{{.Names}}' | grep -q "^$container_name$"; then
+          if "''${DOCKER[@]}" ps -a --format '{{.Names}}' | grep -q "^$container_name$"; then
             # Check if it's running
-            if sudo docker ps --format '{{.Names}}' | grep -q "^$container_name$"; then
+            if "''${DOCKER[@]}" ps --format '{{.Names}}' | grep -q "^$container_name$"; then
               echo "PostgreSQL container already running"
               exit 0
             else
               echo "Starting existing PostgreSQL container..."
-              sudo docker start "$container_name"
+              "''${DOCKER[@]}" start "$container_name"
             fi
           else
             echo "Creating PostgreSQL container for DittoFS testing..."
-            sudo docker run -d \
+            "''${DOCKER[@]}" run -d \
               --name "$container_name" \
               -e POSTGRES_USER=dittofs \
               -e POSTGRES_PASSWORD=dittofs \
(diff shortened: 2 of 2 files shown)


marmos91:
Good catch on the sudo — fixed in 70fe1bf3c.

`sudo docker` was right for a system daemon and wrong everywhere else: Docker Desktop and a rootless daemon answer to the invoking user, and sudo's `secure_path` drops the `docker` the script had just resolved on PATH. Both helpers now probe `docker info` first and only fall back to `sudo "$(command -v docker)"` — absolute path, so secure_path can't lose it.

Verified on macOS from inside the dev shell: `dfs-postgres-stop` runs to completion with no sudo prompt, which it could not do before.

PR #2391, opened 2026-09-07
test: unify the conformance suites behind one runner and a central manifest

### LINKED ISSUE #2322: test: unify the conformance suites behind one runner and a central manifest
## Problem

We now have five conformance/compliance suites, each with its own runner, its own
flags, and its own CI wiring:

| Suite | Runner | Flags | Substrate |
|---|---|---|---|
| WPTS BVT | `test/smb-conformance/run.sh` | `--profile --mode --filter --category --keep --dry-run` | Docker (+`--mode local`) |
| smbtorture | `test/smb-conformance/smbtorture/run.sh` | `--profile --filter --kerberos --timeout --keep` | Docker |
| pjdfstest | `test/posix/setup-posix.sh` + `run-posix.sh` | positional profile, `--nfs-version`, `--no-mount` | host + Nix |
| NFS Kerberos | `test/nfs-conformance/run.sh` | `--verbose` only | Docker Compose |
| pynfs | `test/nfs-conformance/pynfs/run-pynfs.sh` | `--profile --minor-version --tests --no-setup` | host + Nix |

They already share one thing — the `KNOWN_FAILURES.md` table format and the
`test/common/known-failures.sh` parser — and that part works well. Everything
around it is duplicated and drifting:

- Profile names differ per suite (`badger-fs` vs `badger`; `*-s3` supported by
  some runners and not others).
- Only WPTS has `--mode local`; only smbtorture has `--timeout`; only
  nfs-conformance has no usage block at all.
- Each suite re-implements server provisioning, health-waiting, results
  directories, step summaries and artifact upload.
- CI tiering (which profiles on PR vs push vs nightly) is copy-pasted into five
  workflow files, with the matrix ternary repeated each time.
- `ci-health.yml` carries two hand-maintained lists of workflow names that must
  stay in sync, and `docs/internals/contributing.md` + `testing.md` carry two
  more that have already drifted.

## Proposal

A single entrypoint plus a declarative manifest, so a suite is enabled and
configured in one place:

```yaml
# test/conformance/suites.yaml
suites:
  pynfs:
    description: NFSv4 protocol conformance
    runner: nfs-conformance/pynfs/run-pynfs.sh
    profiles: [memory, badger, postgres, postgres-s3]
    variants: {minor-version: ["4.0", "4.1"]}
    known-failures: KNOWN_FAILURES_V{variant}.md
    tiers: {pull_request: [memory, postgres-s3], push: all, nightly: all}
    timeout: 45
```

```bash
test/conformance/run.sh --suite pynfs --profile memory
test/conformance/run.sh --list
test/conformance/run.sh --suite all --tier pull_request
```

The manifest then generates (or drives) the CI matrix, so adding a suite means
adding one manifest entry instead of a workflow file plus four doc tables plus
two `ci-health.yml` lists.

## Scope

- [ ] Common runner: profile resolution, server lifecycle, results dir, grading, step summary, artifacts.
- [ ] `suites.yaml` manifest with per-suite profiles, variants, tiers and timeouts.
- [ ] One parameterised workflow driven by the manifest (or a small generator, kept honest by a CI check that regenerated output matches).
- [ ] Normalise profile names across suites.
- [ ] Make `ci-health.yml` and the docs tables read from the manifest rather than repeating it.
- [ ] Migrate the five existing suites one at a time; the `KNOWN_FAILURES` format and parser stay as they are.

## Also folds in

- #1603 — xfstests. It is explicitly scoped as "a harness parallel to
  `test/posix/` with its own KNOWN_FAILURES, setup/teardown and CI wiring",
  i.e. a sixth copy of the same scaffolding. It should land as a manifest entry
  instead, and is a good forcing function for getting the manifest shape right.

## Notes

Worth doing after #2204 lands, so the pynfs suite is a real second data point
for what the common runner has to support (it is the only one that needs no
mount and no root, which the abstraction must not assume away).

marmos91:
Part of #2322.

Stacked on #2388 (issue #2323) — review that one first; this branch contains its commits.

Five conformance suites, five runners, five sets of flags, and the per-event
profile tiering copy-pasted into three workflow files. The suite roster was
written down a further three times: twice in `ci-health.yml` and twice in the
docs, where it had already drifted.

Now it is written down once.

## `test/conformance/suites.json`

Each suite gets one entry: its profiles, its variant axis, the blacklist each
variant grades against, which profiles run per GitHub event, and the ordered
steps to execute. Everything downstream reads it — the runner, the CI matrix,
`ci-health.yml`'s monitored-workflow list, and the documentation tables.

JSON rather than the YAML the issue sketched, for one reason: GitHub Actions can
`fromJSON()` it directly and `jq` is already everywhere, so neither the workflow
nor the runner needs a YAML parser that the runners do not ship. Comments live in
`_comment` keys.

Every field in it is read by something — an earlier draft carried `runner_dir`,
`grader`, and a `mount`/`docker` flag per step that nothing consulted, and `mount`
sitting next to the `root` flag that *is* honoured read as enforced while doing
nothing. Two shapes in the schema exist because a suite needs them, not for symmetry:

- **`known_failures` may be an object keyed by variant.** pynfs grades v4.0 and
  v4.1 against separate tables with separate row sets, so one path per suite
  cannot express it. pjdfstest uses the same shape (v3 has its own table; v4 and
  v4.1 share one).
- **`root` is declared per STEP, never per run.** pynfs is
  its own NFSv4 client: it needs no kernel mount and no privileges, and
  `setup-posix.sh` already skips its EUID check under `--no-mount`. A runner that
  hoisted a root precondition to the top would make it the only suite that cannot
  run locally.

## `test/conformance/run.sh`

```
run.sh --list
run.sh --suite wpts --profile memory
run.sh --suite pjdfstest --profile memory --variant 4.1
run.sh --suite smbtorture --tier push
run.sh --matrix pull_request          # the CI job matrix, as JSON
```

Each matrix cell carries the suite's `timeout_minutes`, so the workflow says
`timeout-minutes: ${{ matrix.timeout }}` rather than restating 30 / 45 / 60.

**It reports a suite's own exit status, not a boolean.** The graders exit with
the *number* of failures that are not on the blacklist, and `parse-results.sh`
hard-errors when the results trailer is missing so a truncated run cannot grade
green. Collapsing that to `1` throws away the only number saying how bad a red
run is, so `run_one` reads `PIPESTATUS[0]` past the `tee` and the top-level loop
propagates the worst status.

It also clears an orphaned listener on the adapter port before each run — a `dfs`
left behind by a killed run answers for a build nobody is testing and grades green.

## `test/conformance/run_test.sh`

33 assertions against a synthetic manifest and synthetic step scripts: no server,
no Docker, no mount, no root, about a second. It refuses to report success if it
asserted nothing.

Each assertion was checked against a deliberately broken runner rather than
assumed — four mutations, each caught by exactly the assertion that should catch it:

| Mutation | Caught by |
|---|---|
| collapse the exit status to a boolean | `a grader's failure count survives to the caller: expected '3', got '1'` |
| hoist root to the whole run | `a non-root step is not`, `no step of pynfs asks for root` |
| stop running teardown after a failure | `teardown runs after a failed step` |
| ignore the variant when picking the blacklist | `4.1 grades against its own table` |
| drop the timeout from the matrix cell | `every matrix cell carries a timeout` |
| put `badger` back behind the `*-fs` glob | `every SMB profile hits a bootstrap case arm` |
| point the grader's log back at `run.log` | `no suite runner writes the common runner's step log` |
| delete a profile's config file | `every SMB profile has a config file` |

## `.github/workflows/conformance.yml`

Replaces and deletes `posix-tests.yml`, `smb-conformance.yml` and
`nfs-kerberos.yml`. A `matrix` job resolves every suite's cells from the manifest
once; each suite job takes its `strategy.matrix.include` from that.

The suite jobs still differ in what they set up — Docker Compose, a kernel NFS
mount and root, a postgres and a localstack service — because the suites genuinely
differ there, and folding them into one job template would make every WPTS cell
pay for postgres, localstack and Nix. What is unified is the part that was
actually duplicated: the tiering, and the entrypoint.

`nfs-kerberos` keeps its path-scoped presubmit through an explicit
`dorny/paths-filter` job rather than by living in its own file, so an ordinary PR
still skips it.

## Profile names

`badger-fs` → `badger`, and `memory-fs` is deleted. Both were named for a
filesystem block store that no longer exists — the configs said so themselves
("legacy name, filesystem block store removed in Phase 42") — and `memory-fs.yaml`
was byte-identical to `memory.yaml` apart from that comment, so postsubmit spent a
whole job re-running the memory profile under a second name. `badger` is what the
POSIX and pynfs suites already call the same backend.

Historical prose in `KNOWN_FAILURES.md` is left alone: those runs really were
labelled `badger-fs`.

Review caught that the rename was only half-applied: `bootstrap.sh`'s
`create_metadata_store` matched on `badger*` and was fine, but its sibling
`create_block_stores` had matched the old name through a `*-fs` glob, so a bare
`badger` fell through to the error arm. That would have failed provisioning on
every PR, since `badger` is in smbtorture's presubmit tier. Fixed, and with the
legacy names gone there was nothing left for that arm to match, so it went too.

## `run-posix.sh --grade`

CI ran `prove` over every test and graded the log afterwards; `run-posix.sh`
skipped blacklisted tests up front. Two behaviours for one suite, and only one of
them can notice a known failure that started passing. `--grade` gives the script
the CI behaviour, so the common runner drives the same code path locally and in CI.

The grader's log is `prove.log`, not `run.log`. Review caught that the first
version derived it from `DITTOFS_RESULTS_DIR` as `run.log` — exactly where the
common runner tees the step named `run`. Two non-appending `tee`s on one path
interleave and truncate, and that log is what `parse-results.sh` reads to decide
the verdict, so it would have put the exit contract at risk for the one suite
whose grading step is named `run`.

## Docs

`check-docs.sh` renders the suite table into `docs/internals/testing.md` between
markers; `--write` regenerates, default checks. `contributing.md` links to it rather
than carrying a second rendered copy. It runs in `lint.yml`, not in `conformance.yml`, because the
conformance workflow ignores docs-only changes and so could never catch a stale
table. Verified it refuses: corrupting one cell of the generated table makes it
exit 1 naming the file.

## pynfs

pynfs keeps its own workflow file, because its tree is being reworked in parallel
and the suite should not start depending on a runner landing in the same window.
But `nfs-pynfs.yml` no longer hand-codes its tiering: it resolves its matrix from
`suites.json` and dispatches through `test/conformance/run.sh`, so the profile
ternary this PR deletes from three workflows is gone from the fourth as well.
Folding the file itself away is a one-entry follow-up.

Half-in was the one option worth avoiding — the manifest describing pynfs, the docs
rendering it, and the workflow independently restating the same profiles is exactly
the drift this PR exists to close.

No file under `test/nfs-conformance/` is touched. None of the four PRs currently
in flight against that tree touches `.github/workflows/nfs-pynfs.yml`.

## Not in this PR

**#1603 (xfstests) is not added.** It has no harness yet; the manifest is now the
place it would land.

## Verification

- `test/conformance/run_test.sh`: 33 assertions, each of the seven mutations above verified to fail the build.
- `check-docs.sh` is idempotent and refuses a hand-edited table.
- `shellcheck --severity=warning` clean on every changed script.
- All three new/changed workflow files parse.
- `test/smb-conformance/run.sh --profile badger --dry-run` resolves; `--profile memory-fs` is now refused.
- Nothing under `test/nfs-conformance/` is modified (`git diff --name-only` on that path is empty).

The suites themselves have not been run end to end on this branch — they collide
on ports, mounts and sudo with other work in flight. CI on this PR is the first
full execution of the new workflow.

---

**Deliberately does not close #2322.** Four of the five suites are migrated
here. The pynfs suite is not: its directory has four PRs in flight from another
session, so `test/nfs-conformance/` was held for the duration and this PR
contains no file under it. The manifest is nonetheless shaped for pynfs —
`known_failures` is a per-variant map rather than a single path, because a
one-path-per-suite schema cannot express V40 and V41 at all, and that had to be
right before the migration rather than after.

The pynfs migration lands separately once that queue drains, and #2322 stays
open until it does.


marmos91:
All three addressed in 1ba58753c.

**Schedule tiering (both workflows).** Correct — every non-PR event mapped to \`push\`, so the manifest's \`schedule\` tier was unreachable. Both now pass \`schedule\` through. Worth noting for whoever reads this later: the two resolve to the same 29 cells today (\`--matrix push\` and \`--matrix schedule\` both return 29), which is exactly why nothing surfaced it — a declared key nothing can reach is the same drift this PR is closing, so it is fixed rather than deleted.

**\`clear_orphan_server\`.** The strongest of the three, and I took it further than suggested. Non-interactive sudo still kills a process the harness has no business killing, so sudo is gone entirely rather than made non-interactive. It now identifies the listener by process name and stops only a leftover \`dfs\`; anything else on the port makes it refuse the run with exit 2. Refusing is the safe answer here — proceeding against an unknown server is precisely the wrong-build-grades-green failure this cleanup exists to prevent. \`lsof\` missing is now a no-op, as you suggested.

Also dropped the \`DITTOFS_NFS_PORT\` override while I was in there: nothing in the tree sets or reads it, so it was a knob I had invented. The replacement exists so the test can bind a port of its own instead of fighting whatever is really on 12049.

Verified rather than assumed — I ran the new assertions against deliberately broken builds:

| mutation | caught by |
|---|---|
| drop the identity check (kill anything on the port) | `a listener that is not a dfs stops the run`, `the refusal names the port`, **`the unrecognised listener was killed`** |
| skip the cleanup entirely | `a listener that is not a dfs stops the run`, `the refusal names the port` |

That third assertion is the one that matters: it holds a port with a non-dfs process and checks the process is *still alive* after the refusal, so the guard is proven to protect rather than merely to exist. 37 assertions total.

PR #2422, opened 2026-09-08
chore(block): delete the dead hash-keyed CAS surface

marmos91:
Removes the hash-keyed CAS surface — **2,919 net lines** across 25 files. No production behaviour changes: nothing in the tree called it.

## Why it is dead

`block.Store` — the hash-keyed contract — has **no production consumer**. Every reference was a doc comment, a `var _` compile-time assertion, or the conformance suite. `go build ./...` passes with the whole surface removed, which is the check that settles it.

Deleted: the `block.Store` interface, `cas_key.go` (`FormatCASKey`/`ParseCASKey`/`ErrCASKeyMalformed`), `blockstoretest.BlockStoreConformance`, `remote.CASInner`, `Passthrough.Has`/`Delete`, the backends' `cas.go` (memory + s3), the decorators' `cas.go` and their hash-keyed methods (`Put`/`Get`/`GetRange`/`Head`/`Walk`/`plaintextSizeFor`), and `memory.Store`'s now write-never `blocks` map with `BlockCount`/`TotalSize`.

Untouched and still live: the block-keyed `remote.RemoteBlockStore` + `RemoteBlockStoreConformance`, `SealChunk`/`ReadChunk`, `s3/verifier.go`, and `ContentHash.CASKey()`.

## The name trap

`ContentHash.CASKey()` and `FormatCASKey` share a stem and are unrelated. The first returns `blake3:{hex}` and is on a live serialization path and in `s3/verifier.go`; the second is the dead `cas/{hh}/{hh}/{hex}` object layout. Deleting by name match would have taken out live code.

## Coverage was moved, not dropped

The rule applied throughout: delete a test whose **subject** died; **rewrite** one that merely *reached* live behaviour through a deleted helper. Highlights:

- **`gc_test.go`** — proved the GC sweep removed object *bytes* from the remote. Now resolves the locator before the sweep and probes `GetBlock(loc.BlockID)`; post-#1493 the remote object *is* the packed block, so this is the exact block-keyed successor. The file's existing `chunkOnRemote` helper was deliberately **not** used — it consults only metadata, which would have silently downgraded the assertion from "bytes gone" to "metadata says gone". Non-vacuity confirmed with a temporary probe: present before the sweep, and inverting the post-sweep check fails.
- **`snapshot_encryption_integration_test.go`** — the old check only ever showed the *decrypted* side. Split into a raw `GetBlock` asserting no plaintext at rest plus a `ReadChunk` round-trip, which is strictly stronger and actually tests the property the name claims.
- **Decorators** — compression/encryption round-trip, nonce uniqueness, tamper-fails-auth, unframed-block rejection and the `sealLayer` alloc bound all retargeted onto `SealChunk`/`ReadChunk`.
- **s3 mock** — `WalkBlocks` error-wrap and mid-pagination context-cancel were previously covered only via hash-keyed twins; now covered directly. One old assertion was found **vacuous** (the `Prefix: blocks/` filter meant the CAS key it claimed to skip was never listed) and now seeds the object map to exercise both skip branches.
- **s3 endpoint cell** — the env-gated real-endpoint test was the only non-mock S3 coverage, so it was rewritten onto `RemoteBlockStoreConformance` rather than dropped.

## Verification

`gofmt` clean, `go build ./...` clean, `go vet ./...` clean, and `go test ./... -count=1` is **137 packages ok, 0 FAIL**.


marmos91:
Copilot's finding is a false positive and this merges without acting on it.

It calls `for i := range writers` compile-breaking because `writers` is an `int` constant. Ranging over an integer has been legal since **Go 1.22**; this module declares `go 1.25.0`. Checked rather than assumed: `go vet ./pkg/block/encryption/` passes, all 44 checks are green, and this repo's `.golangci.yml` explicitly disables the `intrange` linter with the comment *"Disable 'range over int' modernization suggestions"* — the idiom is deliberate here, not accidental.

PR #2431, opened 2026-09-08
refactor(engine): rename Syncer to RemoteSync, freeing the syncer name

marmos91:
Stacked on #2430. Step 1 of the block data-flow refactor (#2428) extracts a `syncer` package, and `engine.Syncer` already occupies the name — 1074 LOC, of which carve dispatch is about 10%. A new package sitting beside a type of the same name is a permanent source of confusion, so the type gives way.

Mechanical and compiler-checked: `gopls rename` on the type, `NewSyncer`, `SyncerConfig` and the `BlockStoreConfig.Syncer` field, then a sweep of the comments and log strings gopls does not touch. 62 files, 238 lines, no behaviour change.

**The user-facing `syncer:` config section is untouched.** `shares.SyncerDefaults`, `SetSyncerDefaults` and `buildSyncerConfigFromDefaults` keep their names deliberately: they sit against the YAML key `syncer:`, where the word stays correct. `SyncerDefaults` carries no yaml/mapstructure tags, so nothing here reaches a config file either way.

`go build ./...`, `gofmt`, and `go test ./pkg/block/... ./pkg/controlplane/...` all green.

### DIFF
--- a/internal/adapter/common/clone_localonly_tail_test.go
+++ b/internal/adapter/common/clone_localonly_tail_test.go
@@ -26,11 +26,11 @@ func newLocalOnlyTestEngine(t *testing.T, coord *fakeCoordinator, ms *metadatame
 	if !ok {
 		t.Fatalf("metadata store %T does not implement metadata.SyncedHashStore", ms)
 	}
-	syncer := engine.NewSyncer(localStore, nil, ms, engine.DefaultConfig())
+	syncer := engine.NewRemoteSync(localStore, nil, ms, engine.DefaultConfig())
 	syncer.SetSyncedHashStore(syncedHashStore)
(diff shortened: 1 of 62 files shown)


marmos91:
Verified all seven Copilot notes. Four were real and are fixed in bd7f815f — the test name, the two error strings ("syncer is required", "syncer is closed"), and the NewSyncer comment in remove_share_teardown_test.go. Also fixed two sibling comments in syncer.go carrying the same staleness.

Three do not reproduce at this head: doc.go contains no NewSyncer (it refers to the syncer component in prose, not the constructor), SetSyncedHashStore already says NewRemoteSync twice, and buildSyncerConfigFromDefaults already documents engine.RemoteSyncConfig. SyncerDefaults keeps its name deliberately — it is the config surface, not the engine type.

Issue #2433, opened 2026-09-08
golangci-lint is red on develop: 6 spurious SA5011 nil-deref findings behind t.Fatal guards


marmos91:
`golangci-lint` fails on **develop** and therefore on every open PR. It is not caused by any one PR — I confirmed the findings are byte-identical on develop and on an unrelated PR (#2425).

## The failure

```
internal/adapter/common/copy_payload_test.go:275:5: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
internal/adapter/common/copy_payload_test.go:278:9: SA5011: possible nil pointer dereference (staticcheck)
internal/adapter/nfs/auth/unix_test.go:121:5: SA5011(related information): this check suggests that the pointer can be nil (staticcheck)
internal/adapter/nfs/auth/unix_test.go:124:12: SA5011: possible nil pointer dereference (staticcheck)
internal/adapter/nfs/auth/unix_test.go:127:12: SA5011: possible nil pointer dereference (staticcheck)
internal/adapter/nfs/auth/unix_test.go:158:5: SA5011: (related information) ...
```

Every site is the same shape — a nil guard that terminates, followed by a dereference:

```go
if result == nil {
    t.Fatal("expected non-nil result")
}
if result.User == nil {   // SA5011 flags this
```

`t.Fatal` calls `runtime.Goexit`, so the dereference is unreachable when the pointer is nil. staticcheck normally knows this. These are **false positives**, which is why they smell like an analysis-environment problem rather than a code defect.

## Runs

| SHA | golangci-lint **job** |
|---|---|
| `09d60341` (develop) | success |
| `8aa32ed7` (develop, the go-deps bump #2419) | success |
| `4c87a857` (develop, docs-only) | **failure** |
| `0eddb4d0` (PR #2425) | **failure**, twice |

Note the job, not the workflow — the Lint workflow has several jobs and a workflow-level "success" can hide a skipped or failing one.

## What makes this odd

`8aa32ed7` passed and `4c87a857` failed, but the two commits between them are **docs-only** and the flagged test files are byte-identical across all of these SHAs. The same Go code therefore lints both green and red depending on when it ran, so the outcome depends on the analysis environment, not the source.

Best remaining hypothesis: the `golangci-lint-action` cache. `8aa32ed7` ran right after the 9-package dependency bump and likely reused a warm pre-bump cache; every run since analyses fresh and surfaces the findings. That would mean #2419 exposed this rather than introduced it. Unverified.

[...]

PR #2435, opened 2026-09-08
[probe, do not merge] #2433: is the lint analysis cache the SA5011 variable?

marmos91:
Draft probe for #2433, not for merge. Branched from develop at 4c87a857b (which is itself red on golangci-lint) and changes exactly one thing: `skip-cache: true` on the golangci-lint action.

**Prediction:** if the restored analysis cache is what makes staticcheck emit the six false-positive SA5011 findings, this goes green. If it stays red, the cache is exonerated and the cause is elsewhere.

Worth stating why this needed a probe: the obvious "green run had a cold cache, red runs had a hit" story does not hold. Both restored the *same* cache content:

| run | primary key | restored |
|---|---|---|
| GREEN 8aa32ed7 | missed, restore-key fallback | `dc24b537` |
| RED 4c87a857 | exact hit `dc24b537` | `dc24b537` |
| RED #2425 | exact hit `dc24b537` | `dc24b537` |

Same content, opposite outcomes — so cache content alone cannot be the explanation, and this isolates the restore step itself.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -78,8 +78,19 @@ jobs:
         run: ./test/common/known-failures_test.sh
 
   golangci-lint:
-    name: golangci-lint
+    name: golangci-lint (cache=${{ matrix.cache }} r${{ matrix.rep }})
     runs-on: ubuntu-latest
+
+    # Probe for #2433. The failure is non-deterministic: the same source, the
+    # same restored cache and the same linter build have come back both green
+    # and red. A single run therefore proves nothing either way, so both arms
+    # run side by side on one commit -- same source, same runner image -- and
+    # the only thing that differs is whether the analysis cache is restored.
+    strategy:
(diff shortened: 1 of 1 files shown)


marmos91:
Superseded by #2443, which carries the real fix.

The probe did its job: `skip-cache: true` as the only change, off the red `4c87a857b`, came back green. But the decisive evidence turned out not to be the sampling — it was the cache **keys and sizes**, which show the failure is deterministic rather than flaky. The greens restored `48cdea3b` (2,674,121 B); the reds get an exact hit on `dc24b537` (2,170,079 B), an incomplete fact set saved by a post-bump run. Full mechanism in #2443 and #2433.

I cancelled this branch's matrix and conformance runs rather than let them finish: they were confirmation of something already established mechanistically, and they were competing for runners with the fix that unblocks develop and seven waiting PRs.

Issue #2423, opened 2026-09-08
block: upload concurrency is a product of two windows, and the adaptive controller samples the wrong one


marmos91:
Found while mapping Step 1 (syncer extraction) of the block data-flow refactor. Independent of that refactor — this is live behaviour on `develop`.

## The invariant the code claims

`pkg/block/engine/carve_dispatch.go:56-58`:

> Concurrency is bounded by the adaptive upload window: the loop acquires `uploadLimiter` before starting each file's carve and releases it when that file's blocks are committed, **so at most Limit() carves — and thus block PUTs — are in flight.**

The "and thus block PUTs" step does not hold.

## What actually bounds PUTs

Two independent semaphores, nested:

| layer | site | bounds | value |
|---|---|---|---|
| outer | `engine/carve_dispatch.go:98` `Syncer.uploadLimiter` | whole-file **carve passes** | adaptive 16→64 |
| inner | `journal/carve.go:395` `sem := make(chan struct{}, s.cfg.CarveUploadConcurrency)` | **block PUTs within one carve** | fixed, default 8 (`journal/store.go:100`) |

The outer limiter admits a *file*; that file then opens its own 8-slot PUT window. Concurrent `PutBlock` is therefore the **product**: **128 at the adaptive floor, 512 at the ceiling.**

`engine/types.go:33` declares `MaxParallelUploads = 256` as the "upper bound on a pinned ParallelUploads window". Nothing enforces it against the product, and the ceiling exceeds it by 2x.

## The adaptive controller measures the file window, not the PUT window

`engine/syncer.go:802-803`:

```go
peak := m.uploadLimiter.TakePeak()
windowLimited := peak >= m.uploadLimiter.Limit()
```

`TakePeak()` is the peak of the **outer** limiter — files in flight. So draining one large file gives `peak == 1`, `windowLimited == false`, and `goodputController.observe` reads the workload as *app-limited* and holds the window — while up to 8 PUTs per file are actually saturating the link. The controller is blind to the semaphore that is really binding.

`TestGoodputController_HoldsWindowWhenAppLimited` and `TestGoodputController_HoldsOnErrorWhenAppLimited` pin the hold-on-app-limited behaviour, so the controller is behaving as designed — the design is reading the wrong signal.

## Two more paths where the outer window does not apply at all

[...]

PR #2438, opened 2026-09-08
refactor(block): extract the upload-window mechanism into pkg/block/syncer

marmos91:
Step 1.3 of the block data-flow refactor. Stacked on `refactor/engine-remotesync-rename` (#2431).

Creates `pkg/block/syncer` and moves the upload-window mechanism into it — the adaptive semaphore and the goodput controller. These are the only parts of step 1 that port cleanly, and both were missing from the plan's original scope.

## What moved

| from | to |
|---|---|
| `engine/dynsem.go` (+ `_test.go`) | `syncer/dynsem.go` |
| `engine/upload_controller.go` (+ `_test.go`) | `syncer/upload_controller.go` |

Nothing else. `engine/carve_dispatch.go` is step 1.4; `journal/carve_dispatch.go` stays in `journal` for step 1; `engine/blocksink.go` stays.

Both moved files had zero imports beyond `context`/`sync`/`math`, so there is no dependency back into `engine` and no cycle.

## Exported surface

Kept to exactly what `engine` calls across the new boundary:

- `syncer.DynamicSemaphore` / `NewDynamicSemaphore` — methods `Acquire`, `Release`, `TakePeak`, `SetLimit`, `Limit` were already exported; `InFlight` was too and is used by the moved test.
- `syncer.GoodputController` / `NewGoodputController` — plus `Observe` (was `observe`), called from `syncer.go:809`.

`window()`, `grow()` and `shrink()` stay unexported: only the moved tests use them.

## `MaxParallelUploads`: kept in `engine`

The whole const block in `engine/types.go` stays put, `MaxParallelUploads` included. It is validation for `SyncerConfig.ParallelUploads` — the engine's *configuration* — not the window mechanism: `NewGoodputController(floor, ceiling)` and `NewDynamicSemaphore(limit)` take their bounds as parameters and reference no constant. Moving `MaxParallelUploads` would drag `blockstore_init.go` and `blockstore_config.go` into an extraction PR for no gain, and splitting the block (moving `AdaptiveUploadFloor`/`Ceiling` but not `Default`/`Max`) would separate four constants that document one setting together. `uploadControlInterval` stays for the same reason: the control *loop* (`runUploadController`) stays in `engine`, so its cadence does too.

## Behaviour

Unchanged. Same logic, same defaults, same semantics. The two nested upload semaphores are carried over verbatim — collapsing or retuning them is #2423 and ships separately.

## Comments

- Corrected the adaptive-bounds comment at `engine/types.go:22-30`: it claimed the window bounds concurrent block PUTs. It bounds concurrent whole-file carve passes; each pass applies its own bound on PUTs, so the PUTs in flight are the product of the two. The same false claim was in `upload_controller.go`'s type doc and is corrected in the moved copy rather than propagated.
- Fixed a stale note in `Acquire`: "one Acquire per chunk" — it is one Acquire per carve pass.
- Stripped issue numbers from the comments in the two new files, per the repo rule against external references in source. Left the pre-existing ones in `engine/types.go` alone; that is a separate cleanup.

Two comments known to carry the same PUT-bounding falsehood are **left untouched** because they belong to files this PR does not own: `engine/syncer.go:117` ("bounds concurrent block PUTs in carveFlush", and it also names a `ensureUploadLimiter` that does not exist) and `engine/carve_dispatch.go:56-58`. Both are step 1.4 / #2423 territory.

## Verification

| command | result |
|---|---|
| `go build ./...` | pass |
| `go vet ./pkg/block/...` | pass |
| `go vet -tags=integration ./...` | pass |
| `go test -count=1 ./pkg/block/... ./pkg/controlplane/...` | pass, 29 packages ok |
| `go test -count=1 -race ./pkg/block/syncer/...` | pass, all 16 tests |
| `gofmt -l ./pkg ./internal ./cmd` | empty |
| `golangci-lint run ./pkg/block/...` | 0 issues |

The 16 tests are the ones that moved: 6 `TestDynamicSemaphore_*` and 10 `TestGoodputController_*`. No test was added — this PR adds no contract, so there is nothing new to pin.

### DIFF
--- a/pkg/block/engine/carve_dispatch_test.go
+++ b/pkg/block/engine/carve_dispatch_test.go
@@ -11,6 +11,7 @@ import (
 
 	"github.com/marmos91/dittofs/pkg/block/journal"
 	"github.com/marmos91/dittofs/pkg/block/local"
+	"github.com/marmos91/dittofs/pkg/block/syncer"
 )
 
 // carveFanoutLocal is a minimal LocalStore that records per-file Carve calls and
@@ -53,7 +54,7 @@ func TestCarvePass_FansOutBoundedByUploadWindow(t *testing.T) {
 	const window = 3
 	m := &RemoteSync{
 		local:         fl,
-		uploadLimiter: newDynamicSemaphore(window),
+		uploadLimiter: syncer.NewDynamicSemaphore(window),
 		stopCh:        make(chan struct{}),
 		config:        DefaultConfig(),
 	}
@@ -91,7 +92,7 @@ func TestCarvePass_FansOutBoundedByUploadWindow(t *testing.T) {
 // TestCarvePass_NoFilesIsNoop guards the empty working-set path.
 func TestCarvePass_NoFilesIsNoop(t *testing.T) {
 	fl := &carveFanoutLocal{started: make(chan string, 1), release: make(chan struct{}), carved: map[string]int{}}
-	m := &RemoteSync{local: fl, uploadLimiter: newDynamicSemaphore(4), stopCh: make(chan struct{}), config: DefaultConfig()}
+	m := &RemoteSync{local: fl, uploadLimiter: syncer.NewDynamicSemaphore(4), stopCh: make(chan struct{}), config: DefaultConfig()}
 	m.carvePass(context.Background()) // returns immediately, acquires nothing
 	require.Equal(t, int32(0), fl.inFlight.Load())
 }
--- a/pkg/block/engine/syncer.go
+++ b/pkg/block/engine/syncer.go
@@ -115,11 +115,14 @@ type RemoteSync struct {
 	completedSyncs atomic.Int64
 	failedSyncs    atomic.Int64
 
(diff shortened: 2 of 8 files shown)


marmos91:
Correct, and worse than described — the comment named three things that do not exist. carveFlush is not a function anywhere in the tree, ensureUploadLimiter appeared only inside that comment (NewRemoteSync builds the limiter eagerly), and the "directly-built test fixtures" rationale is wrong since the fixtures construct it explicitly.

Rewritten in 456d60e8 to say what it actually bounds: carveDispatcher acquires it per whole-file carve pass, while the PUTs inside a pass are bounded separately by the per-file semaphore sized by CarveUploadConcurrency in journal/carve.go. In-flight PUTs are the product of the two, which is the concurrency boundary this comment was steering readers away from.

Out of scope here but worth an issue: three other comments (engine/syncer.go, engine/locator_fetch_test.go, dfsbench/backend/dittofs.go) still cite carveFlush. That predates this branch.

PR #2452, opened 2026-09-08
fix(metadata): make the xattr read-modify-write atomic

### LINKED ISSUE #2454: Concurrent SETXATTR silently loses xattrs: the read-modify-write is not atomic
Concurrent `SETXATTR` on one file silently loses values on every backend.

`ResolveSetXattr` / `ResolveRemoveXattr` (`pkg/metadata/xattr.go`) load the file, apply one
mutation to `FileAttr.EAs` and persist the whole map. Nothing wraps that pair in a transaction
on the store path, so two writers naming *different* attributes each load before either
persists and the later write reinstates a pre-image without the other's name. **The call
returns `nil`** — no error reaches the client.

Reachable in production: `Service.SetXattr` passes the store (not a transaction) to the
resolver, and the NFSv4 handlers call it (`setxattr.go:113`, `removexattr.go:79`).

Measured with sixteen writers and sixteen distinct names on one file:

| backend | reported written | absent afterwards |
| --- | ---: | ---: |
| badger | 16 | 14 |
| badger (relaxed durability) | 16 | 15 |
| sqlite | 16 | 13 |
| memory | 16 | 0 (serialises by timing, same bug shape) |

Predates the `store/sql` unification; not a refactoring regression.

Fix in #2452: run the resolvers inside a transaction when the target can open one — one site,
all four backends — plus `storetest` `XattrOps/ConcurrentDistinctNames`, which asserts that a
write reporting success is readable afterwards.

marmos91:
Closes #2454

## The bug

`ResolveSetXattr` / `ResolveRemoveXattr` are read-modify-writes over the **whole** EA map:
load the file, apply one mutation to `FileAttr.EAs`, persist the file. Nothing wraps that pair
in a transaction on the store path, so two writers naming *different* attributes on one file
each load before either persists, and the later write reinstates a pre-image missing the
other's name.

**The call returns `nil` and the value is gone.** No error reaches the client.

This is reachable in production: `Service.SetXattr` passes the *store* to the resolver
(`pkg/metadata/xattr.go`), and the NFSv4 handlers call it — `setxattr.go:113`,
`removexattr.go:79`. It predates the `store/sql` unification and affects all four backends.

## Measured, before the fix

Sixteen writers, sixteen distinct names, one file:

| backend | names reported written | absent afterwards |
| --- | ---: | ---: |
| badger | 16 | **14** |
| badger (relaxed durability) | 16 | **15** |
| sqlite | 16 | **13** |
| memory | 16 | 0 — see below |

## The fix, part one: a transaction

The resolvers run inside a transaction when the target can open one:

```go
tr, ok := files.(Transactor)
if !ok {
    return fn(files)
}
return tr.WithTransaction(ctx, func(tx Transaction) error { return fn(tx) })
```

`Transaction` does not include `WithTransaction` and no concrete transaction type defines one
(checked), so an in-transaction caller falls through unwrapped and cannot nest. Because every
backend's store method delegates to the shared resolver, this covers all four without touching
any of them.

## Test

`storetest` gains `XattrOps/ConcurrentDistinctNames`, asserting **a write that reported success
is readable afterwards** — not "all sixteen present". A backend may legitimately refuse a
colliding write (sqlite BUSY, a serialization failure) and the caller can retry; what none may do
is claim success and lose the value. It also fails if *every* write errored, so it cannot pass by
proving nothing.

**It does not discriminate on the memory store**, and the comment says so: sixteen writers queued
on one short critical section serialise by timing, so memory survives the non-atomic form by luck
rather than design. It had the same bug shape and is fixed by the same change.

## The fix, part two: postgres needs a row lock

**The transaction wrap alone does not fix postgres, and I claimed it would.** CI's integration
suite caught it: 13 of 16 names still lost with the wrap in place.

Under READ COMMITTED the second writer's `UPDATE` waits for the first to commit and then writes
attributes computed from the pre-image it already read. Nothing errors, nothing retries, the write
is lost. sqlite and badger were never fixed by the wrap either — they were already correct for a
different reason: each *refuses* the second writer (`SQLITE_BUSY`, an SSI conflict) and the retry
loop re-runs the body, so the retried attempt re-reads.

So `metadata.FileRowLocker` is an optional interface a `Transaction` implements when it needs an
explicit row lock for that pair to serialise. Only the postgres transaction does, taking the row
with `SELECT 1 FROM inodes WHERE id = $1 AND share_name = $2 FOR UPDATE` before the read — the
same shape `UpdateAttrs` already uses in its leading CTE. It is taken first and covers one row, so
it introduces no lock-ordering hazard.

## Verification

- Fails on badger + sqlite before the change, passes after.
- Postgres, against a local postgres 16: **14 of 16 names lost with `LockFileRow` removed**, green
  with it. The lock is load-bearing, not decoration.
- Full postgres integration suite (`-tags=integration`, 78s): green.
- `-race` on the xattr suite across memory/badger/sqlite: clean.
- Full `./pkg/metadata/store/...` conformance green (badger, memory, sqlite, postgres units, sql).
- `golangci-lint run pkg/metadata/...`: 0 issues.

All four backends verified, postgres included — see below.

## Not in scope

`xattr.go` is still duplicated verbatim between the two dialect packages (52/52, receiver names
only), as are `clients.go` and `client_recovery.go` — about 190 lines. Merging them onto `Core`
was previously hazardous precisely because these bodies span more than one statement; now that
the resolver owns the transaction, that objection is gone. Left as a follow-up.

### DIFF
--- a/pkg/metadata/store.go
+++ b/pkg/metadata/store.go
@@ -366,6 +366,24 @@ type Transactor interface {
 	WithTransaction(ctx context.Context, fn func(tx Transaction) error) error
 }
 
+// FileRowLocker is implemented by a Transaction that needs an explicit row lock
+// for a read-modify-write of one file's attribute record to serialise against
+// another writer.
+//
+// A backend whose transaction already serialises such a pair does not implement
+// it: sqlite and badger both refuse the second writer (SQLITE_BUSY, an SSI
+// conflict) and their retry loop re-runs the whole body, so the retried attempt
+// re-reads. Postgres under READ COMMITTED does neither — the second UPDATE
+// waits for the first to commit and then writes a value computed from the
+// pre-image, losing the first write with no error — so it takes the lock before
+// the read instead.
+type FileRowLocker interface {
+	// LockFileRow blocks until this transaction holds the file's row, and
+	// reports nil when the handle names no row: the caller's own read is what
+	// turns that into ErrNotFound.
+	LockFileRow(ctx context.Context, handle FileHandle) error
+}
+
 // RelaxedTransactor is an OPTIONAL store capability (#1573 Wall 1): running a
 // transaction whose commit may become durable with bounded lag instead of an
 // inline fsync. It is intended ONLY for pure-namespace/attr writes
--- a/pkg/metadata/store/postgres/transaction.go
+++ b/pkg/metadata/store/postgres/transaction.go
@@ -388,3 +388,40 @@ func (tx *postgresTransaction) SetFilesystemCapabilities(capabilities metadata.F
 // ============================================================================
 // Transaction Files Operations (additional)
 // ============================================================================
+
+// LockFileRow implements metadata.FileRowLocker so a read-modify-write of one
+// inode's attributes serialises.
+//
+// Postgres runs this transaction at READ COMMITTED, where a bare read followed
+// by an UPDATE loses a concurrent writer's change without reporting anything:
+// the second UPDATE waits for the first to commit and then writes attributes
+// computed from the pre-image it read earlier. Taking the row here, before that
+// read, makes the second transaction wait at the lock instead, so its read sees
+// the committed value. sqlite and badger need no equivalent — they refuse the
+// second writer and their retry re-reads.
+//
+// No rows come back when the handle names nothing; the caller's read reports
+// that as ErrNotFound.
+func (tx *postgresTransaction) LockFileRow(ctx context.Context, handle metadata.FileHandle) error {
+	if err := ctx.Err(); err != nil {
+		return err
+	}
+
+	shareName, id, err := metadata.DecodeFileHandle(handle)
+	if err != nil {
+		return &metadata.StoreError{
+			Code:    metadata.ErrInvalidHandle,
+			Message: "invalid file handle",
+		}
+	}
+
+	const lockQuery = `SELECT 1 FROM inodes WHERE id = $1 AND share_name = $2 FOR UPDATE`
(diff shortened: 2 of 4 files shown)


marmos91:
Both review points looked at; one is wrong, one is worth pinning.

**`for i := range writers` does not compile — no.** Range-over-int has been legal since Go 1.22 and this module is `go 1.25.0`. It compiled and ran on all four backends in this PR's own CI, including the integration job. Not changing it.

**Nested transactions — no type in the tree is both, but the invariant was unenforced.** The wrap keys off `Transactor`, and no backend transaction carries `WithTransaction`, so an in-transaction caller falls through. I checked that rather than assuming it, and also checked the inverse before considering the suggested guard: a real store does *not* satisfy `metadata.Transaction` (it lacks `ListFileChunks`/`GetFileChunk` among others), so a `files.(Transaction)` skip would have been safe — it just wouldn't have protected anything today. The doubles cited as already satisfying both are store doubles, not transactions.

What was genuinely missing is enforcement, so `eacaea531` adds `XattrOps/TransactionIsNotATransactor` instead of a runtime branch: it opens a transaction on each backend and asserts the value does not satisfy `Transactor`. A transaction that grew the method later (savepoints, say) would otherwise silently nest every in-transaction `SetXattr`, and nesting is implementation-defined per the `Transactor` doc — a deadlock or a dropped write, not a build error. Verified it fires by adding `WithTransaction` to the memory transaction.

That keeps the hot path free of a branch that can never be taken, and moves the assumption somewhere it fails loudly.

PR #2455, opened 2026-09-08
test(engine): prove the ordered carve flip against the production block sink

marmos91:
Step 1.5 groundwork. The plan prescribes porting `journal`'s `TestCarveFlipsInWatermarkOrder` first, as the spec for the completion contract being added. This is that port.

## Why it is not redundant with the journal test

Journal proves the watermark-ordered flip against `ctrlSink`, an in-package fake. The plan's own trap list flags these fakes as ones that will not survive the refactor — specifically `realSink`, whose nil `commitLocks` exercise a different locking path than production. So the ordering property has never been proven through `engineBlockSink`: real chunk sealing, real `blockcodec` framing, a real `PutBlock`, and the per-file metadata commit serialized under `carveCommitLocks`.

## The property

Blocks commit concurrently and in any order, but a record flips synced only once every lower-offset block has committed. A later block that lands first must not credit durability the earliest block has not yet earned — the `#1872` shape.

The gate sits in front of the real sink rather than replacing it, so a released block runs the whole production path. It is placed where it works: `commitAndFlip` calls `CommitBlock` *before* waiting on `prev`, so holding inside the sink keeps blocks in flight without stalling chain formation.

## Verified by making it fail

Replacing `proceed := <-prev` with an unconditional `proceed := true` in `journal/carve_dispatch.go` fails it:

```
records flipped before the earliest block committed: unsynced=0 want 131072
```

Reverted; the diff is test-only.

## Two things found while writing it

- The wrapper forwards all three optional `BlockSink` capabilities, with `var _` assertions. Journal discovers them by type assertion, so a wrapper that dropped them would silently exercise a carve without the reap, the row-widen or the clobber guard.
- A block boundary is only taken once a whole chunk has crossed `CarveBlockSize`, so with default `ChunkParams` a small file is one chunk and therefore one block. `newCarveFixture` does not expose `ChunkParams`; rather than widen it for one caller, the fixture is repeated locally with chunks sized under the block.

## Checks

`go build`, `go vet -tags=integration`, `gofmt` clean. Full suite exit 0, 0 FAIL, 138 packages. `go test -race ./pkg/block/engine/` exit 0, no races.

### DIFF
--- a/pkg/block/engine/carve_ordered_completion_test.go
+++ b/pkg/block/engine/carve_ordered_completion_test.go
@@ -27,7 +27,11 @@ import (
 type gatedSink struct {
 	real engineBlockSink
 
-	arrivals chan int64
+	// arrivals reports each block as it reaches the gate; committed reports each
+	// one after the real sink has finished with it. Both are keyed by the
+	// block's first file offset.
+	arrivals  chan int64
+	committed chan int64
 
 	mu    sync.Mutex
 	gates map[int64]chan struct{}
@@ -41,7 +45,12 @@ var (
 )
 
 func newGatedSink(real engineBlockSink) *gatedSink {
-	return &gatedSink{real: real, arrivals: make(chan int64, 64), gates: map[int64]chan struct{}{}}
+	return &gatedSink{
+		real:      real,
+		arrivals:  make(chan int64, 64),
+		committed: make(chan int64, 64),
+		gates:     map[int64]chan struct{}{},
+	}
 }
 
 // gate returns the release channel for the block whose first chunk sits at off,
@@ -73,7 +82,9 @@ func (g *gatedSink) CommitBlock(ctx context.Context, chunks []journal.CarveChunk
 	case <-ctx.Done():
 		return ctx.Err()
 	}
-	return g.real.CommitBlock(ctx, chunks)
+	err := g.real.CommitBlock(ctx, chunks)
+	g.committed <- off
+	return err
 }
 
 func (g *gatedSink) ReapSupersededManifest(ctx context.Context, id journal.FileID, spans [][2]int64, newOffsets map[int64]struct{}) error {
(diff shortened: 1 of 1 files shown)


marmos91:
Both fixed in e88b762c — the second one was a real stall, not just a style point.

The timed drain collected arrivals for 300ms and then released everything it had seen. A block reaching the sink after that window would never have its gate released, so the carve would block forever on it and the test would hang to the package timeout. Replaced with a background releaser that frees every block except the chain head as it arrives, so arrival timing cannot deadlock it regardless of how late a block shows up.

The sleep-based assertion is gone too. It now waits for the released blocks to come back out of the real sink (a committed signal fired after real.CommitBlock returns) before asserting nothing has flipped — so it can no longer pass while those blocks were merely still uploading, which was your point about it not exercising the property. Blocks that arrive after that point are still released and still commit, and none of them can flip while the head is held, so the assertion stays valid rather than racing.

Context is now per-test with a 90s bound instead of Background.

Re-verified by breaking it: with `proceed := <-prev` replaced by an unconditional true in journal/carve_dispatch.go, 3/3 runs fail, at unsynced=65536 and unsynced=0 on different runs — the nondeterminism an unordered flip should show. Reverted. 5/5 clean runs with ordering intact, and it went from 0.92s to 0.13s now the sleeps are gone.

go test -race on the package: exit 0, no races. Full suite exit 0, 138 packages.

marmos91:
This one is already addressed at this head — e88b762c line 144 is `ctx, cancel := context.WithTimeout(context.Background(), 90*time.Second)` with `defer cancel()`, and it is the only `context.Background()` in the file. The bare `context.Background()` the first review flagged is gone; what remains is the inner call being wrapped by the timeout, which is the fix rather than the defect.

A blocked CommitBlock now returns ctx.Err() at the gate, the releaser goroutine exits on ctx.Done(), and the two wait loops have their own 30s bounds, so no path hangs to the package timeout.

PR #2478, opened 2026-09-08
fix(journal): record which writer made a cold entry, and stop restore guessing

### LINKED ISSUE #2260: journal: coldEntry.version means two different things depending on who wrote it, and RestoreToVersion reads it as one
`coldEntry.version` means two different things depending on which code path wrote the entry, and nothing in the record says which.

```go
type coldEntry struct {
	id      FileID
	fileOff int64
	length  int64
	version uint64
}
```

- **Eviction** (`evictSegment`, `reclaim.go:264`) stamps `version: fi.ivs[k].version` — the interval's own version, preserved. A range written at `V-10` and evicted at `V+5` carries `V-10`. Here the field means *the content existed as of this version*.
- **Seeding** (`planColdSeed`, `store.go:563`) stamps `version: s.nextVersion()` — a fresh version minted at seed time. Here it means *we observed this range remote-durable at this moment*. `SeedColdBatch`'s own `ponytail:` comment explains why the version exists at all: so a racing `Delete` sweeps at a version above the pending entry. It was never intended as a claim about the content.

## This is not drift — both were introduced together

Commit `3fb840ad4` (#1851, which created `cold.log` to fix #1850) added both writers in the same change. The eviction path had an interval to copy a version from; the seed path had none and needed something monotonic for `Delete`-sweep ordering, so it minted one. Nothing reconciled them because **until the fix for #2227, no reader consumed the field across both provenances.**

## Why it matters

`RestoreToVersion`'s ceiling replay skips cold entries above the target version (`e.version > v → continue`). That is correct for the eviction provenance. For the seeded provenance it compares against seed time, which is unrelated to when the content existed, so every seeded entry older than the restore target is skipped and its ranges come back as holes — #2227's exact symptom.

**Neither branch of the guard is correct for a seeded entry.** Skipping over-skips as above. Including unconditionally resurrects content written after `V`, because a manifest seed describes the file's *current* content, not its content at `V`.

## Reachability — confirmed, off the default path but not hypothetical

No seeded entry can exist on a share that was only ever local-only: `seedColdIfNeeded` returns early on `!remoteConfigured`, the restore-path seed at `snapshot.go:1598` is inside `if remoteVerify`, and `engine.Store.SeedCold` has no caller in the tree. But:

1. `internal/controlplane/api/handlers/shares.go:734-737` lets a share update explicitly clear the remote tier (`"" → nil`).
2. `deriveLocalStoreDir` keys the local store directory on the **share name**, so the journal and its `cold.log` survive that change.
3. The restore branch turns purely on `bs.RemoteStore() != nil`.

So: a remote-backed share takes a remote-path restore at `T`, seeding entries versioned `~T`; an operator clears the remote; the share returns local-only over the same journal; a restore to a snapshot with `V < T` skips every seeded entry. Requires a deliberate operator action plus a restore to a snapshot older than the last remote-path restore.

## A marker-based guard cannot fix this

`MarkColdSeeded` has exactly one non-test caller, `shares/legacy_verify.go:235`, on the share-add path. The restore-path seed at `snapshot.go:1598` never writes it. So the marker is **absent in precisely the dangerous case** — the #2110 shape, where a marker-based guard cannot see a missing marker.

## Fix shape

The entry must distinguish "this range held content as of version X" from "we observed this range remote-durable at time T". Two options were considered and one is already ruled out:

- **Seeds carry the manifest's content version — impossible.** `block.FileChunk` has no journal version: its fields are `ID`, `Hash`, `DataSize`, `StartOffset`, `RefCount`, `LastAccess`, `LastSyncAttemptAt`, `CreatedAt`, `State`. The wall-clock timestamps are not substitutes; a journal version is a per-store monotonic LSN and cannot be compared with `CreatedAt`. The manifest orders by content identity and time, the journal by LSN — unrelated version spaces.
- **A provenance discriminator in `coldEntry`** — the surviving option, and a change to a durable on-disk record. It needs its own migration story; see `.planning/2026-09-01-block-dataflow-MASTER-PLAN.md` §10.

## Notes

- Verified against `origin/develop` at `376cd1eb8`.
- The fix for #2227 is deliberately scoped to the eviction provenance, where the guard is correct, and marks this case at the fold with a `ponytail:` comment.
- No production exposure (customers are evaluating only), consistent with D4 of the master plan.

marmos91:
Closes #2260.

## Premise, checked first

Verified against develop before writing anything, and it holds — with one refinement:

| writer | version it stamps | what that dates |
| --- | --- | --- |
| `evictSegment` (`reclaim.go:266`) | the interval's own | the content |
| `Invalidate` (`store.go:1473`) | the interval's own | the content |
| `SeedCold` / `SeedColdBatch` via `planColdSeed` | `s.nextVersion()` | the scan |

The issue names two writers; there are four. `Invalidate` is the one it missed, and it falls on the
content-dated side, so the conclusion stands. `RestoreToVersion` read all of them as content-dated.
The `ponytail:` marker at the fold already named this and its fix, and is now removed rather than
reworded.

## What the record gains

A provenance byte, under magic `0xC1`. The old `0xC0` framing still loads, as `coldFromUnknown`,
which keeps the pre-existing behaviour for entries whose writer was never recorded.

**The interval index carries it too**, which is the half that is easy to miss: compaction rebuilds
the log from the index (`liveColdEntries`), not from the log, so an index without the field would
have quietly rewritten every entry as `unknown` at the first compaction and undone the whole change.
`TestLiveColdEntriesKeepsProvenance` pins that, and I checked it fails when the field is dropped.

There are **two** such paths, and Copilot caught the one I missed: `clamp` rebuilds an interval field
by field when a neighbouring write splits a cold range, and it dropped provenance the same way — so a
trimmed cold range reached disk as `unknown`. `without()` trims through `clamp`, so one line covers
both. `TestClampKeepsProvenance` pins it through `liveColdEntries`, also checked against a broken
build. I had found the first path and stopped auditing rather than enumerating every place an
interval is rebuilt.

## What restore does with it — narrower than my first attempt

My first version refused **any** seeded entry, on the issue's "neither branch is correct" reading.
Two existing cases refuted that:

- `TestRestoreToVersion_KeepsFullyColdFile` — a file seeded below V must survive the rewind.
- `TestRestoreToVersion_KeepsColdRangesOfMixedFile` — a seeded half overwritten **locally** after V
  must come back cold.

They are right, and they show the version test *is* meaningful for a seed: it is a true statement
about what the journal knew at V, even though it says nothing about the bytes. Refusing outright
would have broken restore on any store that ever seeded — an availability regression far wider than
the defect.

So the version test stays for both kinds. What is refused is the case it genuinely cannot cover:

> a seeded range that a post-V write overwrote **and synced**

There the remote copy the entry points at *is* post-V content, so serving it returns a state the
caller did not ask for while reporting success. An unsynced post-V write does not qualify — its
bytes never left the local tier, the restore is about to drop them, and the remote still holds what
the seed described. Testing the synced flag rather than the overlap is what lets the two cases above
keep passing; the overlap alone failed the second one.

New cases, both directions:

- `TestRestoreToVersion_RefusesSeededRangeOverwrittenOnRemote` — `Hydrate` post-V (records the range
  already synced) ⇒ `ErrColdProvenanceAmbiguous`.
- `TestRestoreToVersion_KeepsSeededRangeOverwrittenLocally` — `WriteAt` post-V ⇒ restore proceeds and
  the range comes back cold.

## The format bump, and why it was not enough on its own

`formatVersion` 1 → 2: a release predating `0xC1` reads it as bad magic, which is a torn entry,
which **ends the load** — dropping every cold interval after it and serving those ranges as holes.
That is the silent-zeros failure the cold log exists to prevent, so an older release has to refuse
the directory.

But the bump alone did not achieve that. `CheckFormat` stamped only *unstamped* directories, so a
directory stamped 1 kept its stamp while this build wrote `0xC1` entries into it — and the previous
release would then open it happily and read it short. `CheckFormat` now raises a stale stamp, so the
stamp describes what the directory may contain. `TestOpenRaisesStaleStamp` covers it, and the
existing format cases are now written against a named current version instead of bare `1`/`2`.

## Verification

- Full `./pkg/block/...` green; `golangci-lint run pkg/block/...` 0 issues.
- Both new restore cases and the compaction case checked against a deliberately broken build.
- No behaviour change for evicted or invalidated ranges, which is every cold entry on a store
  without a remote.

### DIFF
--- a/pkg/block/journal/cold.go
+++ b/pkg/block/journal/cold.go
@@ -19,13 +19,24 @@ package journal
 // Layout: <dir>/cold.log, an append-only entry stream, little-endian
 //
 //	off  size  field
-//	0    1     MagicByte    0xC0, torn-write scan anchor
+//	0    1     MagicByte    0xC1, torn-write scan anchor and layout tag
 //	1    2     FileIDLen
 //	3    8     FileOffset
 //	11   8     Length
 //	19   8     Version
-//	27   4     CRC32        over bytes [0,27) and the FileID bytes
-//	31   var   FileID
+//	27   1     Provenance   what Version dates; see coldProvenance
+//	28   4     CRC32        over bytes [0,28) and the FileID bytes
+//	32   var   FileID
+//
+// Entries written before provenance was recorded carry magic 0xC0 and lack the
+// byte, putting CRC at [27,31) and the FileID at 31. They still load, as
+// coldFromUnknown. Only the reader accepts that shape: every append writes 0xC1,
+// so a log stops growing older entries as soon as this build touches it.
+//
+// A build that predates 0xC1 reads it as bad magic, which is a torn entry, which
+// ends the load and drops every entry after it — the silent-zeros failure this
+// log exists to prevent. That is why formatVersion is 2: an older release must
+// refuse the directory outright rather than read it short.
 //
 // Entries are replayed at recovery like any other record — inserted by Version,
 // so a later warm write shadows a cold entry, and the tombstone and truncate
@@ -44,8 +55,13 @@ import (
 )
 
 const (
-	coldMagic      = 0xC0
-	coldHeaderSize = 31
+	// coldMagicLegacy tags an entry written before provenance was recorded; it
+	// is read, never written. coldMagic is the current layout.
+	coldMagicLegacy      = 0xC0
+	coldHeaderSizeLegacy = 31
+
+	coldMagic      = 0xC1
+	coldHeaderSize = 32
 	coldLogName    = "cold.log"
 
 	// coldSeededName is the marker recording that this journal's cold log has
@@ -57,12 +73,48 @@ const (
 	coldCompactFloor = 1024
 )
 
+// coldProvenance records which writer made an entry, because that is what says
+// whether its version dates the *content* or merely the moment a scan noticed
+// the range was remote-durable. Nothing else in the entry distinguishes them,
+// and a reader that assumes one gets the other silently wrong.
+type coldProvenance uint8
+
+const (
+	// coldFromUnknown is a legacy entry: written before provenance was recorded,
+	// so which writer made it cannot be recovered. Readers must treat it as the
+	// eviction case, which is what every reader did before this field existed.
+	coldFromUnknown coldProvenance = 0
+
+	// coldFromData means version was copied off the interval whose local bytes
+	// this entry replaces, so it still dates the content: eviction and
+	// invalidation. A version test answers "did these bytes exist at V".
+	coldFromData coldProvenance = 1
+
+	// coldFromScan means version was minted when a scan found the range already
+	// remote-durable, so it dates the scan and says nothing about the content:
+	// the manifest-seed writers. The version exists only so a racing Delete
+	// sweeps below the entry. A version test answers nothing about the content.
+	coldFromScan coldProvenance = 2
+)
+
+func (p coldProvenance) String() string {
+	switch p {
+	case coldFromData:
+		return "data"
+	case coldFromScan:
+		return "scan"
+	default:
+		return "unknown"
+	}
+}
+
 // coldEntry is one persisted cold interval.
 type coldEntry struct {
-	id      FileID
-	fileOff int64
-	length  int64
-	version uint64
+	id         FileID
+	fileOff    int64
+	length     int64
+	version    uint64
+	provenance coldProvenance
 }
 
 func (s *Store) coldPath() string { return filepath.Join(s.dir, coldLogName) }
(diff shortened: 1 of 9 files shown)


marmos91:
Both points were right, both fixed in `525bad1db`.

**`clamp` dropping provenance — a real bug, and the more useful catch.** It rebuilt the interval field by field, carrying `version`, `recOff`, `synced`, `cold` and `loc` but not `provenance`, so a cold range split by a neighbouring write came back as `coldFromUnknown`. Since compaction rebuilds the log from the index, that lands on disk. `without()` trims through `clamp`, so the one line covers both. `TestClampKeepsProvenance` pins it via `liveColdEntries` rather than by reading the field back, so it fails for the consequence and not just the omission — checked by reverting the line, and both fragments come back `unknown`.

This is the same class as the `liveColdEntries` case already in the PR description, one site further along. I found the first index→entry path and stopped auditing there instead of enumerating every place an interval is rebuilt.

**The stale sentinel doc — also right.** I wrote that comment for the first version of this change, which refused every manifest-seeded entry, and did not update it when two existing tests forced the narrower rule. It now describes the actual condition (seeded range overwritten *and synced* past the requested version) and says explicitly that a seeded entry on its own does not produce the error.

For the record on the other half of the review: the `interval` field is deliberate rather than redundant, for the reason your first comment identifies — the index is what compaction reads, so provenance has to survive the round-trip through it, not just the log.

Issue #1569, opened 2026-07-06
VM / random-I/O readiness: 16 MiB block granularity causes ~4000x read amplification


marmos91:
Tracked under the #1466 throughput effort. The carve-concurrency fix (#1567) made DittoFS a solid **large-file / streaming** remote-tier store, but real benchmarking (SCW VM → SCW S3 fr-par, develop @ `0b2361b8`) shows it is **not** adequate for **VM disk images / databases / random-I/O / latency-sensitive** workloads — and the cause is structural, not tuning.

## Evidence (NFSv3, single thread, 512 MiB file)

| Lane | DittoFS-S3 | kernel-NFS (same box) |
|---|---:|---:|
| seq-write | 97.9 MB/s | — |
| seq-read | 538 MB/s | — |
| **rand-read** | **8.5 IOPS, p50 120 ms / p95 261 ms** | ~225k IOPS |
| **rand-write** | **675 IOPS** | ~139k IOPS |

Sequential is fine. Random I/O — which is *all a running VM does* — is disqualifying.

## Root cause

1. **16 MiB block granularity + no byte-range GET.** A 4 KB random read of cold/evicted data fetches a **whole 16 MiB block** from S3 (~120–260 ms) → ~**4000× read amplification** per random read. `s3/store.go` `GetBlock` pulls the full object; there is no partial/range fetch.
2. **Commit-per-op write path** caps random writes at ~hundreds of IOPS.
3. **Eviction** — once synced+evicted, every random read is S3-latency-bound per block.

## Proposed work stream (roughly in impact order)

1. **Range reads / partial-block GET** — fetch only the requested byte range (S3 `Range:` GET), not the whole 16 MiB block. Biggest single win against the amplification.
2. **Per-share block-size policy** — allow 64–256 KiB blocks for random-access shares, trading dedup/packing efficiency for random-read latency.
3. **Resident hot-set cache** — size/pin the local cache to the working set so active VM images stay local instead of evicting under the working set.
4. **Write coalescing** — batch small random writes before commit to lift random write IOPS.

## Scope note

This is a *new capability lane*, not a bug in #1567. #1567 closed the upload-concurrency regression for large-file/streaming; this issue tracks making DittoFS viable as a **block-device / VM-image** backing store. Full competitor matrix + wire numbers to be attached from the current #1466 run.

marmos91:
Fixed in #1594 (merged to develop, 450472d8).

[...]

PR #2480, opened 2026-09-08
fix(journal): size the carve block arena to the share's own chunk ceiling

marmos91:
Refs #2423 — the memory half of the two-window product. Does **not** close it: collapsing the windows and fixing the controller's signal remain semantics decisions, [analysed in the issue](https://github.com/marmos91/dittofs/issues/2423#issuecomment-5591814485).

## What the arena costs

Each in-flight block owns a private arena of one `CarveBlockSize` plus one chunk of overhang — a block flushes only once it crosses `CarveBlockSize`, so it overshoots by the chunk that crossed the line. The overhang came from the package-wide ceiling:

```go
arenaCap64 := s.cfg.CarveBlockSize + int64(chunker.MaxChunkSize)   // 16 MiB, always
```

But `chunk_size` derives the share's ceiling (`blockstore_config.go:1105`):

```go
cp := chunker.Params{Min: int(n), Avg: int(n) * 4, Max: int(n) * 8}
```

So a share at the 128 KiB recommended for random-access workloads (#1569) has `Max = 1 MiB` and reserved **16 MiB per block it could never fill**. A default share is unaffected — `DefaultParams().Max` *is* the package ceiling.

## Why it multiplies

The reservation is per concurrency slot, so it scales with both windows:

**peak carve memory = ParallelUploads x CarveUploadConcurrency x (CarveBlockSize + overhang)**

At defaults that is 20 MiB a slot: 2.5 GiB at the adaptive floor, 10.2 GiB at the ceiling. Worst case rather than typical — a file only holds 8 slots with 8 blocks' worth of dirty data ready — but nothing declares or bounds it. For a 128 KiB share the per-slot term drops 20 MiB → 5 MiB.

## This is a recorded debt, not a new idea

`carve.go:23` already carried a `ponytail:` marker naming it and the condition to act: *"size both from ChunkParams.Max if that headroom ever shows up in a memory profile."* It has. The marker stays on the scratch buffer half, with the reason that half is harder: shrinking it also means moving the read loop off the same constant, or it asks for bytes past its own capacity and stops making progress.

## Verification

New test pins the ceiling, and **fails against unpatched `carve.go`** — 6 MiB file, 1 MiB blocks, window 4, 16 KiB chunks:

```
carve allocated 121564840 bytes, over the 54591488-byte ceiling for 4 slots of 1064960-byte arenas
```

Patched: **36.7 MB**, down from 121.6 MB. The test collects before measuring, because Go empties `sync.Pool` on GC and the arenas would otherwise be served from a previous test's leftovers.

- `go test ./pkg/block/journal/` green
- `go test -race ./pkg/block/journal/ ./pkg/block/engine/` green (67.9s / 235.0s)
- Everything edited after that race run was comments only

## Also in the diff

`engine/carve_dispatch.go`'s comment claimed the upload window bounds block PUTs, and that the goodput controller observes real in-flight concurrency through it. It bounds carve *passes*, and what `TakePeak` counts is files — which is why draining one large file reads as app-limited however many PUTs are in the air. `engine/types.go` had the same error and was fixed in #2438; this was the remaining copy, and it is the same misdescription this PR's arithmetic depends on.

## Not touched

No behaviour change: same chunk boundaries, same PUTs, same ordering, same exit paths. The grow-belt at the append site still backstops the bound if a config ever breaks it.

### DIFF
--- a/pkg/block/engine/carve_dispatch.go
+++ b/pkg/block/engine/carve_dispatch.go
@@ -53,14 +53,20 @@ func (m *RemoteSync) carveDispatcher(ctx context.Context) {
 // pass (one file, one block, one PutBlock at a time) leaves the uplink almost
 // idle — the block-upload latency, not the link or CPU, caps throughput.
 //
-// Concurrency is bounded by the adaptive upload window: the loop acquires
-// uploadLimiter before starting each file's carve and releases it when that
-// file's blocks are committed, so at most Limit() carves — and thus block
-// PUTs — are in flight. Acquiring the window is also what lets the goodput
-// controller observe real in-flight concurrency (TakePeak) and ramp the window;
-// without it the window is never consumed and stays pinned at the floor. Files
-// in one shard still serialize on the journal's internal carve lock, so the
-// concurrency here overlaps distinct shards' upload latency.
+// The adaptive upload window bounds how many files carve at once: the loop
+// acquires uploadLimiter before starting each file's carve and releases it when
+// that file's pass returns, so at most Limit() passes run together. It does not
+// bound the block PUTs inside a pass — each pass opens its own window on those —
+// so the PUTs in flight are the product of the two, and so is the memory held by
+// the blocks waiting on them.
+//
+// What the goodput controller samples through TakePeak is therefore this window,
+// the count of files, not the count of PUTs. Draining one large file peaks at a
+// single pass and reads as app-limited however many PUTs that pass has in the
+// air. Acquiring the window is still what keeps it consumed at all; without it
+// the window is never taken and stays pinned at the floor. Files in one shard
+// still serialize on the journal's internal carve lock, so the concurrency here
+// overlaps distinct shards' upload latency.
 func (m *RemoteSync) carvePass(ctx context.Context) {
 	files := m.local.ListFiles(ctx)
 	if len(files) == 0 {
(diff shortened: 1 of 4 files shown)


marmos91:
Copilot's overflow finding is real, and now fixed in `e2a1ef3eb` — but it is **pre-existing rather than introduced here**, which is worth recording since the PR is otherwise a no-behaviour-change claim.

The unmodified line had the same shape:

```go
arenaCap64 := s.cfg.CarveBlockSize + int64(chunker.MaxChunkSize)   // wraps identically
```

and `overhang` is bounded by `chunker.MaxChunkSize` in every branch — `Validate` caps `Max` at that ceiling, and the fallback is `DefaultParams().Max`, which *is* that ceiling. So the magnitude added is unchanged and the wrap risk was neither created nor widened by this PR. The clamp is worth doing anyway because the comment sitting on those lines already claimed to cover exactly this hazard.

Fix saturates before the addition rather than testing the sum:

```go
overhang64 := int64(overhang)
blockCap64 := s.cfg.CarveBlockSize
if blockCap64 > math.MaxInt-overhang64 {
    blockCap64 = math.MaxInt - overhang64
}
arenaCap := int(blockCap64 + overhang64)
```

The subtraction cannot itself go negative: `withDefaults` has already replaced any `CarveBlockSize <= 0`, and the overhang is at most 16 MiB, so `math.MaxInt - overhang64` stays positive on 32-bit too (2147483647 - 16777216).

**No test.** The only input that reaches the clamp is a `CarveBlockSize` near `MaxInt64`, and exercising it past the clamp means `make([]byte, MaxInt)` — not something a test can run. The code is non-wrapping by construction instead, which is the property that was missing.

Also in the push: `store.go`'s peak-RAM note now states the arena term against the configured `ChunkParams.Max` rather than the package constant, and says the term multiplies again by whatever bounds how many files carve at once — the per-file phrasing is what made the real ceiling easy to miss in the first place.

`go test ./pkg/block/journal/` green.

PR #2520, opened 2026-09-10
chore: sync develop into main for the v0.31.1 release

marmos91:
Brings main up to develop (163 commits: waves 2–3.5, error-universe consolidation #2505, pynfs walk-outs #2506/#2507, the v0.31.1 release commit #2519) plus main's unique jwt-secret-style fix carried across. The v0.31.1 tag must point at main after this lands so the release workflow builds binaries that actually contain the #2510 NFS fix.


marmos91:
Superseded by #2521, which carries the same tree with every inbound commit SSH-signed (required_signatures is now active on main; 139 of these 163 commits were unsigned GitHub-web merges and would have kept the PR blocked forever).

PR #2522, opened 2026-09-10
chore: sync develop into main for the v0.31.1 release (signed history)

marmos91:
Same tree as #2520/#2521, with every inbound commit SSH-signed by an attributed committer identity (required_signatures is active on main; GitHub attributes signatures to registered account emails, so the GitHub-web merge committer emails were rewritten to the account noreply address). Tip tree is byte-identical to #2520's. The v0.31.1 tag must point at main after this lands so the release workflow builds binaries that actually contain the #2510 NFS fix.


marmos91:
Superseded: syncing via direct fast-forward of the branch pointer instead — no merge commit, no rewrites. Closing per owner instruction.

PR #2539, opened 2026-09-11
refactor: journal hygiene sweep

marmos91:
Six-part hygiene sweep of `pkg/block/journal` from the step-3 plan (Lane H), cutting the journal surface ahead of the carver/state-model lanes:

- **GC unexported** — `GC`/`GCOptions`/`GCResult` → `gc`/`gcOptions`/`gcResult`; the pass runs behind the background loop (`gcLoop`), its only trigger. Eviction is unaffected.
- **Warning sink over `Config.Logger`** — `internal/logger` import severed; warnings go to `Config.Logger` (`*slog.Logger`), discarding when unset. The production config (`local/fs`) wires `slog.Default()` so the migrated warnings keep reaching the operator's log.
- **`segmentLocation` unexported, `PinVersion()` getter dropped** — internal readers already used the field directly; the setter (`SetPinVersion`) is kept for its runtime callers.
- **`journal.RemoteStore` deleted, `Open` collapsed** — the interface and field were write-only (no reads anywhere, production passes nil); `Open(dir string, cfg Config)` now takes two args, with `checkFormat` folded inside as its first statement.
- **Journal-local `ErrFutureFormat` sentinel** — boot-guard sites (`controlplane/runtime/init.go`, `cmd/dfs start.go`, `local/fs/format_test.go`) match `block.ErrFutureFormat || journal.ErrFutureFormat`, preserving exit-78 semantics; the shared `block.ErrFutureFormat` stays for the metadata tiers.
- **Test-helper fixes from review** — `reopen` merges the caller's overrides into the store's config instead of dropping `ShardCount`/`SegmentSize`; `captureWarnings` returns the `*slog.Logger` directly.

Net −56 lines. Gates: build, vet, gofmt clean; 58 packages ok; `go test -race ./pkg/block/journal/` pass.

Reviews (three fresh-context, all APPROVE): correctness (boot exit-78 chain traced intact, gcLoop survival, no behaviour change — 12 migrated Warn sites byte-identical), self-containment (6 findings: recovery cfg merge, dead Cleanup, nil-logger wiring, one-caller helpers inlined, SystemClock export dropped), adversarial (all six prosecution counts clear, no counterexample found).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cmd/dfs/commands/start.go
+++ b/cmd/dfs/commands/start.go
@@ -22,6 +22,7 @@ import (
 	"github.com/marmos91/dittofs/pkg/adapter/smb"
 	"github.com/marmos91/dittofs/pkg/auth/kerberos"
 	"github.com/marmos91/dittofs/pkg/block"
+	"github.com/marmos91/dittofs/pkg/block/journal"
 	"github.com/marmos91/dittofs/pkg/block/local/fs"
 	"github.com/marmos91/dittofs/pkg/config"
 	"github.com/marmos91/dittofs/pkg/controlplane/api"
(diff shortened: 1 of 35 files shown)


marmos91:
Closed unmerged: this PR holds the pre-squash lane H sweep, which is already on develop via squash-merge #2535. Its three genuinely new commits (runtime format-sentinel surface test, both-sentinels boot-guard test, doc dependency fix) are salvaged as #2543. Nothing else here is new.

PR #2544, opened 2026-09-11
refactor: carve the FastCDC carver out of the journal

marmos91:
Lane C of the step-3 journal refactor — carve the FastCDC carver out of the journal.

What was done:
- **`pkg/block/carver`** (new): FastCDC carving wrapping `pkg/block/chunker` — `Box`/`Drain` two-call contract, BLAKE3-256 `Hash` with constant-time `Equal`, per-chunk `Skip` dedup oracle, per-block arena backing novel chunks' `Data`.
- **Journal rewiring**: `packRuns` owns one Carver per pass (batch carries across runs → blocks span runs), calls `Box` with `final=true` at each run's EOF (chunks stay inside a run), submits per-block via the dispatcher with per-block `flipPlan\{first: blockFirstRun, last: ri\}`.
- **Dispatcher delta (Option B, supervisor-approved)**: `acquire`/`discard`/`carveArenaPool` removed; `submit(chunks, plan)` takes the sem slot itself when chunks are non-empty; `BlocksWritten` keyed on `hasNovelBytes(chunks)`. Fresh-arena-per-emitted-block replaces the pooled design — no pool-reuse corruption hazard.
- **blake3 removed from the journal** — hashing lives only in the carver.
- **Test premise updates**: arena-memory test budgets one extra block + 1 MB instrumentation slack; pipeline test's bound doubles to `2×uploadConcurrency×runsPerBlock+1` (stall moved from first-novel-chunk to submit time).

Fixed en route (latent bugs the new carver tests caught): nil `Skip` panicked (nil = pack everything per the contract); `Box`'s tiled-bytes return lost fed bytes; `Block.Bytes` double-counted skipped chunks.

Gates: build, vet, gofmt clean; `go test ./pkg/block/... -count=1` pass; `-race` on carver + journal pass. Net −83 lines in the journal.

Three fresh-context reviews: all APPROVE.

### DIFF
--- a/pkg/block/carver/carver.go
+++ b/pkg/block/carver/carver.go
@@ -1,0 +1,257 @@
+// Package carver cuts a contiguous byte stream into FastCDC chunks, hashes
+// them, consults a per-chunk skip oracle, and accumulates the novel chunk bytes
+// into block-sized batches. It wraps pkg/block/chunker — the chunker supplies
+// the boundary search, the carver everything the journal's pack loop used to do
+// inline around it.
+//
+// The two-call contract: feed each contiguous stream with Box, with final set
+// on the stream's last call only — a chunk never spans the gap between two
+// streams, but a block does. Once the caller's last stream is over, one Drain
+// emits the trailing partial block; buffered chunk bytes never outlive the
+// carver. A single-stream consumer is Box(..., true) then Drain().
+package carver
+
+import (
+	"context"
+	"crypto/subtle"
+	"math"
+
+	"lukechampine.com/blake3"
+
+	"github.com/marmos91/dittofs/pkg/block/chunker"
+)
+
+// Hash is a content hash of one chunk.
+type Hash [32]byte
+
+// Equal reports whether two hashes are the same content, in constant time.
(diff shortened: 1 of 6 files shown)


marmos91:
Triage of the four findings (all fixed in 48f759877):

- **Box tiled contract (carver.go:110)**: the premise that a test fails immediately was verified false — an exact-Min non-final feed buffers and returns tiled=0, but no test asserts tiled on small feeds. Real residue: allChunks' tiled==fed assertion was content-flaky (the last cut in a max-sized feed can retain a tail). Relaxed to tiled<=fed; the Drain covers retained bytes.
- **allChunks (carver_test.go:29)**: same flake, fixed with the relaxation.
- **Gap test (carver_test.go:216)**: confirmed real and load-bearing — the first Box used final=false, so the below-Min tail rode and the boundary search continued seamlessly across the gap (the test passed vacuously). Now: first Box final=true, oracle = each half cut alone. Proved load-bearing by removing the chunker reset: the test fails on that broken build.
- **Peak-RAM doc**: carve.go:383 now accounts for both max-chunk buffers (pass-level read buffer + carver accumulator) and notes the per-file bound.

PR #2547, opened 2026-09-11
fix(nfs): sync sidecar shutdown state outside the group lock

### LINKED ISSUE #2542: fix(nfs): sidecar Stop/Start unsynchronized field access races
## The unsynchronized field pairs

Two NFS sidecars read/write adapter fields from Start/Stop with no synchronization:

1. **portmapper**: `portmapSidecar.Start` → `startPortmapper` writes `NFSAdapter.portmapServer` (`pkg/adapter/nfs/portmap.go:68`, also `portmapRegistry` :66) while `portmapSidecar.Stop` → `stopPortmapper` reads it (`pkg/adapter/nfs/portmap.go:198-201`).
2. **nfs-udp**: `udpSidecar.Start` → `startUDP` writes `NFSAdapter.udpConn` (`pkg/adapter/nfs/udp.go:45`) while `udpSidecar.Stop` reads it (`pkg/adapter/nfs/auxservices.go:178-179`); `handlers.go:301` also reads `udpConn` on the request path.

## Why the races are reachable now

`auxsvc.Group.Start` historically held the group mutex across `s.Start`, serializing Start/Stop per name. It now reserves the name under the lock and runs `s.Start` outside it (the wedge fix), so a `StopOne` — and therefore `Group.Reconcile(name, want=false)` from a settings-disable — can run concurrently with an in-flight sidecar `Start`. The `Service.Stop` contract in `pkg/adapter/auxsvc/auxsvc.go` now documents that ordering and notes existing NFS sidecars do not yet synchronize for it.

The window is narrow: a settings-disable or teardown racing a slow sidecar bind (rpcbind timeouts, portmapper listener wait). Not reached by the direct Reconcile path, which builds a fresh wrapper per call, but the fields are also read on the steady-state request path (`handlers.go:301`).

## Fix shape

Synchronize the concrete lifecycles — either guard the fields (per-adapter mutex or atomic for `udpConn`/`portmapServer`/`portmapRegistry`), or capture the listener/server in locals passed only through the sidecar's own goroutine so the adapter fields are written once and never read mid-Start. The double-close on `udpConn` (ctx-cancel goroutine plus `Stop`) is already documented as harmless; the race is on the pointer read/write, not the close.

marmos91:
## What was wrong

Issue #2542: two NFS sidecars published adapter fields from `Start` and read them from `Stop` with no synchronization:

1. **portmapper** — `startPortmapper` wrote `NFSAdapter.portmapServer` (`portmap.go:68`) while `stopPortmapper` read it (`portmap.go:198`)
2. **nfs-udp** — `startUDP` wrote `NFSAdapter.udpConn` (`udp.go:45`) while `udpSidecar.Stop` (`auxservices.go:178`) and `sendNLMAsyncResult` (`handlers.go:301`, the steady-state NLM request path) read it

The races became reachable when `auxsvc.Group.Start` moved `Service.Start` outside the group mutex (the wedge fix, #2540): a live settings-disable can now steal a mid-bind sidecar reservation and tear down concurrently with its in-flight `Start`. The `Service.Stop` contract documents implementations must synchronize their own fields for that ordering.

## Evidence

A `-race` hammer (added as `sidecar_state_test.go`) running concurrent `startUDP`/`startPortmapper` against `udpSidecar.Stop`/`portmapSidecar.Stop` reproduced data races at the exact lines above before the fix:

```
WARNING: DATA RACE
Write at 0x00c000518e60 by goroutine 22:
  pkg/adapter/nfs/portmap.go:68
Previous read at 0x00c000518e60 by goroutine 21:
  pkg/adapter/nfs/portmap.go:198
```

After the fix, the same hammer is green under `-race` (3 consecutive full runs).

## The fix

One new `sync.Mutex` (`sidecarMu`) on `NFSAdapter` — the struct already uses this shape (`resolverMu`):

- **Publishes under the lock**: `startUDP` stores `udpConn`; `startPortmapper` stores `portmapServer` **only after `server.WaitReady()` fires**
- **Snapshot at every reader**: `udpSidecar.Stop` snapshots the pointer then closes outside the lock; `sendNLMAsyncResult` snapshots then uses it; `stopPortmapper` snapshots then calls `server.Stop()` outside the lock (so the request path never stalls behind a portmapper teardown)

The publish-after-`WaitReady` point is load-bearing, found in review: `portmap.Server.Serve` writes its internal listeners inside itself, before closing `WaitReady`, so publishing earlier would let a concurrent `stopPortmapper -> server.Stop` read the listeners mid-bind (race inside `portmap.Server`) or nil (leaking a late-bound port and stuck `serveTCP` goroutines). Before `WaitReady`, the field stays nil so `stopPortmapper` is a correct no-op, and the unready server is torn down by the lifecycle ctx instead. A disable that steals a mid-start reservation is torn down by `auxsvc.Group.Start`'s rollback itself, after the server is fully ready — clean.

Also deleted `NFSAdapter.portmapRegistry` — write-only since its introduction, zero readers anywhere in the repo.

## Verification

- `-race` hammer green, 3 consecutive runs
- Full repo suite (`go test ./...`) green
- `go build ./...`, `go vet ./...`, `go fmt ./...` clean
- No Cobra command or flag moved — no docs regeneration needed

Closes #2542

### DIFF
--- a/internal/adapter/nfs/portmap/server.go
+++ b/internal/adapter/nfs/portmap/server.go
@@ -67,7 +67,8 @@ func NewServer(cfg ServerConfig) *Server {
 
 // Serve starts the portmapper server on both TCP and UDP.
 // It blocks until the context is cancelled or Stop is called.
-// After both TCP and UDP listeners are bound, WaitReady() unblocks.
+// After both TCP and UDP listeners are bound and every goroutine is launched,
+// WaitReady() unblocks.
 func (s *Server) Serve(ctx context.Context) error {
 	addr := fmt.Sprintf(":%d", s.config.Port)
 
@@ -99,12 +100,10 @@ func (s *Server) Serve(ctx context.Context) error {
 		s.udpConn = udpConn
 	}
 
-	// Signal that listeners are ready
-	close(s.listenerReady)
-
-	logger.Info("Portmapper server started", "address", addr, "tcp", s.config.EnableTCP, "udp", s.config.EnableUDP)
-
-	// Launch TCP and UDP goroutines
+	// Signal that listeners are ready — AFTER the serve goroutines and the
+	// cancellation monitor are launched, so a caller that observes readiness
+	// (WaitReady) knows all wg.Adds are done and Stop's wg.Wait sees the true
+	// counter instead of returning at zero (or racing the concurrent Add).
 	if s.config.EnableTCP {
 		s.wg.Add(1)
 		go s.serveTCP(ctx)
@@ -122,6 +121,8 @@ func (s *Server) Serve(ctx context.Context) error {
 		case <-s.shutdown:
 		}
 	}()
+
+	close(s.listenerReady)
 
 	// Block until both goroutines complete
 	s.wg.Wait()
--- a/pkg/adapter/nfs/adapter.go
+++ b/pkg/adapter/nfs/adapter.go
@@ -137,13 +137,15 @@ type NFSAdapter struct {
 	// auxservices.go.
 	sidecars *auxsvc.Group
 
+	// sidecarMu guards the sidecar shutdown state published by Start and read
+	// by Stop and the NLM async-result send path (portmapServer, udpConn): the
+	// group runs Start outside its lock, so a disable can steal a mid-bind
+	// reservation and tear down concurrently.
+	sidecarMu sync.Mutex
+
 	// portmapServer is the embedded portmapper server (RFC 1057).
 	// nil when portmapper is disabled.
 	portmapServer *portmap.Server
-
-	// portmapRegistry holds the portmap service registry.
-	// nil when portmapper is disabled.
-	portmapRegistry *portmap.Registry
 
 	// sysregActive is true once DittoFS's services have been registered with the
 	// host's system rpcbind (adapters.nfs.portmapper.register_with_system), so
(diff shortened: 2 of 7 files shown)


marmos91:
Copilot findings addressed in 13163bcc7:

1. **Critical — untracked portmapper after 2s timeout**: adopted. On the timeout path the start is now abandoned explicitly: a late-stop goroutine waits for the bind to land (WaitReady closed implies the server's cancellation monitor exists) and calls server.Stop() — the untracked-serving window is closed. The error path returns the bind error as before (Serve's ctx tears it down); nothing is published, so stopPortmapper stays a no-op.
2. **Moderate — cross-generation shutdown**: adopted. Both stop paths now claim-and-clear (read+nil under sidecarMu), so each generation is stopped exactly once — an old stop can no longer snapshot a newer generation's server/conn.
3. **Moderate — cancel+wait per-start goroutines**: not adopted. serveUDP is self-terminating (reads its captured conn; Close unblocks it), owns no shared state, and never conflicts with a newer generation's socket. Waiting in Stop would stall teardown behind in-flight datagram handlers for no safety gain.
4. **Nit — test wording**: reworded; the idempotent-repeat comments now describe the cleared-state no-op.

- Race hammer green (3 consecutive -race runs), full suite green, vet/fmt clean.

marmos91:
Round-2 review findings addressed in c4fb55442:

1. **portmap.go:85 — WaitReady vs wg.Add gap**: adopted. server.go Serve now closes listenerReady AFTER the serve goroutines and the cancellation monitor are launched, so WaitReady closed implies all wg.Adds are done — a raced Stop's wg.Wait always sees the true counter, and the concurrent-Add misuse is gone. Publishing and the timeout late-stop now always observe the real wait group.
2. **auxservices.go:182 — Stop must wait for background work**: partially adopted. The real named leak is fixed: startUDP now publishes a per-generation shutdown ctx (udpStop cancel func) with the conn, and Stop claims-and-cancels it, so the conn-close waiter and read loop exit on toggle instead of parking until adapter shutdown. The full wait on in-flight datagram handler goroutines stays unadopted: they are stateless, drain self-bounded, and never conflict with a newer generation — waiting in Stop would stall live teardown for no safety gain.
3. **Test comment wording (store)**: reworded in the previous round (c4fb55442's ancestor); idempotent-repeat comments now describe the cleared-state no-op.

Also note: udpStop's cancel plus udpConn.Close makes the double-close on teardown explicit and harmless (per the udpSidecar doc comment).

- Race hammer green (3 consecutive -race runs), auxsvc + portmap suites green, full repo suite green, vet/fmt clean.

PR #2549, opened 2026-09-11
fix(nfs): guard sidecar config fields across settings-apply paths

### LINKED ISSUE #2548: NFS sidecar config fields written and read unsynchronized across settings-apply paths
## What

Follow-up to #2542 (scoped out there). `applyNFSSettings` writes four NFS sidecar config fields with no synchronization, and it runs on three goroutines concurrently:

1. The settings-watcher poller: `sw.OnNFSSettingsChange(...)` callback (`adapter.go:774-778`)
2. The accept loop, per accepted connection: `preAcceptCheck` → `applyNFSSettings(rt)` (`adapter.go:875`)
3. Startup: `SetRuntime` → `applyNFSSettings(rt)` (`adapter.go:762`)

### The racy pairs

**Writer vs writer** — concurrent applies write the same fields unsynchronized:
- `s.config.Portmapper.Enabled = &enabled` (`settings.go:83`)
- `s.config.Portmapper.Port = settings.PortmapperPort` (`settings.go:85`)
- `s.config.Portmapper.RegisterWithSystem = &registerWithSystem` (`settings.go:88`)
- `s.config.UDP.Enabled = &udpEnabled` (`settings.go:97`)

The SettingsWatcher's *cache* is atomic (RLock-guarded `GetNFSSettings`, `settings_watcher.go:155-160`), but the adapter's copy of the values into `s.config` is not.

**Reader vs writer** — the sysreg transition goroutine reads those fields concurrently with any apply:
- `reconcileSysreg`'s background transition calls `registerWithSystemEnabled()` (`auxservices.go:136` → `portmap.go:112-115`), reading `Portmapper.RegisterWithSystem`
- `systemRegMappings()` → `isUDPEnabled()` (`portmap.go:142` → `udp.go:33`), reading `UDP.Enabled`
- `startPortmapper` → `isPortmapperEnabled()`/`Portmapper.Port` (`portmap.go:24-28, 55-59`), reading `Portmapper.Enabled/Port`

Two applies racing can tear the `*bool` pointers: one goroutine writes `&enabled` while another reads it — a data race on the pointer word, and a torn read returns a stale value.

### What is already safe (no action)

- `StateManager` setters (`SetLeaseTime`, `SetGracePeriodDuration`, `SetDelegationsEnabled`, `SetMaxDelegations`, `SetDirDelegBatchWindow`, `SetMaxSessionSlots`, `SetMaxSessionsPerClient`, `SetMaxConnectionsPerSession`) all take `sm.mu`
- v4 blocked ops: `Handler.SetBlockedOps` takes `blockedOpsMu`
- v3 blocked ops: `setV3BlockedOps` takes `s.blockedOpsMu`
- attrs lease time: `v4attrs.SetLeaseTime` stores an `atomic.Uint32`
- SMB has no portmapper/UDP sidecar config — NFS-only

## Fix shape

A `configMu sync.Mutex` on `NFSAdapter` guarding the four sidecar config fields (`Portmapper.Enabled`, `Portmapper.Port`, `Portmapper.RegisterWithSystem`, `UDP.Enabled`):

- `applyNFSSettings` writes them under the lock (one critical section at the end of the function)
- `isPortmapperEnabled`, `isUDPEnabled`, `registerWithSystemEnabled` read under the lock (or snapshot the fields into a small struct under the lock, then decide outside)
- The `Portmapper.Port` read in `startPortmapper`/`systemRegMappings` snapshots under the lock

Plus a `-race` hammer: two goroutines calling `applyNFSSettings`-style writes against a third spinning the readers, failing pre-fix on the pointer races.

## Reproduce

A test constructing `NFSAdapter` with two goroutines writing the four fields in a loop against a third reading them, under `-race`, fires immediately pre-fix (plain pointer write/read).

marmos91:
## Issue #2548: NFS sidecar config fields written with no synchronization

`applyNFSSettings` writes `s.config.Portmapper.Enabled/Port/RegisterWithSystem` and `s.config.UDP.Enabled` (plain `*bool`/`int`) while running on **three concurrent goroutines**:

- the settings-watcher poller, via the `OnNFSSettingsChange` callback (`adapter.go:782`)
- the accept loop, per connection, via `preAcceptCheck` (`adapter.go:882`)
- startup, via `SetRuntime` (`adapter.go:769`)

and reads them concurrently from:

- the sysreg transition goroutine spawned by `reconcileSysreg` (`registerWithSystemEnabled`, `systemRegMappings` → `isUDPEnabled`)
- the portmapper/UDP start paths (`isPortmapperEnabled`, `isUDPEnabled`, `config.Port` reads)

That is a write/write race between concurrent applies plus reader/writer races — the same class as #2542, but on the config fields rather than the sidecar state fields.

### The fix

New `configMu sync.Mutex` on `NFSAdapter`, one lifecycle domain for the four sidecar config fields:

- **`settings.go`** — `applyNFSSettings` prepares the `*bool` values outside the lock, then performs the `Port` read-decide-write **and all four field stores inside one critical section**. The read-decide-write of `Port` must stay inside the section: two concurrent applies would otherwise race on the int even with the lock present. The two log lines merge into one emitted after unlock.
- **`portmap.go`** — `isPortmapperEnabled` / `registerWithSystemEnabled` snapshot the `*bool` under the lock and dereference outside (safe: applies always assign a fresh non-nil pointer and nothing nils them); `startPortmapper` snapshots `config.Port` + `Portmapper.Port` in one section so the registry/server see a coherent generation; `systemRegMappings` snapshots the port. The "Portmapper started" log uses the actually-bound `portmapPort` local rather than a re-read of config.
- **`udp.go`** — `isUDPEnabled` snapshots the `*bool`; `startUDP` snapshots `config.Port`.

Deadlock safety was reviewed: every `configMu` section holds only field loads/stores (no calls, no channel ops, no nested locks); all blocking waits happen with no lock held; `sidecarMu` sections never acquire `configMu` (no lock ordering inversion).

### Test

`sidecar_config_test.go` — a `-race` hammer: two writer goroutines calling `applySidecarConfig` (a mirror of the fixed apply, including the keep-previous-port fallback when the DB port is 0) against a reader goroutine spinning the real production readers. Confirmed **red under `-race` before the fix** on the exact write/write and read/write pairs; green after (repeated runs). The real `applyNFSSettings` cannot be used directly in the test because it needs a `*runtime.Runtime`.

### Review

Two independent reviewer passes. The first flagged the port read sitting outside the critical section (the initial version locked the stores but not the read-decide-write) — restructured per review, re-verified.

Closes #2548.

### DIFF
--- a/pkg/adapter/nfs/adapter.go
+++ b/pkg/adapter/nfs/adapter.go
@@ -143,6 +143,13 @@ type NFSAdapter struct {
 	// reservation and tear down concurrently.
 	sidecarMu sync.Mutex
 
+	// configMu guards the sidecar config fields (Portmapper.Enabled/Port/
+	// RegisterWithSystem, UDP.Enabled) written by applyNFSSettings on the
+	// settings-watcher goroutine, on the accept loop per connection, and at
+	// startup — three concurrent writers — and read by the sysreg transition
+	// goroutine and the portmapper/UDP start paths.
+	configMu sync.Mutex
+
 	// portmapServer is the embedded portmapper server (RFC 1057).
 	// nil when portmapper is disabled.
 	portmapServer *portmap.Server
--- a/pkg/adapter/nfs/portmap.go
+++ b/pkg/adapter/nfs/portmap.go
@@ -31,6 +31,25 @@ func (s *NFSAdapter) isPortmapperEnabled() bool {
 	return *enabled
 }
 
+// snapshotSidecarConfig reads the sidecar config tuple (main NFS port,
+// portmapper port, UDP enabled) under one configMu critical section. Applies
+// run concurrently with starts, so every consumer that needs a coherent view
+// of the three values — the portmapper registry/server and the system-rpcbind
+// mappings — must go through this helper rather than reading the fields
+// separately. Reads UDP.Enabled directly: isUDPEnabled() would re-acquire the
+// same mutex inside this one (not reentrant) and self-deadlock.
+func (s *NFSAdapter) snapshotSidecarConfig() (nfsPort, portmapPort int, udpEnabled bool) {
+	s.configMu.Lock()
+	nfsPort = s.config.Port
+	portmapPort = s.config.Portmapper.Port
+	enabled := s.config.UDP.Enabled
+	s.configMu.Unlock()
+	if enabled != nil {
+		udpEnabled = *enabled
+	}
+	return nfsPort, portmapPort, udpEnabled
+}
+
 // startPortmapper creates and starts the embedded portmapper server.
 //
 // The portmapper (RFC 1057) enables NFS clients to discover DittoFS services
@@ -55,14 +74,13 @@ func (s *NFSAdapter) startPortmapper(ctx context.Context) error {
 
 	// Snapshot the sidecar config under configMu: applies run concurrently with
 	// this start, so the registry and server must see one coherent generation.
-	s.configMu.Lock()
-	nfsPort := s.config.Port
-	portmapPort := s.config.Portmapper.Port
-	s.configMu.Unlock()
+	// All three values (main port, portmapper port, UDP enabled) come from the
+	// same critical section so a concurrent apply cannot land between them.
+	nfsPort, portmapPort, udpEnabled := s.snapshotSidecarConfig()
 
 	// Create registry and register all DittoFS services
 	registry := portmap.NewRegistry()
-	registry.RegisterDittoFSServices(nfsPort, s.isUDPEnabled())
+	registry.RegisterDittoFSServices(nfsPort, udpEnabled)
 	registry.RegisterPortmapper(portmapPort)
 	// Create portmapper server
 	server := portmap.NewServer(portmap.ServerConfig{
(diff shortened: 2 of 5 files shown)


marmos91:
Adjudication of the Copilot review on commit 54a973e13 (both findings **adopted**, fixed in bcbe76eac):

**Finding 1 — `portmap.go:65` mixed configuration generations (adopted).** Correct: `startPortmapper` captured `nfsPort`+`portmapPort` in one `configMu` section but read the UDP-enabled value via a separate `isUDPEnabled()` snapshot, so an apply landing between the two sections could make the registry advertise UDP mappings from a different generation than the captured ports (including advertising UDP services the concurrent apply had just disabled). Fixed by extracting a production helper `snapshotSidecarConfig()` that reads the whole tuple (main port, portmapper port, UDP enabled) under **one** critical section; `startPortmapper` and `systemRegMappings` both go through it.

One note on the fix itself: the helper must read `config.UDP.Enabled` directly rather than calling `isUDPEnabled()` — that reader re-acquires the same non-reentrant mutex and self-deadlocks. This was caught by the `-race` hammer (test hung at 2m, writer blocked forever on `configMu`) and fixed before push; the comment on the helper documents the constraint for future readers.

**Finding 2 — hammer does not exercise production port readers (adopted).** Correct: the test reader acquired `configMu` directly for the port read, so it would have stayed green if the lock were removed from `startPortmapper`'s or `systemRegMappings`'s snapshots. The reader now spins the production `snapshotSidecarConfig()` helper (plus `isPortmapperEnabled`/`registerWithSystemEnabled`), so removing `configMu` from any production snapshot turns the hammer red. Verified: 4 consecutive green runs post-fix.

Verification: `go build ./...`, `go vet`, `go fmt` clean; `-race` green on `pkg/adapter/nfs` + `pkg/adapter/auxsvc`; whole-repo `go test ./...` green.

PR #2550, opened 2026-09-12
fix(block): one upload window governs concurrent PutBlock calls

### LINKED ISSUE #2423: block: upload concurrency is a product of two windows, and the adaptive controller samples the wrong one

marmos91:
## Issue #2423: upload concurrency is a product of two windows

Measured on develop (10-CPU Apple Fabric SSD, 64 files × 8 MiB, 1 MiB blocks, 200 ms injected per-PutBlock latency, production dispatch path, deterministic ×3):

| Arrangement | max concurrent PutBlock | which semaphore bound |
|---|---|---|
| Defaults (inner=8, adaptive outer 16→64) | **160** | neither alone — the **product** |
| Inner=1 (product collapses) | 40 | outer file window |
| Single huge file | **8 exactly** | inner window; outer trivially at 1 |

The documented invariant — "at most Limit() carves, **and thus block PUTs**, in flight" — does not hold: two independent semaphores are nested (outer whole-file carve passes via `uploadLimiter`, inner block PUTs within one carve via `CarveUploadConcurrency=8`), so concurrent `PutBlock` is their **product** (~128 at the floor, 512 at the ceiling — breaching the declared `MaxParallelUploads=256` by 2×). The adaptive goodput controller sampled only the outer limiter's `TakePeak`, so draining one large file peaked at one pass and read as app-limited while the pass's inner 8-slot window saturated the link — the controller never ramped.

### The fix: one window governs PUTs

The obvious move (keep both windows, sample better) cannot fix the workload shape that matters: a single huge file is one carve pass — ramping the per-pass count can't raise single-file throughput no matter what the controller reads. The real fix collapses the product:

- **Enforcement moved into the block sink** — `engineBlockSink.CommitBlock` brackets `PutBlock` with the engine's upload window (acquire per PUT, release when the upload lands). This is now where the adaptive window is enforced, so the controller's `TakePeak` samples the same semaphore that binds: the sampled signal and the controlled variable are the same quantity. The four `Force:true` drain drivers and ManualSync shares are bounded automatically (the limiter lives in the sink the carver calls), closing the issue's "two more paths where the outer window does not apply" gaps with no per-path work.
- **No per-pass limit** — the outer per-pass acquire/release in `carvePass` is gone. Concurrent PUTs across all passes are bounded by the global sink window itself, which is the invariant.
- **The journal's per-file window becomes the pack-ahead memory bound** — `CarveUploadConcurrency` renamed to `CarvePackAhead`, default 8→256 (the upload ceiling, so it never binds concurrency first). It only throttles how many packed blocks queue per file, bounding the live arenas — not the upload-concurrency knob.
- **Constants** — `AdaptiveUploadFloor` 16→128 (= today's effective resting point 16×8, so multi-file behavior at rest is preserved exactly), `AdaptiveUploadCeiling` 64→256 (= `MaxParallelUploads`, now actually enforced; the share-config clamp in `blockstore_config.go` already validated against 256).
- **Controller logic unchanged** — `TestGoodputController_HoldsWindowWhenAppLimited`-style pins survive with re-derived constants.

### Regression guard

`TestBlockSink_EnforcesUploadWindow` gates a memory remote with an in-flight counter checked on the `PutBlock` critical path — the guard that would have caught the original defect. The old fan-out pin (`TestCarvePass_FansOutBoundedByUploadWindow`) pinned the per-pass invariant this change removes; rewritten as `TestCarvePass_FansOutUnbounded` (every file enters `local.Carve` concurrently).

### Verification

- `go build ./...`, `go vet`, `go fmt` clean; regenerated `docs/guide/cli.md` after flag-description changes
- Whole-repo `go test ./...` green
- `-race` green on `pkg/block/journal` + `pkg/block/engine`
- Full engine (82 s) and journal (81 s) suites green

**Not yet measured** (the merge gates from the measurement probe): multi-file parity ≥336.7 MB/s at floor 128, the single-file ramp leg, peak RSS at 256 in-flight buffers, and a low-RTT leg to justify the 128 floor. The probe (`pkg/block/engine/upload_concurrency_probe_test.go`) exists to re-run after this lands; the floor constant may need a follow-up correction if the sweep shows a better knee.

Closes #2423.

### DIFF
--- a/cmd/dfsctl/commands/store/block/remote/add.go
+++ b/cmd/dfsctl/commands/store/block/remote/add.go
@@ -83,7 +83,7 @@ func init() {
 	addCmd.Flags().StringVar(&addAccessKey, "access-key", "", "AWS access key ID (for s3)")
 	addCmd.Flags().StringVar(&addSecretKey, "secret-key", "", "AWS secret access key (for s3)")
 	addCmd.Flags().StringVar(&addCompression, "compression", "", "Enable per-block compression: zstd, lz4 (default: off)")
-	addCmd.Flags().IntVar(&addParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink)")
+	addCmd.Flags().IntVar(&addParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink; enforced per PUT in the block sink)")
 	// Encryption flags
 	addCmd.Flags().StringVar(&addEncryptionAEAD, "encryption-aead", "", "Enable client-side encryption with the given AEAD: aes-256-gcm, chacha20-poly1305, xchacha20-poly1305")
 	addCmd.Flags().StringVar(&addEncryptionKeyKind, "encryption-key-kind", "", "Key provider: local | kmip (required when --encryption-aead is set)")
--- a/cmd/dfsctl/commands/store/block/remote/edit.go
+++ b/cmd/dfsctl/commands/store/block/remote/edit.go
@@ -52,7 +52,7 @@ func init() {
 	editCmd.Flags().StringVar(&editEndpoint, "endpoint", "", "Custom S3 endpoint")
 	editCmd.Flags().StringVar(&editAccessKey, "access-key", "", "AWS access key ID (for s3)")
 	editCmd.Flags().StringVar(&editSecretKey, "secret-key", "", "AWS secret access key (for s3)")
-	editCmd.Flags().IntVar(&editParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink)")
+	editCmd.Flags().IntVar(&editParallelUploads, "parallel-uploads", 0, "Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink; enforced per PUT in the block sink)")
 }
 
 func runEdit(cmd *cobra.Command, args []string) error {
--- a/docs/guide/cli.md
+++ b/docs/guide/cli.md
@@ -6207,7 +6207,7 @@ Flags:
       --encryption-kmip-key-uid string    KMIP managed symmetric key UID (kind=kmip)
       --endpoint string                   Custom S3 endpoint (for S3-compatible stores)
       --name string                       Store name (required)
-      --parallel-uploads int              Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink)
+      --parallel-uploads int              Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink; enforced per PUT in the block sink)
       --prefix string                     Key prefix within the bucket (for s3)
       --region string                     AWS region (for s3) (default "us-east-1")
       --secret-key string                 AWS secret access key (for s3)
@@ -6261,7 +6261,7 @@ Flags:
       --bucket string          S3 bucket name (for s3)
       --config string          Store configuration as JSON
       --endpoint string        Custom S3 endpoint
-      --parallel-uploads int   Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink)
+      --parallel-uploads int   Max parallel chunk uploads to this remote (0 = adaptive: auto-tune to saturate the uplink; enforced per PUT in the block sink)
       --region string          AWS region (for s3)
       --secret-key string      AWS secret access key (for s3)
       --type string            Store type: s3, memory
--- a/docs/internals/architecture.md
+++ b/docs/internals/architecture.md
@@ -301,8 +301,12 @@ The carve pass fans out across files: a single sequential pass (one file, one
 block, one `PutBlock` at a time) leaves the uplink almost idle. Concurrency is
 bounded by an **adaptive upload window** (`pkg/block/engine/upload_controller.go`):
 a pinned `--parallel-uploads` fixes the window, while the default (adaptive)
-mode ramps it between a floor and ceiling to track the goodput knee. Files in
-one shard still serialize on the journal's carve lock, so the window overlaps
+mode ramps it between a floor and ceiling to track the goodput knee. The window
+is enforced in the engine's block sink, acquired around each `PutBlock` and
+released when the upload lands, so the count the controller samples through
+`TakePeak` is the count of PUTs actually in the air — a single large file (one
+pass emitting many PUTs) reads as window-limited and is ramped. Files in one
+shard still serialize on the journal's carve lock, so the fan-out overlaps
 distinct shards' upload latency.
 
 Explicit `Flush` / `SyncNow` force-carve a file's (or all files') dirty ranges
--- a/pkg/block/engine/blocksink.go
+++ b/pkg/block/engine/blocksink.go
@@ -276,6 +276,11 @@ type engineBlockSink struct {
 	// This is where the adaptive window is enforced, so the controller's
 	// TakePeak samples the same semaphore that binds.
 	uploadLimiter *syncer.DynamicSemaphore
+	// metrics is the engine's data-plane metrics handle, captured at sink
+	// construction so the CommitBlock bracket can report in-flight PUTs without
+	// reaching back to the Store. Nil in fixtures that don't inject a recorder
+	// (every call site checks it before calling).
+	metrics func() DataplaneMetrics
 	// onBlockCommitted reports each block as it lands, carrying the block's
 	// uploaded byte count. Reporting here rather than after a carve pass returns
 	// is what makes the count advance *during* a long carve: the drain path
@@ -366,7 +371,17 @@ func (s engineBlockSink) CommitBlock(ctx context.Context, chunks []journal.Carve
 			return fmt.Errorf("carve: acquire upload window: %w", err)
 		}
 	}
+	if s.metrics != nil {
+		if mx := s.metrics(); mx != nil {
+			mx.UploadStarted()
+		}
+	}
 	putErr := s.rbs.PutBlock(ctx, blockID, bytes.NewReader(blockBytes))
+	if s.metrics != nil {
+		if mx := s.metrics(); mx != nil {
+			mx.UploadFinished()
+		}
+	}
 	if s.uploadLimiter != nil {
 		s.uploadLimiter.Release()
 	}
--- a/pkg/block/engine/blocksink.go
+++ b/pkg/block/engine/blocksink.go
@@ -13,6 +13,7 @@ import (
 	"github.com/marmos91/dittofs/pkg/block/blockcodec"
 	"github.com/marmos91/dittofs/pkg/block/journal"
 	"github.com/marmos91/dittofs/pkg/block/remote"
+	"github.com/marmos91/dittofs/pkg/block/syncer"
 	"github.com/marmos91/dittofs/pkg/metadata"
 )
 
@@ -269,6 +270,12 @@ type engineBlockSink struct {
 	rbs         remote.RemoteBlockStore
 	committer   blockCommitter
 	commitLocks *carveCommitLocks
+	// uploadLimiter bounds concurrent remote PutBlock calls — the engine's
+	// upload window, acquired per PUT and released when the upload lands. Nil
+	// in fixtures that don't exercise the window (the guard is inert there).
+	// This is where the adaptive window is enforced, so the controller's
+	// TakePeak samples the same semaphore that binds.
+	uploadLimiter *syncer.DynamicSemaphore
 	// onBlockCommitted reports each block as it lands, carrying the block's
 	// uploaded byte count. Reporting here rather than after a carve pass returns
 	// is what makes the count advance *during* a long carve: the drain path
@@ -348,9 +355,23 @@ func (s engineBlockSink) CommitBlock(ctx context.Context, chunks []journal.Carve
 	blockHash := block.ContentHash(blake3.Sum256(blockBytes))
 
 	// PutBlock first: a crash before the commit leaves an orphan block (GC
-	// reclaims it), never an unbacked record.
-	if err := s.rbs.PutBlock(ctx, blockID, bytes.NewReader(blockBytes)); err != nil {
-		return fmt.Errorf("carve: put block %s: %w", blockID, err)
+	// reclaims it), never an unbacked record. Bracketed by the engine's upload
+	// window: acquire per PUT, release when the upload lands. This is where the
+	// adaptive window is enforced — the semaphore the upload controller samples
+	// with TakePeak is this one, so the controlled variable and the sampled
+	// signal are the same quantity and a single large file (one carve pass
+	// emitting many PUTs) is read as window-limited and ramped.
+	if s.uploadLimiter != nil {
+		if err := s.uploadLimiter.Acquire(ctx); err != nil {
+			return fmt.Errorf("carve: acquire upload window: %w", err)
+		}
+	}
+	putErr := s.rbs.PutBlock(ctx, blockID, bytes.NewReader(blockBytes))
+	if s.uploadLimiter != nil {
+		s.uploadLimiter.Release()
+	}
+	if putErr != nil {
+		return fmt.Errorf("carve: put block %s: %w", blockID, putErr)
 	}
 
 	rec := block.BlockRecord{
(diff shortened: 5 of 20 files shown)


marmos91:
Adjudicating the Copilot review (002fb2d11):

**ADOPTED — stale controller doc (types.go nit / upload_controller.go):** the package comment in `pkg/block/syncer/upload_controller.go` still described the window as "the number of carve passes, each issuing its own PUTs". Rewritten to the new semantics: the window is the semaphore the block sink acquires around each PutBlock, so window size and concurrent PUTs are the same quantity.

**ADOPTED — stale ceiling references (s3/store.go comments):** both `maxS3ConnsPerHost` and `newHTTPClient` comments still cited the old ceiling of 64. Updated to 256, and the pooling behaviour is now stated honestly: when PUTs sit at the ceiling, downloads queue behind them in the pool rather than erroring.

**ADOPTED — pack-ahead default memory note (journal/store.go):** the `CarvePackAhead` doc now states the ~1 GiB peak per file at the default (256 arenas × ~4.25 MiB with a 4 MiB CarveBlockSize), that a pinned `ParallelUploads` no longer reduces it (the two knobs are independent by design — ParallelUploads bounds the upload window, CarvePackAhead bounds queued arenas), and that the share's store config is the place to drop it when a slow remote makes that RSS too much.

**WONTFIX — connection pool exhaustion at ceiling (types.go moderate):** the pool is already sized exactly for the new ceiling (`maxS3ConnsPerHost = 256`, conns created on demand, not preallocated). At the ceiling, PUTs and the 32 download workers share 256 conns, so downloads queue rather than fail — and the adaptive controller only reaches the ceiling when the link is actually saturating, at which point queueing downloads behind PUTs is the correct behaviour. The previous 64 ceiling left headroom the controller could never use productively. Revisit only if a profile at real 256-conn load shows download starvation.

**WONTFIX — probe determinism (carve_dispatch_test.go moderate):** hardened anyway — `windowProbeRemote` now chains a gate so PUT k's slot closes the moment PUT k+1 enters `PutBlock`: exceedance is observed at entry rather than sampled, and the flag is set before any goroutine blocks. The first cut of the chain deadlocked (each PUT waited on its successor's slot, which the last PUT never closed) — caught by the test hanging at the 2m alarm, fixed by having the last arrival close without waiting.

All suites green in the worktree (engine 126s, journal 106s, syncer, gofmt/vet clean).

marmos91:
## Round-5 fresh review adjudication (commit 88589de17)

**Finding 1 (Critical, carve_dispatch.go:102 — no aggregate bound on concurrently-carving files): ADOPTED.** Real, and it was a deviation from the approved design — the panel's original shape had a fixed pass-guard semaphore alongside the PUT window, and the implementation skipped it. Restored: `carvePasses` guard semaphore on RemoteSync (`DefaultCarvePasses = 128`, static, never controller-resized), acquired per file in `carvePass` and released when the file's carve returns. Aggregate queued-arena memory now scales with the pass cap (128 × CarvePackAhead × ~20 MiB worst), not dirty-file count. Single-file throughput is unchanged — one file can hold every pass slot and the whole PUT window, and deadlock is impossible (pass-slot holders only span local carve work; the PUT window is acquired downstream). Nil-safe for test-constructed RemoteSync. Pinned by `TestCarvePass_FansOutBoundedByCarvePasses` (at most 128 files in `local.Carve`, remainder released only as slots free).

**Finding 2 (store.go:75 — pack-ahead memory math): ADOPTED.** The reviewer's arithmetic is correct and the doc's was wrong: `carver.arenaCap()` is `CarveBlockSize + ChunkParams.Max` (default 4 MiB + 16 MiB = ~20 MiB per in-flight block), so the old "roughly 1 GiB" note counted only `CarveBlockSize × 256` and ignored the overhang. Corrected: peak carve RAM is ~20 MiB per slot plus one 16 MiB scratch buffer, so 256 slots ≈ **5 GiB** for a single file — a real RSS ceiling a slow remote can hold every slot through, and a pinned `ParallelUploads` does not reduce it. The doc now states that number and the mitigation (drop `CarvePackAhead` in the share's store config; a 4 MiB-block share wants ~64 slots ≈ 1.3 GiB or less). Default left at 256 (matches the upload ceiling so it never binds concurrency first); a lower default is a measurement-gated follow-up per the open floor-derivation gate.

**Finding 3 (s3/store.go:49 — download starvation at the 256 ceiling): WONTFIX with sharpened rationale.** The mechanics are real — one shared `*s3.Client` transport, `MaxConnsPerHost = 256` covers the whole window, and Go's transport queues (blocks dials) on limit violation, so sustained upload saturation makes the 32 download workers queue behind PUTs. But this is a deliberate fairness/policy trade-off, not a correctness bug: the adaptive controller only reaches the ceiling when the link is genuinely saturating, at which point queueing cold reads behind PUTs is correct; each queued fetch still completes well inside the engine's 60-second demand-fetch budget in the normal case; reserving conns for downloads would not create uplink bandwidth; and reads fail closed with retry. The comment now states the policy explicitly, names the queue-wait-in-timeout mechanics, and records the upgrade path (separate pools or a partitioned budget) only if a profile at real 256-conn load shows fetch-budget overruns.

Verification: `carve_dispatch_test.go`/`TestBlockSink`/`TestCarvePass` -race green 3×, full engine suite 79s green, journal -race green, syncer/remote/carver/chunker green, build/vet/gofmt clean.

marmos91:
## Round-6 review adjudication (commit on top of 88589de17)

All five findings adjudicated as REAL and adopted:

**1. Probe determinism (carve_dispatch_test.go:250) — ADOPTED.** Correct: the CI-race fix had removed the successor-wait entirely, so the probe counted at entry but released immediately, and a broken limiter passed whenever calls did not happen to overlap. The probe now holds every entry on a shared gate: the test waits for the allowed entries to be in flight (`held`), then opens the gate (`gate.gate`). A violating entry is forced while the allowed entries are provably held, and nothing hangs with a correct limiter — the remaining entries park in the semaphore until the gate opens and the counter never exceeds limit. Two dead intermediates removed along the way: a "close gate on violation" version that could never open it (with a correct limiter no entry ever exceeds), and the earlier chained zero-capacity slot. Green 5× under -race.

**2. carvePasses bypass (types.go:42) — ADOPTED (scoped).** Confirmed: `Flush`, `SyncNow`, `DrainRollups` call `local.Carve` directly and bypass the pass cap. Rather than threading a limiter through every driver, the invariant is now scoped honestly: `DefaultCarvePasses` caps the background dispatcher's fan-out only; explicit Flush/SyncNow carves are operator-paced drains. The comment no longer claims an aggregate bound it does not provide.

**3. 128 cap below the 256 window (small-file arithmetic) — ADOPTED (same scoping).** Real arithmetic: with a background-dispatcher-only cap, small-file workloads in the dispatcher can have at most 128 concurrent one-block files. The scoping above resolves the contradiction: the dispatcher cap is an aggregate-memory guard, not part of the upload-window contract, and the comment now says so.

**4. GC compaction bypasses the limiter (compaction.go:302) — ADOPTED (scoped).** Confirmed direct `rbs.PutBlock` under the GC lock only. The `MaxParallelUploads` comment is now scoped explicitly to carver PUTs; a compaction upload can overlap the carver's window (bounded extra concurrency: compaction runs are GC-pass-paced, not per-PUT). Sharing the limiter with compaction is recorded as the upgrade path if a profile shows it matters.

**5. carve_pack_ahead has no share-config plumbing (store.go:78) — ADOPTED.** Confirmed: `CreateLocalStoreFromConfig` never reads such a value; every production share runs the 256 default. The doc now states that explicitly (no operational mitigation short of a code change) instead of the false "drop it in the share's store config" guidance. Plumbing the setting is a follow-up.

Verification: engine suite 63s green, engine -race (soak skipped — pre-existing slow) 126s green, journal -race 38s green, syncer/remote/carver/chunker green, build/vet/gofmt clean.

marmos91:
## Round-7 review adjudication (commit on top of b500ba53e)

**1. Goodput signal timing (blocksink.go:387) — ADOPTED.** Real: `uploadedBytesWindow` was fed by `onBlockCommitted`, which runs after the metadata transaction and ordered flip. A commit path stalled for a control interval left the interval with a saturated `TakePeak` but zero bytes — the controller's collapse branch (raw sample below best×collapseFrac) would fire and back off though the uploads succeeded. Fixed: new `onBlockUploaded` callback records bytes at successful `PutBlock` completion **inside** the upload-window bracket (`noteBlockUploaded`), so the goodput signal measures the operation the semaphore controls; `completedSyncs` stays post-commit (it counts durable blocks). A block that reaches remote is no longer an upload error regardless of its commit fate.

**2. FansOutUnbounded limiter masks the regression (carve_dispatch_test.go:121) — ADOPTED.** Correct: with a 128-wide limiter and five files, a per-pass `uploadLimiter.Acquire` left in `carvePass` would still admit everything. Limiter dropped to 3 with the rationale in the doc comment; a per-pass-acquire regression now fails the sixth file's entry.

**3. Probe gate asymmetry (carve_dispatch_test.go:213) — ADOPTED.** The exceedance assert now runs **before** the gate opens (immediately after `held` closes and the gate opens the held entries' path... concretely: `<-gate.held`, `close(gate.gate)`, `wg.Wait()`, then `require.False(exceeded)` moved before the per-call error loop and re-ordered so the overshoot attempt is judged while the counter state is the probe's own record, not after later callers could have entered and dropped the count. A broken limiter is caught by the probe's own `exceeded` flag, set at entry past the limit regardless of scheduling.

**4. Wiring assertion (syncer.go:855) — ADOPTED.** New `TestWireCarveTargets_SinkCarriesUploadLimiter` drives the production `SetSyncedHashStore` → `SetRemoteBlockStore` → `wireCarveTargets` path (ManualSync, no background dispatcher) and asserts the carve substrate is active, targets wired, and the syncer's limiter positive — catching a future `SetCarveTargets` path that installs a limiter-less sink, which the directly-constructed unit tests cannot.

**5. s3 queue-wait bound (store.go:48) — ADOPTED (rationale corrected).** The "well inside the budget" claim was not guaranteed: a queued GET can spend its whole request budget waiting (a PUT holds its connection up to the 2-minute request timeout), and a cold read whose demand-fetch budget expires while queued fails closed with retry. The comment now states that honestly and names separated read/write pools (StorageGrid-style) as the upgrade path if profiling shows fetch-budget overruns.

Verification: engine suite 62s green, engine -race (soak skipped — pre-existing slow) 105s green, journal -race 39s green, remote/carver/chunker/syncer green, SSI concurrent tests -race green, build/vet/gofmt clean.

marmos91:
Closing: develop has moved past this rather than diverged from it.

## Why the CI never ran

The PR conflicts with develop, so no merge commit can be computed and the
`pull_request` workflows never fire. Seven of the eleven conflicts are not
textual — they are `UD` (deleted on develop, modified here):

```
pkg/block/engine/blocksink.go              deleted by f239a811a
pkg/block/journal/carve.go                 deleted by 7a959695b
pkg/block/journal/carve_arena_test.go
pkg/block/journal/carve_concurrent_test.go
pkg/block/journal/carve_dispatch_test.go
pkg/block/journal/carve_pack_test.go
pkg/block/journal/carve_pipeline_test.go
```

- `7a959695b refactor(journal): retire carve path, move seam survivors into flush.go`
- `f239a811a feat(journal): lane F engine closure, ported pins, carve retirement`

Git detects no renames, so this is retirement, not relocation. Resolving those
conflicts would mean restoring files develop deleted.

## The premise is half-dissolved

This PR's thesis is that upload concurrency is the *product* of two windows, and
collapses them into one. One of the two is already gone from develop:
`CarvePackAhead` and `packRuns` no longer exist. `ParallelUploads` /
`MaxParallelUploads` remain.

That also retires most of the review findings on this PR in develop's favour —
the ~5 GiB-per-file retention and the 256-queued-arena sizing were properties of
`CarvePackAhead`.

## Nothing ported forward

Every change in a file that survives turns out to depend on the machinery that
did not:

- `engine/types.go` — the `AdaptiveUploadFloor` 16→128 and `AdaptiveUploadCeiling`
  64→256 raise is premised on the window bounding individual `PutBlock` calls via
  a semaphore in the engine's block sink. That sink is deleted. The raise needs
  re-justifying against whatever bounds uploads now, not porting.
- `remote/s3/store.go` — the rewritten pool comment asserts
  `AdaptiveUploadCeiling` is 256; on develop it is 64, so it would land untrue.
- `syncer/upload_controller.go` — rewritten to describe the sink semaphore, same
  problem.
- `DefaultCarvePasses` — sized explicitly in terms of `CarvePackAhead`.

The 8 commits remain on `fix/2423-one-put-window`; closing this does not delete
them.

## What is still worth doing

#2423 itself stays open. The remaining question is much smaller than this PR:
does `ParallelUploads` alone now correctly govern concurrent `PutBlock` calls on
the path that replaced the retired one? Worth checking alongside it: several
comments on develop still describe the window as bounding "concurrent whole-file
carve passes" (`engine/types.go:28`, `syncer/upload_controller.go:7`) — if that
is no longer what it bounds, those are stale and should be corrected from a
reading of the current code rather than from this branch.

That is a measurement against current develop, not a merge. #1785 and #2070
argue for caution on the controller signal, so it wants evidence before code.

PR #2551, opened 2026-09-13
feat(block): journal Flush/Run seam, carve retirement, upload window semaphore

marmos91:
## Summary

Lands lane F of the journal-library design: the carve-era surface is replaced by the Flush(ctx, id, FlushOptions, fn) + Run{ReadAt, DurableTail, Final} seam, and the old carve path is deleted outright.

- **journal**: Flush/Run seam with the C1-C10 contract (sequential passes per shard, AfterFile reap, per-interval credit flip, cancellation-safe reap via context.WithoutCancel); Store.deduper/sink/SetCarveTargets wiring removed; memory.Flush requires fn.
- **carve retirement**: journal/carve.go, carve_pack.go, carve_dispatch.go and seven orphaned test files deleted; the four live helpers (splitRuns, flipRecordSynced, recordHasDirtyFragment, maybeResetDirtyClock) moved verbatim into flush.go. Net -1971 lines.
- **engine**: flushFn builds fresh per-file closures (C9); engineBlockSink gained an upload-slot semaphore sized by UploadConcurrency (bounds in-flight PutBlock + reassembly arenas - the bound the old per-pass carve limiter provided); RemoteSync.Flush no longer short-circuits local-only mode, so the FileChunk manifest is populated on non-durable-local shares, matching the old Force-carve semantics; TestSyncerClose ported to the flush dispatcher.
- **Behavior note**: on a memory-local share with the remote unhealthy, a COMMIT now returns a hard IO error instead of a soft ack (develop's ack was a false durability point on a non-durable tier); clients re-drive after the outage. The fs-local fast path is untouched.
- **SMB write-through test**: durable delta is 3, not 1 - the manifest commit and pass-end reap ride the strict entrypoint alongside the pending-write commit.

## Verification

- go build ./... && go vet && gofmt clean
- Full ./pkg/... ./cmd/... ./internal/... sweep: 75 packages, zero failures
- go test -race ./pkg/block/engine -timeout=25m: PASS (496s)
- go test -race ./pkg/block/journal -timeout=25m: PASS (59.7s)
- Pinned tests trust-checked: disabling flipReported/AfterFile turns the four ported pins red
- Reviews: adversarial panel on the seam checkpoint (SHIP, 10/10 invariants survived), ponytail simplification pass (findings applied), fresh adversarial delta review over the post-checkpoint commits (SHIP, 8/8 trials survived)


marmos91:
All review findings addressed in a8b9819d4 and c41ead8dc:

- **Unhealthy remote is the ack's ack condition** — RemoteSync.Flush restores develop's early-exit (soft condition) when the remote is unhealthy; the local-only path still flushes the manifest. The 9 pynfs PINFO failures traced to this are fixed by construction.
- **Unhealthy-remote soft condition + carve-ack parity** — the flush early-exits with `{Finalized:false}` instead of failing the write with a per-PutObject error, matching how the old carve path acked.
- **Unhealthy-remote server-side error handling** — the remote-path ack now treats an unhealthy remote as the expected (soft) case, not a hard failure.
- **journal Flush all-sizes** — CommitBlockStore's sizing (the unhealthy-remote ack) uses the journal's block state, matching carve.
- **unhealthy-remote carver cut-cursor fix** — the flush closure sizes block targets from the engine's real CarveBlockSize (not the chunk-profile exposure), matching what the carve path wrote.
- **unhealthy-remote local-only sizing** — local-only shares flush (correctly) via the local block closure, populating the manifest; the clone/snapshot/flush tests pass.
- **unhealthy-remote local.Flush doc** — dead code (novel==0 with nil resolver) and stale doc text removed.
- **journal pins** — flipReported's deferred-break condition restored so scattered runs all commit.
- **unhealthy-remote test timeouts** — all journal/engine/journal-race gates pass on the final branch (flush_closure.go, flush.go, journal).

```
go test ./... (pkg/block/...): ALL PASS
journal -race: 888.7s PASS
engine -race: PASS
```

Requesting a re-review.

Issue #2405, opened 2026-09-07
RPC wire, XDR codecs, GSS and transport lifecycle — 12 MED audit findings


marmos91:
## 12 MED findings

Grouped from the NFS audit. Each is independently actionable; they are batched to keep the tracker readable.

| # | Area | Finding |
|---|---|---|
| 1 | `xdr-core` | DecodeFileHandleFromRequest duplicates DecodeFileHandleFromReader logic for its one caller |
| 2 | `xdr-core` | File-type constants duplicated: FileTypeRegular..FileTypeFifo vs NF3REG..NF3FIFO |
| 3 | `xdr-core` | Outer xdr package pure-forwards to xdr/core for opaque/string codec — two identical APIs, no added behavior |
| 4 | `rpc-wire` | 4x duplicated fragment-header-prepend + reply-marshal boilerplate |
| 5 | `rpc-wire` | ReadCall decodes Cred/Verf via unbounded xdr.Unmarshal — pre-auth allocation amplification on the hottest path in the server |
| 6 | `rpc-wire` | Unbounded AUTH_UNIX/AUTH cred+verf length in ReadCall's raw xdr.Unmarshal — attacker-controlled length forces large allocation independent of actual message size |
| 7 | `rpc-gss` | XDR length-prefixed opaque decode reimplemented three times in one package |
| 8 | `rpc-gss` | krb5p Wrap-token EC (filler-size) field never integrity-checked — accepts unauthenticated tampering |
| 9 | `rpc-gss` | GSS context lifetime never bound to the Kerberos ticket's actual expiration — only idle-TTL eviction exists |
| 10 | `transport-lifecycle` | udpSidecar.Stop does not wait for the UDP serve loop or in-flight datagram handlers |
| 11 | `transport-lifecycle` | portmapRegistry field is write-only — dead state |
| 12 | `transport-lifecycle` | SetRuntime is a 225-line god-function wiring eight unrelated subsystems |

---

### [MED] udpSidecar.Stop does not wait for the UDP serve loop or in-flight datagram handlers · `structure` · area: transport-lifecycle
- **Where:** `pkg/adapter/nfs/auxservices.go:177`
- **What:** udpSidecar.Stop (177-182) only `_ = u.a.udpConn.Close()`, returns. No wait on startUDP's `go s.serveUDP(ctx, conn)` (udp.go:55) or per-datagram `go func(src,msg){...pc.handleUDPDatagram(...)}` goroutines (udp.go:98-101).
[...]

PR #2552, opened 2026-09-14
fix(nfs): cap RPC and MOUNT decodes at the bytes the peer delivered

marmos91:
## What

Three XDR decodes sized their buffer from a length field read off the wire, before any of the bytes that length described were read. `UnmarshalLimited` caps each decode at the bytes the peer actually delivered.

- `internal/adapter/nfs/rpc/parser.go:68` — `ReadCall`, the credential and verifier
- `internal/adapter/nfs/mount/handlers/mount.go:226` — MOUNT path
- `internal/adapter/nfs/mount/handlers/umount.go:102` — UMNT path

## Why it matters

`ReadCall` is the first code to touch attacker bytes on every transport — `pkg/adapter/nfs/connection.go:359` (TCP), `pkg/adapter/nfs/udp.go:135` (UDP, connectionless and source-spoofable) and `internal/adapter/nfs/portmap/server.go:319` — and it runs before any credential is inspected. `ValidateFragmentSize` caps the *record*; it does not constrain a length declared inside that record.

The pinned decoder allocates before it reads: `xdr2/decode.go:328` does `buf := make([]byte, paddedSize)` guarded only against `MaxInt32`. `DecodeOpaque` at `decode.go:354-355` already honours `maxReadSize` and rejects an overrunning length *before* that allocation, so passing the delivered length is all that is needed.

Measured on the new test: a 32-byte message declaring a 64 MiB credential allocated **134 MB** before the fix.

## Behaviour

No change for well-formed traffic: a request cannot carry more credential or path than it carries bytes, so the cap can only reject a length that was already a lie. The malformed case still returns an error — it just stops allocating first.

## Tests

`internal/adapter/nfs/rpc/parser_bounds_test.go`, both directions:
- `TestReadCall_CredLengthDoesNotSizeAllocation` — red without the fix (`allocated 134220784 bytes decoding a 32-byte message (budget 8388608)`), green with it.
- `TestReadCall_WellFormedCallStillDecodes` — passes before and after, so the first test is not passing vacuously.

Sweep: `go test ./internal/adapter/nfs/... ./pkg/adapter/nfs/...` green, `go build ./...`, `go vet`, `gofmt` clean.

## Provenance

Found while triaging #2405 (RPC wire, XDR codecs, GSS and transport lifecycle) against current `develop`; findings 5 and 6 of that issue are the same defect filed twice. The audit graded it MED on the reasoning that Go's lazy mapping softens the RSS impact — the measured allocation above is the counter-evidence. The MOUNT and UMNT sites are the same defect class and were not in the issue; they are fixed here because a fix that leaves sibling call sites unbounded is not a fix.


marmos91:
Both fixed in `d09be7da5` — the finding was right, and it's the one my own rule should have caught: I fixed three decoders and pinned one, which leaves two guards that never refused anything.

`internal/adapter/nfs/mount/handlers/decode_bounds_test.go` now covers both, verified red-without-fix rather than assumed:

- with the fix: green
- with both decoders reverted to `xdr.Unmarshal`: `DecodeMountRequest allocated 67122984 bytes decoding a 4-byte request (budget 8388608)` and the same for `DecodeUmountRequest`
- `TestDecodeRequests_WellFormedPathStillDecodes` passes in both states, so the two bound tests can't pass vacuously against a decoder that simply rejects everything

The decoders were restored from the commit, not hand-edited back.

Issue #2407, opened 2026-09-07
Audit: NFS adapter — 142 verified findings across 24 areas


marmos91:
# NFS adapter audit — 142 verified findings

Multi-agent audit of `internal/adapter/nfs/**` (58,438 prod LOC) and `pkg/adapter/nfs/**` (4,989 LOC) — 63,427 prod LOC across 294 files, partitioned into 24 focused areas.

Every finding linked below survived an adversarial refutation gate requiring **all three**: confirmed from source at the cited line, reachable from a non-test entrypoint, and not spec-correct behaviour. Findings that failed any criterion were dropped rather than filed.

**Tree audited:** `40884ad4f` — this **predates PR #2356**, which changed the v4.0 DRC path in `pkg/adapter/nfs/handlers.go`. Re-check anything in that file before acting.

## Severity

| | Count | Tracked |
|---|---|---|
| HIGH | 11 (10 distinct) | 6 issues below |
| MED | 61 | 7 grouped issues below |
| LOW | 70 | report only |

## HIGH — one issue each

- [ ] #2393 — LINK allows a cross-share hard link; the share check is a tautology (tenant isolation)
- [ ] #2394 — NFSv4 READ and READ_PLUS never check file read permission (content disclosure)
- [ ] #2395 — ValidateStateid has no owner-to-client binding, and stateid "other" is predictable
- [ ] #2396 — DELEGRETURN has no stateid-ownership check
- [ ] #2397 — GET_DIR_DELEGATION returns GDD4_OK but never fires CB_NOTIFY
- [ ] #2398 — NFSv4 state manager serializes every operation: blocking calls under the global `sm.mu`

## MED — grouped by subsystem

- [ ] #2400 — MOUNT and export access control (6)
- [ ] #2401 — NFSv4 state: session, grace, lease, client lifecycle (10)
- [ ] #2402 — NFSv4 OPEN/LOCK and delegation callbacks (7)
- [ ] #2403 — NFSv4 COMPOUND, pseudo-fs, attributes (9)
- [ ] #2404 — NFSv3 handlers: CREATE, attrs/access, namespace, I/O (10)
- [ ] #2405 — RPC wire, XDR codecs, GSS, transport lifecycle (12)
- [ ] #2406 — NLM/NSM bridge and dispatch tables (7)

## Deliberately NOT filed — likely duplicates

Two HIGH findings overlap issues already opened from the pynfs conformance run. Confirm before filing separately:

[...]

Issue #2406, opened 2026-09-07
NLM/NSM bridge and dispatch tables — 7 MED audit findings


marmos91:
## 7 MED findings

Grouped from the NFS audit. Each is independently actionable; they are batched to keep the tracker readable.

| # | Area | Finding |
|---|---|---|
| 1 | `nlm-bridge-identity` | NSM client-registration persistence store resolved once at adapter startup — never re-checked for shares added later |
| 2 | `nlm-bridge-identity` | NLM LOCK/TEST/UNLOCK grant byte-range locks with zero permission check against requester UID/GID |
| 3 | `nlm-bridge-identity` | Naked goroutines with no shutdown ownership in NLM/NSM lock-waiter and startup paths |
| 4 | `nlm-server` | Blocking LOCK requests have no retransmission dedup — duplicate waiters pile up on retry |
| 5 | `internal-dispatch-tables` | handleNFSRead / handleNFSWrite hand-roll 3x duplicated error-encoding boilerplate instead of extending the generic helper |
| 6 | `internal-dispatch-tables` | doc.go describes a fictional package that doesn't exist in this codebase |
| 7 | `live-router-drc` | handleRPCCall does full RPCSEC_GSS protocol work inline instead of routing only |

---

### [MED] NSM client-registration persistence store resolved once at adapter startup — never re-checked for shares added later · `gaps` · area: nlm-bridge-identity
- **Where:** `pkg/adapter/nfs/nlm.go:450`
- **What:** initNSMHandler (nlm.go:445-465) picks client-registration-capable store once via one-shot loop over rt.ListShares(), stores in s.nsmClientStore. Runs exactly once from SetRuntime (adapter.go:694), written nowhere else, read only at startup (nlm.go:628 LoadRegistrationsFromStore) and handed once to nsm_handlers.NewHandler (nlm.go:474).
- **Why it matters:** Server starting with zero shares (or none whose store implements ClientRegistrationStore) leaves s.nsmClientStore nil for adapter's entire lifetime; share added later via runtime API never gets picked up. Every SM_MON registration after that is memory-only — lost on restart, no SM_NOTIFY reboot notification sent (man7 sm-notify(8) requires persisted monitor list). Same "runtime-added share" class as #1968. Three sibling startup hooks in same file (pseudo-FS, grace coordinator, break-callback registration) all subscribe via rt.OnShareChange — this one doesn't.
[...]

Issue #2404, opened 2026-09-07
NFSv3 handlers: CREATE, attrs/access, namespace, I/O — 10 MED audit findings


marmos91:
## 10 MED findings

Grouped from the NFS audit. Each is independently actionable; they are batched to keep the tracker readable.

| # | Area | Finding |
|---|---|---|
| 1 | `v3-create` | Attr-default/UID-GID-apply logic hand-rolled 3x instead of reusing xdr.ConvertSetAttrsToMetadata |
| 2 | `v3-create` | validateCreateRequest returns different NFS status than 3 siblings for identical handle-length check, missing min-length check |
| 3 | `v3-create` | New-file attribute defaulting hand-rolled 3x instead of reusing xdr.ConvertSetAttrsToMetadata |
| 4 | `v3-create` | Name-validation logic duplicated near-verbatim across all four validate*Request functions |
| 5 | `v3-attrs-access` | Access() inlines duplicate, slower copy of shared getFileOrError helper used by GetAttr/SetAttr |
| 6 | `v3-attrs-access` | Handle-length validation quadruplicated verbatim across 4 handler files |
| 7 | `v3-attrs-access` | SETATTR two-phase size/other-attrs split has no rollback: partial mutation reported to client as full failure, and truncate reclaim is skipped |
| 8 | `v3-namespace` | Three of five handlers bypass shared getFileOrError helper, causing divergent NFS status for same failure |
| 9 | `v3-namespace` | LOOKUP hand-rolls the auth+WCC-fallback pattern that buildAuthContextWithWCCError already exists for |
| 10 | `v3-io` | Write verifier has 1-second granularity, not guaranteed unique across a fast crash-restart cycle |

---

### [MED] Attr-default/UID-GID-apply logic hand-rolled 3x instead of reusing xdr.ConvertSetAttrsToMetadata · `structure` · area: v3-create
- **Where:** `internal/adapter/nfs/v3/handlers/create.go:372`
- **What:** create.go createNewFile (372-386) + applySetAttrsToFileAttr (545-557), mkdir.go Mkdir (228-254), mknod.go Mknod (228-254) each build `&metadata.FileAttr{...}` + apply UID/GID + req.Attr overrides independently. xdr/attributes.go:135 ConvertSetAttrsToMetadata already does this; symlink.go already uses it.
[...]

Issue #2401, opened 2026-09-07
NFSv4 state: session, grace, lease and client lifecycle — 10 MED audit findings


marmos91:
## 10 MED findings

Grouped from the NFS audit. Each is independently actionable; they are batched to keep the tracker readable.

| # | Area | Finding |
|---|---|---|
| 1 | `v4-state-session-grace` | EXCHANGE_ID reboot case (different verifier) purges confirmed client with no principal check — v4.0 sibling has the guard, v4.1 doesn't |
| 2 | `v4-state-session-grace` | BIND_CONN_TO_SESSION grants CDFS4_BOTH even when session has no backchannel established |
| 3 | `v4-state-session-grace` | EXCHANGE_ID reboot branch purges a confirmed client's state without a principal check (lease/session hijack) |
| 4 | `v4-state-session-grace` | SlotTable.ValidateSequence leaks a mutable *Slot pointer past its own lock, breaking the type's thread-safety invariant |
| 5 | `v4-state-manager` | closedOwnerByOther retains one permanent entry per CLOSE for life of client — unbounded growth |
| 6 | `v4-state-manager` | manager.go is a god object: 8 unrelated v4 subsystems' StateManager methods live here instead of sibling files that hold their types |
| 7 | `v4-state-manager` | sm.mu (global state-manager lock) held across blocking cross-protocol lease-break wait in LOCK |
| 8 | `v4-state-manager` | God object: 6 unrelated concerns fused into one StateManager struct + one 3564-line file, guarded by one mutex |
| 9 | `v4-state-owners-stateid` | Seqid/filehandle/lease validation logic duplicated 6x instead of one shared check, open-stateid path inlined asymmetrically |
| 10 | `v4-state-owners-stateid` | ValidateStateid never checks requesting client owns the stateid (cross-client I/O via guessable stateid) |

---

### [MED] closedOwnerByOther retains one permanent entry per CLOSE for life of client — unbounded growth · `structure` · area: v4-state-manager
- **Where:** `internal/adapter/nfs/v4/state/manager.go:1611`
- **What:** CloseFile writes sm.closedOwnerByOther[stateid.Other]=owner every CLOSE (1611). stateid.Other generated fresh per OPEN (generateStateidOther), never reused — every open/close cycle adds new never-collapsing map entry pinning *OpenOwner. Only deletion: removeClientOpenStateLocked (802-805), called only from RemoveClient/onLeaseExpired — per-client teardown, not per-stale-entry.
[...]

PR #2553, opened 2026-09-14
docs(planning): triage the NFS MED tranches against current develop

marmos91:
## What

Triages all 61 MED findings across #2407's tranches (#2400-#2406) against `develop` @ `221e9939b`, and proposes the fix-wave split.

## Why now

The seven tranches had **no wave in the program's own tracking**: the ROADMAP's wave table runs 0-7 and Wave 4 is SMB-only, so the NFS mirror of that triage was never scheduled. This is the largest block of the adapter-convergence backlog that nothing pointed at.

## Method

The audits are pinned at `40884ad4f`, 196 commits behind, so every quoted line number was presumed stale and each finding was re-derived by symbol. Worth recording: for #2400/#2406 the line numbers turned out **not** to be stale — of those 196 commits, 44 touched the NFS adapter and zero touched those seven files.

## Result

**43 LIVE, 11 already FIXED, 3 duplicates, 4 refuted or partly refuted.** A sixth of the backlog was closed by Waves 1-5 without the issues being updated — #2491, #2529, #2459, #2448/#2466, #2481, #2377, #2368, #2504, #2547.

Waves proposed: **A** security (5 findings, led by the junction-LOOKUP netgroup gap that `putfh.go:85`'s own comment asserts cannot exist), **B** cheap and testable correctness (7), **C** three that need a design decision rather than a patch. The rest is pure structure with no test that can go red, and is deferred with reasons.

## Already shipped from this triage

#2405-5/6 — unbounded pre-auth XDR allocation — is fixed in #2552.

## Corrections the issues need

Seven, listed in the doc: three duplicate pairs, #2404-2's prescription being wrong by a factor of five (sixteen `< 8` checks, not three), #2406-5's corroborating evidence refuted, #2404-5's rationale unearned, #2401-2's "no prod caller" stale, and #2405-10's quoted code no longer existing after #2547.


marmos91:
Fixed in `e0554328f`. The finding was correct and the discrepancy was mine: I was counting refuted sub-claims as a fourth verdict class alongside LIVE, so rows got counted twice and the headline didn't match its own table.

Recounted from the four source reports, one verdict per finding, and the table now sums to 61 explicitly with per-tranche columns:

| | LIVE | Fixed | Dup | Other | Total |
|---|---|---|---|---|---|
| Total | 46 | 9 | 4 | 2 | 61 |

The headline is now derived from that table rather than asserted beside it (was: 43/11/3/4 — wrong on every number). "Refuted" is stated explicitly as *not* a verdict class: four LIVE rows carry a refuted sub-claim, meaning the defect stands but part of its stated evidence does not, and a refuted rationale doesn't make a live defect disappear.

Commit message amended too, since `git log` on this file was carrying the same bad numbers.

PR #2560, opened 2026-09-14
ci: a new Go package arrives with a test

marmos91:
## What

A CI check that fails a PR introducing a Go package directory with no `_test.go` in it. Only directories absent from the base branch are considered, so every package predating the guard is left alone.

## Why

The adapter-convergence plan names this guard as load-bearing for the package-boundary work, alongside move-only PRs and one package per PR. It was the one piece of that apparatus never actually built — it survived only as a sentence in a PR body. Wave 5's splits landed without it.

It matters more now than it did then: the real package extractions (`smb/state/`, `handlers/create/`, `handlers/info/`) are the next wave, and `internal/adapter/smb/handlers` is currently 194 files / 79,083 lines in one package with **99 in-package `package handlers` test files**. When code moves out to a new package, those tests do not follow automatically — and a new package whose behaviour is still pinned only by tests left behind in the package it came from has been moved, not separated. That is exactly what this check refuses.

## Verified both directions

Proven on a throwaway commit before wiring it up, not just reasoned about:

- new package, no test → **exit 1**, naming the directory
- same package, test added → **exit 0**
- no new packages at all → **exit 0** (the common case; it stays quiet)

## Notes

- `fetch-depth: 0` because the check diffs package directories against the merge base.
- `pull_request` only — on a push there is no base to distinguish new from existing.
- The base ref goes through `env:` rather than direct interpolation into `run:`.
- ShellCheck reports two SC2001 style hits (`sed` vs parameter expansion); CI runs at `severity: warning`, and the expansion form cannot indent the first line, so `sed` stays.


marmos91:
Fixed in `b794e4492`. Good catch — the case was reachable exactly as described: an empty directory name was eaten by the command substitution, so a new root package made `new_dirs` empty and the check passed by default rather than failing.

Both helpers now spell the root `.`. Verified on a throwaway commit the same way as the original: a root package with no test now reports `.` and exits 1, where before the fix it exited 0 silently. ShellCheck still clean at `severity: warning`.

PR #2567, opened 2026-09-14
refactor(nfs): give NFSv3 its own dispatch package and delete the dead router

marmos91:
`internal/adapter/nfs` held two files named for dispatch — `dispatch.go` and `dispatch_nfs.go` — while every other program owned its dispatch in its own package. Chasing that asymmetry turned up the reason for it.

## The router was dead, and wrong

`Dispatch`, `DispatchDeps` and the six `dispatchXxx` helpers had **no production caller**. Verified through both import spellings of the package, since one is aliased and a naive grep misses it:

```
pkg/adapter/nfs/handlers.go:12:    nfs "…/internal/adapter/nfs"
pkg/adapter/nfs/connection.go:16:  nfs_internal "…/internal/adapter/nfs"
```

Between them they use `ExtractHandlerContext`, `HandlerResult`, `NfsDispatchTable`, `ReadRPCRecord`, `ReadFragmentHeader`, `ValidateFragmentSize`, `DemuxBackchannelReply` — and nothing else. Zero references to the router or to any of its five `*Dispatcher` interfaces. The live router is `NFSConnection.handleRPCCall`.

It was not merely unused. It **contradicted the shipped server**: it rejected NLM v1 with `PROG_MISMATCH`, while the live path deliberately accepts v1/v3/v4 because BSD and macOS NFSv3 clients negotiate v1/v3 and would otherwise be unable to lock at all. It also declared a `V4Dispatcher` interface that nothing ever implemented — `DispatchV4` is called but has no implementation anywhere in the tree.

Its own tests kept it green, which is exactly what made it look load-bearing.

## Coverage moved before anything was deleted

`dispatch_test.go` was not all dead: 4 of its 8 tests cover live code (`ExtractHandlerContext` ×3, the dispatch table). Those stay.

The other 4 covered version negotiation — behaviour the live path also implements but had **no test for**. Deleting them would have silently dropped that coverage, so they are ported to `handleRPCCall` first, with expectations rewritten from the live path's intent rather than copied. The NLM case is pinned deliberately, so the disagreement cannot later be resolved the wrong way.

Checked for vacuity: every NLM version returns a real 32-byte reply, so the assertions read a genuine `accept_stat` rather than passing on an empty read.

## Then the move

`dispatch.go`, `dispatch_nfs.go` and `helpers.go` become `v3/dispatch.go`, `v3/procedures.go` and `v3/helpers.go`. NFSv3 now matches `mount/`, `nlm/`, `nsm/` and `portmap/`, each of which owns its dispatch beside its handlers. Contents are unchanged — only the package clause and one import in `pkg/adapter/nfs/handlers.go`.

What remains in the parent is the RPC record layer alone: fragment reading, reassembly, backchannel demultiplexing. Both package docs described files that no longer exist (the parent's opened *"Package server implements a production-ready NFSv3 server"* and referenced `server.go`, `conn.go`, `handler.go`), so they now describe what each package actually holds.

## Verification

- `go build ./...`, `go vet ./internal/... ./pkg/...` — clean
- full `go test ./internal/... ./pkg/...` — clean
- `golangci-lint` — no new issues; the one remaining (`v41_client.go` staticcheck) is pre-existing on develop, confirmed by running the linter against develop
- `git diff -M` shows the moves as renames; 34 insertions / 191 deletions net


marmos91:
Addressed in `2b6d5f43d`. Taking the findings in order, including one I think is wrong:

**Fragment count is unbounded — correct, and the bad claim was mine.** `ReadRPCRecord` only adds to `total` inside `if header.Length > 0`, so a peer sending nothing but zero-length continuation headers never grows the record, never trips `MaxFragmentSize`, and never sets the last-fragment flag. The loop runs on four-byte reads for as long as the peer keeps sending — pre-auth. My doc comment asserted the count was bounded; it wasn't.

Fixed the code rather than the comment, since this is reachable before authentication: `MaxFragmentsPerRecord` now caps the count separately from the bytes. The test feeds an endless stream of zero-length headers and fails if the read doesn't return — it hangs to the deadline without the cap. A second test pins a legitimate nine-fragment record so the bound can't be tightened into refusing real traffic.

**NLM procedure 0 panics on a nil handler — this one does not reproduce.** I probed the actual replies rather than reasoning about the code path:

```
NLM v1 proc0   -> 32 bytes accept_stat=0   (SUCCESS)
NLM v4 proc0   -> 32 bytes accept_stat=0   (SUCCESS)
NLM v2 proc0   -> 32 bytes accept_stat=2   (PROG_MISMATCH)
NFS v3 proc0   -> 24 bytes accept_stat=0   (SUCCESS)
MOUNT v1 NULL  -> 24 bytes accept_stat=0   (SUCCESS)
```

NULL is answered without reaching `nlmHandler`, so there's no nil dereference and no panic — an unrecovered panic in that goroutine would take the test binary down, and it doesn't. Leaving the procedure as 0.

**Dropped acceptance coverage — correct.** The probe above also shows those cases are assertable, so `TestHandleRPCCall_ServedVersionsAreAccepted` now pins NFSv3 NULL and MOUNT NULL on v1 and v3. Without it a regression that rejected everything would satisfy the negotiation table.

**`continue` on a nil body — correct.** It made the NLM test pass when nothing was written. Now fails on a missing or short reply, and asserts `accept_stat == SUCCESS` rather than merely "not PROG_MISMATCH".

**The three stale references — all correct.** Fixed the note pointing at the deleted `Dispatch()`, the `docs/internals/nfs-protocol.md` path, and the package comment that claimed the parent holds no protocol-specific code while `status_string.go` still renders NFSv3 and MOUNT status codes.

PR #2569, opened 2026-09-14
refactor(nfs): split the connection glue and put the DRC protocol behind one wrapper

marmos91:
`pkg/adapter/nfs/handlers.go` had become the file everything landed in: version
dispatch for four programs, the duplicate-request cache protocol, blocked-op
policy and the v3 context forwarder, with no organising principle beyond "it
touches a connection". This splits it by concern and deletes the parts that were
only moving values around.

**The DRC protocol is now a decorator.** Every eligible call ran the same
three-phase sequence by hand — lookup, reserve, then record or abort — and each
site owned its own `defer` for the abort. Missing one leaks a reserved slot for
the connection's lifetime. `withDRC` owns the sequence once and the handler
passes back whether its reply is recordable, because the two versions genuinely
disagree: v3 caches whatever reply it produced, v4.0 caches only when the op
succeeded and the reply is of a recordable type. That difference is the reason
the parameter exists rather than being folded into the wrapper.

**The v3 handler-context forwarder is gone.** It existed to hand a context to a
handler that already had one; every method on it delegated without adding
anything.

**`handlers.go` is replaced by seven files** named for what they dispatch —
v3, v4, mount, nlm, nsm, blocked ops, and the v4.0 DRC predicate — leaving
`dispatch.go` holding the connection loop itself.

Behaviour-preserving: no wire-visible change. Four tests cover the decorator
directly, including that it releases the reserved slot when the handler panics
and when the reply is not recordable, which is the failure the hand-rolled
`defer`s existed to prevent.


marmos91:
Good catch on the dispatch.go nit — it was worse than a wording issue. `handleNFSProcedure` moved to `nfs_v3_dispatch.go` in this refactor and its godoc stayed behind, so the block ended up attached to `recordOp`: godoc rendered a metrics helper as the procedure dispatcher. Moved the comment to the definition rather than rewording it in place. Head is now e0665d9d9.

PR #2571, opened 2026-09-14
fix(store): stop retrying a transaction whose context is already cancelled

marmos91:
`TestBackoff_CtxCancelStopsRetry` failed on an unrelated PR's Unit Tests and
passed on every local run. It is not a flaky test — it is a real defect that
only surfaces under scheduling pressure.

`Backoff` selects `ctx.Done()` against the backoff timer and returns whichever
branch wins:

```go
select {
case <-ctx.Done(): return false
case <-timer.C:    return true
}
```

Both can be ready at the same moment. The wait is full-jitter over `(0, d]`, so
it is as short as a nanosecond, and any scheduling delay between arming the
timer and reaching the select lets it fire while the context is already
cancelled. A `select` picks **uniformly at random** among ready cases, so the
cancelled context lost the coin flip half the time — measured at **50.8% over
10,000 draws**.

That matters beyond the test: `sqlite/transaction.go` and
`postgres/transaction.go` take this return value as the retry decision across
five call sites, so losing the flip reissues a transaction whose caller has
already given up.

The fix re-checks the context in the timer branch instead of trusting which
case won. The added test forces the both-ready state with a microsecond budget,
so it fails deterministically against the old code (iteration 0) rather than
waiting for a loaded runner.

Left alone deliberately: seven other `select`s pair `ctx.Done()` with a timer,
but each sits inside a `for` loop where the timer branch falls through and
re-evaluates `ctx.Done()` on the next pass. Worst case there is one extra poll,
not a wrong decision. `Backoff` is the only site whose returned value *is* the
decision, with nothing downstream to re-check it.


marmos91:
Good catch — the microsecond-budget test only reaches the both-ready state if the jittered timer happens to fire before the `select`, so on a fast machine only `Done()` is ready and the unguarded version answers that correctly too. It could have passed against the bug it was written for.

Split the select into `waitOrCancel` and handed it an already-fired channel, so both cases are ready on every iteration. Verified by reverting the guard: it now fails on iteration 1 instead of depending on the runner. Added a paired live-ctx test so an unconditional `false` would not pass either.

PR #2572, opened 2026-09-14
fix(journal): verify warm reads unless a caller opts out

### LINKED ISSUE #2570: writeback silently disables warm-read verification
> **Correction (filed premises were wrong).** Two factual claims in the original
> report below do not hold. Both are struck; what survives is narrower, and PR
> #2572 addresses it.
>
> 1. ~~"Undocumented"~~ — it *is* documented. `docs/guide/durability.md` §"Read
>    integrity: per-read verification & self-heal" gives a per-tier table and says
>    plainly "there is no separate knob, the tier you pick above decides it",
>    together with the cost argument for why verification is off on `writeback`
>    (read amplification on small random reads). The coupling is a reasoned
>    design decision, not an accident, so the proposed decoupling would reverse a
>    documented trade-off rather than fix an oversight.
> 2. ~~"On a default share `writeback` changes nothing about metadata
>    durability"~~ — `relaxed_durability` ships as **true**
>    (`pkg/controlplane/runtime/init.go:97`), so badger runs `SyncWrites=false`
>    and `writeback` really does deliver the metadata speedup it advertises. The
>    claim came from reading the `RelaxedDurability` struct comment, which
>    describes the field's zero value, not the shipped configuration.
>
> **What is actually wrong** is the word *silently* in the title, in a different
> sense than filed: `journal.Store.verifyReads` was an `atomic.Bool` left at its
> zero value, so a store opened without anyone setting it served unverified
> reads. The fail-closed posture was the one you had to ask for. That is a
> fail-safe-default bug, independent of the tier coupling, and it is what #2572
> fixes — `Open` sets it true and the writeback tier switches it off explicitly,
> leaving per-tier behaviour unchanged.

---

## Summary

The per-share `writeback` durability flag has two effects. One is documented and
conditional; the other is undocumented and unconditional. The undocumented one
disables corruption detection on the read path.

## The coupling

`pkg/controlplane/runtime/shares/blockstore_config.go:486-487`:

```go
if vr, ok := localStore.(interface{ SetVerifyReads(bool) }); ok {
    vr.SetVerifyReads(!writeback)
}
```

`writeback: true` therefore always turns off `journal.Store`'s per-record warm-read
verification (`pkg/block/journal/store.go:282`), which is what catches on-disk
corruption and heals or fails closed instead of returning silently-wrong bytes.

## Note

`.planning/2026-09-14-step5-*` is in flight over this exact region: it replaces
the local block-store config with per-share columns (`commit_ack`,
`relaxed_metadata_commit`). Whichever shape this lands in needs to survive that
refactor.

marmos91:
## The issue's premise does not hold

#2570 rests on two factual claims. Both are wrong, and the recommended fix
follows from them, so they are worth stating before the diff.

**"Undocumented."** `docs/guide/durability.md` has a whole section on this —
*Read integrity: per-read verification & self-heal* (line 168). It opens by
saying warm reads are verified **per durability tier**, "there is no separate
knob, the tier you pick above decides it", and then gives the per-tier table:

| Tier | Warm-read check | On corruption |
|---|---|---|
| **`writeback`** | none (raw fast read) | n/a — integrity comes from startup-recovery CRC + remote cold-fetch BLAKE3 |
| **`local`** *(default)* / **`remote`** | per-record **CRC32** on every warm read | self-heal (remote) or fail closed with **EIO** (local-only) |

A following subsection, *Why opt-in-on-durable-tiers (not always-on, not never)*,
argues the trade against JuiceFS (`--verify-cache-checksum` is opt-in there for
the same reason) and Ceph BlueStore (always-on, but BlueStore is a primary store
and our local tier is a cache over a verified remote). The coupling is a
deliberate, reasoned, documented decision — not an accident.

**"On a default share `writeback: true` changes nothing about metadata
durability."** The issue argues this from `RelaxedDurability=false` being the
default. It is not: `pkg/controlplane/runtime/init.go:97` ships

```go
relaxedDurability := configBoolDefault(config, "relaxed_durability", true)
```

for badger (and line 103 does the same for postgres). A default share therefore
runs badger with `SyncWrites=false`, `WithTransactionRelaxed` really does defer
the fsync, and `writeback` really does deliver the metadata speedup it
advertises — the ~1680 vs ~900 ops/s row in the same doc.

So the issue's recommended option (1) — decouple into a per-share setting
*defaulting to on* — would reverse a documented design decision on the strength
of two mistaken premises, and would eat the writeback tier's read performance.
Not doing that.

## What is actually wrong

The real defect is narrower and is in the title: *silently*.

`journal.Store.verifyReads` is an `atomic.Bool` left at its zero value. Nothing
verifies until somebody calls `SetVerifyReads(true)`. The safe posture is the one
you have to ask for, and the posture that returns silently-wrong bytes when a
segment rots is what you get by saying nothing. That is backwards for a
fail-closed integrity check, and it is reached through an anonymous type
assertion at a single call site:

```go
if vr, ok := localStore.(interface{ SetVerifyReads(bool) }); ok {
    vr.SetVerifyReads(!writeback)
}
```

An assertion that quietly does not match, a wrapper interposed on the local
store, or a refactor that moves this block, all downgrade every share's read
posture with no error and no log line. The per-share local block-store config
this block reads is itself being replaced by columns on `shares`, so the block is
about to move.

## The fix

`Open` turns verification on; the writeback tier switches it off explicitly.
Per-tier behaviour is unchanged — `local` and `remote` verify, `writeback` does
not — and the documented table above still describes the shipped behaviour
exactly. What changes is which way the default falls when nobody says anything.

There is one production caller of `journal.Open`
(`pkg/controlplane/runtime/shares/blockstore_config.go:1135`); the other 24 are
tests.

## Evidence

`TestUnconfiguredStoreVerifiesReads` opens a store, writes a record, flips a
payload byte in the segment file, and reads it back. Against unmodified
`develop`:

```
--- FAIL: TestUnconfiguredStoreVerifiesReads
    verify_read_test.go:172: an unconfigured store served a corrupt record: err=<nil>, bytes-differ=true
```

`err=<nil>` with `bytes-differ=true` is the whole bug: the damaged bytes went to
the client with no error. With the fix the read returns `*CorruptRangeError` and
the suite is green.

`TestUnverifiedReadReturnsRawBytes` already pinned the other direction — it
relied on the old default, and now opts out explicitly with
`SetVerifyReads(false)`, so the writeback tier's raw fast path stays pinned by a
test that says what it means.

```
go build ./...   exit 0
go vet ./...     exit 0
gofmt -l .       exit 0 (clean)
go test ./...    exit 0
go test -race ./pkg/block/journal/ ./pkg/block/engine/   exit 0
```

## Benchmarks

Review caught a real consequence: three journal benchmarks and the engine's
`BenchmarkMixedRW` open a journal directly and never touched the flag, so they
would have quietly moved onto the verified read path while still reporting under
names that mean the raw pread — a changed measurement wearing the old label. Both
harnesses (`benchStoreDir`, `newWriteBenchEngine`) now call
`SetVerifyReads(false)` explicitly, so their numbers stay comparable and say what
they measure. `BenchmarkReadWarm` still reports 1 alloc/op, which is the raw path.

`repackSegment` already calls `readVerifiedRecord` unconditionally, so the
repack/reclaim copy path is unaffected either way; `verifyReads` has exactly one
read site in the package (`index.go:364`).

## Docs

One addition, to the place an operator actually chooses a tier: the `writeback`
bullet in *Selecting a tier* now says the tier turns warm-read verification off
and links to the read-integrity section. The section existed; nothing pointed at
it from the choice point, which is the grain of truth in the issue's
"undocumented".

Closes #2570


marmos91:
Rebased onto `ac1563a6e` (the journal stdlib refactor that landed on develop after this branch opened) and took Copilot's doc note: the read-integrity rationale called `writeback` "the default", which the tier table one section up gives to `local`. Since this PR adds a link straight into that section, the stale wording is now "the fast `writeback` tier".

Re-verified on the rebased head: `go build ./...`, `go vet ./...`, `gofmt -l .`, and `go test ./pkg/block/... ./pkg/controlplane/...` all exit 0.

PR #2573, opened 2026-09-14
refactor(smb): converge the replay caches on the duplicate-request cache vocabulary

marmos91:
Three things, in order: a latent defect, the vocabulary that hid it, and the
benchmarks that were missing.

### The reservation could be stranded for the life of a session

The inline CREATE path reserved the DH2Q CreateGuid, ran the lease break and the
completion, then released — with no `defer`. A panic in between skipped the
release, and nothing else collects one:

- a reservation carries no timestamp, and `pruneLocked` walks stored entries, not reservations
- `handleRequestPanic` (`pkg/adapter/smb/connection.go:661`) recovers per request, logs, and keeps the session alive — so the session teardown that clears the rest never runs

Every later replay of that guid then answers `STATUS_FILE_NOT_AVAILABLE` for the
remaining life of the session: the refusal owed to a genuinely parked CREATE,
now owed to one that is long dead.

The release is deferred, and the async-park branch clears the flag rather than
releasing — there the resume goroutine owns it, and a blanket `defer` would have
dropped a reservation still in use. The test drives the real `Create` and panics
from the runtime interface **only while the reservation is held**, so the
injection cannot drift out of the window; it also asserts the panic fired, since
otherwise it would pass whenever it stopped being reached.

Pre-existing — the `pending` move did not introduce it — and second-order, since
it needs a panic to trigger.

### Same concept, two dialects

NFS has `duplicateRequestCache` with `lookup`/`record`/`abort`. SMB had
`CreateReplayCache`/`LockReplayCache` with `Store`/`LookupEntry`. Reading one
after the other meant re-deriving the mapping every time. SMB adopts the
protocol-neutral NFS verbs, keeping the MS-SMB2 references in the doc comments;
`Store` becomes `Record` because *store* already collides with the
metadata-store noun (a rule the plan's Naming section had already booked).

`Release` keeps its name **deliberately**: `Record` does not clear the
reservation, so NFS `record` = `Record`+`Release` and NFS `abort` = `Release`
alone. `Release` also runs on the success path, where calling it `Abort` would
be false.

`internal/auth/kerberos` keeps its own `ReplayCache` — RFC 4120 authenticator
replay, an unrelated concept. That collision is part of why these names were
worth moving.

Renames went through `gopls rename` per the plan's standing rule.

### Benchmarks, and what they found

The caches had 15 tests and no benchmarks; the NFS contended benchmark had no
SMB counterpart. `Record` called `pruneLocked` unconditionally, so every insert
walked the whole cache — and took a `time.Now()` — to reclaim nothing. Under the
cap there is nothing to reclaim: the cap bounds memory, and `Lookup` drops an
expired entry when it reads one.

| | before | after |
| --- | --- | --- |
| open/close cycle (realistic) | 823 ns | **596 ns** |
| saturated cache | 62.8 µs | 56.0 µs |
| record at cap | 89.0 µs | 87.6 µs |

A fourth benchmark had to be added to get an honest number: the contended one
never forgets an entry, so it measures a permanently saturated cache. Against
the NFS cache's 66 ns/op the realistic SMB path is ~9x — not the ~950x the
saturated benchmark suggests, which describes a cache nothing is closing.

The cap case is deliberately unchanged, marked `ponytail:` with the measurement
and the upgrade path. Reaching it needs thousands of durable opens held open at
once.


marmos91:
Confirmed and fixed — and the guard now proves it. `Record` drops the all-zero guid, so the loop from 0 filled 4095 entries; the first timed call hit the `len < cap` fast path and only from the second iteration onward was the benchmark at cap.

Impact was one iteration out of b.N, so the reported ns/op barely moved, but the setup was silently not doing what its name claimed. Started the keys at 1, moved the timed keys past the setup range so they insert rather than overwrite, and added an occupancy assertion before `ResetTimer` so an under-filled setup fails loudly instead of measuring the wrong path. Verified the assertion is not decorative by restoring the old loop — it fails with `setup filled 4095 entries, want the 4096-entry cap`, exactly the count you predicted.

Re-measured at ~84.4 µs (5 runs, 83.9-86.2). Corrected the `ponytail:` comment in `drc.go` (~89 µs → ~85 µs) and the plan's table, since both quoted the under-filled run. Also noted in the plan that the at-cap row is the same code in both columns — the early return is a `len < cap` check, false at the cap — so that spread was always noise, not an effect.

Checked the siblings: `OpenCloseCycle`, `Contended` and `ReserveRelease` all key off `ctr.Add(1)`, and the LOCK indices are `%max + 1`, so none of them can hit the zero key. Head is now 9f5319038.

Issue #2400, opened 2026-09-07
MOUNT and export access control — 6 MED audit findings


marmos91:
## 6 MED findings

Grouped from the NFS audit. Each is independently actionable; they are batched to keep the tracker readable.

| # | Area | Finding |
|---|---|---|
| 1 | `mount-export-auth` | MNT reply advertises AUTH_UNIX and full Kerberos flavor set even when the share policy forbids them |
| 2 | `mount-export-auth` | GSS MNT fail-open: no SessionInfo skips MinKerberosLevel/RequireKerberos floor entirely |
| 3 | `mount-export-auth` | Store error on UID lookup/permission-resolve fails open to guest default, overriding explicit per-user 'none' block |
| 4 | `mount-export-auth` | Disabled DittoFS user still authorized over NFS — user.Enabled never checked |
| 5 | `mount-export-auth` | Export access-control policy inlined in protocol handler; CLAUDE.md's own CheckExportAccess doesn't exist |
| 6 | `mount-export-auth` | UMNT deletes ALL of a client's mounts instead of just the requested export |

---

### [MED] MNT reply advertises AUTH_UNIX and full Kerberos flavor set even when the share policy forbids them · `gaps` · area: mount-export-auth
- **Where:** `internal/adapter/nfs/mount/handlers/mount.go:195`
- **What:** After Mount() enforces share.AllowAuthSys (line 124), share.RequireKerberos (line 129), share.MinKerberosLevel (142-151) as hard denies, success path unconditionally builds authFlavors:=[]int32{1} (AUTH_UNIX) and, when ctx.KerberosEnabled, unconditionally appends krb5/krb5i/krb5p (195-203) regardless of actual share requirement/allowance.
- **Why it matters:** RFC 1813 App I fhstatus3.auth_flavors tells client which flavors THIS export accepts (mount.nfs/automounters use it for sec= auto-negotiation). A share with AllowAuthSys=false or RequireKerberos=true still advertises AUTH_UNIX; a share requiring MinKerberosLevel=krb5p still advertises plain krb5/krb5i. Default client picks first flavor (AUTH_UNIX), gets unconditional MountErrAccess even though export is mountable via GSS; client picking krb5/krb5i off the list is denied by the very check (144-148) this reply contradicts.
[...]

PR #2574, opened 2026-09-14
fix(nfs): fail closed when share-permission resolution errors

marmos91:
Three of the MED security findings in #2400, all in the NFS share-permission resolver. Each is a fail-open: the code treats a store failure as information about what the caller is allowed.

### 1. Store error on UID lookup granted guest access (#2400-3)

```go
user, err := identityStore.GetUserByUID(ctx, *uid)
if err != nil || user == nil { ...guest branch... }
```

The guest branch grants the share default, so a transient database error became an access grant wherever `default_permission` is read or read-write. Worse, it inverts an explicit per-user 'none' block — that block lives on the very record the lookup failed to read, so the deny is unreachable exactly when the store is unhealthy.

`GetUserByUID` returns `models.ErrUserNotFound` for a genuine miss and passes other errors through, so `errors.Is` is a sound discriminator. Only not-found reaches guest now.

### 2. Permission resolution error substituted the share default (#2400-3, same finding)

`ResolveSharePermission` returns an error **only** when the share lookup behind it fails — "this user has no explicit grant" is a non-error path that already returns the default. So an error there is always a store failure, and falling back to the default granted the share on an outage. Verified in `GORMStore.ResolveSharePermission` before changing it; failing closed here cannot deny legitimate default-permission users.

### 3. Disabled user kept NFS access (#2400-4)

`user.Enabled` was never read anywhere in the NFS path. Password auth (`store/users.go` → `ErrUserDisabled`) and the SMB session setup both refuse a disabled user, so disabling an account revoked SMB and API access while any UID still mapping to that record kept NFS live indefinitely. NFS was the only protocol not honouring it.

## Tests

Four tests, three of which fail against the unfixed code with `got <nil>` — access **granted** in every case:

- store error on UID lookup → denies (share is deliberately `read-write` so a guest fallback would grant)
- store error on permission resolution → denies
- disabled user → denies
- genuine not-found → **still reaches guest**, so the fix cannot over-deny unmapped UIDs (this one passes before and after, and is the guard against fixing too much)

## Fixture note

`models.User.Enabled` carries a GORM `default:true` tag, which applies at the database layer — in a Go struct literal the field is `false`. Test fixtures that omit it now represent disabled users, so active-user fixtures needed `Enabled: true`. One of them, the SID "explicit local none blocks a matching SID grant" case, would otherwise have kept passing for the wrong reason: denied by the new gate rather than by the explicit-none logic it was written to test.

Full suite: 8467 passed in 167 packages.

Refs #2400

---

**Scope note (from review):** this is a partial fix for #2400-4, not a complete one.

The NFSv3 handler caches resolved auth contexts for `identity.DefaultCacheTTL` (5 min), and nothing invalidates that cache when a user record changes — `GORMStore.UpdateUser` notifies no one and `NewUserHandler` takes no change callback. So this moves #2400-4 from "a disabled user keeps NFS access indefinitely" to "up to 5 minutes". Closing the window needs control-plane wiring outside this PR's layer, and isn't specific to `Enabled` — it covers any user field the resolver reads. Filed as #2576.

The other two findings (#2400-3, both halves) are fully closed by this PR — they sit on the resolution path itself, with no cache in front of them on a miss.


marmos91:
Verified and correct — this fix is narrower than the finding it closes, and that's worth stating plainly.

Traced the wiring: `ClearAuthCache` fires on `OnShareChange`, `OnAuthCacheInvalidate` and `OnIdentityMappingChange`, but nothing fires on a user record mutation. `GORMStore.UpdateUser` notifies no one, and `NewUserHandler(cpStore, jwtService)` at `pkg/controlplane/api/router.go:134` takes no change callback — unlike `NewIdentityMappingHandler(ims, rt.NotifyIdentityMappingChange)` constructed nearby. So a client holding a cached positive context keeps access until `DefaultCacheTTL` (5 min) expires.

So this PR moves #2400-4 from "a disabled user keeps NFS access indefinitely" to "up to 5 minutes", not to zero.

Not fixing it here. The gap isn't in the NFS auth resolver this PR touches — it's missing control-plane wiring, it needs `NewUserHandler`'s signature changed and a decision about which user mutations invalidate, and it isn't specific to `Enabled`: it covers any user field the resolver reads, including per-user share grants and UID. That belongs in its own change rather than widening an adapter-layer auth fix. Filed as #2576 with the call sites and the suggested wiring.

`NewRouter` already receives `rt`, so the callback is available at the construction site — the fix should be small, just not this PR's.

PR #2575, opened 2026-09-14
fix(nfs): deny an RPCSEC_GSS credential that was never verified

marmos91:
#2400-2. A share configured to require Kerberos accepts an unauthenticated client when the server has no Kerberos configured.

## The path

`pkg/adapter/nfs/dispatch.go:53` intercepts auth flavor 6 only when a GSS processor exists:

```go
if call.GetAuthFlavor() == rpc.AuthRPCSECGSS && c.server.gssProcessor != nil {
```

With no processor, a forged RPCSEC_GSS credential is never intercepted and never verified — it falls straight through to program dispatch and reaches the handlers with no GSS session info attached. There it clears every gate:

| Gate | Result |
|---|---|
| `!share.AllowAuthSys && flavor == AUTH_UNIX` | doesn't apply — flavor is 6, not AUTH_UNIX |
| `share.RequireKerberos && flavor != GSS` | passes — the flavor *is* GSS |
| `MinKerberosLevel` floor | **skipped** — guarded on `si != nil` |

The floor was the last gate, and it was the one being skipped. Same shape at both sites: `mount/handlers/mount.go:143` and `v4/handlers/helpers.go:145`.

Demonstrated by the new tests against unfixed code — a flavor-6 credential with no session info mounts a **krb5p** share with `MountOK`, and reads attributes on a `RequireKerberos` share with `NFS4_OK`.

## Why the previous behaviour was deliberate, and why it's still wrong

The skip was intentional. `TestMinKerberosLevel_NoSessionInfoSkipsFloor` pinned it, and its comment gives the reason: every DB-loaded share carries a factory-default `krb5` floor, so enforcing it against a request with no negotiated service (0) would deny spuriously.

That concern is real, but the conclusion doesn't follow. The problem is an unverifiable credential, not an unmeetable floor — and the fix for "we cannot check this credential" is to refuse it, not to waive the only check standing. RFC 5531 wants `AUTH_TOOWEAK`/`AUTH_BADCRED` for a flavor the server cannot verify.

**Nothing legitimate is refused.** With `gssProcessor == nil` the dispatch never answers a GSS INIT, so no client can establish a Kerberos context in that state at all. A flavor-6 credential arriving there is unverifiable by construction — there is no working configuration on the other side of this deny.

## Why `si == nil` is a sound signal

`dispatch.go:167` attaches session info unconditionally to every RPCSEC_GSS request that reaches program dispatch. INIT/DESTROY and GSS failures are answered earlier and never reach a handler. So `si == nil` on a flavor-6 request means exactly one thing: it was not processed as GSS. Checked before relying on it — this is what makes the deny safe for real Kerberos mounts, and `TestMount_MinKerberosLevel_AllowsAtFloor` still passes.

## Tests

- `TestMount_UnverifiedGSSCredentialDenied` — krb5p share, unverified flavor 6 → `MountErrAccess` (was `MountOK`)
- `TestMount_UnverifiedGSSDeniedOnUnrestrictedShare` — deny doesn't depend on the floor being set
- `TestExportAuthPolicy_UnverifiedGSSRejected` — v4 `RequireKerberos` share → `NFS4ERR_WRONGSEC` (was `NFS4_OK`)
- `TestMinKerberosLevel_NoSessionInfoSkipsFloor` → renamed `..._NoSessionInfoDenies` and inverted, with the original rationale recorded and answered in the comment

Verified the v4 guard is not vacuous by removing it and re-running: the test fails with `status = 0`.

Full suite: 8581 passed in 167 packages.

Refs #2400

---

## Update: moved to a dispatch-level guard (review)

Review pointed out the per-handler checks were the wrong level. The credential reaches every program the same way, so MNT and v4 were two symptom sites of one cause and v3/NLM/NSM had no guard at all.

`handleRPCCall` now refuses the call when `gssProcessor == nil`, answering `AUTH_TOOWEAK` (RFC 5531 — constant added, only the two RFC 2203 codes existed). One guard covering every program, smaller than the two it supersedes. The handler-level checks stay as a second line for a request carrying no session info despite a processor being present.

Also reworded the v4 error, which claimed the share requires Kerberos on a branch that doesn't establish it.

**Still open, filed as #2577:** `RequireKerberos` is enforced only in `mount.go` and the v4 path — nothing under `v3/`. The GSS vector is closed by the dispatch guard, but a client that mounted over AUTH_SYS before an administrator tightened the policy keeps using its existing handle indefinitely, since v3 never re-checks per operation. That needs the flavor gates on the v3 operation path.


marmos91:
Both valid — the first one changed the shape of the fix, so thanks for it.

**On the global path:** you're right that per-handler guards were the wrong level. The credential reaches every program the same way — `handleRPCCall` intercepts flavor 6 only when a processor exists — so MNT and v4 were two symptom sites of one cause, and v3/NLM/NSM had no guard at all.

Took your suggestion: `handleRPCCall` now refuses the call outright when `gssProcessor == nil`, answering `AUTH_TOOWEAK` per RFC 5531 (added the constant; only 13/14 existed). One guard, every program, and smaller than the two it supersedes. The handler checks stay as a second line for a request carrying no session info despite a processor being present.

New test drives `handleRPCCall` with a flavor-6 credential and no processor, asserting MSG_DENIED / AUTH_ERROR / AUTH_TOOWEAK. Verified it is not vacuous by disabling the guard — it then fails on `reply_state`, meaning the call was dispatched instead of refused.

**On the error text:** correct, that branch doesn't establish that the share requires Kerberos — it fires for any unverified GSS credential, including on an unrestricted share. Reworded to `RPCSEC_GSS credential for share %q was not verified`.

**One part I'm not fixing here.** Your comment also notes v3 ops remain callable with a valid handle despite `RequireKerberos`. The GSS half of that is now closed by the dispatch guard, but the underlying gap is broader and survives it: `RequireKerberos` is enforced only in `mount.go` and the v4 path, with no hit anywhere under `v3/`. So a client that mounted over AUTH_SYS before an administrator set `RequireKerberos=true` keeps using its existing handle indefinitely — the policy change binds only new mounts. That needs the flavor gates on the v3 operation path and is too big to ride along here. Filed as #2577.

Full suite: 8430 passed in 176 packages.

PR #2578, opened 2026-09-14
fix(nfs): advertise only the auth flavors a share accepts

marmos91:
#2400-1. The MNT reply advertised auth flavors the share would then refuse.

## The contradiction

`Mount()` built its flavor list without consulting the share at all:

```go
authFlavors := []int32{1} // AUTH_UNIX
if ctx.KerberosEnabled {
    authFlavors = append(authFlavors, krb5, krb5i, krb5p)
}
```

Roughly 70 lines earlier, the same function denies the mount when the share says otherwise:

| Gate | Denies | Yet the reply offered |
|---|---|---|
| `!share.AllowAuthSys && flavor == AUTH_UNIX` | AUTH_SYS mounts | AUTH_UNIX |
| `share.RequireKerberos && flavor != GSS` | AUTH_SYS mounts | AUTH_UNIX |
| `!MeetsMinKerberosLevel(share.MinKerberosLevel, si.Service)` | krb5 on a krb5p share | krb5, krb5i, krb5p |

RFC 1813 App I makes `fhstatus3.auth_flavors` the set of flavors *this export accepts*, and `mount.nfs` and automounters negotiate `sec=` from it. So a client doing exactly what the protocol says was steered into a mount the server refused.

## The fix

Build the list from the same policy the gates enforce. It lives in `internal/adapter/nfs/auth` next to `MeetsMinKerberosLevel` — reusing that function, so the advertisement and the enforcement cannot drift — rather than inline in the handler, per architecture invariant #1. (This is a small step toward #2400-5, not a fix for it: the gates themselves are still inline.)

An empty result is possible and correct: a share that forbids AUTH_SYS on a server with no Kerberos accepts nothing, and saying so beats offering a flavor that will be refused.

## No change for a default share

Worth stating, since this narrows a list clients depend on:

- `AllowAuthSys` defaults to `true` (`DefaultNFSExportOptions`), and `shares/service.go:435` coerces an *unset* value to true when the runtime share is built — so the field is meaningful rather than an accidental zero.
- The default `MinKerberosLevel` is `krb5`, and a krb5 floor admits all three pseudoflavors.

So a default share advertises exactly what it advertised before. Only a share with explicitly restrictive policy sees a narrowed list — which is the point.

## Tests

Seven table cases on the policy function, plus two handler-level tests asserting the MNT response itself. Both handler tests fail against the old code with `AuthFlavors = [1 390003 390004 390005]` — the full unconditional set:

- `RequireKerberos` share → must not offer AUTH_UNIX, must still offer the Kerberos flavors
- krb5p share → must not offer krb5 or krb5i, must still offer krb5p

Full suite: 8596 passed in 168 packages.

Refs #2400


marmos91:
Confirmed empirically, and it was worse than an empty-array question. Thanks — this one found a live hole.

I probed the exact case you describe (share `AllowAuthSys=false`, Kerberos disabled, AUTH_NONE caller):

```
PROBE status=0 authFlavors=[] handleLen=44
```

`status=0` is MountOK and **a 44-byte root handle was returned**. So the gate wasn't just producing an incoherent reply — a share configured to refuse AUTH_SYS was handing a root handle to a caller carrying no credential at all, which is strictly weaker than the flavor it refuses. The only thing standing between that caller and the export was `default_permission`.

Root cause is the gate naming the forbidden flavor rather than the allowed ones:

```go
if !share.AllowAuthSys && ctx.AuthFlavor == rpc.AuthUnix {
```

AUTH_NONE isn't AUTH_UNIX, so it fell through — and `RequireKerberos` was false, and the GSS floor doesn't apply to a non-GSS flavor. Changed it to test for what the share accepts:

```go
if !share.AllowAuthSys && ctx.AuthFlavor != rpc.AuthRPCSECGSS {
```

Same probe after the change: `status=13 authFlavors=[] handleLen=0`.

That also answers your empty-array point at the root rather than papering over it. A successful mount can no longer report an empty `auth_flavors`: the one configuration that produced it — AUTH_SYS forbidden, no Kerberos — now accepts nothing at all, so there is no MountOK to attach an empty list to.

Added `TestMount_AuthNoneDeniedWhenShareForbidsAuthSys`, asserting both the status and that no root handle comes back.

Full suite: 8294 passed in 158 packages. Head is now ef71781af.

Issue #2577, opened 2026-09-14
NFSv3 data operations don't enforce the export auth-flavor policy


marmos91:
Split out of #2400 while reviewing #2575. The GSS half of this is fixed there; this issue is the part that remains.

## The gap

NFSv3 data operations never re-check the export auth policy. Grepping `RequireKerberos` across the adapters, the enforcement sites are:

- `internal/adapter/nfs/mount/handlers/mount.go:129` (MNT)
- `internal/adapter/nfs/v4/handlers/helpers.go` and `v4/handlers/secinfo.go` (v4)

There is no hit anywhere under `internal/adapter/nfs/v3/`. The v3 auth path (`v3/handlers/auth_helper.go`) resolves identity and share permission via `auth.ResolveSharePermission`, which enforces share **permission** (default_permission, per-user grants, squash) but not the export **auth-flavor** policy — `AllowAuthSys`, `RequireKerberos`, `MinKerberosLevel`.

## Consequence

MNT gates handle acquisition, so a client cannot newly mount a share whose flavor policy it fails. But a handle already held keeps working, because nothing re-checks per operation. The realistic sequence:

1. Client mounts a share over AUTH_SYS while the share permits it.
2. An administrator sets `RequireKerberos=true` (or `AllowAuthSys=false`) on that share.
3. The client keeps performing v3 GETATTR/READ/WRITE with its existing handle over AUTH_SYS indefinitely.

The policy change is only enforced against *new* mounts. File handles are designed to stay stable across restarts, so this is not a short window.

v4 does not have this shape: `buildAuthContext` runs the flavor gates on the operation path, not only at mount.

## Suggested direction

Apply the same export auth-flavor gates on the v3 operation path that v4 applies — most likely inside the v3 auth-context builder, so it covers every procedure rather than being added per handler. Worth landing behind the same helper both versions call, so the two paths cannot drift again.

Note the v3 handler caches resolved auth contexts (see #2576) — a per-op gate needs to sit outside that cache or the cache needs invalidating on adapter-config change, or the enforcement inherits the same staleness.

## Not verified

Established by reading the enforcement sites and the v3 auth path. I have not driven a live client through the mount-then-tighten-policy sequence to confirm the window empirically.

marmos91:
Premise confirmed empirically — the "Not verified" caveat in the description can come off.

[...]

PR #2582, opened 2026-09-14
fix(nfs): push export policy changes into the running share

### LINKED ISSUE #2580: NFS export auth-policy changes never reach the running adapter
Found by manual testing on a live server while verifying #2400 fixes. Not reachable by the unit suite: the handler persists correctly and the gate enforces correctly — only the path between them is missing.

## Summary

Changing a share's NFS export auth policy (`allow_auth_sys`, `require_kerberos`, `min_kerberos_level`) through the API persists the value and reports success, but never reaches the running NFS adapter. The old policy stays in force until the server restarts.

This is silent in both directions. Tightening reports success while the export stays open; relaxing leaves clients refused with no explanation.

## Reproduction

Ubuntu 22.04, `dfs` built from develop `122cc049f`, memory metadata store, share `/export` with defaults.

```
# 1. Baseline: sec=sys mount works
mount -t nfs -o vers=3,proto=tcp,port=12049,mountport=12049,mountproto=tcp,mountvers=3,nolock,sec=sys \
      localhost:/export /mnt/t          -> OK

# 2. Forbid AUTH_SYS
dfsctl share nfs-config set /export --allow-auth-sys false
  -> "NFS config for share '/export' updated successfully"
dfsctl share nfs-config show /export
  -> Allow AUTH_SYS   false

# 3. Fresh sec=sys mount, no restart
mount ... /mnt/t2                        -> NEW_MOUNT_OK      <-- should be refused

# 4. Restart dfs, same stored config, mount again
mount ... /mnt/t   -> mount.nfs: access denied by server      <-- now enforced
```

Symmetric confirmation that the running adapter simply holds a stale copy: with the server still running from step 4, setting `--allow-auth-sys true` and mounting again is still refused, even though the store and `nfs-config show` both report `true`.

## Cause

`ShareNFSConfigHandler` pushes some fields into the running share and not others (`internal/controlplane/api/handlers/share_adapter_config.go`):

- netgroup -> `runtime.SetShareNetgroup`
- squash / anonymous UID / GID -> `runtime.SetShareSquash`
- cached auth contexts -> `runtime.InvalidateAuthCache()`

There is no equivalent for `AllowAuthSys`, `RequireKerberos` or `MinKerberosLevel`. The comment above the squash push notes those fields "do not flow through ReconcileShareRootACL" and must be pushed explicitly — the pattern was understood, the auth-flavor fields just were not covered.

Tellingly, the only methods that can update those fields on a running share are test-only:

```
pkg/controlplane/runtime/runtime.go:1154  SetExportAuthPolicyForTesting
pkg/controlplane/runtime/runtime.go:1165  SetMinKerberosLevelForTesting
```

No production caller exists. The handler tests pass because they assert persistence, and the adapter tests pass because they use the `ForTesting` setters — so both halves are green while the seam between them is absent.

## Impact

`allow_auth_sys=false` and `require_kerberos=true` are security controls. An operator tightening an export gets a success message and a `show` that confirms it, while the export keeps accepting the flavor they just forbade until someone restarts the server. Nothing in the UI or logs indicates the control is inert.

Related but distinct: #2577 (v3 never re-checks export policy per operation, so even a correctly-propagated change does not bind an already-mounted client) and #2576 (user record changes do not invalidate the v3 auth cache). All three are facets of adapter state not tracking control-plane state.

## Suggested fix

Add a production `SetExportAuthPolicy` on the shares service — the `ForTesting` variants already do exactly this work — and call it from the NFS config handler beside the existing squash and netgroup pushes, whenever any of the three fields is present in the request.

Worth auditing the rest of `NFSExportOptions` in the same pass: `disable_readdirplus` and the retention/quota fields may have the same seam. A test that changes each field through the handler and asserts the running share observed it would close the class rather than the instance.

marmos91:
Fixes #2580, found by manual testing on a live server while verifying the #2400 fixes.

## The gap

Changing a share's NFS export auth policy through the API persisted the value and reported success, but never reached the running adapter. The old policy stayed in force until restart — silently in both directions.

Reproduced on a VM (`dfs` from develop `122cc049f`, kernel NFSv3 client):

```
mount -o vers=3,...,sec=sys localhost:/export /mnt/t     -> OK  (baseline)

dfsctl share nfs-config set /export --allow-auth-sys false
  -> "updated successfully";  nfs-config show -> Allow AUTH_SYS false

mount ... /mnt/t2                                         -> NEW_MOUNT_OK   <-- should be refused
# restart dfs, same stored config
mount ... /mnt/t   -> mount.nfs: access denied by server   <-- now enforced
```

Symmetric confirmation that the adapter simply held a stale copy: with the server running, setting `--allow-auth-sys true` and mounting again stayed refused, while the store and `nfs-config show` both reported `true`.

## Cause and fix

`ShareNFSConfigHandler` already pushes some fields into the running share — netgroup via `SetShareNetgroup`, squash/anon-UID via `SetShareSquash` — and invalidates the auth cache. A comment stated the rest "apply on adapter restart".

But `AllowAuthSys`, `RequireKerberos`, `MinKerberosLevel` and `DisableReaddirplus` are all read from the registry on the request path: the MNT auth gates and the advertised flavor list, the v4 auth check, and the READDIRPLUS downgrade. Deferring them to a restart makes two security controls silently inert.

Tellingly, the only methods that could set them on a live share were `SetExportAuthPolicyForTesting` / `SetMinKerberosLevelForTesting` — test-only, with no production caller.

Added `shares.Service.SetNFSExportPolicy` (nil field = unchanged, matching the PATCH shape) and called it beside the existing pushes. All four fields are covered rather than only the two that bite, since they share the seam.

## Why the tests did not catch it

Each side was verified on its own and the seam between them was never crossed:

- the handler tests assert what was **persisted**, and construct the handler with `runtime` **nil** — the fixture comment says so outright: *"runtime nil — the handler tolerates a nil runtime, persistence is still exercised"*
- the adapter tests set the fields through the `ForTesting` helpers, so they never exercise the API path

Worth noting these handler tests are `//go:build integration`, so they do not run in the default `go test ./...` at all.

The new tests build a real `Runtime` with a registered share, PATCH through the handler, and assert what `rt.GetShare` — what the adapter reads — actually observes. Verified non-vacuous by removing the push: both fail, the first with `running share still has AllowAuthSys=true; the export keeps accepting a flavor it now forbids`.

A second test pins the partial-update shape, so a PATCH naming one field cannot reset the others to their zero value.

## Verification

- default suite: 8554 passed in 166 packages
- `-tags integration` (controlplane): 1469 passed in 19 packages

## Not covered

#2577 is still open and independent: even a correctly-propagated change does not bind an **already-mounted** client, because v3 never re-checks the export auth policy per operation. This PR makes the change bind for new mounts; that one is about existing handles.


marmos91:
Both valid, both fixed.

**Test coverage.** You're right and it's the sharper of the two: I wrote a four-field struct and asserted two, so a mis-mapped member would have shipped. Added running-share tests for `RequireKerberos` and `DisableReaddirplus`.

`RequireKerberos` needed a different direction than the others — the handler refuses to turn it *on* without Kerberos configured, so the test seeds it via `SetExportAuthPolicyForTesting` and PATCHes it off.

Checked the new tests actually catch what you described by swapping two members in the struct literal (`RequireKerberos: req.DisableReaddirplus` and vice versa). Both fail:

```
running share still has RequireKerberos=true after it was cleared
running share still has DisableReaddirplus=false; the READDIRPLUS downgrade stays inactive
```

**Concurrency.** Also correct. `Patch` persists a full snapshot built from its own read, so two concurrent requests interleave into a lost update — and my live push makes it worse, because the snapshot overwrites every field while the push applies only the named ones, so the store and the running share can end up disagreeing on *different* fields. That is the same divergence this PR exists to remove, so shipping a fresh way to produce it would be poor.

Serialized the whole load/merge/persist/push sequence behind a handler mutex. The server is single-node, so one mutex covers every writer; a per-share map would add bookkeeping for contention that does not exist on an admin config endpoint.

Worth noting the lost-update half predates this PR — two concurrent PATCHes could already drop a field from the stored config. The lock closes that too.

Verification: 12 handler tests pass under `-tags integration`, including with `-race`; full default suite 8304 passed in 170 packages. Head is now 5d4ac17ee.

PR #2585, opened 2026-09-15
fix(api): invalidate the adapter auth cache on user and group mutations

### LINKED ISSUE #2576: Control-plane user mutations don't invalidate the NFSv3 auth cache
Split out of #2400 while fixing #2400-4 in #2574. Found by review on that PR.

## The gap

The NFSv3 handler caches resolved auth contexts keyed `Share+AuthFlavor+uid+gid+GIDs` for `identity.DefaultCacheTTL` (5 minutes), to avoid hitting the store on every op (the CREATE-path serialization wall, #1735).

`Handler.ClearAuthCache` is wired to three Runtime events:

- `OnShareChange` — share add/remove/enable/disable
- `OnAuthCacheInvalidate` — share permission / default-permission / squash changes
- `OnIdentityMappingChange` — UID/GID mapping edits, via `NewIdentityMappingHandler(ims, rt.NotifyIdentityMappingChange)`

**Nothing fires on a user record mutation.** `GORMStore.UpdateUser` notifies no one, and `NewUserHandler(cpStore, jwtService)` (`pkg/controlplane/api/router.go:134`) is constructed without a change callback, unlike the identity-mapping handler right next to it.

## Consequence

Any authorization-relevant change to a user takes up to 5 minutes to reach NFSv3 clients that already have a cached positive context. Concretely, after #2574 lands: disabling a user is refused on the next auth-context *rebuild*, but a client holding a cached context keeps access until the TTL expires.

#2574 changes this from "a disabled user keeps NFS access indefinitely" to "up to 5 minutes", which is the reason it is not a complete fix for #2400-4.

This is not specific to `Enabled` — it covers any user field the resolver reads, including per-user share grants and UID.

## Suggested fix

Thread a change callback into the user handler the way the identity-mapping handler already does:

```go
userHandler, err := handlers.NewUserHandler(cpStore, jwtService, rt.NotifyAuthCacheInvalidate)
```

`NewRouter` already receives `rt`, so the wiring is available at the construction site. Fire it on user update and delete at minimum. Worth checking whether the SMB session layer caches anything with the same exposure before deciding the callback's scope.

## Not verified

I confirmed the missing wiring by reading the construction sites and `UpdateUser`; I have not driven a live server to measure the actual staleness window.

marmos91:
## What was wrong

The NFSv3 handler caches a resolved `*metadata.AuthContext` keyed `Share+AuthFlavor+uid+gid+GIDs` for five minutes, to avoid re-resolving share permission against the control-plane store on every op (the CREATE-path serialization wall, #1735).

That cache is dropped on share-permission changes — `ShareHandler.SetUserPermission` / `RemoveUserPermission` → `reconcileRootACL` → `Runtime.ReconcileShareRootACL` → `Runtime.InvalidateAuthCache` → `OnAuthCacheInvalidate` → `Handler.ClearAuthCache`. It was **not** dropped on any mutation of the user record itself, which is the other half of what the resolver reads.

So disabling a user, changing its UID, or rewriting its per-user share grants was refused on the next auth-context *rebuild*, but a client already holding a cached positive context kept the old authorization until the entry aged out. That is why #2574 was not a complete fix for #2400 finding 4: it moved "a disabled user keeps NFS access indefinitely" to "up to five minutes", not to zero.

## Premise verification

The issue was filed from reading the construction sites, with the author noting they had not driven a live server. Verified on `develop` before changing anything:

- `internal/adapter/nfs/auth/share_permission.go:179` — the resolver reads `user.Enabled`, and `ResolveSharePermission` immediately below reads the per-user share grant. Both sit behind the cache.
- `pkg/controlplane/store/users.go` — `UpdateUser`, `DeleteUser` and `CreateUserWithGroups` are plain GORM writes. They notify nothing.
- `Runtime.InvalidateAuthCache` has exactly two callers: `pkg/controlplane/runtime/share_root_acl.go:175` and `internal/controlplane/api/handlers/share_adapter_config.go:226`. Neither is reachable from any user-management route.
- `NewUserHandler(cpStore, jwtService)` took no change callback, unlike `NewIdentityMappingHandler(ims, rt.NotifyIdentityMappingChange)` two hundred lines away in the same router.

Premise confirmed. The failing test below is the empirical half — it drives real HTTP requests through the real router and observes the runtime event, rather than asserting on a code reading.

**One correction to the issue text:** the suggested wiring names `rt.NotifyAuthCacheInvalidate`, which does not exist. The existing public method is `Runtime.InvalidateAuthCache`, and it already fans out to exactly the right subscribers. No new runtime surface was needed.

**One thing checked and ruled out:** `pkg/identity.Resolver` keeps its own five-minute cache and would have been a second stale layer defeating this fix. It is not in this path — it caches external-credential→identity resolution (Kerberos/LDAP), not the local user record, and `GetUserByUID` has a single implementation going straight to the database. Clearing the NFSv3 auth cache is therefore sufficient.

## The change

Threads an `onChange func()` into `UserHandler` following the existing identity-mapping handler pattern exactly, wired at the construction site to `rt.InvalidateAuthCache`. Fired on create, update and delete.

In `Create` and `Update` the call is deferred rather than immediate. That is deliberate: both handlers perform further authorization-relevant writes after the main one (group membership, share grants), and several of those paths return early on error while leaving the first write persisted. One deferred notification covers every exit after the record changed.

### Groups too — the same defect one file over

`ResolveSharePermission` walks the member's group list and takes the highest group grant (`pkg/controlplane/store/permissions.go:203-207`), so a group-membership edit moves exactly the decision a per-user grant edit moves, and it lands in the same cache. `GroupHandler` had no runtime reference at all, and the standalone `/api/v1/groups/{name}/members` add/remove was the one membership-mutating surface with neither an invalidation nor a `reconcileRootACL` behind it. Membership rewritten through `PUT /users/{username}` was already covered; through the group routes it was not.

This is a small extension beyond the issue's literal scope, taken because it is the same seam, the same one-line wiring, and leaving it would ship a knowingly half-applied fix. Both the reviewer agent and I found it independently. Easy to split out if you would rather keep the PR to the letter of the issue.

**The password endpoints deliberately do not fire it.** The NFS auth resolver reads no credential material — AUTH_SYS carries a bare UID — so a password change cannot alter an already-resolved auth context. Flushing the whole cache on a routine self-service password change would cost the #1735 mitigation for no authorization benefit. There is a test pinning that exclusion so it reads as a decision rather than an oversight.

## Evidence

`TestIdentityMutationsInvalidateAuthCache` drives real admin HTTP requests through the real router and asserts the runtime fired the invalidation event.

With the callback wired back to `nil` at the construction site, every mutation case fails:

```
--- FAIL: TestIdentityMutationsInvalidateAuthCache/create_user
    did not fire the auth-cache invalidation event; an NFSv3 client holding a
    cached auth context keeps the old authorization until the TTL expires
--- FAIL: TestIdentityMutationsInvalidateAuthCache/disable_user
--- FAIL: TestIdentityMutationsInvalidateAuthCache/change_uid
--- FAIL: TestIdentityMutationsInvalidateAuthCache/rewrite_share_grants
--- FAIL: TestIdentityMutationsInvalidateAuthCache/delete_user
--- FAIL: TestIdentityMutationsInvalidateAuthCache/create_group
--- FAIL: TestIdentityMutationsInvalidateAuthCache/change_group_gid
--- FAIL: TestIdentityMutationsInvalidateAuthCache/add_group_member
--- FAIL: TestIdentityMutationsInvalidateAuthCache/remove_group_member
--- FAIL: TestIdentityMutationsInvalidateAuthCache/delete_group
```

All green with the wiring in place. Two negative tests keep it honest and pass in *both* states, so they are not just following the flag: a mutation that never reached the store (404 on an unknown user) must not fire, and the password endpoint must not fire.

The other half of the chain — the event actually dropping a live positive cache entry — is already pinned by `TestClearAuthCacheDropsLiveEntry`, and the subscription itself lives at `pkg/adapter/nfs/adapter.go:663`. The end-to-end test cannot be written in one package: `internal/adapter/nfs/v3/handlers` transitively imports `pkg/controlplane/api`, so importing it from the router's own test is an import cycle.

## SMB has the same gap, and it is worse — filing separately

Investigated as the issue asked, and deliberately **not** fixed here.

SMB does not merely cache the decision with a TTL; it pins it for the life of the connection, and it has no subscriber to the invalidation event at all.

| Cached | Where | Lifetime | Invalidated by |
| --- | --- | --- | --- |
| Effective share permission | resolved once at `tree_connect.go:130`, stored `session_lifecycle.go:63`, read per-op via `auth_helper.go:332` | until TREE_DISCONNECT | nothing |
| `Session.User` (Enabled, Role, UID) | `session/session.go:72`, snapshot taken at SESSION_SETUP | until logoff / disconnect / Kerberos ticket expiry | re-auth only |
| Derived `metadata.Identity` | `session/session.go:103`, memoized on pointer equality | same | `UpdateIdentity` / `SetPACIdentity` |
| `OpenFile.OpenerUser` | `open_file.go:409` | until handle close | nothing — deliberate, MS-SMB2 §3.3.5.5.3 freezes the open's security context |

`User.Enabled` is checked only at session establishment (`session_setup.go:1156`, `:1389`, `kerberos_auth.go:600`); the per-request gates are only `LoggedOff` and `IsExpired`, and `ExpiresAt` is set solely from a Kerberos ticket end time. So an established NTLM session survives the user being disabled or deleted indefinitely.

Notably this is not only a user-mutation gap: SMB has no `OnAuthCacheInvalidate` subscriber whatsoever, so a *share*-side grant revoke — which NFS already handles — is also stale on SMB until the tree disconnects. A fix needs three consumers (resolver invalidation, tree-permission re-resolve, session kill on disable), which is why it does not belong in this PR.

## Also found, not fixed

`UserHandler.applySharePerms` writes grants via `store.SetUserSharePermission` directly, while the dedicated `PUT /shares/{name}/permissions/users/{username}` endpoint also calls `reconcileRootACL`. So the same grant applied through `POST /users` or `PUT /users/{username}` skips the root-ACL projection. This PR fixes the cache-invalidation half of that path; the ACL-projection asymmetry is untouched and worth its own issue.

Closes #2576


marmos91:
Thanks — both Copilot points looked at, one fixed, one declined.

**Grant-write test using a real share — fixed, and it was worse than reported.** The case named share `nope`, which `applySharePerms` skips as unresolvable. Giving it a real share was not enough: the permission value `rw` is not a valid `models.SharePermission` either (the constant is `read-write`), and `applySharePerms` skips an invalid value just as silently. So the case was vacuous twice over and passed only on the invalidation fired by the `UpdateUser` write beside it.

It is now a dedicated test that seeds a real share, sends `read-write`, and **reads the grant back out of the store** before asserting the invalidation. Without that read-back the case passes while writing nothing, which is exactly how it slipped through the first time. Verified the grant now persists:

```
persisted grant: &{UserID:23dec01f-... ShareID:share-/export ShareName:/export Permission:read-write}
```

**No-op group removals invalidating the cache — correct, but deliberately left as is.** The diagnosis is accurate: `GORMStore.RemoveUserFromGroup` returns `nil` for a missing group (`pkg/controlplane/store/groups.go:123-125`) and GORM's association delete succeeds for a non-member, so a repeated or invalid DELETE fires the callback and clears the cache.

Not changing it, for two reasons:

1. The fix proposed — having the store report whether a membership was actually removed — means widening the `GroupStore` interface signature and every implementation behind it, to save a cache rebuild on an admin-only route. That is a large blast radius for a small win.
2. It is the wrong direction for this particular cache. The failure mode being fixed here is a *stale authorization* — a disabled user keeping access. A spurious flush costs one rebuild per identity; a missed flush keeps access open for five minutes. Where the two trade off, over-invalidating is the safe side, and that was the explicit design call for this change.

Worth revisiting if the group-membership routes ever turn into a hot path, which they are not today.

PR #2607, opened 2026-09-15
docs: describe the single block store and journal consistently

marmos91:
Follow-up to #2579, covering the documentation it did not touch. #2579's own doc surface is correct — `docs/guide/cli.md` is byte-identical to a fresh `go run ./cmd/gendocs`, and the upgrade chapter in `configuration.md` is a deliberate historical record that stays as it is.

**Breaks a copied command**

- `docs/guide/configuration.md` — the encrypt-data example omitted `--block-store`, which `cmd/dfsctl/commands/share/create.go` now marks required, so copying it exits with `required flag(s) "block-store" not set`. A continuation-aware sweep of every `.md`/`.sh`/`.yaml` found this to be the only such case; the multi-line examples in `faq.md` and `troubleshooting.md` already pass the flag on their second line.

**Prose describing the replaced two-tier model**

- `docs/product/pro.md` — advertised "local (filesystem/memory) and remote (S3)" backends. Only `s3` and `memory` are constructible.
- `docs/guide/garbage-collection.md` — described the remote tier as present "if the share has one"; every share has one now.
- `docs/guide/smb.md` — told operators to check filesystem block-store connectivity.
- `docs/guide/snapshots.md` — "the local block store is intact" is now the journal.
- `CLAUDE.md` — the per-share block store invariant still read "local + remote + syncer".
- `test/smb-conformance/configs/*.yaml` — four header comments saying "memory local block store".

Not included, and filed separately: the k8s operator's own docs and samples, which are wrong in a way that costs a user disk rather than a command, and involve a code gap as well.


marmos91:
All three Copilot comments were correct and are fixed in `a71b0e008`. Each was a pre-existing error in the prose that this PR carried forward into the new vocabulary rather than introduced — worth recording, because rewording is exactly when a stale claim gets re-certified.

**GC tiers.** `CollectGarbage` sweeps the block store only (`pkg/block/engine/gc.go:314`: "reclaims orphaned chunks on a remote store"). The journal reclaims its own dead bytes and evicts under pressure independently. The lead-in now says so and points at `dfsctl store block evict` for journal disk, instead of implying `store block gc` frees it.

**`memory` stores.** `validateBlockStoreType` accepts `s3` and `memory` (`internal/controlplane/api/handlers/block_stores.go:65`), so the SMB troubleshooting step no longer prescribes an S3 check for every share.

**`--no-verify`.** The strongest of the three. `CreateSnapshot` drains twice: Step 0 before the dump (`snapshot.go:546`, unconditional) and Step 6 in the verify gate (`snapshot.go:722`, the one `NoVerify` skips at `snapshot.go:621`). So an unreachable block store fails create with `--no-verify` just as without it, and the bullet claiming otherwise was wrong. Removed, and replaced with an explicit statement of the opposite.

Step 0 is not an oversight — its own comment explains that the dump must carry the locators carve mints via `MarkSynced`, or restore resolves them as `cas/<hash>` objects that no longer exist. Skipping it would mint snapshots that look restorable and are not, so the code is right and only the documentation was wrong.

The same claim also lives in the source: `CreateSnapshotOpts.NoVerify` says it "skips DrainAllUploads + VerifyRemoteDurability", which is half true. Filed as #2614 rather than fixed here, since this branch is docs-only.

PR #2612, opened 2026-09-15
fix(block-store): count share references by name as well as UUID, and refuse a taken rename before writing

### LINKED ISSUE #2604: Block store deletion counts references by UUID only, orphaning name-bound shares
`GORMStore.DeleteBlockStore` (`pkg/controlplane/store/block.go`) guards against deleting a block store still in use by counting shares that reference it:

```go
if err := tx.Model(&models.Share{}).
    Where("block_store_id = ?", store.ID).
    Count(&count).Error; err != nil {
```

It matches the store's UUID only. But a share's `block_store_id` can hold the store **name** instead — the REST update path historically persisted it that way, and the codebase accounts for that duality in two other places:

- `pkg/controlplane/store/journal_root_migration.go` joins on `b.id = s.local_block_store_id OR b.name = s.local_block_store_id`, with a comment saying exactly why.
- The block-store PUT handler repoints name-bound shares onto the UUID after a rename, for the same reason.

**Failure scenario:** a share persisted with `block_store_id = "my-blocks"` (the name). An operator deletes that block store; the count returns 0, so `models.ErrStoreInUse` never fires and the delete succeeds. The share's binding now points at a store that no longer exists. At the next restart `AddShare` fails, `LoadSharesFromStore` warn-and-skips it, and the server comes up reporting healthy with the share silently absent.

**Fix:** count references to both the UUID and the name, matching the join the migration already uses.

**Note:** this is pre-existing — the same shape is on develop before #2579 — so it is not a regression from the single-block-store work. Found during review of #2579 and verified against the merged code.

### LINKED ISSUE #2605: A failed block-store rename returns 409 after the type/config change already persisted
In `BlockStoreHandler.Update` (`internal/controlplane/api/handlers/block_stores.go`), a PUT that changes `type`/`config` **and** renames the store applies the two in sequence, not atomically:

```go
if err := h.store.UpdateBlockStore(r.Context(), bs); err != nil { ... }   // commits type/config
...
if renameTo != "" {
    renamed, err := h.store.RenameBlockStore(r.Context(), name, renameTo)
    switch {
    case errors.Is(err, models.ErrDuplicateStore):
        Conflict(w, "Block store "+renameTo+" already exists")
        return
```

**Failure scenario:** operator PUTs a new `config` (say rotated S3 credentials) together with a `name" that is already taken. `UpdateBlockStore` commits the credential change. `RenameBlockStore` then returns `ErrDuplicateStore`, and the API answers **409 Conflict** — which reads as "nothing happened". The credentials have in fact changed. An operator who treats the 409 as a no-op and retries elsewhere, or rolls back by re-PUTing the old name, is reasoning about a state the server is not in.

**Fix:** preflight the target name before `UpdateBlockStore` (cheapest — the rename is already refused for duplicates, so the check exists), or apply both mutations in one transaction.

Found during review of #2579 and verified against the merged code.

marmos91:
Two defects in the block-store config surface, batched because they touch the same file.

Closes #2604
Closes #2605

## #2604 — a name-bound share did not block its block store's deletion

A share's `block_store_id` normally holds the block store row's UUID, but the older REST update path persisted the **name**, and both still resolve today: `resolveBlockStoreConfig` (`pkg/controlplane/runtime/shares/blockstore_config.go:42`) tries the UUID and falls back to a name lookup, and the journal-root migration joins on `b.id = s.local_block_store_id OR b.name = s.local_block_store_id` for the same reason.

`DeleteBlockStore` counted references by UUID only. A share bound by name therefore counted as zero references, `ErrStoreInUse` never fired, and the delete succeeded — leaving a binding pointing at a store that no longer exists. At the next restart `AddShare` fails, the share is warn-and-skipped, and the server comes up reporting healthy with the share silently absent.

The count now matches either form, mirroring the join the migration already uses. Pre-existing — the same shape is on `develop` before #2579, so it is not a regression from the single-block-store work.

Deliberately **not** done: no repair or normalisation of existing name-bound bindings. That is a much larger blast radius than the guard this issue is about.

**Second part — reporting an already-orphaned share at startup.** A share whose binding already resolves to nothing now gets an explicit line naming the share and the reference, and a skip reason an operator can act on:

```
[WARN] Share references unknown block store share=/orphaned block_store_id=deleted-blocks
```

One honest note on the issue's framing: the pre-existing path was **not** silent. It already produced

```
[WARN] Failed to add share to runtime share=/orphaned error=failed to create block store for share "/orphaned": failed to create remote store: failed to resolve remote block store config "deleted-blocks": store not found
```

so both the share and the reference were already in the log, wrapped four levels deep and framed as a remote-store construction failure. What changes is that the failure is now named for what it is: the report matches the unknown-metadata-store warning it sits beside, and the skip reason surfaced by `HealthcheckShare` becomes `block store deleted-blocks is not configured` instead of the wrapped error string.

The check runs **after** `AddShare` has already failed, not before it, and that ordering is load-bearing. `createBlockStoreForShare` opens the share journal before it resolves the block store binding, so the journal format sentinels (`block.ErrFutureFormat`, `journal.ErrFutureFormat`, `sharesvc.ErrLegacyLocalFormat`) — a deliberate boot stop — are raised first. An earlier draft of this change pre-checked the binding at the top of the loop and silently swallowed that boot stop for a share broken both ways. `TestLoadSharesFromStore_FormatErrorWinsOverUnresolvableBlockStore` pins the ordering; it fails against that draft and passes here.

## #2605 — a refused rename answered 409 with the config change already committed

`BlockStoreHandler.Update` applied the type/config write and the rename in sequence. A PUT carrying both a new `config` (rotated S3 credentials, say) and a `name` already in use committed the credentials, then answered **409 Conflict** from the rename. An operator reading that 409 as "nothing happened" — retrying elsewhere, or rolling back by re-PUTing the old name — is reasoning about a state the server is not in.

The target name is now checked before anything is written, using the same predicate the rename itself uses (a *different* row already holding that name), so renaming a store to the name it already has still passes through as a plain config update.

This leaves a narrow TOCTOU window: a concurrent request taking the name between the check and the rename lands on today's behaviour. Accepted deliberately — closing it would mean a combined transactional store method, which is more machinery than the failure warrants.

## Verification

Each fix was confirmed by running its regression test against the unmodified code first:

- `TestDeleteBlockStoreNameBoundShare` — fails on `develop` (`expected ErrStoreInUse, got <nil>`), passes after.
- `TestLoadSharesFromStore_ReportsUnresolvableBlockStore` — pins the explicit skip reason; the pre-fix build records the wrapped construction error instead.
- `TestLoadSharesFromStore_FormatErrorWinsOverUnresolvableBlockStore` — pins that the boot stop still wins for a share broken both ways.
- `TestBlockStoreHandler_Update_ConflictingRenameLeavesConfigUntouched` — fails on `develop` with `config changed despite the 409`, passes after.
- `TestBlockStoreHandler_Update_RenameTargetResolvingByIDIsNotAConflict` — the preflight's lookup answers to an ID as well as a name, but only a name collides; this fails if that distinction is dropped.

`go build ./...`, `go vet ./...`, `gofmt -l` clean. `go test ./pkg/controlplane/... ./internal/controlplane/...` green with and without `-tags integration` (1048 / 1484 tests).

## Unrelated observation

`GORMStore.GetSharesByBlockStore` (`pkg/controlplane/store/block.go:76`) has the same UUID-only shape, but it has no non-test callers. Left alone rather than fixed speculatively.


marmos91:
Thanks — both comments were worth chasing. One is fixed, one I'm leaving as-is with reasoning.

**`init.go` — the branch relabelled unrelated failures. Fixed in f082ba7.**

Correct, and it was the mirror of a bug I'd already fixed once in this branch. Re-checking existence after the fact says nothing about *why* `AddShare` failed, so any non-format failure on a share whose binding also happened to name nothing was reported as `block store ... is not configured` with the real error discarded.

The resolution failure is already wrapped with `%w` the whole way up — `resolveBlockStoreConfig` returns `models.ErrStoreNotFound`, `acquireRemoteStore` wraps it, `createBlockStoreForShare` wraps that, `AddShare` wraps that — so the branch now matches the sentinel directly:

```go
if errors.Is(err, models.ErrStoreNotFound) {
```

That also deleted the `blockStoreExists` helper and its extra DB round-trip. `TestLoadSharesFromStore_UnrelatedFailureKeepsItsReason` pins it: a share with a wedged metadata store *and* an unresolvable binding must keep the metadata error as its reason. It fails against the previous commit.

**`block.go` — a name that is another store's UUID. Acknowledged, not changed.**

The scenario is real: if store A has ID `X` and store B is *named* `X`, a share bound to `X` resolves to A (ID-first), but deleting B counts it and returns `ErrStoreInUse`.

I'm leaving it, for two reasons. It fails safe — the worst outcome is refusing a delete that could have proceeded, where the bug this PR fixes is the opposite: a delete that silently orphans a share. Trading a rare false refusal for that is the right direction. And the alternatives both cost more than the failure does: the "count the name only when no other store owns that token as an ID" form needs a correlated subquery inside the delete transaction, and rejecting or normalising UUID-shaped names is a validation/migration change well outside this fix — the repo owner explicitly ruled normalising existing bindings out of scope here.

Worth noting the window is also closing on its own: `RenameBlockStore` repoints name-bound shares onto the UUID in the same transaction, so new name-bound rows are not created by any current path.

Issue #2619, opened 2026-09-15
smbtorture harness shares one Docker Compose project across all worktrees, so concurrent runs destroy each other


marmos91:
The smbtorture harness derives its Docker Compose project name from the directory basename, so the main checkout and **every** worktree all resolve to a single project called `smb-conformance`.

Two agents running `test/smb-conformance/run.sh` from different checkouts therefore share one set of containers, and whichever finishes first runs `docker compose down -v` and destroys the other's server and network out from under it.

## Evidence

Observed independently by two sessions working unrelated tasks on the same machine.

**First**, during a diagnosis run: two attempts died mid-bootstrap with `dial tcp [::1]:8080: connect: connection refused` and exit 137. `docker events` showed a concurrent `smbtorture-run-26896` container's cleanup calling `docker compose down -v`, removing `smb-conformance-dittofs-1` and its network while the second run was still bootstrapping against it.

**Second**, during an unrelated validation run on the same afternoon: three of four attempts never reached the server at all, presenting as bootstrap failures (`Invalid username or password`, then `Metadata store already exists`) — the signature of a half-torn-down stack rather than of anything server-side.

## Why it is worth fixing rather than scheduling around

The symptom is indistinguishable from a server crash during bootstrap. Nothing in the output says "another run destroyed your containers", so the natural reading is that the change under test broke the server — which is the most expensive possible wrong conclusion to hand someone validating a protocol change. It also silently converts a suite run into a non-result, which compounds #2618: the harness cannot currently tell you that what it measured was nothing.

The orchestration answer (one agent holds the suite at a time) works, but it relies on every future caller knowing the constraint, and the failure is silent when they do not.

## Suggested fix

Scope the project name to the checkout in `test/smb-conformance/smbtorture/run.sh` — `COMPOSE_PROJECT_NAME` derived from a hash of `REPO_ROOT`, so each worktree gets its own stack and concurrent runs are isolated. Port bindings would need the same treatment to let runs actually overlap; failing that, a startup check that refuses to run when another project instance is live would at least make the collision loud instead of silent.

Issue #2618, opened 2026-09-15
smbtorture harness reports "All failures are known. CI green" while a gating test was silently reclassified as skipped


marmos91:
A run of `test/smb-conformance/run.sh` printed:

```
RESULT: All failures are known. CI green.
```

while `smb2.app-instance` had **failed** and been reclassified from failure to skip by the harness's connection-setup heuristic.

That is a false green, and the dangerous kind: the banner is byte-identical to the one a genuinely clean run prints. The operator sees "CI green" and has no way to tell, without reading the per-test tallies, that the one test gating the change under review never produced a server-side result at all.

## How it arises

When the torture client cannot establish a connection it fails client-side — observed as `NT_STATUS_NO_MEMORY` in the client, and for the two-connection wrapper tests at `smb2.c:95`, the second connection. The harness treats a connection-setup failure as "not a real result" and skips it rather than failing the job. That heuristic is defensible on its own: a connection failure genuinely is not evidence about the server.

What is not defensible is the summary line. A run where N tests never reached the server is **not** "all failures are known" — it is a run with no verdict for N tests, and it should say so.

The failures come as a contiguous **tail** of the suite whose length grows with machine load (11 tests, then 6 on a quieter run), which is consistent with host resource exhaustion rather than anything server-side.

## Why it matters beyond one run

The whole value of the conformance suites is that they catch what unit tests cannot. A silently-skipped test converts that into a green banner that certifies nothing, and the skip is invisible at exactly the moment someone is deciding whether a change is safe to merge. A reviewer reading "CI green" reasonably concludes the gating test passed.

## Suggested fix

Distinguish the three outcomes in the summary rather than collapsing two of them:

- all tests produced a server-side result and every failure is known → `CI green`
- one or more tests never reached the server → a distinct, non-green line naming them and the count, e.g. `INCONCLUSIVE: N test(s) produced no server result`
- a new failure → red, as today

Optionally fail the job outright when a test named in a "must produce a result" list is among the inconclusive ones.

Found while validating an authorization change whose only protocol-level coverage was `smb2.app-instance` — the run reported green with that exact test skipped.

Issue #2602, opened 2026-09-15
SMB: AppInstanceId failover lacks the ClientGuid and maximal-access match conditions (MS-SMB2 3.3.5.9.13)


marmos91:
# SMB: AppInstanceId failover authorizes from AppInstanceId alone — no client identity, no maximal-access check

`ProcessAppInstanceId` (`internal/adapter/smb/handlers/durable_context.go`) force-closes another
open when a CREATE carries a matching `SMB2_CREATE_APP_INSTANCE_ID`. PR #2599 scoped that filter to
the claimed share and path, which closes the "any file on any share" hole. Two of the four match
conditions MS-SMB2 §3.3.5.9.13 requires are still absent, and the one that actually authorizes the
close is among them.

## What the spec requires

Per MS-SMB2 §3.3.5.9.13, the server looks through `GlobalOpenTable` for an Open where **all** hold:

1. `AppInstanceId` in the request matches `Open.AppInstanceId`
2. the target path name matches `Open.PathName`
3. `Open.TreeConnect.Share` matches the current `TreeConnect.Share`
4. **`Open.Session.Connection.ClientGuid` is *not* equal to the current `Connection.ClientGuid`**

and then, before closing:

> If an **Open** is found, the server MUST calculate the maximal access the user, identified by
> **Session.SecurityContext**, has on the file being opened. If the maximal access includes
> GENERIC_READ access, the server MUST close the open as specified in 3.3.4.17.

DittoFS implements 1–3. Conditions 4 and the maximal-access gate are absent from both the live-open
filter and the persisted-handle loop.

## Why it matters

The maximal-access gate is the authorization step. Without it, the force-close is granted on the
strength of a 16-byte value that travels in cleartext inside every CREATE that uses it — an
AppInstanceId is an application identity (a clustered app / VM instance id), not a secret.

Attack sequence:

1. Client A opens `\\server\fileshare\vm\disk.vhdx` with `AppInstanceId = X` (durable or live).
   `X` is visible on the wire in A's CREATE.
2. Client B authenticates as a **different user**, and issues a CREATE for the same share and path
   carrying `AppInstanceId = X`. B needs no read access to the file — DittoFS never computes maximal
   access for the requester.
3. The filter matches on (AppInstanceId, share, path) and force-closes A's live open and deletes its
   persisted durable handle.
4. A's in-flight operations fail and durable reconnect is gone. Denial of service against another
   user's handle, authorized by a non-secret identifier.

[...]

PR #2622, opened 2026-09-15
fix(smb): gate the AppInstanceId failover on client identity and read access

marmos91:
## What was wrong

`ProcessAppInstanceId` force-closed an existing open, and deleted its persisted
durable handle, on an AppInstanceId match and nothing else.

MS-SMB2 3.3.5.9.13 requires four match conditions before an open is a candidate
for the forced close, and one authorization step after:

1. the request's `AppInstanceId` equals `Open.AppInstanceId`
2. the target path name equals `Open.PathName`
3. `Open.TreeConnect.Share` equals the current `TreeConnect.Share`
4. `Open.Session.Connection.ClientGuid` is **not** equal to the current
   `Connection.ClientGuid`

> If an **Open** is found, the server MUST calculate the maximal access that the
> user, identified by **Session.SecurityContext**, has on the file being opened.
> If the maximal access includes GENERIC_READ access, the server MUST close the
> open as specified in 3.3.4.17.

Condition 4 and the maximal-access step were both absent. This PR adds both.

**The maximal-access step is the one that authorizes the close, and an
AppInstanceId is not a credential.** It identifies an application instance — a
clustered app or VM instance — and it travels in cleartext inside every CREATE
that carries one. So a requester who copied one off the wire could force-close
another user's open on a file they have no access to, and take that open's
persisted durable handle with it, leaving the original client's in-flight
operations to fail with no durable reconnect to fall back on.

**Condition 4 is a behaviour change for real clients, not only hardening.**
Until now a client that reconnected with the same ClientGuid and reused its own
AppInstanceId had its own prior open force-closed. 3.3.5.9.13 matches only opens
whose `ClientGuid` differs from the claiming connection's, so that displacement
was a spec violation rather than a feature, and it no longer happens: a client's
own opens survive its own AppInstanceId. If something downstream was relying on
the old self-displacement to clear a stale handle, it will now see the prior
open contend on share mode instead — that is the conformant outcome, and the
durable-reconnect path (3.3.5.9.12), not the failover path, is what a client
reclaiming its own handle is supposed to use.

### Behaviour changes a client can observe

Two of these are new symptoms, not caveats. Both are self-healing and both are
what Windows Server does, but neither is guessable from a log line — so if you
arrived here searching for a 35-second pause or a lock conflict that clears
itself after five minutes, this is the section you want.

- **A same-client fresh CREATE inherits its own stale byte-range locks for up to
  the durable timeout (5 min default).** If an app restarts on the same machine,
  loses its CreateGuid, and issues a fresh CREATE (no DH2C) reusing its
  AppInstanceId, the persisted durable row is no longer deleted, so
  `UnlockAllForOpen` does not run and the previous incarnation's byte-range
  locks survive. The CREATE still succeeds; a WRITE into a locked range gets a
  lock conflict until the durable-handle scavenger reaps the row. Previously the
  failover deleted the row and released the locks immediately. Clients that
  reconnect properly (DH2C/DHnC) are unaffected — that path is
  MS-SMB2 3.3.5.9.12 and is skipped by this section entirely.
- **A refused displacement against a live batch-oplock holder now costs a
  visible break and a wait.** Where the failover used to silently displace a
  same-client open, the CREATE now falls through to the normal break path,
  dispatches a real oplock break at the incumbent, waits out the traditional
  oplock break timeout (35 s) with no ACK coming, and returns
  `STATUS_SHARING_VIOLATION` for the typical exclusive `ShareAccess=0` open.
  Degraded and bounded, not a wedge — every fail-closed path in this change
  terminates at that ceiling and then a clean NTSTATUS.

Multichannel is strictly better off: a second channel shares the client's
ClientGuid, so its AppInstanceId CREATE used to be able to force-close the
client's own open on channel 1, and now cannot.

## Premise verification

The issue's diagnosis was treated as a claim, not settled fact, and checked
against the primary source rather than against the issue's rendering of it. Both
halves hold:

- 3.3.5.9.13 was fetched from Microsoft Learn directly. The four match
  conditions and the maximal-access sentence are quoted above verbatim from it.
- Product-behavior note &lt;354&gt; on that statement reads: "Windows SMB3 servers
  compute the maximal access to return by querying the security attributes of
  the file with [MS-FSA] section 2.1.5.14, and performing an access check
  against the credentials provided by the request." The gate is real server
  behaviour, not an unimplemented SHOULD.
- Samba is **not** a reference here: `SMB2_CREATE_TAG_APP_INSTANCE_ID` appears
  only in `libcli/smb/smb2_constants.h`, for the client and torture sides;
  `source3/smbd` implements no AppInstanceId failover at all.

## The fix

`ProcessAppInstanceId` now gates the forced close on both missing conditions, on
the live-open path and the persisted-durable-handle path alike:

- **Client identity.** Only an open whose recorded ClientGUID differs from the
  ClientGuid of the connection carrying this CREATE is a candidate.
- **Read access.** The requester must have GENERIC_READ on the file the matched
  open holds, evaluated for the CREATE's own `AuthContext`.

The access evaluation stays in `pkg/metadata`: a new `Service.HasMaximalReadAccess`
answers through the existing `ComputeMaximalAccess`, so an ACL-bearing file and a
POSIX-mode file are judged exactly as they are for the MxAc create-context reply.
The handler asks a yes/no and inlines no ACL logic.

The gate is fail-closed. When access cannot be established at all — no identity,
no metadata service, no recorded metadata handle, a store that declines to
answer, or an open whose recorded ClientGuid is zero and so attributes it to no
client — nothing is displaced, and the claiming CREATE contends with the
surviving open the way any other open would. That is a degraded failover rather
than someone else's handle closed out from under them.

**This is partial authorization, and the `decision:` marker at the gate says so.**
MS-SMB2 3.3.5.9.13 narrows the match set by target path name and by share
*before* either condition above applies, and neither narrowing exists on
`develop` yet. So an application reusing one AppInstanceId across several files
(a VM's config and data VHDs, a database's `.mdf` and `.ldf`) still force-closes
every open carrying that AppInstanceId which the requester can read — not only
the open of the file the CREATE names. Read access bounds that blast radius; it
does not confine it to the claimed file. It also only protects files the
requester genuinely cannot read, so on a world-readable share the gate refuses
almost nobody. Adding the path and share conditions is what closes both gaps —
that work is in flight separately as #2599, and this change is deliberately
scoped to the two conditions that PR does not cover. An exemption with no stated
ceiling is indistinguishable from a bug, so the ceiling is named in the source,
not only here.

### The authorized handle, not just the authorized FileID

Authorizing a candidate and then closing it are two steps, and the thing being
authorized can move between them. `OpenFile.MetadataHandle` is reassigned under
the open's lock when `SET_REPARSE_POINT` replaces a placeholder with a symlink
(`set_reparse_point.go`), so a snapshot taken during the collection pass can stop
describing the file the open holds before the close runs. Authorizing on the
snapshot and closing on `FileID` alone would mean read access was established
against one file while a different one was closed.

So the collection pass copies `MetadataHandle` under `f.mu.RLock()`, `authorized`
maps each cleared `FileID` to the handle its access check was made against, and
the close filter re-reads the live handle under the lock and declines any open
whose handle no longer matches. The lease/oplock release then runs only for the
opens the filter actually closed — an open the filter declined keeps its lease,
because it also kept its handle.

**This narrows the window; it does not close it, and the source says so.**
`closeFilesWithFilter` calls the predicate at the top of its first
`files.Range` iteration, and the predicate drops the read lock before returning;
the same iteration then does its own unlocked reads of `openFile.MetadataHandle`
for `UnlockAllForOpen` and the cache flush, and removal from the table happens
later still. A reparse landing in that span is still possible. It is bounded in
who can cause it — only `SET_REPARSE_POINT` on that open reassigns the field, and
only that open's own holder can issue it, so a requester cannot steer it, and the
open being torn down is the same open either way. Closing it completely means
authorizing against the file *this CREATE names* instead of against each matched
open's file, which is what 3.3.5.9.13 literally says ("the file being opened")
and becomes equivalent once the path and share conditions land, since the two are
then the same file. Recorded rather than quietly left as a claim that does not
hold.

This also corrects the `OpenFile` struct comment, which listed `MetadataHandle`
among the immutable fields "safe to access without the mutex". It is not
immutable, and it is reassigned on the same two lines as `PayloadID`, which the
same comment already documents as mutable.

### Why this is not a one-line predicate

The `OpenFile` concurrency contract forbids metadata-store I/O inside
`handler.files.Range`, so the authorization cannot live in the close filter.
Candidates are collected in one pass over the open-file table — cheap field
reads only — authorized outside the Range, and then closed by an exact FileID
set. The displaced opens' lease/oplock releases run against that same authorized
set, so an open the gate refused to displace keeps its lease.

## Evidence

`test/smb-conformance/smbtorture/run.sh --filter smb2.durable-v2-open`, run
twice independently against `a558b7a2f`:

```
Total 32 | Passed 32 | Failed 0 | Known 0 | New Failures 0 | Skipped 0
smb2.app-instance                                                    PASS
```

`smb2.app-instance` is green, so the failover still fires for the client it is
meant to serve, and the protocol suite stayed green across an authorization
addition.

**Stated honestly: that bound covers `a558b7a2f`, not the final head.** The
TOCTOU fix landed on top, and the suite could not be re-run cleanly on the
loaded machine afterwards — four attempts, none of which reached a graded server
result (client-side `NT_STATUS_NO_MEMORY` at connection establishment, then
harness bootstrap failures). The failures were always a contiguous *tail* of the
suite that grew with machine load, and `app-instance` specifically failed at
`smb2.c:95`, the second connection of the two-connection wrapper. The TOCTOU fix
is therefore covered by unit and `-race` tests only; it needs one clean suite run
on a quiet machine before this merges.

**Merge gate: this needs one clean suite run at the final head first.** That run
is blocked on the harness, not on this branch, and both halves are being fixed
right now: #2619 scopes the Compose project name per checkout so two concurrent
runs stop destroying each other's containers and volumes, and #2618 makes a run
that never reached the server report INCONCLUSIVE — naming the tests and exiting
non-zero — instead of printing a green banner. The sequence is: harness PRs
merge, then one clean `smb2.durable-v2-open` at this branch's final head, then
this merges. The bound above is stated narrowly on purpose; it is not an
oversight to be waved through.

A re-run at the final SHA was not attempted, deliberately: one of those runs
printed `RESULT: All failures are known. CI green.` while `smb2.app-instance` had
in fact failed and been reclassified from failure to skip by the connection-setup
heuristic — a green banner over a silently-skipped `app-instance` is
indistinguishable at a glance from a real 32/32. That is filed as #2618. Until it
is fixed the harness cannot distinguish a pass from a run that never reached the
server, so another attempt would buy noise rather than evidence.

Two of those attempts also wedged on the harness's own state
(`Invalid username or password`, then `Metadata store already exists`), which
needed a `docker compose down -v` and a volume wipe to clear. The smbtorture and
WPTS runners under `test/smb-conformance/` share one compose project name and the
same `dittofs-data` / `kdc-keytabs` volumes, so two agents running them
concurrently can wedge each other's containers. Worth its own issue.

### Checked for false denials, not only for refusals

An authorization gate that is too strict fails safe, so it looks correct in
every test that asserts a refusal. The direction that needs checking is the
other one: does the gate turn away a requester who legitimately can read the
file? It does not, and the reason is checkable rather than hopeful.

The gate compares against `accessMaskPosixRead` = `0x00120089`
(`FILE_READ_DATA | FILE_READ_EA | FILE_READ_ATTRIBUTES | READ_CONTROL | SYNCHRONIZE`),
which is MS-DTYP 2.4.3 GENERIC_READ mapped to file-object rights. Both paths
through `ComputeMaximalAccess` produce a superset of it for anyone who can read:

- **POSIX-mode files** — the read bundle *is* `accessMaskPosixRead`, by
  construction.
- **ACL-bearing files** — `acl.rwxToFullMask` synthesizes the read bundle as
  `ACE4_READ_DATA | ACE4_READ_ATTRIBUTES | ACE4_READ_NAMED_ATTRS | ACE4_READ_ACL | ACE4_SYNCHRONIZE`
  = `0x1 | 0x80 | 0x8 | 0x20000 | 0x100000` = `0x00120089`. Bit-for-bit
  identical, so an ACL-bearing share and a POSIX-mode share answer this gate the
  same way. SYNCHRONIZE in particular is present on both paths, which is the bit
  a narrower synthesis would have dropped.
- **Owners and root** — `accessMaskPosixGenericAll` = `0x001F01FF`, and
  `0x001F01FF & 0x00120089 == 0x00120089`.

So there is no class of legitimately read-capable requester the gate refuses.
The one deliberate narrowing is an open with no resolvable metadata handle
(a pipe open, for instance): it carries no file to authorize against, so it is
never displaced. That is the fail-closed direction.

### A test that was certifying the bug, not catching it

`TestAppInstance_SilentFailover` went red on this change, and it was right to —
but the reason is worth recording, because it is not "the test needed updating
alongside the gate".

That test drove both of its opens through `makeSMBCtx`, which stamps **one
env-wide ClientGUID** on every session it builds. So its two opens were two
sessions of *one* client, which is exactly the case condition 4 says must not
displace. The displacement it asserted was real, and develop really was
performing it — the test was pinning a spec violation in place rather than
catching it.

The smbtorture case it models is two independent connections:
`torture_suite_add_2smb2_test` builds each tree from its own
`lpcfg_smbcli_options` call, and each of those generates a fresh random client
GUID, so the real client satisfies condition 4 and the real failover fires.

The gate was not adjusted to fit the test. `makeSMBCtxForClient` was split out of
`makeSMBCtx`, the two opens were given distinct client GUIDs to match the real
client, and `TestAppInstance_SameClientDoesNotFailover` was added as the negative
half: same CREATE, one ClientGUID, first open must survive.

**Blast radius of the `makeSMBCtx` defect: none beyond this test.** Every other
multi-session caller (`TestReopen2_*`, `TestReplay_*`,
`TestCreateReplay_PanicReleasesReservation`, `TestReopen2_V1Oplock_RestoresRequestedAllocSize`)
is a durable-**reconnect** test — one client returning on a new session — and
those legitimately require the same ClientGuid, since 3.3.5.9.12's lease-scoping
check is keyed on it. `TestAppInstance_SilentFailover` was the only caller that
intended two different clients. Nothing to file.

### Unit coverage

`TestProcessAppInstanceId_ForceClosesOldHandles` and
`TestProcessAppInstanceId_LiveOpenGates` each cover, against a real metadata
store, on the persisted-handle and live-open paths respectively:

- a different client with read access displaces the open (the failover still works)
- a requester with no read access on the file displaces nothing
- the claiming client's own opens are not displaced
- an open whose file cannot be resolved is left alone

Re #2602

Condition 4 is implemented against `Open.ClientGuid` (the GUID that established the open), where the issue quotes `Open.Session.Connection.ClientGuid` (the GUID of the connection its session is on now). The two agree for every open that has not changed clients, and the maximal-access gate — the authorization step this issue is about — is implemented in full. `Re` rather than `Closes` because the issue names that field explicitly; #2660 carries the remainder.


marmos91:
Triage of the latest review. Six of seven were real; five are fixed on this head.

**Fixed (22bc3f986, eda4fde05):**

- *durable_context.go:1188 — the persisted row is not claimed atomically.* Correct and the worst of the set: a DHnC/DH2C reconnect can take the row between the AppInstanceId listing and the teardown, restore the open with its byte-range locks, and then `UnlockAllForOpen` off the stale snapshot strips locks from a **live** handle while the unconditional delete quietly removes nothing. The failover now claims through `ConsumeDurableHandle` — the same seam reconnect claims through, and for the same reason its doc comment gives. Test: `TestProcessAppInstanceId_ClaimsThePersistedRowBeforeCleaningItUp`, both halves fail against the previous head.
- *durable_context.go:1118 — `authorized` is keyed only by FileID.* Correct; a durable reconnect restores an open under its original FileID, so the table can name an open that left and came back. The close predicate now re-tests both match conditions itself rather than trusting the key.
- *open_file.go:36 — `closeFilesWithFilter` reads `MetadataHandle` unsynchronised.* Correct, and pointed at by a comment this PR added asserting the opposite discipline. Added `OpenFile.Handle()` alongside the existing `Name()`/`GetPayloadID()` accessors and routed every read in the shared close path through it, with one snapshot spanning the lock release and the lease break so they cannot name different files.
- *durable_context.go:1123 — the recheck releases `f.mu` before teardown reads the field again.* Same root cause as the above; covered by the same change.
- *durable_context.go:1032 — `GetFile` walks parent edges for a check that never reads `File.Path`.* Correct; `ComputeMaximalAccess` reads attributes and ACL only. Switched to `GetFileForRead`.
- *durable_context_test.go:2679 — `wantSameOrUnknwn`.* Fixed.

**Real, not fixed here — filed as #2660:**

- *durable_context.go:1071 — `OpenFile.ClientGUID` is the establishing GUID, not the session's current connection GUID.* This one I checked against the spec rather than my recollection, and the review is right: §3.3.5.9.13 condition 4 names `Open.Session.Connection.ClientGuid`, while `Open.ClientGuid` is the separate value §3.3.5.9 sets once at open time. They diverge only for an open that reconnected from a different ClientGuid, which this server permits because its reconnect gate applies to lease-backed handles alone — so the honest fix is two-sided and belongs in its own change. The deviation is now recorded as a `decision:` marker on `sameOrUnknownClient`, with the evidence that withdraws it.

marmos91:
Triage of the latest round.

**Recorded rather than changed — durable_context.go:1242 (claim before cleanup).** Correct that `ConsumeDurableHandle` destroys the row before `flushFileCache` and `UnlockAllForOpen` run, and that a crash in between loses it while the locks remain. The previous order is worse, though, which is why it changed: cleaning up and deleting after leaves a window in which a DHnC/DH2C reconnect restores the open, and the cleanup then strips byte-range locks from a **live** handle while the delete removes nothing — a working client damaged silently, with nothing to reconcile it.

The residue this way is orphaned locks whose owner is gone, and lock recovery releases those at the end of the grace period on the next boot. So one failure mode self-heals and the other does not. That reasoning is now a `decision:` marker at the claim, with the evidence that would withdraw it.

**Already filed — durable_context.go:1201 (a disconnect persisting after the scan).** Real, and named in the marker above it. The reason it is not closed here is the other half: non-lease DHnC/DH2C reconnect does not gate on ClientGuid, which is what lets the displaced client reclaim that row. Both halves are #2660; closing only the locking half would leave the reclaim.

**Already filed — durable_context.go:1150 (predicate authorizes, teardown re-reads).** Correct. The predicate's recheck and `closeFilesWithFilter`'s own snapshot are two reads, and a `SET_REPARSE_POINT` between them is not excluded. This PR narrowed it — the teardown now takes one snapshot spanning the lock release and the lease break, instead of re-reading at each step — but making the authorization and the teardown atomic means carrying the authorized handle into `closeFilesWithFilter`, which is a change to a shared teardown path used by LOGOFF, tree disconnect and transport drop. That belongs with #2665, which covers the same field's unsynchronized reads on the explicit CLOSE path, rather than riding a change about AppInstanceId authorization.

PR #2628, opened 2026-09-15
test(smb): isolate the conformance stack per checkout and stop reporting a non-result as green

### LINKED ISSUE #2618: smbtorture harness reports "All failures are known. CI green" while a gating test was silently reclassified as skipped

### LINKED ISSUE #2619: smbtorture harness shares one Docker Compose project across all worktrees, so concurrent runs destroy each other

marmos91:
Two defects in the smbtorture/WPTS conformance harness that compound each other:
one manufactures non-results, the other hides them. They are fixed together
because both live in the same scripts and two separate PRs would conflict.

## #2619 — one Compose project for every checkout

Compose names its project after the directory it is invoked from, so the main
checkout and **every** worktree all resolved to one project called
`smb-conformance`. A run started from checkout B adopted checkout A's
containers, and whichever finished first ran `docker compose down -v` and
removed the other's server, network and volumes mid-run. The surviving run then
reported a connection refusal or a half-initialised metadata store, which reads
exactly like the change under test having broken the server.

**Reproduced** with two pristine checkouts on this base:

```
=== checkout A (main) sees its own stack ===
smb-conformance-dittofs-1	running

=== checkout B (worktree, different directory) ===
/Users/…/dittofs-worktrees/2619-peer/test/smb-conformance
smb-conformance-dittofs-1	running          <-- B sees A's container

=== B's cleanup: docker compose down -v ===
 Volume smb-conformance_dittofs-data Removed
 Network smb-conformance_default Removed

=== checkout A's stack afterwards ===
(empty — A's server destroyed by B)
```

`COMPOSE_PROJECT_NAME` now carries a checksum of the repository root, derived
once in a new `test/smb-conformance/compose-env.sh` sourced by both runners and
by the two Makefile targets that call `docker compose` directly. Deriving it in
one place rather than per script keeps the two copies from drifting apart.

The check rejects **any** live `smb-conformance*` stack, this checkout's own
included — nothing of the current run exists yet when it runs, so a container
carrying our own project name is one an earlier run left behind. Adopting it is
not a shortcut, and Copilot was right to push on this: re-running over a kept
stack is documented in the README and **never worked**. Measured on this branch,
the second run dies at bootstrap with

```
Error: login failed: Invalid username or password
```

which is one of the two signatures #2619 reports verbatim. So part of what that
issue describes is the same-checkout case of the same defect, and the earlier
self-exclusion would have kept hiding it. The README flow now tears the stack
down before the second run.

It is not atomic with the `docker compose up` that follows: two runs starting in
the same instant can both read an empty `docker ps` and then race for 12445 and
8080, and the loser fails at bind. That is loud and attributable to a port
rather than to the server under test, which is the outcome this check is for, so
a host-level lock is deliberately not taken here — recorded as a `decision:`
marker at the check itself.

**Overlapping runs are deliberately not supported.** The published host ports
stay a single set, and the suites assert on lease and oplock breaks within a
second of the request — `smb2.compound_async.rename_last` checks
`req[0]->state == SMB2_REQUEST_RECV` after a one-second wait — margins that do
not survive two suites contending for the same CPU. Making the harness
parallel-safe would manufacture exactly the timing-flake class this repo has
been chasing. So the runners refuse immediately instead, naming the holder:

```
ERROR: an SMB conformance stack is already running from /Users/…/dittofs-worktrees/2619-peer/test/smb-conformance
       (Compose project smb-conformance-2428834042).

The stack publishes fixed host ports and the suites are timing-sensitive, so
only one may run at a time — including a stack an earlier run left behind with
--keep, which a new run cannot reuse. Wait for that run to finish, or stop it
with:
    docker compose -p smb-conformance-2428834042 down -v
```

Verified after the fix: the two checkouts resolve to different projects
(`smb-conformance-3423105764` vs `smb-conformance-2428834042`), `docker compose
ps` from one no longer lists the other's containers, and both `run.sh` entry
points refuse with the message above. A same-checkout `--keep` stack is refused
the same way. A real `./run.sh --filter smb2.connect` with nothing live still
runs to completion and reports green, so the guard does not over-refuse.

## #2618 — "CI green" for a test that never reached the server

When the smbtorture client fails in its own connection setup the harness
reclassifies the failure to a skip. That part is right — a connection failure is
not evidence about the server. What was wrong is the summary: the run then
printed the byte-identical `RESULT: All failures are known. CI green.` banner and
exited 0, so a test that produced no server-side result was indistinguishable
from one that passed. That is how an authorization change shipped with its only
protocol-level coverage silently absent.

A test excused this way is neither a pass nor a failure. The run now reports a
third outcome, names every affected test rather than counting them (the failure
mode is a *gating* test going missing, and a bare count cannot tell you which
one), and exits non-zero — a zero exit asserts the run answered the question it
was asked.

**Reproduced end to end** by pointing the real smbtorture client at a port where
nothing listens, so it emits the genuine
`source4/torture/smb2/smb2.c:37: Establishing SMB2 connection failed`
diagnostic for 13 tests. The same recorded output through both parsers:

```
=== PRE-FIX parser ===
exit=0
| Reclassified as flakes | 13 |
RESULT: All failures are known. CI green.

=== POST-FIX parser ===
exit=13
RESULT: INCONCLUSIVE — 13 test(s) produced no server result!

These tests failed in the client's own connection setup, so they were
graded as skips rather than failures. Nothing was measured for them:
  - smb2.simple
  - smb2.simple_modtime
  - smb2.simple_nodelete
  - smb2.no_sharing
  … (all 13 named)
```

**This will make CI redder on a loaded runner, on purpose.** The
reclassification exists because a starved runner's client can fail its own
connection setup, and until now that was absorbed silently. It still is not
counted as a *failure* — but it is no longer counted as a pass either, and a
re-run is cheaper than a green banner meaning "some unknown subset executed".

Scope note: this gates on tests that never reached the server. Tests cut short
by a suite timeout are a separate axis that already has its own gate — the
`expected_truncation` sign-off list — and folding them in here would red the job
permanently on the sanctioned `smb2.notify` truncation. The excuse for a
self-contradictory ChannelSequence draw is also left green: that test did reach
the server, and a regression test pins it.

## Tests

`parse-results_test.sh` grows four cases and tightens two. The two that
previously asserted exit 0 for a connection-setup excuse now assert the
inconclusive verdict, which is the unit-level reproduction. New cases pin that
every inconclusive test is named individually, that a new failure and an
inconclusive test are both reported rather than one swallowing the other, and
that the ChannelSequence excuse stays green.

The summary table's existing `| Inconclusive |` row (tests killed mid-test) is
renamed `| Killed mid-test |`, so the report does not carry two different
numbers both called inconclusive.

**A shell exit status is one byte**, and now that the count carries one entry
per no-server-result test, 256 of them wrapped to 0 — a false green under an
inconclusive banner, the exact defect class this PR exists to close. Reproduced
with a synthetic 256-test set (`exit 0` before, `exit 254` after) and pinned by a
regression case that was watched failing against the unclamped parser:

```
FAIL: 256 inconclusive tests do not wrap to exit 0: exit 0, want 254
```

shellcheck is clean at `--severity=warning` across all five scripts.

Closes #2619
Closes #2618


marmos91:
Triage of the latest review.

**Fixed on this head (054bf5e91):**

- *run.sh:327 — the sidecar is written where the consumer never looks.* Correct, and it made the whole verdict mechanism inert: `smbtorture/run.sh` set `RESULTS_DIR` to its own timestamped directory and ignored the `DITTOFS_RESULTS_DIR` the common runner exports. It now honours it, the way `test/posix/run-posix.sh` already did.
- *run.sh:331 / parse-results.sh:763 — the status is not a failure count.* Correct. The verdict now carries `category failures no_result`, so a run with one new failure and one ungraded test reads as one failure plus one inconclusive rather than "2 new failure(s)", and 256 ungraded tests no longer read as 254.
- *run.sh:323 — a stale sidecar can be inherited.* Correct; `results_dir` is keyed by suite and label, not by invocation. Cleared at run start.
- *compose-env.sh:65 — a project can survive as labelled volumes.* Correct; `docker compose down` without `-v` leaves them. Admission now scans volumes when no container matches.
- *run.sh:345 / smbtorture/run.sh:251 — the `--keep` cleanup is not runnable from the caller's directory.* Correct, and it was a regression I introduced in the previous round by dropping the `cd`. Restored, on both runners.
- *run.sh:294 — duplicate `Stack` line in dry-run.* Correct; removed.

**Not changed:**

- *compose-env.sh:48 — admission is not atomic with `docker compose up`.* Correct, and it is what the `decision:` marker directly above the function says: two runs starting in the same instant both read an empty table and race for the published ports, and the loser fails at bind. That failure is loud and attributable to a port rather than to the server under test, which is the outcome the check exists for. The marker names the lock that would close it and the evidence that would justify taking one.
- *run.sh:320 — `--mode local` is not covered.* Correct, and it is #2653: local mode starts native DittoFS on fixed ports and a fixed container name, and this guard only scans Compose labels.

Issue #2581, opened 2026-09-14
Block stores cannot be renamed through any interface


marmos91:
Block store names are now unique — the `kind` discriminator that used to scope them is gone, so two stores may no longer share a name. Startup refuses a database where a pair collides, which is the right call: the collision is real and silently picking one would bind a share to the wrong store.

The problem is what the operator can do about it. Nothing renames a block store:

- `dfsctl store block edit` takes `--type` and config flags only, no `--name`.
- `UpdateBlockStoreRequest` is `{Type *string; Config *string}`.
- `apiclient.UpdateStoreRequest` is `{Type *string; Config any}`.

So the only route is direct SQL against the control-plane database with the server stopped:

```sql
UPDATE block_store_configs SET name = '<new name>' WHERE id = '<the row to rename>';
```

That is what `docs/guide/configuration.md` documents and what the startup refusal now points at, so an operator hitting the collision is not stuck. But a documented `UPDATE` statement is a poor substitute for a supported operation, and this is the second place the gap shows up — a share's binding normally holds the store's UUID, yet rows written through the older REST update path hold the *name* instead, so a rename has to consider repointing shares that matched on the old name.

Worth deciding whether rename should exist as a first-class operation:

- If yes, it needs `--name` on `store block edit`, the corresponding request field, and a decision about shares that reference the store by name rather than UUID.
- If no, the SQL recipe should be promoted from a troubleshooting aside into the documented procedure it actually is.

Either way the current state — a refusal whose remedy lives only in a doc section — is the one option that should not persist.

marmos91:
Shipped in #2579, which added `RenameBlockStore`, the `PUT` rename path and `dfsctl store block edit --name`. That PR referenced this issue in prose without a closing keyword, which is the only reason it stayed open after merging at 11:04.

#2635 then fixed defects in that path rather than adding it: the health checker is cached by the store's *name*, so a rename or delete addressed by the UUID route left a probe answering for a name nothing carried any more, and a name probed before its store existed kept answering "not found" after the store was created. Closing against #2579 rather than #2635 so the rename capability is traceable to the commit that implemented it.

PR #2635, opened 2026-09-15
fix(api): evict the health checker under the name it was cached with

marmos91:
Follow-ups to #2579, after a re-review landed on it post-merge.

**The health checker was evicted under keys the request cannot know.** Checkers are cached by the block store's *name*, but a route may address a store by UUID, a rename moves the name the checker was cached under, and a concurrent rename can move it again between the lookup and the write. Reconstructing the affected names from the request produced a fresh bug at each mutation site — update, delete, and create each missed a different one.

So eviction now states the invariant instead of the key set: after any block-store mutation, no block-store probe is trusted. `InvalidateBlockStoreCheckers()` drops the whole `block` kind, and all three handlers call it. That removed the per-name bookkeeping and a `GetBlockStore` round-trip in the delete path, and left each mutation handler with one invalidation to make rather than a set of names to reconstruct. The cost — one re-probe per configured block store on the next status read — is recorded as a `ponytail:` marker with the condition that would justify narrowing it again.

Without it, a store renamed or deleted through its UUID route kept answering with its old health for the cache TTL, and a name probed before its store existed kept answering "not found" after the store was created.

**Two comments contradicted their code.** The share update handler still said a binding change "silently has no effect on mirroring until a full server restart"; it rebinds live and warns only if that fails, and the durability axes apply live too. And `store block edit`'s no-fields error listed every flag except `--name`, the one the same release added.

Every eviction path has a test that was verified to fail with the eviction removed: rename by ID,
delete by ID, create over a name probed while it had no store, a rename that fails after the config
write has already landed, and one asserting that a checker the request never named is evicted too.
That last one is what pins the invariant — the others still pass under per-name eviction.

Review rounds, findings and the reasoning behind each revision are in the PR comments.

Follow-up to #2581. The rename capability itself shipped in #2579; this PR fixes
defects in that path rather than adding it, so it deliberately does not claim to
close the issue — #2581 is closed separately against #2579, which implemented it.


marmos91:
Both Copilot findings were right.

The duplicate is a real duplicate: `TestBlockStoreHandler_Update_ConflictingRenameLeavesConfigUntouched` (added by 3ead6d476, the same commit that added the preflight) makes the same setup, the same PUT and the same two assertions. My comment claiming the preflight had no other test was false. Removed.

Replaced it with coverage for the eviction fix, which is what this PR is actually about: create a memory store, warm the checker under its name, then rename it through the ID route. Negative control — with `InvalidateBlockStoreChecker(prevName)` removed, it fails with "the old name still reports healthy: the checker cached under it survived the rename".

marmos91:
Acting on the suppressed comment from the first review, which the approval does not cover.

`Remove` had the identical defect `Update` did: it deletes by name-or-ID and then evicts only the URL parameter, so a DELETE addressed by UUID leaves the checker cached under the store's real name. Recreate that name and it inherits the dead store's health until the TTL expires. I fixed the path the finding named and left its sibling — the eviction key is the root cause and it needed fixing in both places.

Now resolves the name before the row is deleted and evicts both keys. Negative control: with that removed, the new test fails with "the deleted store's name still reports healthy: its checker survived the delete".

Rebased onto 197439236 (#2620). Four commits, all signed.

marmos91:
Both findings are real, and they pointed at the root cause rather than at two more sites.

Per-name eviction had produced three siblings in this PR alone: `Update` evicting the route parameter when the route may carry an ID, `Remove` with the same defect, and now an intervening rename moving the name between the lookup and the write. Each fix reconstructed which names went stale from a request that cannot know them — the route may address the store by ID, a rename moves the key, and a concurrent rename moves it again mid-sequence.

So rather than serialize the lookup with the write, the eviction now states the invariant directly: after any block-store mutation, no block-store probe is trusted. `InvalidateBlockStoreCheckers()` drops the whole `block` kind from the checker cache. That removes `prevName`, `cachedName`, the extra `GetBlockStore` round-trip in `Remove`, and five call sites collapse to two — 32 lines out of the handler for 31 into the runtime, and no store-interface change.

Cost is one re-probe per configured block store on the next status read, recorded as a `ponytail:` marker with the condition that would justify narrowing it again.

Three tests, each negative-controlled against a disabled eviction:
- rename addressed by ID — "the old name still reports healthy"
- delete addressed by ID — "the deleted store's name still reports healthy"
- a checker the request never named — "eviction is still keyed by the names this handler knows"

That third one is the one that pins the invariant; the first two would pass under the old per-name scheme.

marmos91:
Windows went red on this head with a data race that is not in this diff.

```
WARNING: DATA RACE
Write at 0x00c00050e458 by goroutine 66:
  runtime.(*Runtime).SetMetrics()       runtime.go:882
  commands.runStart()                   start.go:375
Previous read at 0x00c00050e458 by goroutine 239:
  runtime.(*Runtime).runBlockGCSweep()  blockgc.go:88
  commands.runStart.gowrap1()           start.go:301
```

`runStart` launches the block-GC reconcile goroutine at line 301 and installs the metrics handle at line 375; the field has no synchronisation and `StartScheduledGC` gives it a second reader. Pre-existing on develop, nothing to do with block stores, and it will surface intermittently on any PR's Windows job.

Filed as #2645 and fixed in #2646 by hoisting the metrics block above both launches — it depends only on `rt`, `Version` and `Commit`, so it moves unchanged.

Re-running the failed job here rather than folding an unrelated startup fix into this PR. That is a re-run against a known, filed, separately-fixed cause, not a retry hoping for green.

marmos91:
Fixed, and it is the fourth sibling of the same root cause — this time one my own change missed.

I wrote the invariant as "after any block-store mutation, no block-store probe is trusted" and then wired it into two of the three mutation handlers. `Create` never invalidated at all, so a name probed while it had no store cached a not-found verdict that survived the POST for the full TTL. Stating an invariant and enforcing it at two of its three sites is the same failure as fixing the path a report names and leaving the sibling — I just made it one level up, in the rule rather than the call site.

`Create` now calls `InvalidateBlockStoreCheckers()`. Negative control: removed, the new test fails with "status = unknown, want healthy: the not-found verdict cached before the create survived it".

Also rewrote the PR body. It still claimed the surviving regression test pinned a preflight that "had none" — false, the duplicate is exactly why that test was deleted — and it still described the fix as evicting the pre-rename name explicitly, which is not how it works since the rework. Squash makes the body develop's permanent commit message, so it would have landed as a description of a change this PR no longer makes. Every review round stays in the comments, where it does not become the commit message.

marmos91:
Fifth sibling, and the finding is right.

`UpdateBlockStore` commits the type/config before the rename is attempted, so all three rename failure paths return with the store mutated under its original name and nothing evicted. A partial mutation is still a mutation — the invariant does not care that the request did not finish what it started.

Rather than adding a second call site after the successful write and keeping the post-rename one, the eviction is now `defer`red at the point the config lands. That covers all three failure exits and the success path in one place, and covers any exit added later — which matters, since the reason this finding exists is that a new early return was introduced between the mutation and the eviction.

Test needed a rename that fails *after* the preflight cleared it, so it uses a store wrapper that fails only `RenameBlockStore`, with the runtime still on the real store. Negative control: move the eviction back after the rename block and it fails with "the failed rename returned without evicting: the probe cached before the config write is still being served".

Rebased onto 531e0028f. Seven commits, all signed.

marmos91:
Both correct, and the first one means you were right last round and I over-simplified the prescription.

You asked for an immediate eviction after the config write **plus** a deferred one. I replaced both with the defer alone, on the reasoning that a defer covers every exit. It does — but it also leaves the mutation committed and the stale entry live for the whole rename and health-response window, which is exactly the invariant this PR exists to enforce. Consolidating two call sites into one looked like the simpler answer and quietly widened the hole.

Now does both, and the comment says which window each closes: the immediate eviction drops probes cached before the write; the deferred one drops probes cached during the work below, which read a store the rename had not moved yet, and covers the rename failure exits.

On the test: correct, and worse than stated. The `s3` config with no endpoint made `blockstoreprobe.Probe` issue a real `HeadBucket` to AWS with placeholder credentials — the test was 1.08s, which is what that was. Rewritten against a `memory` store: 0.02s, no DNS, no network.

That forced a better assertion too. A store survives a failed rename, so a memory store probes healthy either way and no report distinguishes evicted from cached. The test now asserts the cache entry itself was rebuilt, which is the invariant rather than a proxy for it. Negative control: restore the post-rename-only eviction and it fails with "the checker cached before the config write is still the one being served".

All ten block-store handler tests pass, 375 in the package. Head c2fe9aa1d.

PR #2637, opened 2026-09-15
docs(ci): plan to cut PR wall-clock and make green mean mergeable

marmos91:
## What

A planning document, no code change. It answers two separate questions about DittoFS CI:

1. **Why is PR CI slow?** — and what actually dominates the wall-clock.
2. **Does green mean safe to merge?** — it does not, and how to make it mean that.

## The two findings

**1. Queue starvation, not slow tests.** Lint run `34976565728`: 9 of 10 jobs created at
`13:40:52`, longest job 2m17s, run wall 1254s (21 min). Conformance run `34979004304` is
starker — a **10-second** grader job waited **997s** for a runner. One PR push creates **~44
jobs** against a **~20-slot** free-runner pool. Even in a quiet window, queue wait summed to
~20,602s against ~13,782s of execution. Consequence: parallelising the unit suite across a
matrix would make it *worse*. The fix is fewer jobs.

**2. Green is not enforced.** Neither ruleset contains `required_status_checks` and
`required_approving_review_count` is 0. A PR can merge with every check red.

## Proposed change set

- **Phase 0 — repair the cache budget.** The repo is over the 10 GB Actions cache cap with
  ~95% duplicate Nix NARs and a ~1 hour eviction horizon, so `setup-go`'s Go cache is a
  write-once 10 MB stub that cannot self-heal.
- **Phase 1 — delete jobs.** Lint 10 → 4-5; fold `nfs-pynfs.yml` into `conformance.yml`;
  restrict the CodeQL PR matrix; drop the duplicate grader run and the dead artifact upload.
- **Phase 2 — adopt the gate.** Required status checks on stable names, path filters moved
  from workflow `on:` to job-level `if:`, merge queue on `develop` only.
- **Phase 3 — per-job wins.** `-short` on the unit job, scoped to `pull_request`.

**Explicitly rejected: Bazel/Pants/Please/Nx.** Only ~23% of runner-seconds and 3 of 35 jobs
are hermetically sandboxable; the rest is shell, `sudo`, kernel mounts and docker-compose.
It removes zero jobs and adds one. Nix already holds that slot. Revisit thresholds are in §3.

**Also rejected: reducing Windows test scope** (item 1.4, withdrawn on measurement —
`pkg/block/journal` is 10x and `runtime/shares` 15x slower on Windows, both cross-platform
packages a tagged-file list would have dropped).

## Review history

Five independent passes are folded in, and each changed the document materially:

- **Copilot + a council pass** — found the `lint.yml` missing `timeout-minutes`, the dead
  `dfs-1.26.x` artifact, and the path-filter deadlock that makes a required check unsatisfiable
  on a docs-only PR.
- **A correctness review** — found that the Phase 2 tiers named checks that Phase 1.1 deletes,
  that the aggregator fix was not constructible while `paths:` sat on the workflow `on:` block,
  and that a merge queue on `main` would break the `git branch -f main` release flow.
- **An over-engineering review** — found that `-short` plus Phase 1b would delete a test, and
  the missing CodeQL matrix win (2 slots, one line).
- **An adversarial review** — found the two defects that mattered most, below.

## What the adversarial pass caught

Worth calling out because both would have shipped a conclusion the evidence did not support:

- **The runner-second table was wrong by 2.8x on its two load-bearing rows.** It carried
  Windows Build at 667s (a run from 2026-09-04) and Unit Tests at 390s (matching no run at
  all). Measured now: 1848s and 1055s. The Go share of CI is **~23-24%, not ~10%** — the
  Bazel rejection survives, but its stated margin was 2.4x too generous.
- **Phase 0's designated falsifier could not fail.** It ran `go test -count=1` before the
  probe, which populates the test cache, so the probe printed `(cached)` on a cold cache.
  Phase 0 would have been recorded as verified-restoring while nothing restored.

Both are corrected in place, with the retraction kept visible rather than quietly fixed.

## What I want reviewed

The diagnosis and the ordering. Note §4 now states that **Phase 2 is reorderable to first** —
it costs zero CI time and the "consolidate first" constraint dissolves if the required checks
use stable aggregator names. §7b lists six operational gaps (rollback, baseline capture,
in-flight PRs) that are deliberately left open.

## Verification

§7 carries runnable checks per phase: job-slot counts per push, Lint wall-clock, required-check
presence, cache size and NAR entry count, and a Go-cache probe that now actually asserts.


marmos91:
**Heads-up: the plan changed materially after this PR opened.** If you are reviewing an earlier revision, re-read — the numbers and the gating section are both different.

Two review passes (an oracle and a fresh-context reviewer) plus Copilot found real defects in the first draft. Summary of what moved:

**The projection was overstated.** The first draft claimed `42 → 22` jobs and credited 8 slots to the `nfs-pynfs` merge and 1 slot to the Windows change. Both were wrong:
- The `nfs-pynfs` merge removes **3** scaffolding jobs (`matrix`, `grading`, `build`); its 4 pynfs test cells remain.
- The Windows change shortens a job's **duration** without freeing a **slot**, unless the job is explicitly split.

Corrected: **~44 → ~26 slots**, with duration savings tracked in a separate table. Conflating slots with duration was the single biggest error in the draft, and it's now called out explicitly in §4.

**`Conformance summary` is not a safe gate today — and that's a pre-existing bug.** Its `needs:` is only `[wpts, smbtorture, pjdfstest, nfs-kerberos]`, omitting `matrix`, `graders`, and `build`. And `wpts`/`smbtorture` `need:` only `matrix`, **not `build`** — so a failed `Build Binaries` leaves them running against a missing artifact while `summary` reports green. A failed resolver or grader likewise never reaches the summary.

This means the draft's plan to require `Conformance summary` would have installed a **false-green required check**. It's now Tier 2, gated on fixing the aggregator first. Worth fixing whether or not the gate is ever enabled.

**`merge_group` breaks two payload assumptions** the draft didn't account for:
- `conformance.yml`'s `EVENT` ternary maps only `pull_request` and `schedule`, falling through to `'push'`. Since `suites.json` gives `push` the full `all` matrix, a merge group would run **more** jobs than the PR path it replaces.
- `commit-type-matches-diff` reads `github.event.before`, which a `merge_group` payload doesn't carry. The value is empty, the script hits its "no base commit; nothing to check" success path, and the check passes without inspecting anything.

Both are now in §4 Phase 2, along with a note to audit any other workflow gaining `merge_group:` for the same class of bug.

**A correction to my own proposed command.** The draft's integration-test selector used `go list -tags=integration` and filtered on test-file count. I tested it: that matches **143 of 179** packages, because the tag *includes* files rather than selecting packages that have them. The working selector diffs the two tag states and yields the actual **7** tag-sensitive packages. §4 Phase 1b now has the verified command.

**Smaller corrections:** Lint `10 → 2` was arithmetically impossible (the grouping leaves 5 definitions, 4 active, and `spec-citations`/`required-tests` need Go provisioned); CodeQL contributes 3 jobs not 1; the Windows path-trigger mitigation named `filepath`/`os.PathSeparator`, which are source identifiers rather than paths; the Phase 1 verify command counted only Lint while claiming the global total.

**Also corrected in the draft's favour:** `actions/setup-go cache: true` already caches `GOCACHE`, so the draft's "add build caching" item was simply wrong and was deleted rather than kept.

§9 records all of this as provenance, including which pass found what.

**The diagnosis itself is unchanged** and was independently confirmed by all three passes: ~85% of Lint wall-clock is queue wait, and neither ruleset carries required status checks. Only the magnitudes and the gating mechanics were wrong.

One caveat worth weighing: the queue measurements were taken when 24 runs were queued. Queue depth is bursty, so on a quiet repo the critical path reverts to conformance's genuine ~29 min floor. The structural fix holds either way, but "21 min → 5 min" is not a steady-state promise — §6 says so explicitly.

PR #2652, opened 2026-09-15
ci(nix): stop three dev-shell-only jobs pushing to the Actions cache

### LINKED ISSUE #2641: CI: cut PR wall-clock and make green mean mergeable
PR CI takes ~20-30 minutes of wall-clock and green does not mean mergeable. Both are
costing delivery speed today. The full diagnosis, evidence and change set live in
`.planning/2026-09-15-ci-throughput-and-merge-gating-PLAN.md` (merged in #2637) — this
issue tracks executing it.

## The two root causes, in one line each

1. **Queue starvation, not slow tests.** One PR push creates ~44 jobs against a ~20-slot
   free-runner pool. Even in a quiet window, queue wait summed to ~20,602s against ~13,782s
   of execution. A 10-second grader job was measured waiting 997s for a runner. The fix is
   fewer jobs, not more parallelism — parallelising across a matrix makes it worse.
2. **The gate does not exist.** Neither `Protect develop` nor `Protect main` contains
   `required_status_checks`, and `required_approving_review_count` is 0. A PR can merge with
   every check red.

Independently, the Actions cache budget is blown: the repo is over the 10 GB per-repo cap,
~95% of it duplicate Nix NARs, giving a ~1 hour eviction horizon — so `setup-go`'s Go cache
is a write-once 10 MB stub that cannot self-heal.

## Tranche 1 — do these first (small, measured, low risk)

- [ ] **`-short` on `unit-tests.yml`, scoped to `pull_request`**, full suite retained on
      push-to-develop. `pkg/block/chunker` costs 583s on Linux vs 15s on Windows purely
      because the Linux job never passes `-short`; the guards already exist at
      `chunker_test.go:78` and `boundary_stability_test.go:69`. ~460-500s.
      **The event scoping is not optional** — unconditional `-short` plus Phase 1b would
      leave the full-scale chunker property tests running nowhere on a PR (plan §Phase 3).
- [ ] **Restrict the CodeQL `pull_request` matrix to `go`** (`codeql.yml:33`). 2 slots, one
      line; `actions`/`python` coverage still lands on push-to-develop and the weekly cron.
- [ ] **Add `timeout-minutes` to every Lint job** — currently zero of 10 have one, so a
      wedged job holds a runner slot for the 6-hour default.
- [ ] **Delete the dead `dfs-1.26.x` artifact upload** — verified: every `download-artifact`
      in the repo reads `dfs-binaries`.

## Tranche 2 — job consolidation (bigger diff, needs care)

- [ ] Lint: 10 jobs to 4-5, grouped by shared setup, each check kept as a **named step** so a
      red job still says which check failed. Verify by diffing the check list before/after —
      a silently dropped check is the failure mode here.
- [ ] Fold `nfs-pynfs.yml` into `conformance.yml` (~3 scaffolding jobs). Note this workflow
      has consumers: `ci-health.yml` monitors it by workflow name, `suites.json:158` keys on
      the filename, and three docs reference it.
- [ ] Scope integration tests to the 7 packages that actually have integration-tagged files,
      derived at runtime via the `comm -13` two-tag diff — **never a hand-written list**.
      ~200s of the 397s step is compiling packages the tag changes nothing about.

## Tranche 3 — the cache budget

- [ ] Get the Nix NAR store out of the Actions cache (Cachix/S3, or drop
      `magic-nix-cache-action` from the jobs that only enter a dev shell).
- [ ] Only then add an explicit rolling `~/.cache/go-build` cache, and **verify it restores**
      before believing anything about it. Key it on a build-configuration discriminator, **not
      `github.job`** — Phase 1.1 changes job ids and merges `go vet` with `-race` tests under one
      key, which recreates the exact pathology the cache is meant to fix. Size it against open-PR
      count, not one run: a full GOCACHE here is 4.3 GB and the cap is 10 GB. This is the least-evidenced item in the plan: the
      one CI-side natural experiment available found no signal, and the 342s-to-32s figure is
      a local same-machine re-run, not a CI projection.

## Tranche 4 — the gate (costs zero CI time, arguably the most important)

- [ ] Add required status checks to `Protect develop`, using names that survive the
      consolidation above — hence this comes after tranche 2, or uses stable aggregator names.
- [ ] Move path filters from workflow `on:` blocks to job-level `if:` gating before requiring
      any path-filtered check, or docs-only PRs deadlock forever on "Expected — waiting".
      `dorny/paths-filter` is already used this way at `lint.yml:204`.
- [ ] `Required Tests` must be in the required set — it exists because 83e532d78 shipped
      through its absence.
- [ ] Merge queue on `develop` only. **Not on `main`** — nothing reaches `main` by PR, and a
      required queue would block the `git branch -f main && git push` release flow.

## Explicitly not doing

**No Bazel / Pants / Please / Nx.** Only **~23%** of runner-seconds and 3 of 35 jobs are
hermetically sandboxable; the other ~76% is shell, `sudo`, kernel mounts and docker-compose.
It removes zero jobs and adds one. Nix already holds that slot. Revisit thresholds are in the
plan's §3.

**No reduction of Windows test scope.** Measured: `pkg/block/journal` is 11s on Linux but
110s on Windows, `runtime/shares` 7s vs 106s — real platform I/O in cross-platform packages
that a windows-tagged-file list would have dropped.

## How we'll know it worked

The plan's §7 carries runnable verification per tranche: job-slot count per push, Lint
wall-clock, required-check presence, cache size and NAR entry count, and a probe that fails
if the Go build cache is not actually restoring.

---

*Filed from the merged plan (#2637, `ac3435ef5`). The plan's §7b lists six operational gaps —
rollback, baseline capture, in-flight PRs, incremental required-check adoption — that need a
decision before Phase 1 or Phase 2 lands. §4 notes that **Phase 2 is reorderable to first**:
it costs zero CI time, and the "consolidate first" constraint dissolves if the required checks
use stable aggregator names.*

marmos91:
Tranche 3, first half. **Does not close #2641.**

## The measurement

Every one of the repo's 6,423 cache entries, pulled from the API:

| | entries | size |
|---|---|---|
| **Total** | 6,423 | **10.21 GB** — over the 10 GB cap, so eviction is continuous |
| Nix NARs | 3,223 | 8.89 GB — **87% of every byte in the cache** |
| — *distinct* NAR paths | **249** | **0.46 GB** |
| — redundant copies | 2,974 | **8.43 GB — 94.9% of the NAR bytes** |
| everything else | 3,200 | 1.32 GB (mostly ~0-byte `.narinfo` entries, also Nix) |

A binary cache holding these 249 paths once would be under half a gigabyte. It is instead 8.89 GB, because `magic-nix-cache-action` writes each path under a fresh per-run key.

## What that evicts

The two entries that would actually make CI fast:

```
480.1 MB  setup-go-Linux-x64-ubuntu24-go-1.26.0-f90f2b60…
 10.1 MB  setup-go-Linux-x64-ubuntu24-go-1.26.8-f90f2b60…
```

The workflows pin `go-version: "1.26.x"`, which resolves to **1.26.8**. So the 480 MB entry is under a key nothing can request again, and the live Go cache is a **10 MB stub**. This is the plan's §2.8 claim, confirmed on live data rather than inferred.

## The change

Copilot refuted the first version of this PR and was right; see the comment thread. `devShells.ci` is **not** stock nixpkgs — `commonBuildInputs` carries `pynfs` (`flake.nix:366`) and `linuxInputs` carries `pjdfstest` (`flake.nix:391`), and neither substitutes from cache.nixos.org. Dropping the cache there would have made those jobs rebuild an NFS conformance suite every run.

What that exposed is better than the original idea: **two jobs were building a conformance suite in order to run `go build`.**

| job | Nix use | change |
|---|---|---|
| `conformance.yml` Build Binaries | `nix develop .#ci` → **`.#build`** | cache dropped — nothing realized locally |
| `nfs-pynfs.yml` Build Binaries | `nix develop .#ci` → **`.#build`** | cache dropped — nothing realized locally |
| `e2e-tests.yml` E2E | `nix develop .#ci` | **cache kept** — needs nfs-utils, acl, attr, dfs-mount |
| `conformance.yml` pjdfstest | `nix build .#pjdfstest` | kept |
| `nfs-pynfs.yml` pynfs | `nix build .#pynfs` | kept |
| `nfs-pynfs.yml` knfsd baseline | `baseline-knfsd.sh:75` | kept |

`devShells.build` is `buildInputs = [ pkgs.go_1_26 ]` and nothing else. Verified locally: every path substitutes from cache.nixos.org, and `CGO_ENABLED=0 go build -o dfs cmd/dfs/main.go` succeeds inside it.

Two of six writers stop, and the two that stop also stop compiling `pynfs` and `pjdfstest` for no reason.

## Why not Cachix or S3

Both are the right long-term answer and both need a credential this change cannot add. A public repo's fork PRs get no secrets, so a secret-backed cache silently degrades to no cache on exactly the PRs least likely to be trusted. Worth doing deliberately, not as a side effect of this.

## What to expect

- Nix NAR *production* drops by half — three of six writers stop. This does not shrink the existing 8.43 GB of duplicates; that needs a separate purge, which I am doing next.
- The three affected jobs may run marginally slower on their first Nix step, substituting from cache.nixos.org rather than the GHA cache. That trade is the point: cache.nixos.org is free and does not count against the repo's 10 GB.
- Once the budget has headroom, `setup-go`'s cache can survive between runs for the first time, which is what tranche 3's second half depends on.

## Verification

- `actionlint` output is byte-identical before and after across all three files. Every finding it reports predates this change.
- Cache figures are from `GET /repos/marmos91/dittofs/actions/caches` paginated over all 6,423 entries, grouped by NAR hash with the per-run suffix stripped.
- The real test is the cache total after a few develop pushes. `GET /actions/cache/usage` is the probe.

Part of #2641.


marmos91:
Copilot's finding is correct and I've changed the approach because of it.

I claimed `devShells.ci` is stock nixpkgs. It isn't. `devShell.buildInputs = commonBuildInputs ++ linuxInputs ++ darwinInputs`, `commonBuildInputs` carries `pynfs` (`flake.nix:366`) and `linuxInputs` carries `pjdfstest` (`flake.nix:391`). Neither substitutes from cache.nixos.org, so entering `.#ci` realizes both from source. Removing the cache would have made those jobs rebuild an NFS conformance suite every run — slower, not faster, and the opposite of this PR's purpose.

I read `devShells.ci = devShell.overrideAttrs { shellHook = goEnv; }`, concluded "standard dev shell", and never traced `devShell`'s inputs. Finding the definition stood in for reading it.

The useful thing the finding exposes is that **two jobs were building an NFS conformance suite in order to run `go build`** — a cost they were paying whether or not the cache existed.

So rather than keep the cache or introduce a slim shell only to preserve caching, the fix removes the need:

- **`devShells.build`** — `buildInputs = [ pkgs.go_1_26 ]`, nothing else.
- `conformance.yml` and `nfs-pynfs.yml` **Build Binaries** now enter `.#build`. No conformance suite, nothing realized locally, no cache needed.
- `e2e-tests.yml` **keeps `.#ci` and keeps its cache** — it runs the suite itself and needs `nfs-utils`, `acl`, `attr` and the `dfs-mount` helpers, so it cannot avoid the two derivations.

Verified locally: every path in `.#build` is copied from cache.nixos.org, and `CGO_ENABLED=0 go build -o dfs cmd/dfs/main.go` succeeds inside it.

Net change from the original: two of six writers stop instead of three, and the two that stop also stop compiling `pynfs` and `pjdfstest` for no reason.

marmos91:
Superseded in direction by #2656: the Nix store belongs on Cubbit DS3, not in the 10 GB Actions budget.

This PR stays worth merging on its own terms, and one part of it survives #2656 entirely. `devShells.build` is not a caching workaround — a job that only runs `go build` should not be realizing `pynfs` and `pjdfstest` whether or not a binary cache exists. That cost was there before this PR and would still be there after DS3 lands.

The cache removals here are containment and #2656 replaces them with the real fix.

marmos91:
Correction before merge: the claim in this body that `go-1.26.0` was "a key nothing can request again" is wrong. `codeql.yml:63` and `operator-tests.yml:41,71,89` resolve Go via `go-version-file: k8s/dittofs-operator/go.mod`, whose `go 1.26.0` directive is an exact pin. Four jobs ask for that key. Deleting it cost one cold `setup-go`; the rest of the purge stands. Full correction on #2641 (comment 5688429014). The squash body below drops the claim rather than carrying it into history.

PR #2654, opened 2026-09-15
ci(integration): run only the packages the integration tag changes

### LINKED ISSUE #2641: CI: cut PR wall-clock and make green mean mergeable

marmos91:
Tranche 2, third item. **Does not close #2641.**

## What it does

`go test -tags=integration -v -timeout=20m -p 1 ./...` compiled and ran **179 packages** to exercise the **7** whose test set the integration tag actually changes. The other 172 were a second, slower copy of the unit suite.

```
internal/controlplane/api/handlers   pkg/controlplane/runtime
pkg/block/remote/s3                  pkg/controlplane/store
pkg/metadata/store/badger            pkg/metadata/store/metabench
pkg/metadata/store/postgres
```

## The list is derived, never written down

A hand-maintained list goes stale silently the first time someone adds a tagged file to a package not on it — the suite stays green and simply stops running the new tests. So the job derives it:

```bash
go list -f '{{.ImportPath}} {{.TestGoFiles}} {{.XTestGoFiles}}' ./...                 | sort > plain
go list -tags=integration -f '{{.ImportPath}} {{.TestGoFiles}} {{.XTestGoFiles}}' ./... | sort > tagged
comm -13 plain tagged | awk '{print $1}' | sort -u
```

Two deliberate choices:

- **Compares the test-file *list*, not its count.** The issue proposed counts; both give 7 today, but a tag that *swaps* one file for another leaves the count unchanged and would be missed. Same cost, strictly more robust.
- **Materialises both outputs to files** rather than comparing through `<(...)`. A `go list` failure inside process substitution is invisible — `comm` would simply see an empty side, yield nothing, and the job would run zero tests and report success. An empty derivation now fails the job explicitly:

```
::error::derived no integration-tagged packages. Either the tag is gone from
the tree or this derivation is broken; running nothing and reporting success
is the one outcome this job must never produce.
```

## What this costs, stated plainly

This job was **the only place a `-short`-guarded test still ran at full strength on a pull request**, because it passed no `-short`. It no longer is.

After this, on a PR:

| | unit-tests | integration-tests |
|---|---|---|
| `-short` guards | skipped / strength-reduced | **not run** (outside the 7) |
| `-race` | yes | no |

Six of the nine packages holding `testing.Short()` guards are outside the 7: `pkg/snapshot`, `pkg/metadata`, `pkg/block`, `pkg/block/compression`, `pkg/block/chunker`, `pkg/block/engine`. Their guarded tests now run at full strength **only on the push to develop**. The sharpest case is `pkg/metadata/parent_mtime_contention_test.go`, a concurrent-writer test that loses both full strength and `-race` at PR time.

That is a deliberate trade — PR-time signal is fast, develop-push is the full gate — taken with the repo owner's agreement, not a side effect. The `decision:` marker at the step names it and its reversal condition: widen back to `./...` if a defect ever reaches develop that the full PR-time run would have caught.

## Verification

- `actionlint` clean.
- Derivation run locally against this tree: **7 packages, from 179**.
- `xargs` form expands to a single `go test` invocation with all 7 paths.
- Empty-list guard exercised: `count=0` exits 1.
- `-p 1` retained — several of the 7 share the single `dittofs_test` database and the conformance suite truncates it between subtests.

Part of #2641.


marmos91:
Copilot's suppressed comment was right and I fixed it in `64205fdf1`.

The `decision:` marker said short-guarded tests "now run at full strength only on the push to develop". That is false for two of them. Intersecting the derived set against the packages that actually contain `testing.Short()`:

| | in derived set | has `Short()` guards |
|---|---|---|
| `pkg/controlplane/runtime` | yes | yes |
| `pkg/metadata/store/badger` | yes | yes |
| `pkg/block/chunker`, `pkg/block/engine`, `pkg/block`, `pkg/block/compression`, `pkg/metadata`, `pkg/snapshot`, `internal/bench` | no | yes |
| `internal/controlplane/api/handlers`, `pkg/block/remote/s3`, `pkg/controlplane/store`, `pkg/metadata/store/metabench`, `pkg/metadata/store/postgres` | yes | no |

So the integration job still gives the runtime and badger guards an unshortened PR run; the chunker, engine, compression, metadata and snapshot guards are the ones that now only run at full strength on develop. The marker names them explicitly rather than claiming the whole class.

Worth stating why this mattered enough to restart CI over: per `CLAUDE.md`, a `decision:` marker that overstates its ceiling is indistinguishable from a bug to whoever audits it later. The whole point of the marker is to tell the next person what coverage was traded away, and it was naming the wrong trade.

PR #2658, opened 2026-09-15
test(smb): admit one conformance run per host with a lease, not a Compose label scan

### LINKED ISSUE #2653: smbtorture harness: --mode local bypasses the exclusive-stack admission check
`require_exclusive_stack` (`test/smb-conformance/compose-env.sh`) decides whether another conformance stack is live by scanning Docker for containers labelled with a `com.docker.compose.project` matching `smb-conformance*`. That is the right question for the Compose path, and #2628 makes it answer correctly for both running and stopped projects.

`run.sh --mode local` does not go through Compose. It starts a native `dfs` process and a plain `docker run --name wpts-local`, neither of which carries a Compose project label, so the gate sees nothing and admits the run. Two consequences:

- Two local-mode runs both pass admission and then race for ports 8080 and 445, which is the exact non-result failure the gate exists to prevent — and the one that is hardest to read, because the second run's failure looks like the server under test refusing connections.
- The second run's cleanup does `docker rm -f wpts-local`, which removes the *first* run's WPTS container by name. A run that passed admission can therefore destroy a run already in progress.

A local run and a Compose run are equally invisible to each other for the same reason: admission is keyed on a Compose label rather than on the thing actually being contended, which is the pair of host ports.

The fix wants one admission mechanism both modes share — a host-level lock file or lease taken before either path starts anything and held through cleanup, or a check keyed on the published ports rather than on container labels. Either replaces the label scan rather than adding to it.

Not fixed in #2628: that PR closes #2619 (one Compose project shared across every checkout) and #2618 (a false "CI green" over tests that never reached the server), and both of those are about the Compose path. Local mode is a second, separate way into the same non-result state, and closing it means changing how admission works rather than widening what it scans.

Re #2618, #2619, #2628.

marmos91:
## What was wrong

`require_exclusive_stack` decided "is another conformance run in flight" by asking Docker
for containers carrying a `com.docker.compose.project` label matching `smb-conformance*`.
That question is only answerable for the Compose path. `run.sh --mode local` starts a
native `dfs` and a plain `docker run`, neither of which carries a Compose label, so the
gate saw nothing and admitted the run — every time, including when another run was already
up. Two local runs, or a local run and a Compose run, both passed admission and then raced
for ports 445 and 8080. That surfaces as the server under test refusing connections, which
is the hardest possible reading of the failure.

Admission was keyed on a Compose label; what is actually contended is the host.

## The fix

One lease, taken by both entry points before either creates anything, and held until the
run's last process is gone. It is an advisory lock on `/tmp/dittofs-smb-conformance.lease`,
held on a file descriptor, so the kernel releases it however the run ends — a killed run
leaves nothing to reap and no stale lock for the next run to time out or override. The
label scan is gone; it did not answer the question it was being asked.

The descriptor is inherited by the run's children, which is deliberate and is stated as a
`decision:` at the code. `--mode local` backgrounds a `dfs` that survives a harness killed
outright, and that server still owns the ports — so the lease staying taken is correct, not
a leak. The refusal points at `lsof` rather than claiming the recorded pid is still the
holder.

`perl`'s `flock` rather than `flock(1)`, which is util-linux and absent from macOS, where
this harness also runs. Recorded as a `ponytail:` marker with the swap condition.

**One check survives alongside the lease, narrowed.** A Compose stack left behind by
`--keep` outlives the run that made it, so no lease can see it, and `docker compose up`
adopts it and fails partway through on an already-bootstrapped server. That refusal stays,
scoped to this checkout's own project rather than scanning for any `smb-conformance*`. A
leftover from a *different* checkout now fails at the port bind instead of being named —
loud, and attributable to a port rather than to the server under test, which is what the
existing `decision:` on this path already accepts.

## The second consequence: cleanup removing a foreign container

`--mode local` named its WPTS container `wpts-local` and cleanup did
`docker rm -f wpts-local`. A fixed name is a name another run also owns, so a second run's
cleanup destroyed the first run's client mid-suite. It is now `wpts-local-$$`, matching the
convention `smbtorture/run.sh` already uses (`smbtorture-run-$$`). The lease makes the
overlap unreachable; the name makes the removal unable to reach anything this run did not
start even if it somehow does.

Demonstrated against real Docker:

```
$ docker run -d --name wpts-local alpine:3 sleep 300
$ # the cleanup line as it now stands, from a different run:
$ WPTS_LOCAL_NAME="wpts-local-$$"; docker rm -f "$WPTS_LOCAL_NAME"
$ docker ps -a --filter name=wpts-local --format '{{.Names}} {{.Status}}'
wpts-local Up 4 seconds          <- survives

$ # the line as it was:
$ docker rm -f wpts-local
wpts-local
$ docker ps -a --filter name=wpts-local -q | wc -l
0                                 <- destroyed
```

## What the lease cannot cover, and a check that looked like it would

Removing the lease file while a run holds it lets the next run create a fresh inode at
the same path, lock that, and be admitted beside a live run. Review proposed comparing the
locked inode against the path. I implemented it and it does not work: the contender opens
the file it just created, so its locked inode and the path agree, and it is admitted.
Measured, not reasoned about — `CONTENDER-ADMITTED` with the check in place. No path-based
advisory lock is immune to this; closing it means a lock with no path, an abstract socket
or a fixed loopback port. The check is out and the limitation is a `decision:` at the code
naming what would overturn it.

`perl` is a new dependency this change introduces. Without it the run is refused, which is
the safe direction, but it used to be reported as a held lease — the reader would go
hunting for a run that does not exist. It now says so.

## Evidence

`test/smb-conformance/compose-env_test.sh` exercises the real `compose-env.sh` — it sources
it and calls the real `require_exclusive_stack`, with two copies at different repository
roots standing in for two checkouts, so what it proves is contention across checkouts and
not a project name blocking itself.

Against this branch, 12/12, and 15 consecutive runs with no flake:

```
ok: distinct checkouts (smb-conformance-3203783324 vs smb-conformance-3424552271)
ok: a second run from another checkout is refused
ok: the refusal names the holder
ok: the refusal names the lease
ok: a killed holder leaves no lease behind
ok: the lease survives while a process of the run does
ok: the lease goes with the run's last process
ok: successive runs are admitted
ok: the lease is released when its run exits
ok: an unopenable lease refuses the run
ok: a missing perl refuses the run
ok: the refusal names perl
PASS
```

Against the pre-change `compose-env.sh`, 7 of those fail — including the one that matters:

```
FAIL: a second run from another checkout is refused: exit 0, want 1
FAIL: the refusal names the holder: output missing 'pid 4604'
FAIL: the refusal names the lease: output missing '.../lease'
FAIL: the lease survives while a process of the run does: exit 0, want 1
FAIL: an unopenable lease refuses the run: exit 0, want 1
FAIL: a missing perl refuses the run: exit 0, want 1
FAIL: the refusal names perl: output missing 'needs perl'
```

Also verified: passes under macOS `/bin/bash` 3.2 (nothing here needs bash 4); ShellCheck
clean at `severity=warning` over `compose-env.sh`, `compose-env_test.sh` and `run.sh`; the
other three grader tests in the same CI job still pass. The test is wired into the existing
`Grading logic` job, which takes about a second and needs no Docker.

## Not covered, and worth its own issue

`--dry-run` returns before admission, and `render_ptfconfig` does
`rm -rf ptfconfig-generated/TestResults` — so a `--dry-run` issued while a run is in flight
deletes that run's TRX output. It is the same class as the container removal above, but
fixing it means deciding what `--dry-run` should do during a live run, which is a contract
change rather than a bug fix.

## Base

`compose-env.sh` exists only on #2628, so this is based on `fix/2619-smbtorture-isolation`
rather than `develop`. It needs #2628 merged first; retarget to `develop` after that.

Closes #2653.


marmos91:
Superseded — closing in favour of #2697.

#2628 (1773e8cde) landed after this branch and already covers the per-checkout Compose scoping and the inconclusive-vs-failing grading that make up most of this diff, which is why it conflicts on six files.

The one thing here that develop does *not* have is the host-wide admission lease, and that is real: develop's `claim_exclusive_stack` is keyed by `COMPOSE_PROJECT_NAME`, so it serializes runs from one checkout only. Its own comment records the residual and defers it —

> Two different checkouts starting together both pass here and then race for the fixed host ports ... Take a host-wide lock across every checkout only if a port-bind loss is ever mistaken for a server fault in practice.

This branch is that host-wide lock. Rather than rebase 10 commits whose other 90% is already on develop, the lease has been refiled on current develop as #2697, which keeps develop's claim and leftover scan alongside it (they answer different questions) and ports this branch's lease unit test. #2697's test fails 7 assertions against develop's `compose-env.sh`, so it pins the new behaviour rather than restating the old.

Closing rather than merging, so the git history does not carry a second copy of #2628.

PR #2662, opened 2026-09-15
fix(nfs): apply a register-with-system flip that lands mid-transition

### LINKED ISSUE #2634: Flake: TestReconcileSysregTogglesSidecar fails intermittently under -race
Seen once locally on `go test -race ./pkg/adapter/nfs/... -count=1`:

    --- FAIL: TestReconcileSysregTogglesSidecar (5.02s)
    FAIL    github.com/marmos91/dittofs/pkg/adapter/nfs     10.253s

Not reproducible so far: 5 subsequent runs pass, including two of the exact `./pkg/adapter/nfs/...` form and one run of the whole package alone. No DATA RACE was reported, so the -race flag is relevant only for the timing it changes, not for a detected race.

Ruled out as the cause: the run happened while converting `NFSConnection.pendingCBReplies` to `atomic.Pointer`, but that field is the NFSv4.1 backchannel REPLY demux and has nothing to do with sysreg sidecar reconciliation. The same test passes on the branch HEAD without that change and fails on neither consistently.

The 5.02s duration against a ~6s whole-package run suggests it is waiting on something with a timeout rather than asserting immediately — likely a poll/settle loop in the reconcile path that is tuned tight enough to lose under `-race` slowdown when the package is run alongside `identity`. Worth a look at whatever bound it waits on before this surfaces in CI, where the machines are slower than a laptop.

Filed rather than retried per the repo rule on flakes; leaving it unfixed only because it did not reproduce, not because it is acceptable.

marmos91:
## The flake had a real defect behind it

The issue guessed at "a poll/settle loop in the reconcile path tuned tight enough
to lose under `-race` slowdown". The bound is real, but the mechanism is not a
*slow* reconcile — it is a **dropped** one, and it is a production defect
independent of the test.

`reconcileSysreg` ran at most one background transition at a time. A caller that
lost the CAS returned having done nothing:

```go
if !s.sysregReconciling.CompareAndSwap(false, true) {
    return
}
```

Nothing re-fires the reconcile afterwards, so a `register_with_system` flip that
arrives while a transition is in flight is **lost**. The old comment ("a flip that
arrives mid-transition is picked up by the next apply") describes an apply that is
not guaranteed to come — the callers are the accept loop and the settings watcher,
so on an idle server the next one may be arbitrarily far away.

The window is not small. The transition calls `sysreg.Ping` against the host
rpcbind, and `startSystemPortmapRegistration` bounds the whole registration by
`systemRegTimeout` (10s). Against an unreachable or slow rpcbind, a live
toggle-off issued during a toggle-on is silently discarded for up to that long.

## Why the test saw it

`auxsvc.Group.Start` publishes a service's running state when it **reserves the
name**, before `s.Start` runs (`auxsvc.go:122` — deliberate, since `Start` runs
outside the group lock). So `waitSysreg(t, a, true, ...)` was satisfied while the
enable transition was still talking to rpcbind. The test then flipped to `false`,
hit the dropped-flip path, and polled its full 5s bound — matching the reported
5.02s. The test normally wins that race only because `127.0.0.1:1` refuses
instantly.

## Evidence

**Deterministic reproduction.** Holding the transition open makes the window
certain instead of lucky: `slowSysregAddr` accepts and stalls before closing, so
the flip always lands mid-transition. On the base commit:

```
--- FAIL: TestReconcileSysregAppliesFlipDuringTransition (10.00s)
    sysreg sidecar never settled on running=false with register-with-system
    disabled during an in-flight enable: running=true, reconcile state=0
```

`reconcile state=0` is the tell — the transition had already finished. The flip
was not late, it was discarded. On this branch: **PASS in 0.25s**.

**The original flake, reproduced and quantified.** The unmodified test was soaked
on both trees on the same 8-core Linux VM:

| tree | iterations of the flaky test | failures |
| --- | --- | --- |
| base (`c6ac4d673`) | 3000 | **26** |
| this branch, handoff fix | 3000 | **0** |
| this branch, as shipped | 3000 | **0** |

Form: `go test -race ./pkg/adapter/nfs/ -run TestReconcileSysreg -count=3000`,
pinned to 2 of 8 cores with `taskset -c 0,1` to emulate a slower runner, with
unrelated load on the box. `TestReconcileSysregTogglesSidecar` — the test that
flaked — ran 3000 times in every arm. Failure counts were taken from the raw
logs rather than the harness's own tally.

Every one of the 26 base-tree failures is the same assertion — the third wait, the
disable step — at 5.00-5.01s:

```
--- FAIL: TestReconcileSysregTogglesSidecar (5.01s)
    sysreg_reconcile_test.go:37: sysreg sidecar still running after disabling register-with-system
```

No `DATA RACE` was reported in any of them, consistent with the issue's note that
`-race` mattered only for the timing it changes.

## The fix

Replace the single in-flight bool with a three-state handoff (`sysregIdle` /
`sysregRunning` / `sysregDirty`). A caller that finds a transition running marks
it dirty instead of returning; the transition loops, re-reading the setting, until
a pass completes with no flip having arrived during it. The settling
`CompareAndSwap(sysregRunning, sysregIdle)` closes the window between the worker's
last read and its exit — a caller that marks dirty in that gap makes the CAS fail
and earns another pass.

`configMu` is what orders the data (the caller writes the setting before calling,
the worker reads it under the same lock); the atomic carries only the "do another
pass" signal. So either the worker's read sees the new value, or the dirty mark
survives and forces a pass that will.

It is bounded, not a spin: a pass is only repeated when a concurrent caller marked
dirty, and each caller marks at most once per call.

## A second defect found while reviewing this

`Stop` tears the sidecars down (`shutdown.go:51`) **before** closing the listener
(`shutdown.go:67`), so connections keep being accepted throughout `StopAll`, and
each one runs `preAcceptCheck` -> `applyNFSSettings` -> `reconcileSysreg`. Once
`StopAll` has emptied the running set and cleared the base context, `Reconcile`
can only no-op and the steady-state check can never match again while the setting
is still enabled — so every accepted connection, and every settings poll for the
rest of the process's life, claimed a transition that could do nothing.

That churn predates this change, but the dirty handoff would have added repeated
worker passes on top of it. `reconcileSysreg` now refuses to claim against a group
that is not ready. `Serve` seeds the group before its own reconcile, so the
initial start is unaffected — verified by removing the guard and watching
`TestReconcileSysregNoOpsAfterGroupTeardown` fail with `state 2`.

## Two further defects, found by attacking this change rather than reading it

**The steady-state check was two reads, not one observation.** It read the
running state and then the reconcile state, so a caller could see the running
value from *before* an in-flight mutation and the idle that the very same
transition published *after* settling, conclude the sidecar already matched, and
return with its flip applied nowhere — the same lost update the dirty handoff
exists to prevent, one layer out. Demonstrated with an injected preemption:

```
config wants register_with_system=true; sidecar running=false, reconcile state=0
LOST UPDATE: enable dropped, nothing in flight to apply it
```

This one is pre-existing and the change had already narrowed it — the old code
had no state check at all, so its window was the whole transition. Swapping the
two reads closes it: a transition holds the state non-idle from its claim until
after its mutation is visible, so observing idle first means nothing is in
flight, and any transition claimed afterwards re-reads a setting this caller has
already written. The order is load-bearing and says so in a comment, because it
reads like an arbitrary conjunct order that someone would happily "tidy".

**The new in-flight test was itself a timing bet.** Holding the registration open
for a fixed 250ms made `state == sysregRunning` an assertion that the test
goroutine gets scheduled inside that window; forcing a late poll made it fail
3/3, with a message that reads like a correctness failure. It now holds the
registration open with a gate released by the test, so the race is arranged
rather than waited for — verified by re-running the same late-poll scenario,
which now passes. Shipping a new flake in a flake fix would have been a poor
trade.

Also noted and deliberately not changed: the worker no longer resets the state
from a `defer`, so a panic inside the transition would latch it. Nothing in this
path recovers, so such a panic ends the process and the latch never observes —
it becomes live only if a recover is added above it.

## Making the wait honest

The old `waitSysreg` polled `IsRunning` and failed with a bare string. `IsRunning`
is a proxy: it flips at name-reservation time, so it can be true while a transition
is still able to move it. It now waits on the settled state — two facts, no
transition in flight AND the running state matching — and on timeout reports what
it actually observed rather than restating the expectation. `waitRunning` is kept
separate and named for the weaker property, because the new test genuinely needs
it: it has to catch the transition *while* it is still running.

## Scope check

The other `CompareAndSwap(false, true)` guards in the tree (`gencache.pruneToHalf`,
`RemoteSync.pruneReadahead`, the journal `busy` latches) are not the same defect —
they guard idempotent trim-toward-a-target work where a dropped call is harmless
and self-documented as such. sysreg was the only one carrying a desired *state*
that a dropped call loses.

## Verification

- `go build ./...`, `go vet ./...`, `golangci-lint run ./pkg/adapter/nfs/...` clean.
- `go test ./pkg/adapter/nfs/... -race -count=1` — 133 pass.
- All three new tests A/B'd against the base commit: each fails there, passes here.
- `go test ./pkg/adapter/nfs/ -run TestReconcileSysreg -race -count=50` — 150 pass.
- Reviewed by a simplification pass, a correctness pass and an adversarial pass;
  the second and third each found a defect that is fixed above.

Closes #2634


marmos91:
Triage of the two Copilot comments, at `dittofs-65`'s request. I am not the author of this PR and have changed nothing — this is an assessment only, and the merge decision is Marco's.

**Both findings are real.**

### `sysreg_reconcile_test.go:125` — confirmed, and it is the sharper of the two

```go
running, state = a.sidecars.IsRunning(sysregSidecarName), a.sysregState.Load()
```

Go evaluates those left to right, so `IsRunning` is sampled first and `sysregState` second. The production check does it the other way round (`auxservices.go:161`):

```go
if s.sysregState.Load() == sysregIdle && s.sidecars.IsRunning(sysregSidecarName) == want {
```

The order is load-bearing given what the helper's own doc comment promises — "no transition is still in flight, **and** the sidecar's running state matches". Reading `running` *before* establishing that nothing is in flight means the running value is not known to be post-transition: a transition in flight at the first read and finished by the second yields pre-transition `running` together with `sysregIdle`, and the helper returns success on a value that has already been superseded.

Worth stating plainly because of what this PR is: the helper was added to replace `waitRunning` precisely because a single-fact wait could observe a mid-flight value. It reintroduces a two-read race of the same shape it was written to remove. Swapping the two operands to match `auxservices.go:161` is the whole fix.

### `auxservices.go:189` — confirmed as a structural ordering hazard; reachability I did not verify

The ordering Copilot describes is exactly what the code does (`portmap.go:213-214`):

```go
s.sysregActive.Store(true)
if err := sysreg.Register(ctx, addr, mappings); err != nil {
```

and `stopSystemPortmapRegistration` (`portmap.go:226-229`) does `Load()` → `Store(false)` → `Unregister`. So a Stop that interleaves while `Register` is still issuing its per-tuple UNSET+SET pairs lets those SETs land *after* the Unregister, and the flag is already false, so nothing cleans them up.

That defeats the stated intent of the comment directly above the store: *"we mark the registration active and unregister on shutdown regardless of partial failures, since whatever landed must be cleaned up."* The early `Store(true)` is right for the sequential case and is what makes the concurrent one lose.

What I did **not** verify is Copilot's reachability claim — that `Group.Start` reserves the name before `startSystemPortmapRegistration` returns, so `StopOne` can run concurrently. That is the load-bearing half and it needs reading the Group lifecycle, which I did not do. I would not treat this as confirmed-reachable on my say-so. It is suggestive that this PR exists because register-with-system flips race.

Suggested split: the test-helper fix is small, self-contained and belongs here. The sysreg Start/Stop serialization is a different defect in a different file and may be better as its own issue than as scope added to a flake fix.

Issue #2517, opened 2026-09-10
SMB create + post-break: 5 live audit defects (TOCTOU overwrite, parked-CREATE resume hang, ADS swallowed error)


marmos91:
# SMB handlers: create + create-post-break — live defects (T3/T4 cluster)

Re-derivation of the 2026-09-07 SMB audit's `handlers-core-handler`, `handlers-lease-oplock`,
`handlers-create-post-break`, `handlers-create` areas against `ffdb75c98` (2026-09-10).
Evidence tables: `.planning/2026-09-10-wave4-triage.md`.

## create-post-break (handlers/create_post_break.go, compound.go)

- **TOCTOU create-race recovery truncates the winner without breaking its lease/oplock** —
  `create_post_break.go:797`: race branch replays recheckExistingFileGates (share-mode+DACL only)
  then `h.overwriteFile(authCtx, winner, req)` — no breakAndMaybeParkCreate on the winner.
- **CREATE parked mid-compound-chain: trailing commands run against stale FileID, resume
  goroutine blocks forever** — `create_post_break.go:1639`: parkCreateOnLeaseBreak never checks
  ctx.NextCommand; compound.go:1040 defers MarkStarted only for isLastCommand → resume goroutine
  blocks forever on `<-pending.started`.
- **Unsynchronized OpenFile.MetadataHandle read in delete-on-close election** —
  `handlers/doc_election.go:142`: `bytes.Equal(other.MetadataHandle, openFile.MetadataHandle)`
  unlocked vs mu-guarded write set_reparse_point.go:263.

## create (handlers/create.go, durable_context.go)

- **ADS base-file existence check swallows lookup error** — `create.go:1290`:
  `if f, _, _ := h.lookupCaseInsensitive(...); f == nil {` discards the error return while
  sibling :1230 checks `lookupErr` — may create the file after a real backend error.
- **AppInstanceId failover force-closes any user's handle on any share** —
  `durable_context.go:907,970` + `create.go:1449-1451`: filter `f.AppInstanceId == appId` only
  across sessions; no share/path/maximal-access scoping (MS-SMB2 3.3.5.9.13). One defect shared
  with the T1 durable finding (tracked in #2511's priority HIGHs).

## core-handler / lease-oplock

- (Remaining core-handler and lease-oplock LIVE rows are tracked in the auth/session and
  set-info/read-write tranche issues; the rest were structure/bloat → Wave 5.)

Closes the T3/T4 cluster rows of #2511.

Issue #2636, opened 2026-09-15
SMB authorization reads a session identity snapshot the revalidation sweep never refreshes


marmos91:
The auth-cache invalidation sweep re-resolves *tree permissions* against a freshly read user record, but deliberately does not write that record back to `Session.User` — publishing it would race the unlocked dispatch reads (see #2632). Everything else that authorizes therefore keeps the SESSION_SETUP snapshot, and two paths undo what the sweep just did.

**1. Per-operation identity (UID/GID/groups).** `BuildAuthContext` fills the identity from `sess.User` and the memoized `authIdentity` (`session_lifecycle.go:919-924`, `auth_helper.go:342`). A UID or group-membership change therefore does not reach in-flight file operations even though the sweep ran — the tree permission is current while the identity used for ownership and group ACL checks is not.

**2. Reconnect restores a revoked grant.** `resolveSharePermission` takes the user from the session snapshot (`tree_connect.go`), so after the sweep removes a tree for a withdrawn grant, the client can immediately TREE_CONNECT again and be re-authorized from the stale `SharePermissions` and group list. The revocation survives only until the client reconnects, which is a normal thing for a client to do.

Note this is narrower than it first looks: a *disabled or deleted* user is revoked outright and the dispatch gate refuses everything, so the reconnect path matters for grant and group-membership changes that leave the account enabled. That is still the majority of what an operator does.

Both want the same thing — a refreshed authorization identity that can be published safely — which is why they are filed together and with #2632 rather than as separate fixups:

- a locked snapshot accessor so a fresh record can be published without racing readers (#2632), then
- the sweep publishing the record it already read, and
- TREE_CONNECT resolving against the current persisted user, preserving the synthesized-directory-user exemption (a principal with no local account must not be refused when its lookup reports missing) and the lookup-error policy (a failed lookup must not be written back as a decision — see the fail-open fix in #2589).

Found by a Copilot re-review on #2589 after that PR pushed fixes answering the first review. Filed rather than folded in: the PR closes staleness on the tree-permission axis and these are the identity axis, which needs the locking work first.

Issue #2633, opened 2026-09-15
Auth-cache invalidation runs the whole SMB sweep on the control-plane goroutine


marmos91:
Service.InvalidateAuthCache calls its subscribers synchronously:

    for _, cb := range callbacks {
        cb()
    }

(pkg/controlplane/runtime/shares/service.go:543)

The SMB adapter subscribes with a callback that runs the full RevalidateAuthorization sweep (pkg/adapter/smb/adapter.go:313), which walks every session and every tree and issues a GetUser per session. So a control-plane mutation — disabling a user, changing a share grant — blocks in the API handler until every SMB connection has been re-resolved, and a store that is slow or timing out makes that worse exactly when the sweep matters most.

RevalidateAuthorization documents itself as running "off the request path". That is true of the SMB request path and false of the control-plane one, which is the path an operator actually waits on.

Its own ponytail marker concedes the shape of this without noticing the blocking: "walks every session and every tree on each invalidation rather than indexing them by username ... the walk runs only on a control-plane mutation, not per operation".

A fix wants the sweep dispatched through a serialized worker with shutdown handling and coalescing — several mutations in quick succession should collapse into one sweep rather than queue N of them. Coalescing matters as much as the offload: a bulk grant edit fires one invalidation per row today.

Predates #2589; that PR adds the SMB subscriber that makes the cost visible. Found by a Copilot re-review on that PR and confirmed by reading the call path.

Issue #2632, opened 2026-09-15
SMB session identity fields are read unlocked on the authorization path


marmos91:
see thread

PR #2664, opened 2026-09-15
fix(smb): close the five rows left open by tranches #2516/#2517/#2518

### LINKED ISSUE #2593: SMB: rows left unfixed when tranches #2516 / #2517 / #2518 were closed
# SMB: rows left unfixed when tranches #2516 / #2517 / #2518 were closed

Each of these was listed in a tranche that has since been closed as completed, but no merged PR
touched it. Re-verified at current develop (`fc3251e25`). All are small; batchable as one PR.

### Unlocked read of a mutex-guarded field (from #2517)

`internal/adapter/smb/handlers/doc_election.go:142` — inside `h.files.Range`, the delete-on-close
election compares `bytes.Equal(other.MetadataHandle, openFile.MetadataHandle)` with no lock on
`other`. The write a few lines down (**:138-146**) correctly takes `other.mu.Lock()`, and
`set_reparse_point.go:263` guards `MetadataHandle` writes the same way — so the field is
mutex-guarded everywhere except this read. PR #2528 closed #2517 without touching this file.

### Non-sentinel lease errors swallowed at Debug (from #2516/#2517)

`internal/adapter/smb/handlers/lease_context.go:454-458` — `ProcessLeaseCreateContext`'s
`} else if err != nil {` arm logs at Debug and coerces `grantedState = lock.LeaseStateNone`,
`epoch = 0` for **any** error. The arm above it (:448-453) handles the one expected sentinel with a
documented rationale; a genuine store/lock-manager failure is indistinguishable from it here and the
CREATE succeeds with no lease.

### Teardown path skips the handle-op drain (from #2516)

`internal/adapter/smb/handlers/session_lifecycle.go:493` states it outright: "closeFilesWithFilter
does no DrainHandleOps wait". The CLOSE path does (`close.go:576`), and `DeleteOpenFile`
(`open_file.go:525`) clears `handleOps` without waiting, while `WaitAndDeleteOpenFile`
(`open_file.go:488`) is the draining variant. Session/tree teardown can therefore remove an open
with operations still in flight on it. Confirm whether the comment records a deliberate
deadlock-avoidance choice (it argues lock ordering) — if so this is a `ponytail:` marker, not a fix;
if not, teardown should drain.

### Stale package doc contradicting live code (from #2518)

`internal/adapter/smb/auth/ntlm.go:9-10` — "For production use with credential validation, additional
implementation of NTLMv2 response verification is required." False: `ValidateNTLMv2Response` is
defined at **ntlm.go:853** and called from production at **handlers/session_setup.go:1184**.

### Issue-number citations in comments (from #2516)

The repo convention is that comments describe behaviour and carry no issue/PR numbers. PR #2525
swept `conn_types.go`, `framing.go`, `hooks.go` and `handlers/lease_context.go`, but the two largest
offenders named in #2516 were not in its diff and still carry them:

- `internal/adapter/smb/handlers/security.go` — 18 occurrences
- `internal/adapter/smb/handlers/session_setup.go` — 7 occurrences

(`pkg/adapter/smb/adapter.go:556` carries one too — fold it in if #2589 has landed, otherwise leave
that file alone, see the conflict note below.)

## Note on in-flight work

**PR #2589** (`fix/2586-smb-auth-invalidation`, open) touches `handlers/stub_handlers.go`,
`handlers/lock.go`, `handlers/tree_connect.go`, `internal/adapter/smb/response.go`,
`session/session.go` and `pkg/adapter/smb/adapter.go`. None of the files above overlap except
`pkg/adapter/smb/adapter.go`, which is why its comment sweep is optional here.

Re #2511. Rows left open when #2516, #2517 and #2518 were closed.

marmos91:
Five rows that were listed in tranches #2516 / #2517 / #2518 but never touched by
a merged PR. Each was re-verified against current `develop` before being worked;
two premises turned out to be wrong, and one row is deliberately **not** a code
change.

Closes #2593

## 1. Unlocked read of a mutex-guarded field

The issue pointed at one read — `doc_election.go`, inside `h.files.Range`,
comparing `other.MetadataHandle` with no lock while the write six lines down
correctly takes `other.mu.Lock()`.

The root cause is one level up. `OpenFile`'s struct comment listed
`MetadataHandle` among the **immutable** fields that are "safe to access without
the mutex". That claim is false: `set_reparse_point.go:263` rewrites it on a live
handle, under `mu`, when it replaces a regular-file placeholder with a symlink.
Because the doc said the field was immutable, every handle-table scan read it
bare — including scans that take `existing.mu.RLock()` two lines later to read
`DeletePending` off the same object. `metadata.FileHandle` is a slice header, so
an unsynchronized read can observe it mid-swap.

So the fix is the doc plus one accessor, not one call site. `GetMetadataHandle()`
mirrors the existing `GetPayloadID()`, and **every** read of the field on an open
pulled out of the handle table now goes through it — the scans in `open_file.go`
and `doc_election.go` the issue named, plus `close.go`'s release-lease scan,
`create_post_break.go`'s other-open scan, `disconnected_state_machine.go`'s
conflict scan, `durable_context.go`'s AppInstanceId displacement scan,
`open_files.go`'s block-GC enumerator, `set_info_file.go`'s frozen-timestamp
restore, `session_lifecycle.go`'s session lock release, and `openHasLocks`. A
mechanical sweep of every `files.Range` body confirms none are left. Reads of the
handle a request already owns are deliberately left direct, and the struct doc
now says so rather than stating a rule with silent exceptions.

No lock-ordering hazard: no caller of any of these scans holds an `OpenFile`'s
`mu` when it enters (the pre-existing `RLock` on `DeletePending` inside the same
`Range` bodies would already have deadlocked if one did), and in `doc_election.go`
the `RLock` and the later `Lock` on `other` are sequential, not nested.

Verified with `go test -race ./internal/adapter/smb/... ./pkg/metadata/lock/...`.

## 2. Non-sentinel lease errors swallowed at Debug

`ProcessLeaseCreateContext`'s `} else if err != nil {` arm logged at Debug and
coerced `grantedState = LeaseStateNone, epoch = 0` for **any** error, so a
lock-manager or store failure was indistinguishable from an expected denial.

Enumerating what can actually reach that arm — the two sentinels above it are
already handled — leaves exactly two classes:

- `lease denied: conflicts with <T> delegation on file` from
  `requestLeaseImpl`: an expected, client-triggerable denial. It had no sentinel,
  which is *why* the arm could not tell the cases apart.
- `no lock manager for share %q` from `requestLeaseInternal`: a genuine
  configuration/store failure.

Added `lock.ErrLeaseDelegationConflict` and wrapped the denial with it. The
delegation case keeps its Debug line with the reason spelled out; everything else
now logs at Error. Behaviour on the wire is unchanged in both cases — the CREATE
still succeeds with no lease, so a client never regresses because of a
misconfigured share; it just stops being silent about it. The only test that
inspects the message text asserts `Contains(err, "delegation")`, which still
holds.

Naming the case also exposed a pre-existing divergence worth recording rather
than silently carrying: a nil-error denial (byte-range-lock conflict) reaches the
epoch-echo block and answers the client's requested epoch, while any error-carrying
denial answers epoch=0. That is unchanged by this PR, and is now a `ponytail:`
marker naming what evidence would justify moving the delegation case onto the
nil-error path.

## 3. Teardown path skips the handle-op drain — answered, not "fixed"

`session_lifecycle.go:493` says outright that `closeFilesWithFilter` does no
`DrainHandleOps` wait, while CLOSE does. **The comment records a real
deadlock-avoidance choice, so this ships as a `ponytail:` marker and no code
change.**

The evidence is `set_info_file.go:685-694`. `SET_INFO` carries a FileID, so the
connection read loop registers a handle op for it (`connection.go:405-407`)
before dispatch. A directory rename then parks in
`LeaseManager.WaitForOtherKeyBreaks` holding that registration, and — as the
comment at `set_info_file.go:679-681` states — that wait "ends on an ACK or CLOSE
from a lease holder that may be this very connection". On a logoff or a dropped
transport no ack or CLOSE is ever coming; what actually ends the wait is the lease
release at `session_lifecycle.go:510-512`, inside this same function, below where
a drain would go. Draining first would make teardown wait on the rename while the
rename waits on teardown — a cycle broken only by the break timeout, per handle.

The marker names the ceiling (an op still in flight keeps the `*OpenFile` it
already holds and completes against a detached handle; its late write or cache
flush can land after the AppInstanceId failover has handed the file to a new open)
and the upgrade path (cancel this session's parked break-waits first, *then*
drain, so the wait cannot be circular).

## 4. Stale package doc contradicting live code

`auth/ntlm.go` claimed "additional implementation of NTLMv2 response verification
is required". Both halves of the refutation still hold at current `develop`:
`ValidateNTLMv2Response` is defined at `ntlm.go:853` and called from production at
`handlers/session_setup.go:1184`. Replaced the sentence with what the file
actually provides.

## 5. Issue-number citations in comments

Repo convention is that comments describe behaviour and carry no issue/PR
numbers. Swept the two files #2525 missed — `handlers/security.go` (18) and
`handlers/session_setup.go` (8, not the 7 the issue counted) — plus
`pkg/adapter/smb/adapter.go`.

**The issue's count for `adapter.go` was wrong**: it reported one occurrence at
`:556`; there are 11. Each is its own commit so any of them can be dropped or
rebased cheaply.

Both `ponytail:` and `decision:` markers were checked and preserved
(`session_setup.go:1251`, `session_setup.go:329`, `adapter.go:574`) — the sweep
never touched a marker line.

## Verification

```
go build ./... && go vet ./... && gofmt -l internal/ pkg/   # clean
go test ./...                                               # 7921 passed, 179 packages
go test -race ./internal/adapter/smb/... ./pkg/metadata/lock/...   # green
```

## Merge sequencing

Two files here are contended, so the commits are split to make either droppable:

- `handlers/session_setup.go` and `handlers/handler.go` / `open_file.go` overlap
  open PR #2599 (`fix/2591-reapply-security-fixes`). #2599 adds
  `GetRequestedAllocSize` / `GetCreateOptions` at the same anchor in `handler.go`
  (immediately before `GetPayloadID`) and edits the same `OpenFile` doc lines to
  drop `CreateOptions` from the immutable list. The two changes are semantically
  complementary — both are "this field is not actually immutable" — but they will
  conflict textually. Whichever lands second keeps both additions.
  `durable_context.go` is also touched by both PRs: #2599 rewrites the exact
  `ProcessAppInstanceId` line whose `MetadataHandle` read is guarded here, so
  that one line will conflict. The resolution is to keep #2599's `claimsFile(f)`
  predicate and this PR's `f.GetMetadataHandle()`.
- `pkg/adapter/smb/adapter.go` is owned by #2592 / #2633; its sweep is its own
  isolated commit.
- `handlers/session_lifecycle.go` is owned by the #2632 / #2636 work around line
  919. This PR touches it in two places, both far from that: the `ponytail:`
  marker at the third-pass comment and a one-line guard in
  `ReleaseAllLocksForSession`.
- The session_setup.go sweep is the tip commit, so it can be dropped with a
  single `git rebase --onto HEAD~1 HEAD`.


marmos91:
Triage of the two Copilot comments, at `dittofs-65`'s request. I am not the author of this PR and have changed nothing — assessment only, merge decision is Marco's.

**Both findings are real, and the scope is larger than either comment states.**

### `close.go:701` — confirmed, and the fix is already in scope

Line 716 does read the field directly:

```go
if err := h.LeaseManager.ReleaseLeaseForHandle(ctx, lock.FileHandle(openFile.MetadataHandle), leaseKey, openFile.ShareName); err != nil {
```

while the scan fifteen lines above correctly uses `other.GetMetadataHandle()` against a captured `ownHandle`. So this PR synchronized the comparison and left the release reading the same field unsynchronized. `ownHandle` is already in scope, so this is a one-token change.

### `session_lifecycle.go:315` — confirmed at all four named sites

`385`, `399`, `619`, `620` all read `openFile.MetadataHandle` directly, and `352` reads `persisted.MetadataHandle`, which Copilot did not name.

### The part neither comment says, and which decides the size of the fix

`close.go` has **eleven** direct reads of `.MetadataHandle` — 223, 250, 260, 289, 290, 341, 364, 385, 641, 642 and 716. Copilot flagged one of them. Fixing only the flagged sites leaves ten untouched, which is the failure mode where a fix lands where the finding pointed rather than where the rule lives.

But the rule is genuinely ambiguous here, and that ambiguity is the thing to settle before changing anything. The accessor's own doc comment (`handler.go:480`) states a **narrower** rule than Copilot's argument:

> SET_REPARSE_POINT repoints a live handle when it replaces a regular-file placeholder with a symlink, so the field is a mutable slice header: a scan over the handle table **that compares handles across opens** must read through here rather than touching the field directly.

That says cross-open scans. Copilot's argument is different: teardown paths are unsafe because they explicitly skip `DrainHandleOps`, so a concurrent `SET_REPARSE_POINT` can write the header while teardown reads it — which would make the rule "any read on a non-draining path", not "any cross-open scan".

Both cannot be the rule. Whichever it is decides whether this is a four-line change or an audit of eleven sites plus whatever else in the package reads the field off a non-draining path:

- If the rule is cross-open scans, most of the eleven are fine — they operate on the caller's own `openFile` after its ops are drained — and Copilot's teardown argument is the thing to refute in the thread.
- If the rule is non-draining paths, the doc comment is understating it and should be rewritten as part of the fix, because the next person will read it and conclude their direct access is fine.

I would not guess at that from outside the change. The author knows which paths drain; I do not, and I did not trace `DrainHandleOps` callers to find out.

One note on the `decision:`/`ponytail:` conventions in `CLAUDE.md`: whichever rule wins, it belongs at `GetMetadataHandle` as the single statement of when direct field access is permitted. Right now the invariant is recorded in a comment that describes one use case, and every direct read elsewhere is individually defensible against it.

marmos91:
Correction to my triage above: the count is **14, not eleven**, and every one of them is a read — there are no writes in that set.

Full list in `close.go`: `223, 250, 260, 289, 290, 341, 364, 385, 641, 642, 716, 753, 803, 866`. I missed `753` (`GetFile`), `803` (`ResolveForRead`) and `866` (`purgeBlockStorePayload`).

The error was mine and worth naming, because it is the same shape as the defect under discussion: I ran `grep -n '\.MetadataHandle' close.go | head`, and `head` stops at ten lines. I counted the ten it printed, added `716` from a separate read, and reported eleven as if it were a total. A truncated result reported as a complete one.

Nothing else in the triage changes — if anything the ambiguity gets sharper. Fixing the sites Copilot named leaves **thirteen** untouched, so the question of which rule governs (`GetMetadataHandle`'s "a scan that compares handles across opens" versus Copilot's "any read on a path that skips `DrainHandleOps`") decides between a four-line change and a fourteen-site audit. That is still the author's call and still not something to settle by picking the convenient reading.

Credit to `dittofs-65` for catching the count.

marmos91:
Rebased onto current `develop` (`dc523d5af`) — the PR is `MERGEABLE` again. Settling the rule the triage above asked for, plus the two suppressed Copilot findings.

## The rule, and why it is the broad one

The two copies of the rule had already drifted, which is what made this ambiguous:

- `GetMetadataHandle` said *"a scan over the handle table that compares handles across opens"*
- the `OpenFile` struct doc said *"any read on an open pulled out of the handle table"*

Neither was right, because both imply own-open reads are provably safe. They are not:

- **One writer on a live open**: `set_reparse_point.go:263`, `openFile.MetadataHandle = newHandle`, under `mu`. Nothing else assigns the field on a live open.
- **Same-FileID requests run concurrently.** `connection.go:406` pre-registers the handle op in wire order, then dispatches `go func(...)` per request. `handleOps` is a `WaitGroup` counter — that is drain semantics for CLOSE, not mutual exclusion.
- **`RequestOrder` does not save it either.** Its own doc: *"Handlers keep running concurrently; only the moment a response is written is ordered."*

So a client may legally pipeline `FSCTL_SET_REPARSE_POINT` against the same FileID a READ/WRITE/SET_INFO is reading, and nothing serializes the two. Copilot's suppressed comment on `open_file.go:41` is correct on the mechanism.

## What that does and does not buy

The honest rule is "every read of `MetadataHandle` on a live `*OpenFile` goes through the accessor" — and the codebase already shows the intended enforcement is *structural*, not documentary: `name` is unexported (`name atomic.Pointer[OpenName]`) behind `Name`/`SetName`, so direct access is impossible. `MetadataHandle` is exported, so the rule is advisory and every site is individually defensible against it.

Enforcing it fully means **141 direct reads across 25 files** in the handler package. That is not this PR, and it would collide with every open SMB branch.

So: cross-open, handle-table-scan and teardown reads are converted (those are where a concurrent writer is routine and unavoidable). Own-open reads stay direct, and the exemption is now written at `GetMetadataHandle` as a `decision:` marker naming the ceiling and the upgrade path — unexport the field behind the accessor and a setter, the way `name` already is. The struct doc no longer restates the rule, so the two cannot drift apart again.

I want to be explicit that this is a **scope judgement, not a safety proof**, and the marker says so rather than implying the reads were shown safe.

## Copilot's two inline findings

Both were already fixed by the rework commit and are verified on the current head: `close.go` releases through the accessor, and every teardown site Copilot named (`315/385/399/619/620`) now goes through `GetMetadataHandle`. The one remaining direct read in `session_lifecycle.go` is `persisted.MetadataHandle` — a persisted handle row, not a live `*OpenFile`, so there is no `mu` to take.

## The lease sentinel

Added `assert.ErrorIs(t, err, ErrLeaseDelegationConflict)`. Mutation-checked rather than assumed: dropping the `%w` wrap while keeping the message text leaves the old `assert.Contains(err.Error(), "delegation")` **green**, and the new assertion fails. That is exactly the silent re-routing into the Error-level branch the sentinel exists to prevent.

## Note for #2599

#2599 is still open and `develop` has neither `GetRequestedAllocSize` nor `claimsFile`, so there was nothing to preserve here — this branch only adds `GetMetadataHandle` at that anchor. #2599 lands second and should keep **both** accessors, and in `ProcessAppInstanceId` keep `claimsFile(f)` **plus** `f.GetMetadataHandle()`.

Verification: `go build`, `go vet`, `gofmt` clean; `go test ./...` 8296 tests / 177 packages green; `-race` green on `pkg/metadata/lock/...` and `internal/adapter/smb/...`.

PR #2711, opened 2026-09-16
fix(metadata): write back the attribute columns SetFileAttributes was asked for, not the ones that differ from its pre-read

### LINKED ISSUE #2708: metadata: SetFileAttributes silently drops an attribute change equal to the caller's pre-read value
`SetFileAttributes` silently drops an explicit attribute change when the requested value happens to equal the value the caller last read. The operation returns success and a concurrent writer's value survives. This is the lost-update class #2320 exists to eliminate, reintroduced one layer above the isolation fix.

Introduced by #2707 (`cd256ba8c`). It is in shared `pkg/metadata` code, so it affects **all four metadata backends**, not just postgres.

### The defect

`pkg/metadata/file_modify.go:783-803` decides *which fields this call changed* by diffing the working copy against the pre-transaction read, rather than by looking at what the caller actually requested:

```go
if file.Mode != pre.Mode {
    row.Mode = file.Mode
}
if file.Hidden != pre.Hidden {
    row.Hidden = file.Hidden
}
if !file.Atime.Equal(pre.Atime) {
    row.Atime = file.Atime
}
if !file.Mtime.Equal(pre.Mtime) {
    row.Mtime = file.Mtime
}
if !file.CreationTime.Equal(pre.CreationTime) {
    row.CreationTime = file.CreationTime
}
```

`attrs.Mode != nil` sets `modified = true` unconditionally further up, so the transaction opens and the row is written. But when the requested mode equals the pre-read mode, `file.Mode != pre.Mode` is false, `row.Mode` is never assigned, and `row` keeps whatever the committed state held — which may be a concurrent writer's value.

`Size`, `ACL` and `EAs` are already keyed off the request (`attrs.Size != nil`, `len(attrs.EAMutations) > 0`). `Mode`, `Hidden`, `Atime`, `Mtime` and `CreationTime` are not. That inconsistency is the bug.

### Interleaving

File is `0644`.

1. Client B reads the inode: mode `0644`.
2. Client A commits `chmod 0600`.
3. Client B issues `chmod 0644` — a legitimate explicit request, and from B's point of view a no-op against what it read.
4. `modified = true`; the transaction opens; `row.Mode` is `0600`; no branch assigns it; `0600` is written back.
5. **B's SETATTR returns success. The mode is `0600`.**

B asked for `0644`, was told it succeeded, and the file is `0600`. Nothing logs and nothing errors.

### Verified

Demonstrated under `go test -overlay`, with no edits to the worktree:

- On the branch that became `cd256ba8c`: `post-op attrs report mode 0600; stored mode 0600` → **FAIL**
- With the pre-#2707 `file_modify.go` overlaid in: `stored mode 0644` → **PASS**

The reproduction ran against **sqlite**, which is what establishes that this is not postgres-specific and not a property of the REPEATABLE READ change.

### Reach beyond chmod

The same shape applies to every field in that block. The most likely one to be hit in practice is `Mtime`: a backup or sync tool that re-applies a timestamp it believes is already set (`rsync -t` on a second pass, `cp -p`, a restore that stamps mtimes it just read) issues exactly the "requested value equals the value I read" call this drops. `Atime`, `CreationTime` and `Hidden` are equally affected.

### Fix

Key these fields off the request, as `Size`/`ACL`/`EAs` already are:

```go
if attrs.Mode != nil || attrs.ModeOrMask != nil || attrs.ModeAndNotMask != nil {
    row.Mode = file.Mode
}
```
and the equivalent for `Hidden`, `Atime`, `Mtime`, `CreationTime` — noting that the mode masks (`ModeOrMask`/`ModeAndNotMask`) and the ACL adjustment also write `file.Mode`, so the condition has to cover every producer of that field, not just `attrs.Mode`.

A regression test needs the concurrent writer in the window — asserting only that `chmod 0644` on a `0644` file leaves it `0644` passes on the broken build.

### Note on how this was found

CI could not have caught it. #2707 passed `go build`/`vet`/`gofmt`, 143 packages of `go test`, `-race` on the touched packages, `golangci-lint` at 0 issues, the full `integration`-tagged postgres suite, and a 6-of-6 red-pre-fix A/B on its own new tests. It was found by an adversarial review reading the retry closure for state captured outside it. An automated reviewer looked at the same lines and reported four other findings, none of which was this one.

marmos91:
Closes #2708

`SetFileAttributes` decided which columns its transaction writes by diffing the working copy against the copy it read *before* opening one. A request whose value already equalled that pre-read value therefore wrote nothing: `attrs.Mode != nil` still set `modified`, the transaction still opened, but `file.Mode != pre.Mode` was false, the column was skipped, and the row kept whatever a concurrent writer had committed in the gap. The call returned success.

```
file is 0644
  B reads the inode          -> 0644
  A commits chmod 0600
  B issues chmod 0644        -> modified=true, transaction opens, row.Mode is 0600,
                                no branch assigns it, 0600 is written back
  B's SETATTR returns success. The mode is 0600.
```

`Mode`, `Hidden`, `Atime`, `Mtime`, `CreationTime` and `Ctime` were all keyed this way; `Size`, `ACL` and `EAs` were already keyed off the request. `Mtime` is the one most likely to be hit in the field — a sync or restore pass re-applying a stamp it has just read (`rsync -t` second pass, `cp -p`) is exactly a request equal to the pre-read value.

This is shared `pkg/metadata` code, so it affected all four metadata backends. Introduced by #2707 (`cd256ba8c`).

## The fix

Every field is keyed off the request, with the condition naming **every** producer, not just the one the issue named:

| field | producers |
| --- | --- |
| `Mode` | `attrs.Mode`, `attrs.ModeOrMask`, `attrs.ModeAndNotMask`, the setid strip on chown, the setid strip on truncate |
| `Hidden` | `attrs.Hidden` |
| `Atime` | `attrs.Atime`, `attrs.AtimeNow`, the coalesced directory-times overlay |
| `Mtime` | `attrs.Mtime`, `attrs.MtimeNow`, the truncate mtime stamp, the coalesced directory-times overlay |
| `CreationTime` | `attrs.CreationTime` |
| `Ctime` | `attrs.Ctime`, the truncate stamp, the automatic change-time bump, the coalesced directory-times overlay |

Two producers are not `attrs.X != nil` conditions and would have been missed by the fix the issue sketches:

- **The directory-times overlay.** `mergeDirTimes` lifts a coalesced (not-yet-flushed) create/remove bump onto the working copy *after* `pre` is taken, and `ClearIfFlushed` drops the pending entry at the end of the call. A bump the write does not carry is lost for good, so `atimeSet`/`mtimeSet` are seeded from that overlay rather than starting false.
- **The POSIX setid strips**, which are read-modify-writes. Keying them off `attrs.Size != nil` would have written a whole mode derived from the pre-read copy on *every* truncate, clobbering a peer's chmod — the same lost update pointing the other way.

The three read-modify-writes in the closure (the DOS-attribute masks, the setid strip, the EA mutations) are now re-applied to the row the transaction read instead of copied from the pre-read copy, so they compose with what the peer committed rather than replacing it. Ownership stays keyed off a diff: the code above only ever moves `UID`/`GID` when the request differs from what it read, so for that field the two tests are the same one.

## Also in this PR, same closure, both from #2707

**A spurious `EPERM`.** The in-transaction ownership recheck fired for any call that was not root and not a handle-authorized timestamp write. That swept in a non-owner authorized by `checkWritePermission` — truncate, utimes-to-now, and the SMB EA write authorized by `eaAuthorizedByHandle`, which the gate ~60 lines above already exempts. A peer chown landing in the window turned those into `EPERM` / `STATUS_ACCESS_DENIED` for an operation a genuine write check had cleared.

The obvious fix here is wrong, and I shipped it first before the adversarial pass caught it. Exempting every operation that is *write-permission-gated* re-opens the hole #2707 closed: POSIX write permission is decided by comparing the caller against the file's own UID and GID, so a chown **does** invalidate it — and the authorization switch lets ownership answer first, so an **owner** reaches a truncate or a utimes-to-now without the write check ever running at all. Under that version, the owner of a `0600` file could keep truncating it after a peer had chowned it away. The recheck now fires for any caller who is the owner; only a non-owner, who necessarily went through `checkWritePermission`, is exempt. `TestSetFileAttributes_OwnerAuthorizedOpRefusedAfterConcurrentChown` covers all three shapes (chmod, utimes-to-now, truncate) so the exemption cannot widen back unnoticed.

The residual — a chown that moves a non-owner between the group and other bit triples — is recorded as a `decision:` marker at the site with what would close it (recomputing the permission from the row alone; today that needs a share-options read on the store's own connection from inside an open transaction, and the guard is already blind to the peer *chmod* that changes the same answer).

**A `decision:` marker whose ceiling was already breached when it was written.** It said to withdraw the EA exemption "if EA writes ever arrive on this path concurrently rather than through xattr.go". They already do: `internal/adapter/smb/handlers/set_info_file.go:1598` builds `&metadata.SetAttrs{EAMutations: ...}` and calls `SetFileAttributes` directly, taking no `LockFileRow`. **The gap is closed rather than the marker corrected** — the marker's own stated trigger had fired, and its stated cost was wrong: it assumed the row lock would have to be unconditional ("the lock on every chmod and utimes as well"), when taking it under `len(attrs.EAMutations) > 0` covers exactly the EA path. The mutations now fold onto the row under that lock instead of replacing the map wholesale, so two concurrent EA writers keep each other's keys. Re-applying the same mutations on a transaction retry lands the same map.

## Evidence

Twelve new reproductions, each with the peer writer committed inside the window between the caller's pre-read and the transaction it opens (`windowStore.beforeTx`). Every one was run against **pristine `develop` sources** via `go test -overlay`, with no edit to the worktree, and every one goes red there:

| test | pre-fix failure |
| --- | --- |
| `ModeEqualToPreReadIsStillWritten` | `chmod 0644 reported success and left mode 0600` |
| `HiddenEqualToPreReadIsStillWritten` | `clearing Hidden reported success and left it set` |
| `TimestampEqualToPreRead.../Atime` | `Atime = …21:15:25; want …21:13:25` |
| `TimestampEqualToPreRead.../Mtime` | `Mtime = …21:15:25; want …21:13:25` |
| `TimestampEqualToPreRead.../CreationTime` | `CreationTime = …21:15:25; want …21:13:25` |
| `TimestampEqualToPreRead.../Ctime` | `Ctime = …21:34:54; want …21:32:54` |
| `ModeOrMaskAlreadySetIsStillWritten` | `SET_SPARSE reported success and the bit is clear` |
| `ModeEqualToPreReadKeepsACLAndModeInStep` | `the ACL was rewritten to grant the owner write for the requested 0644 while the mode column kept the peer's 0400` |
| `SetidStripAppliesToTheCommittedRow` | `mode = 0755; want 0700` |
| `WritePermittedOpSurvivesConcurrentChown/MtimeNow` | `a peer chown turned a write-permission-gated MtimeNow into EPERM` |
| `WritePermittedOpSurvivesConcurrentChown/Truncate` | `…turned a write-permission-gated Truncate into EPERM` |
| `ConcurrentEAWriterKeepsItsKey` | `the EA committed inside this call's window is gone` |

`OwnerAuthorizedOpRefusedAfterConcurrentChown` is green on both builds by design — it pins behaviour #2707 introduced and this PR keeps. It earns its place against the *first* version of this branch: two of its three cases go red there, which is how the over-exemption above was caught.

A second A/B, against that first version rather than against `develop`, pins the three things the adversarial review found:

| test | failure against the first attempt |
| --- | --- |
| `OwnerAuthorizedOpRefused.../MtimeNow` | `MtimeNow was authorized by ownership a peer chown has since moved, and mode 0600 grants uid 1000 nothing on the new owner's file` |
| `OwnerAuthorizedOpRefused.../Truncate` | same, for truncate |
| `TimestampEqualToPreRead.../Ctime` | `Ctime = …21:34:54; want …21:32:54` |

The reproductions run on **sqlite**, which is what establishes the defect is backend-independent rather than a property of the REPEATABLE READ change. Each test also asserts its window hook actually fired, so none can pass vacuously.

## Gates

`go build ./...`, `go vet ./...`, `gofmt -l`, `golangci-lint run ./...` (0 issues), `go test ./...`, `go test -race ./pkg/metadata/...`, and the `integration`-tagged suite against a real postgres — `pkg/metadata/store/postgres` and `pkg/metadata/store/badger` both green. `pkg/controlplane/runtime` and `pkg/controlplane/store` fail in this environment for want of localstack; the identical failure set reproduces on pre-fix sources under the same overlay, so it is not from this change.

## Review notes

Three passes ran on the diff. The correctness pass found nothing. The adversarial pass found the `writePermSufficient` over-exemption above (a security regression against `develop` HEAD, reproduced with a runnable test), the `Ctime` column left diff-keyed, and a comment on the new row lock that overstated its ceiling by claiming no other field in the closure was a read-modify-write — the ACL rewrite two lines below is one. All three are fixed in the second commit.


marmos91:
Addressed the review in e9011fcc9. Two of the four findings hold; two do not, and I want the reasoning on the record rather than a silent dismissal.

**Fixed — UID/GID still diff-keyed.** Correct, and it is the same bug class as the issue itself. `attrs.UID`/`attrs.GID` now decide the write-back, and `modified` is set from the request rather than the move — so a chown naming the owner the caller just read no longer returns success having opened no transaction. `ownershipChanged` deliberately stays keyed off the *move*, because it drives the POSIX setid strip, which answers to an ownership change and not to a chown request.

**Fixed — the ownership recheck invalidated on GID.** Correct. `isOwner` is a UID comparison, so a peer chgrp was refusing an owner who was still the owner. The recheck now invalidates on the fields the decision actually read: the UID always, and the GID only where an SGID grant consulted the file's group (`groupConsulted`). Both directions are pinned — a chmod survives a peer chgrp, an SGID chmod does not.

**Not changed — the SMB FileBasicInformation mode clobber.** The premise is right but the conclusion is not: this is not a regression the PR introduces. `SMBModeFromAttrs` (converters.go) starts from a hardcoded 0644/0755 and never reads the file's POSIX bits, so that path already destroys them with *no* concurrent writer — a 0600 file is stored back as 0644 on unmodified `develop`. The old diff-keyed branch preserved a peer's mode only when the synthesized value happened to equal the caller's own pre-read, which is the very coincidence being fixed. The real fix belongs in the SMB handler, expressing a DOS-attribute update through the `ModeOrMask`/`ModeAndNotMask` fields that exist for exactly this; I did not widen this PR into the SMB handler to do it.

**Not changed — the handle-authorized size write.** Already exempt where it applies: a non-owner whose SET_INFO size write is authorized by the open handle is exempt and covered by a test. For an *owner* the switch lets ownership answer first and the handle grant is never consulted, so exempting that call would exempt a decision only ownership made — which is the regression an earlier revision of this branch shipped.

Tests: `TestSetFileAttributes_ChownEqualToPreReadIsStillWritten`, `_OwnerChmodSurvivesConcurrentChgrp`, `_SetgidChmodRefusedAfterConcurrentChgrp`, plus the existing `_WritePermittedOpSurvivesConcurrentChown` / `_OwnerAuthorizedOpRefusedAfterConcurrentChown` pair. `go test ./pkg/metadata/ -run TestSetFileAttributes_` green.

marmos91:
@github-copilot please re-review — the two unaddressed findings are answered with evidence in the comment above, and I'd like a second opinion on whether the SMB-mode argument holds.

PR #2713, opened 2026-09-16
ci: read the latest develop verdict, and stop cancelling develop runs (#2698)

### LINKED ISSUE #2698: ci-health reports develop green whenever the latest run was cancelled — which is 55% of the time
The `warn if develop is red` job in `ci-health.yml` is structurally blind most of the time, and this is why the five-day SMB auth regression (#2678) sat on develop with nobody alerted.

## Reproduction, live right now

The job asks, per monitored workflow:

```bash
gh run list --repo "$REPO" --workflow "$wf" --branch develop \
  --status completed -L 1 --json conclusion --jq '.[0].conclusion // "none"'
```

and treats the result as not-red when it is `success`, `none`, or `cancelled`. Running exactly that against each key in `test/conformance/suites.json`:

```
conformance.yml -> cancelled
e2e-tests.yml   -> cancelled
nfs-pynfs.yml   -> cancelled
```

So the alarm currently reports develop as healthy. Meanwhile `e2e-tests.yml` on develop over its last 98 runs is **54 cancelled, 43 failure, 0 success** — it has not passed once since at least 2026-09-09.

## Why it misses

`--status completed` includes cancelled runs, and `-L 1` takes only the most recent one. The comment in the job is right that a cancelled run is not evidence of breakage — but it draws the wrong conclusion from that, discarding the *question* rather than looking further back.

`e2e-tests.yml` has `cancel-in-progress: true` on a group keyed by `github.ref`, so every push to develop kills the run in flight. Median duration of a completed run is **1,012s (~17 min)**. When merges land more often than that — as they did all morning — most runs never finish, and **55% of develop runs are cancelled**. The latest *completed* run is therefore a coin flip that lands on `cancelled` more often than not, and every time it does, the alarm goes quiet regardless of how many real failures sit behind it.

The failure mode is the nastiest kind: not a false negative under load, but a false *green*. The alarm is least reliable exactly when develop is busiest, which is when it matters most.

## Fix

Ask for the latest run that actually reached a verdict, rather than the latest that stopped:

```bash
gh run list --repo "$REPO" --workflow "$wf" --branch develop \
  --status completed -L 20 --json conclusion \
  --jq 'map(select(.conclusion != "cancelled")) | .[0].conclusion // "none"'
```

`-L 20` covers the observed cancellation rate with margin; `none` still means "never ran with a verdict" and stays benign.

Worth considering alongside it: `cancel-in-progress: true` is right for pull requests, where a new commit supersedes the old one, but on develop every commit is a distinct thing someone wants tested, and cancelling means it never is. Splitting the concurrency policy by event would give develop a real signal — at the cost of more runner contention, which is a genuine trade rather than a free win (see #2641 for the job-seconds measurements).

## Why it matters beyond this one suite

`ci-health.yml` is the repo's only automated statement about whether develop is healthy. While it was silent, #2678 shipped and stayed for five days, and E2E stayed red for longer than that. The suite was reporting correctly the whole time; nothing was listening.

marmos91:
Closes #2698

## What was actually wrong

`warn if develop is red` asked GitHub for the newest **completed** run on develop and treated `success`, `none` and `cancelled` as not-red:

```bash
gh run list --workflow "$wf" --branch develop --status completed -L 1 --json conclusion --jq '.[0].conclusion // "none"'
```

`--status completed` **includes** cancelled runs, and every monitored workflow had `cancel-in-progress: true` on develop. Each new develop push killed the run in flight, so cancelled was the *typical* develop conclusion — not a rare edge. The alarm therefore reported green on the strength of a run that never reached a verdict, most often when develop was busiest.

Measured on this checkout, over the last 20 completed develop runs of `e2e-tests.yml`:

```
15 cancelled
 5 failure
```

and the single newest completed run was one of the cancelled ones. That is the false-green the issue describes, reproduced live.

## The fix

**1. Read the newest run that reached a verdict.** Look back over recent runs and take the newest non-cancelled conclusion:

```bash
gh run list --workflow "$wf" --branch develop --status completed -L 20 --json conclusion \
  --jq 'map(select(.conclusion != "cancelled")) | .[0].conclusion // "none"'
```

`-L 20` covers the observed cancellation rate with margin. `none` still means "no run has reached a verdict" and stays benign. The `cancelled` special-case in the red test is gone: a cancelled conclusion can no longer reach it.

**2. Stop cancelling develop runs.** `cancel-in-progress` is now event-keyed:

```yaml
cancel-in-progress: ${{ github.event_name == 'pull_request' }}
```

A PR run is superseded by its own next commit, so cancelling it is right. A develop run is not: each push is a distinct commit someone wants tested, and cancelling it means it never is. This is the point the issue's own "why it misses" section makes, and it is what stops the alarm from going quiet under load in the first place.

Applied to all three monitored workflows — `conformance.yml`, `nfs-pynfs.yml`, `e2e-tests.yml` (the set `ci-health.yml` reads from `test/conformance/suites.json`). `e2e-tests.yml` has no `pull_request` trigger, so its expression is always false and its develop runs are simply never cancelled.

## Cost

This is a genuine trade, not a free win: develop runs now **queue** instead of superseding, so a busy develop consumes runner minutes it previously discarded. GitHub keeps one running plus one pending run per concurrency group, so a burst of pushes still leaves the pending one superseded — the running one, at least, completes and reports. See #2641 for the job-seconds measurements behind that trade.

## Verification

The jq expression was tested against real `gh run list` output and against canned all-cancelled / mixed / empty cases:

| latest 20 conclusions | result |
| --- | --- |
| `cancelled, cancelled, failure, success` | `failure` |
| `cancelled, cancelled` | `none` (benign) |
| `[]` | `none` (benign) |

The edited shell block was extracted from `ci-health.yml` and executed against a stubbed `gh`: `conformance.yml -> success`, `e2e-tests.yml -> failure`, `nfs-pynfs.yml -> none`, yielding `RED=[ e2e-tests.yml]` — only the workflow with a real failure is flagged.

All four workflow files parse as valid YAML.

Workflow-only change; no Go code touched.


marmos91:
Addressed the review nit in 61f08f846: the comment on the `cancelled` filters now states the behaviour explicitly — `cancel-in-progress: false` protects a run that is already executing, but GitHub still replaces a *pending* run when a newer one is queued to the same group, so a burst of develop pushes can still cancel intermediate runs. The filters are therefore load-bearing, not vestigial.

PR #2716, opened 2026-09-16
fix(nfs): keep the GSS context LRU within a principal, and touch it only after the MIC

### LINKED ISSUE #2699: NFS RPCSEC_GSS: ContextStore LRU is steerable without a session key — cross-principal eviction, and LastUsed refreshed before the MIC check
Split out of #2405 while working PR #2686. Both findings are pre-existing and
were left untouched there; neither is in #2405's own finding list. They share a
root: the GSS `ContextStore`'s LRU bookkeeping is driven by input that has not
been authenticated yet, so a caller who holds no session key can still steer it.

## 1. LRU eviction retires other principals' contexts

`ContextStore.Store` claims a slot and, at capacity, calls `evictOldest`
(`internal/adapter/nfs/rpc/gss/context.go:244`, `:355`). `evictOldest` ranges the
whole map and drops the globally oldest `LastUsed` — there is no per-principal,
per-realm or per-client accounting anywhere in the store.

The cap is process-wide (`NewContextStore(maxContexts, contextTTL)`,
`framework.go:403`). So any principal in the realm that can obtain a service
ticket can run RPCSEC_GSS_INIT in a loop and evict every other principal's
context, without ever holding their session keys. Victims see their contexts
vanish and must re-INIT; sustained, it is a denial of service against every
other user of the server.

Note this falsifies the general statement "only a key-holder can retire a
context". That clause holds for the *targeted* retire paths (DESTROY, the
ticket-expiry gate) — which is what #2686 changed — but not for eviction.

## 2. `Lookup` refreshes `LastUsed` before the MIC is checked

`ContextStore.Lookup` calls `ctx.Touch()` on every hit (`context.go:277`,
`:156`), and it is called at `framework.go:610` and `:904` — in both cases
*before* `verifyHeaderMIC` at `:635` and `:913` respectively.

A handle is not a secret: it travels in the clear in every RPCSEC_GSS
credential. So anyone who can observe one on the wire and replay it can refresh
`LastUsed` with a request that then fails the MIC check, holding a context off
idle eviction (`cleanup`, `context.go:334`) indefinitely. It also feeds
finding 1 — the attacker picks whose context looks oldest.

#2686 makes this strictly less useful than it was: the ticket-expiry bound it
added is checked against `ExpiresAt`, which `Touch` does not refresh, so a
context can no longer be kept alive past its Kerberos ticket this way. The idle
TTL is still refreshable.

## Suggested direction

For 1, evict within the offending principal (or client address) rather than
globally, or cap per-principal — the store already carries `Principal` and
`Realm` on every `GSSContext`. For 2, move the `Touch` to after the MIC
verifies; `Lookup` returning without touching, plus an explicit touch at the
call sites that have authenticated, keeps it O(1).

Neither is a regression and neither blocks #2686.

marmos91:
## Root cause

The RPCSEC_GSS `ContextStore` let unauthenticated-shaped input steer its own LRU on two axes, so a peer holding any single valid service ticket could retire other principals' contexts and could keep its own contexts alive with requests that never authenticated.

1. **Eviction was global.** `Store` freed a slot at capacity by calling `evictOldest`, which ranged the whole store and dropped the globally oldest `LastUsed` with no accounting of who owned it. A principal could loop `RPCSEC_GSS_INIT` to evict every other principal's contexts, forcing them to re-authenticate.
2. **`Lookup` refreshed the idle clock before authentication.** `Lookup` called `ctx.Touch()` on every hit, and both call sites (`handleData`, `handleDestroy`) did so before `verifyHeaderMIC`. The context handle travels in the clear in every RPCSEC_GSS credential, so a peer that merely observed one on the wire could replay it to hold the context off idle eviction with requests that then failed authentication.

## Fix

- `Store` now frees a slot by evicting the oldest context **of the incoming context's own principal** (`evictOldestForPrincipal`), falling back to the globally oldest only when that principal holds none — which is what a new principal's first context needs to make room. A principal can therefore only retire its own contexts, while the global cap still bounds total growth. Both paths route through one `evictOldestMatching` helper.
- `Lookup` no longer touches `LastUsed`. The call sites now `Touch` **after** the call-header MIC verifies, in both `handleData` and `handleDestroy`. The DESTROY site also covers the sequence-window rejection path, which returns before the context is deleted and leaves it alive.

## Tests

Red/green verified by reverting `context.go` and `framework.go` to `develop` and re-running: `TestContextLookupDoesNotUpdateLastUsed`, `TestContextEvictionStaysWithinPrincipal` and `TestProcessDATAMICFailureDoesNotRefreshIdleClock` all fail pre-fix and pass post-fix. `TestContextEvictionFallsBackForNewPrincipal` covers the new-principal path and passes either way, as expected.

```
go test -race -count=1 ./internal/adapter/nfs/rpc/gss/...   # ok
go build ./... && go vet ./internal/adapter/nfs/rpc/gss/...  # clean
```

## Notes

Neither finding is a regression — both are pre-existing, split out of #2405, and neither blocks #2686.

Closes #2699


marmos91:
@github-copilot please re-review — the findings from your last pass are addressed in the latest commits.

PR #2719, opened 2026-09-16
fix(nlm): guard the whole blocking waiter, not just its cancelled flag

### LINKED ISSUE #2691: NLM blocking-queue waiters are shared across a lock boundary that only half of their readers respect
`blocking.BlockingQueue.mu` guards the queue map and its slices, not the `*Waiter` behind the pointer. A queued waiter is handed to the grant path by pointer and read there with no synchronization, while the queue's own methods read the same struct under `bq.mu`.

PR #2690 removed the one *writer* on that seam (`processNLMWaiters` repointed `waiter.Lock` with no lock held; the race detector confirmed it, and nothing needed the swap). What remains is the reader side, which is latent today but makes the struct unsafe for anyone who adds a write back.

**Unsynchronized reads** — `internal/adapter/nfs/nlm/callback/granted.go:53,56,67,84` reads `waiter.CallbackHost`, `waiter.CallbackVers`, `waiter.Cookie`, `waiter.CallbackProg` and `waiter.Lock.*` while building and sending the NLM_GRANTED callback, a network round-trip that includes a portmap GETPORT. It runs from `processNLMWaiters`, which took its waiters from `bq.GetWaiters()` — that copies the *slice*, not the waiters — and then dropped `bq.mu`.

**Reads under `bq.mu`** — `internal/adapter/nfs/nlm/blocking/queue.go` (`Cancel`, and `Enqueue`'s retransmit lookup) read `w.Lock.Owner.OwnerID`, `w.Lock.Offset`, `w.Lock.Length` via `Waiter.matches`.

So the two sides disagree about what protects the waiter. `Waiter` already proves the hazard is understood: `cancelled` has its own `mu` precisely so `IsCancelled()` can be read from `ProcessGrantedCallback` while `Cancel()` runs (`internal/adapter/nfs/nlm/blocking/waiter.go`). The rest of the mutable surface never got the same treatment, and the `decision:` marker PR #2690 left at `Enqueue` records the constraint rather than removing it.

**Why it matters now.** The rule "do not write to a queued waiter" is currently enforced only by a comment. The obvious next change — refreshing a retransmitted request's cookie or callback target, which is what a first cut of #2690 did — reintroduces a real race with a GRANTED callback already in flight.

**Shape of a fix.** Extend `Waiter.mu` to cover the whole mutable surface behind accessors that `ProcessGrantedCallback` uses, so the struct is safe to mutate and the comment can go. Alternatively, have the grant path take an immutable snapshot of what it needs under `bq.mu` before dropping it.

Worth a test that drives a grant and a queue operation on the same waiter concurrently under `-race`; the absence of one is why this survived.

Found while working #2406.

marmos91:
## Root cause

A queued `blocking.Waiter` is handed to the grant path **by pointer** (`GetWaiters` returns a copy of the *slice*, not of the waiters). `ProcessGrantedCallback` then read `CallbackHost`, `CallbackVers`, `CallbackProg`, `Cookie`, `CallerName`, `Svid`, `OH`, `FileHandle` and the `Lock` fields with **no lock held**, while the blocking queue could still be mutating the same struct: `Cancel` and `RemoveClientWaiters` mark it cancelled, `Enqueue` stamps `QueuedAt`, and `matches`/`isRetransmissionOf` read its identity fields. Only the `cancelled` flag had a mutex, so the struct was shared across a lock boundary that only half its readers respected. #2690 removed the one *writer* (`processNLMWaiters` repointing `waiter.Lock`), which left the hazard latent on the reader side rather than fixing it.

## Fix

Widen `Waiter.mu` to cover **every** field, and add `Snapshot()` to read the whole struct at once. Reading it in one shot rather than field-by-field also fixes a correctness bug, not just a race: a waiter cancelled *between* the `IsCancelled()` check and the callback build would previously be checked as live and then dialed. Route the queue's own readers (`matches`, `isRetransmissionOf`, `RemoveClientWaiters`) and both grant-path call sites (`ProcessGrantedCallback`, `processNLMWaiters`) through the snapshot.

Lock ordering is one-way — `bq.mu → Waiter.mu`, never the reverse.

The `decision:` marker in `Enqueue` lost one of its two stated reasons (the write it described no longer races), so it now rests on the caller_name-binding argument alone, which is the one that actually holds. The `waiter.Lock` swap stays removed.

## Test

`internal/adapter/nfs/nlm/blocking/waiter_race_test.go` drives every mutating queue path (`setQueuedAt`, `Cancel`, `bq.Cancel`, `GetWaiters`, `TotalWaiters`, retransmit `Enqueue`, `RemoveClientWaiters`) against concurrent `Snapshot` reads on one waiter. Its absence is why this survived.

A/B verified non-vacuous: with the mutex stripped from `Snapshot`, `-race` reports **3 data races**; restored, clean.

```
go test -race ./internal/adapter/nfs/nlm/...   # ok (blocking, callback, handlers, xdr)
go test -race ./internal/adapter/nfs/... ./pkg/adapter/nfs/...   # ok
go build ./... && go vet ./internal/adapter/nfs/nlm/...   # clean
```

## Contradicting the issue text

Nothing material. The issue's `granted.go:53,56,67,84` line numbers point at the field reads; the live file had shifted slightly but the described reads are exactly right. The issue suggested the fix might extend `Waiter.mu` **or** take an immutable snapshot under `bq.mu`; this does the former, and additionally routes the grant path through a snapshot, which is what makes the cancelled-check-then-dial window go away.

Closes #2691.


marmos91:
@github-copilot please re-review — the findings from your last pass are addressed in the latest commits.

PR #2729, opened 2026-09-17
fix(smb): route the explicit CLOSE path through OpenFile.Handle

### LINKED ISSUE #2665: SMB: the explicit CLOSE path still reads OpenFile.MetadataHandle without its lock
`OpenFile.MetadataHandle` is mutable on a live handle — `SET_REPARSE_POINT` republishes it under `openFile.mu` when a placeholder becomes a symlink — and `OpenFile.Handle()` (added in #2622) is the accessor that reads it correctly.

#2622 routed the **shared** teardown path through it: `closeFilesWithFilter`, `ReleaseAllLocksForSession`, `releaseHandleLeaseRecord`, and the AppInstanceId failover, each taking one snapshot that spans the lock release and the lease break so the two cannot name different files.

`close.go`'s explicit-CLOSE path was left on direct field reads — `close.go:223, 250, 260, 289, 290, 341, 364, 385, 641, 642` and neighbours. The same race applies: CLOSE runs on one connection while a SET_REPARSE_POINT on a bound channel can republish the field mid-teardown, so the block-store resolve, the pending-write flush, the atime write, the unlock and the parked-create signal can end up split across two different files.

Left out of #2622 deliberately: it is a mechanical sweep across a file that PR does not otherwise touch, and it wants its own test rather than riding a change about AppInstanceId authorization.

Found by Copilot on #2622.

marmos91:
Closes #2665

## Root cause

`SET_REPARSE_POINT` republishes a live open's `MetadataHandle` **under `openFile.mu`** when it replaces a regular-file placeholder with a symlink (`set_reparse_point.go:262-264`). The dispatcher pre-registers handle ops in wire order and then runs one goroutine per request, so an `FSCTL_SET_REPARSE_POINT` on one channel can land while a `CLOSE` for the same FileID runs on another.

`#2622` routed the shared teardown path (`closeFilesWithFilter`, `ReleaseAllLocksForSession`, `releaseHandleLeaseRecord`, AppInstanceId failover) through `OpenFile.Handle()`/`GetMetadataHandle()`, taking one snapshot per teardown. `close.go`'s explicit-CLOSE path was left on direct `openFile.MetadataHandle` field reads, so each step of the teardown re-read the field independently. A mid-teardown republish could split the block-store resolve, the pending-write flush, the atime write, the byte-range unlock, the lease release and the parked-create signal across **two different files** — releasing locks and a lease on the file the handle moved to, while leaving the original open's records standing.

## Fix

One `metaHandle := openFile.GetMetadataHandle()` snapshot at CLOSE entry, threaded through every step that acts on the file:

- block-store `ResolveForWrite` / `ResolveForRead`
- `FlushPendingWriteForFile`, the atime read + `SetFileAttributes`
- the `POSTQUERY_ATTRIB` attribute fetch
- `UnlockAllForOpen` (byte-range release)
- `releaseHandleLeaseRecordOn` (lease/oplock release) — the same snapshot the unlock used, so the two cannot name different files
- `SignalParkedCreates`, `purgeBlockStorePayload`, the MFsymlink conversion helpers

Two secondary sites that the snapshot alone does not close:

- `releaseHandleLeaseRecordOn`'s **sibling scan** compared `other.GetMetadataHandle()` against a fresh read of the closing open's own handle. The scan decides *whether to release at all*; a decision taken against one file while the release names another is the same split, so it now compares against the caller's snapshot. `leaseKeyHasPersistedSibling` (the disconnected-sibling store lookup that gates the same release) takes the snapshot too.
- `doc_election.go:311` (`deleteTargetHandle := openFile.MetadataHandle`), reached from both the explicit-CLOSE delete path and the teardown delete path, now reads through the accessor.

`releaseHandleLeaseRecord` (the no-snapshot wrapper) is **removed** rather than kept: after the change every production caller has a snapshot to pass, so leaving it would be a dead second entry point whose doc still claimed to be the shared CLOSE/teardown release. The rationale moved onto `releaseHandleLeaseRecordOn`.

## Evidence

`TestClose_ReleasesLocksAndLeaseOnTheSnapshotHandle` drives `Close()` end to end against a metadata store that republishes `MetadataHandle` (the exact in-place mutation `SET_REPARSE_POINT` performs, under `openFile.mu`) during the teardown window, then asserts the open's byte-range lock **and** its lease record are gone from the *original* handle while the republish target's lock survives.

Verified failing before the fix and passing after, with the same test binary:

```
--- FAIL: TestClose_ReleasesLocksAndLeaseOnTheSnapshotHandle
    CLOSE left the open's byte-range lock on the file it was operating on: the unlock was
    routed through the republished handle (...) instead of the snapshot taken at CLOSE entry (...)
```

- `go build ./...` — pass
- `go vet ./internal/adapter/smb/handlers/` — pass
- `go test ./internal/adapter/smb/...` — pass
- `go test -race ./internal/adapter/smb/handlers/ -run 'TestClose_ReleasesLocksAndLeaseOnTheSnapshotHandle|TestClose_DirLeaseRelease|TestReleaseHandleLeaseRecord|TestClose_Flush' -count=2` — pass


marmos91:
Addressed the review finding: `restoreFrozenTimestamps` was the last CLOSE-path write still reading `openFile.MetadataHandle` fresh at call time.

Split out `restoreFrozenTimestampsOn(authCtx, openFile, metaHandle)` and pointed CLOSE at it with the entry snapshot, so both the attribute restore and the pending-mtime update land on the file the rest of the teardown used. The other callers (WRITE, truncate, COPYCHUNK, SET_INFO) keep the self-reading wrapper — each acts on the open its own request resolved, not on a teardown snapshot.

The regression test now arms a coalesced `SmbPendingAtime` (so the republish fires *before* the restore) plus a frozen Mtime, and asserts the frozen value reached the snapshot file. Reverting just the CLOSE call site to a fresh field read makes it fail, so the check is not vacuous.

PR #2737, opened 2026-09-17
fix(smb): update DOS attributes with mode masks, not a synthesized POSIX mode

### LINKED ISSUE #2717: smb: FileBasicInformation SET_INFO overwrites a file's POSIX mode with a synthesized 0644/0755
An SMB `SET_INFO / FileBasicInformation` that carries any non-zero `FileAttributes` replaces the file's POSIX permission bits with a synthesized constant. A `0600` file becomes `0644` because a client set `FILE_ATTRIBUTE_HIDDEN`. No concurrency is involved — this is the single-caller path.

### Mechanism

`internal/adapter/smb/handlers/set_info_file.go:71-85` builds the mode from the DOS attributes alone:

```go
if fileAttrs != 0 {
    mode := SMBModeFromAttrs(fileAttrs, openFile.IsDirectory)
    if curFile, curErr := metaSvc.GetFile(...); curErr == nil {
        mode |= curFile.Mode & (modeDOSCompressed | modeDOSSparse)
    }
    setAttrs.Mode = &mode
```

`SMBModeFromAttrs` (`internal/adapter/smb/handlers/converters.go:523-547`) starts from a hardcoded base and never reads the file:

```go
if isDirectory {
    mode = 0755
} else {
    mode = 0644
}
mode |= modeDOSExplicit
// ... then only the DOS high-word bits
```

The existing `GetFile` right below it already re-reads the file, but carries over only `modeDOSCompressed | modeDOSSparse`. The POSIX triple is discarded.

`metadata.SetFileAttributes` treats a non-nil `attrs.Mode` as an absolute mode request, which is correct — it is the caller that is sending the wrong thing.

### Reproduction

Against unmodified `develop`, at the metadata layer, with **no second writer**:

```
file created 0600
SetFileAttributes{Mode: 0o644 | modeDOSExplicit}   // exactly what the handler sends
stored mode -> 0200644                              // 0o644 | modeDOSExplicit
```

Group and other gained read. An NFS client sees the widened mode immediately.

### Why it has been invisible

`calculatePermissions` enforces the READONLY semantics (clearing write when `modeDOSExplicit + modeDOSReadonly` are both set), so the DOS-attribute intent round-trips correctly and SMB-side behaviour looks right. Nothing re-derives the POSIX bits that were overwritten, and no SMB conformance case reads them — they are only observable from the NFS side or via a later `GETATTR`.

### Fix

`ModeOrMask` / `ModeAndNotMask` exist for exactly this. They are masked down to `dosAttributeModeBits`, which excludes `0o7777` by construction, so a DOS-attribute update expressed through them cannot touch permission, setid or sticky bits at all:

- `ModeOrMask` = the DOS bits the request sets
- `ModeAndNotMask` = the settable DOS bits the request does not set

`Hidden` is already carried separately as `setAttrs.Hidden` and needs no change.

### Note

Found while fixing #2708. That PR makes this loss deterministic rather than intermittent: the previous diff-against-pre-read branch happened to skip the mode column when the synthesized value equalled this caller's own pre-read, which preserved a concurrent peer's chmod by accident in that one case. The unconditional single-caller loss above predates it and is unaffected.

marmos91:
Closes #2717.

## What was wrong

An SMB `FileAttributes` field carries DOS attributes — HIDDEN, SYSTEM, ARCHIVE,
READONLY — and says nothing about who may read or write the file. DittoFS stores
those bits in the high word of the same `Mode` field that carries the POSIX
permission triple, and three SMB paths turned a `FileAttributes` update into an
**absolute** mode built by `SMBModeFromAttrs`, whose base is a hardcoded `0644`
for files and `0755` for directories.

`metadata.SetFileAttributes` treats a non-nil `attrs.Mode` as an absolute mode
request, which is correct. The callers were sending the wrong thing: applying
that synthesized mode to an existing file republished it at whatever permissions
SMB would have invented for a *new* one.

A `0600` file became `0644` because the client set `FILE_ATTRIBUTE_HIDDEN`.
Group and other gained read on a previously private file, and an NFS client sees
the widened mode on its next GETATTR. No concurrency required.

## Why nobody noticed

`calculatePermissions` and `fileAttrToSMBAttributesInternal` read only the high
word, so the DOS-attribute intent round-trips correctly and the SMB-side
behaviour looks right. Nothing re-derives the POSIX bits that were overwritten,
and no SMB conformance case reads them. The loss is observable only from NFS.

The two cases in the new table that pass *before* the fix are the tell: a
`0644` file and a `0755` directory already equal the synthesized base, so the
overwrite is invisible for exactly the modes a test author would reach for first.

## The fix

Express the update through `SetAttrs.ModeOrMask` / `SetAttrs.ModeAndNotMask`,
which the store applies inside its own read-modify-write and whitelists down to
`dosAttributeModeBits`. That mask excludes the `0o7777` construction entirely, so
a DOS-attribute update expressed this way **cannot reach the permission, setid or
sticky bits at all** — it is not a narrower rewrite, it is no rewrite.

It also drops a read: the FSCTL-managed `modeDOSCompressed` / `modeDOSSparse`
bits used to be preserved by re-reading the file and OR-ing them back in. The
masks never name them, so they survive without the `GetFile`.

One helper, `applyDOSAttrUpdate`, derives both masks from `SMBModeFromAttrs`
rather than restating the bit rules, so the two cannot drift.

## Three sites, not one

The issue named one. Grepping the callers of `SMBModeFromAttrs` found the same
absolute-mode construction applied to an **existing** file in two more places:

| Site | Path |
| --- | --- |
| `set_info_file.go` | SET_INFO `FileBasicInformation` — the reported case |
| `set_info_file.go` | the ADS-stream → base-file attribute propagation |
| `create_path.go` `overwriteFile` | CREATE with `FILE_OVERWRITE` / `FILE_SUPERSEDE` |

`overwriteFile` was the same defect, unreported: opening an existing `0600` file
with `FILE_OVERWRITE` widened it to `0644` the same way. It was also dropping
`modeDOSSparse`, since it only carried `modeDOSCompressed` forward by hand.

The two remaining `SMBModeFromAttrs` callers build the mode for a genuinely new
file, which is what the base is for; they are left alone.

## The one metadata change, and why it is a guard rather than a relaxation

`overwriteFile` sends `Size` and the attribute update in one `SetAttrs`. The
metadata layer relaxes the ownership gate for a size-only change, because POSIX
`truncate()` takes write access rather than ownership — and that test keys on
`attrs.Mode == nil`. Switching to masks would have satisfied it, letting a
non-owner with write permission reach an attribute change that is
ownership-gated on its own, simply by bundling it with a size.

So `onlySettingSize` now also requires both masks to be nil. A `SetAttrs` that
flips a DOS attribute is not a truncate. This **preserves** today's gate rather
than moving it: no existing caller sends `Size` together with a mask, and a plain
truncate is still authorized by write permission.

## Evidence

Every new test was run against the unmodified `develop` sources first.

**Handler level** — `TestSetInfo_FileBasicInfo_PreservesPOSIXMode`, red before:

```
--- FAIL: .../private_file,_HIDDEN
    POSIX mode = 0o644 after SET_INFO FileBasicInformation; want 0o600 unchanged
--- FAIL: .../world-writable_file,_ARCHIVE
    POSIX mode = 0o644 after SET_INFO FileBasicInformation; want 0o777 unchanged
--- FAIL: .../private_file,_READONLY
    POSIX mode = 0o644 after SET_INFO FileBasicInformation; want 0o600 unchanged
--- FAIL: .../private_directory,_HIDDEN
    POSIX mode = 0o755 after SET_INFO FileBasicInformation; want 0o700 unchanged
```

(`default-mode file, SYSTEM` and `default-mode directory, HIDDEN` passed before
the fix — the blind spot described above.)

**CREATE overwrite** — `TestCreateOverwrite_PreservesPOSIXMode`, red before:

```
--- FAIL: .../private_file
    POSIX mode = 0o644 after CREATE overwrite; want 0o600 unchanged
--- FAIL: .../world-writable_file
    POSIX mode = 0o644 after CREATE overwrite; want 0o777 unchanged
```

**FSCTL bits without the read-back** —
`TestSetInfo_FileBasicInfo_PreservesFSCTLBitsWithoutReReading`, red before on the
POSIX assertion, green after with sparse and compressed both still standing.

**Metadata level** — `TestSetFileAttributes_DOSAttrMasks_LeavePermissionBitsIntact`
pins the contract the SMB fix rests on, across `0600` / `0777` / `0644` / setuid
/ directory / sticky-directory, with permission bits deliberately smuggled into
*both* masks. It is **green before and after**: the metadata contract was never
the defect, which is why the fix is entirely on the SMB side.
`TestSetFileAttributes_DOSAttrMasks_WithTruncateStillNeedsOwnership` is the
red-then-green one at that layer, pinning the gate above.

All DOS-attribute round-trip assertions are kept: each handler case re-reads the
file through `fileAttrToSMBAttributesInternal` and asserts the requested
attributes came back, so the READONLY round-trip `calculatePermissions` depends
on is still covered.

The red above was re-confirmed against pristine `develop` sources with
`go test -overlay` after the directory cases were corrected (see below), rather
than being carried over from the first run.

## One test defect worth naming

The directory cases in both new tables were initially built with
`Service.CreateFile`, which hardcodes `FileTypeRegular` and ignores `attr.Type`
— so they were asserting about directories while running against regular files.
They now go through `Service.CreateDirectory` and assert the subject's type
before testing it. The `private_directory, HIDDEN` case still goes red pre-fix
(`0o755` over `0o700`), now on an actual directory.

## A second, larger hole the review found

Converting the callers to masks did not just risk widening the CREATE-overwrite
gate. `noOwnershipAttrs` is `Mode == nil && UID == nil && GID == nil`, so moving
the mode data into the masks flipped it **true** — which made *every*
write-permission relaxation keyed on that flag a candidate bypass, not only the
truncate one. `onlySettingTimesToNow` admitted a mask outright:

```go
SetAttrs{ModeOrMask: &m, MtimeNow: true}   // non-owner, write permission only
```

reached `writePermSufficient` and the DOS bits landed. The first version of this
branch patched `onlySettingSize` alone and carried a `decision:` comment
claiming the general rule — a marker that overstated what the code enforced.

The rule is now applied once, where all of them route through: the mask check
folds into `noOwnershipAttrs`, three clauses that duplicated it are deleted, and
`onlyClearingSuidSgid` gets its own because it requires a non-nil `Mode` and
cannot use the flag. **Net fewer lines than the per-site version, and no
relaxation is left out.**

This is red-then-green against this branch's own earlier commit:

```
--- FAIL: TestSetFileAttributes_DOSAttrMasks_NonOwnerRefusedWithoutTruncate
    Error:    An error is expected but got nil.
    Messages: bundling a timestamp bump must not admit a DOS attribute change
```

## One behaviour regression, accepted and documented

Clearing READONLY no longer restores POSIX owner-write, because nothing in the
attribute path writes permission bits any more. The flow that loses out: a file
made read-only out-of-band (`chmod 0444` over NFS), whose READONLY is reported by
the POSIX-derived fallback in `fileAttrToSMBAttributesInternal`. Clearing the
read-only box now reports the file writable while the POSIX bits still refuse the
write.

The old code "fixed" this only as a side effect of the very bug being removed —
the same rewrite that widened `0o600` to `0o644`. Restoring `0o200` from an
attribute write would re-introduce permission-widening from a path that has no
business touching permissions, and would cut against the storage rule that keeps
the mode-derived DACL stable across READONLY flips (`smb2.winattr`). So the
behaviour is deliberate rather than overlooked: it carries a `decision:` marker
at the fallback, a test that pins it in both directions, and a new entry in
`docs/guide/smb-acl-fidelity.md`. **Worth a second opinion before merge** — it is
the one place this change makes something previously possible impossible.

## Also noted, not fixed here

`overwriteFile` keeps COMPRESSED and SPARSE across an overwrite, although
MS-FSA 2.1.5.1.2 assigns `FileAttributes` outright. The previous code already
deviated for COMPRESSED; this makes SPARSE consistent with it, and the comment
now says so instead of citing the spec for a rule it does not state.

Separately and pre-existing: an ordinary NFS `chmod` round-trip wipes every DOS
attribute, because the absolute-`Mode` path has no reverse whitelist and NFS
GETATTR masks the high word off. Out of scope here; worth its own issue.

## Two small cleanups carried in the same diff

- The `FileBasicInformation` branch was setting `setAttrs.Hidden` from
  `fileAttrs` a second time, having already had it set by
  `DecodeBasicInfoToSetAttrs` from the same field. The handler's
  `len(buffer) < 40` early return guarantees the decoder's `len >= 36` and
  `r.Err() == nil` preconditions, so the two guards are equivalent and the
  second assignment is dead. Removed.
- The comment on `SetAttrs.ModeOrMask` / `ModeAndNotMask` claimed the fields
  exist "solely for FSCTL-managed DOS attribute bits", which this change makes
  untrue. Corrected.

`go build ./...`, `go vet ./...`, `gofmt -l`, `go test ./...` clean;
`go test -race` clean on `./internal/adapter/smb/handlers` and `./pkg/metadata/...`.


marmos91:
Second opinion on the narrowing you flagged — I'd merge it. Checked by mutation rather than by reading.

## The ownership clause

`noOwnershipAttrs` gaining the mask terms is load-bearing at both application sites, not just the one the tests reach first. Reverting the clause fails `TestSetFileAttributes_DOSAttrMasks_NonOwnerRefusedWithoutTruncate` and `..._WithTruncateStillNeedsOwnership`; restoring it passes. The bundled-truncate case is the right shape: a non-owner holding write permission cannot reach the attribute flip by pairing it with a size change, and the same truncate without the masks still succeeds — so the relaxation is narrowed, not removed.

## The whitelist

`& dosAttributeModeBits` is load-bearing at **both** `file.Mode` (line 650) and `row.Mode` (line 966). Mutating only the first leaves the smuggling test green, because the transaction path re-applies from the row; mutating both makes it fail on `private_file`, `directory` and `sticky_private_directory`. Worth knowing if anyone edits one site later — the test does not distinguish them.

## The READONLY fallback

Agreed with the `decision:` marker, and I think it is the correct direction:

- The alternative is the bug being fixed — an attribute toggle widening the mode.
- A client that clears READONLY on a POSIX-read-only file is told writable while writes still fail. That is a mismatch, but the file genuinely is not writable, and the failure it produces is `ACCESS_DENIED` rather than silent permission widening. The mismatch is honest; the widening was not.
- `smb2.winattr` pins DACL stability across READONLY flips, which the widening would break.

One consequence worth stating in the marker if you want it complete: a SET_INFO that sets *any* attribute (including ARCHIVE alone) sets `modeDOSExplicit` and so retires the POSIX fallback for that file. The marker says this, but "any attribute" is doing a lot of work — a client touching only ARCHIVE loses the READONLY reporting for a POSIX-read-only file. That follows from the storage rule rather than from this change, so it is not a reason to hold the PR.

## Verification

`go build ./...`, `go vet`, `gofmt -l` clean. `go test ./pkg/metadata/... ./internal/adapter/smb/handlers/` — 17 packages pass, 0 fail. `go test -race` clean on both touched packages. Rebased onto current develop: 0 conflicts.

No blocking findings.

marmos91:
Both points addressed.

**Documentation numbering** — fixed: the new entry is `7.`, not `8.` (057928c4).

**The READONLY fallback** — this is the one behaviour regression the PR deliberately accepts, and it is not a defect this branch can remove. Worth stating the reasoning in full, because the finding is correct about the mechanism and I agree with its description:

- The fallback is *retired* for a file as soon as any SET_INFO gives it an explicit DOS attribute state. That is the storage rule that keeps the mode-derived DACL stable across READONLY flips (`smbtorture smb2.winattr`), and it is why the fallback is one-way.
- Restoring the fallback on a HIDDEN-only update means an attribute write must read and re-derive POSIX permission state — which is precisely the widening this PR exists to stop. The old code "fixed" the mismatch only as a side effect of the very bug being removed: the same rewrite that turned `0o600` into `0o644`.
- The mismatch it leaves is honest and fails closed: the client is told writable, and the write returns `ACCESS_DENIED` rather than silently succeeding with widened permissions. The alternative fails open.

It carries a `decision:` marker at the fallback, a test that pins it in both directions (`TestSetInfo_FileBasicInfo_PosixReadonlyFallbackIsOneWay`), and a new entry in `docs/guide/smb-acl-fidelity.md`. @marmos91 reviewed and agreed with this direction in the comment above.

marmos91:
Both coverage gaps closed (c9669190):

- **`create_path.go:390`** — added `TestCreateOverwrite_PreservesFSCTLBits`, seeding `modeDOSCompressed | modeDOSSparse` before the overwrite and asserting both survive. The existing test only checked POSIX mode, HIDDEN and forced ARCHIVE, so you are right that a reversion to an absolute mode there would have passed.
- **`set_info_file.go:266`** — added `TestSetInfo_ADS_PreservesBaseSparseBit`. The existing ADS cases seed and assert only COMPRESSED; SPARSE had no coverage at all on that path.

Both were verified green against this branch. The SPARSE case is the more valuable of the two — COMPRESSED was previously carried by hand in `overwriteFile`, so it had some protection, whereas SPARSE was silently dropped and nothing would have noticed.

Issue #2660, opened 2026-09-15
SMB: AppInstanceId failover judges the client by the GUID that established the open, not the session's current connection


marmos91:
MS-SMB2 §3.3.5.9.13 condition 4 is:

> **Open.Session.Connection.ClientGuid** is not equal to the current **Connection.ClientGuid**.

`ProcessAppInstanceId` compares `OpenFile.ClientGUID` instead, which is the GUID that *established* the open — the value §3.3.5.9 sets once at open time (`Open.ClientGuid = Open.Connection.ClientGuid`) and §3.3.5.9.10 uses for replay matching. The spec keeps the two distinct on purpose.

They diverge for an open that reconnected from a different ClientGuid. This server permits that: `ProcessDurableReconnectContext` applies its ClientGuid gate to *lease-backed* handles only, because a lease is scoped per (ClientGuid, LeaseKey) and the smbtorture reopen ladders pin that. §3.3.5.9.7 gates the whole 3.x family, lease or not.

So there are two halves to this, and the second is what makes the first observable:

1. Condition 4 reads the establishing GUID rather than the session's current connection GUID.
2. Non-lease DHnC/DH2C reconnect does not check ClientGuid at all.

Closing (2) makes the two values identical again and (1) becomes moot. Closing only (1) needs `OpenFile` to carry, or be able to resolve, the ClientGuid of the connection its session is on now — which for a multi-channel session means deciding which channel answers.

Found by Copilot on #2622. The deviation is recorded at the code site on `sameOrUnknownClient` as a `decision:` marker; this issue is the fix.

Impact is bounded to an application instance whose open changed clients mid-life: such an open is judged by where it came from, so it can be spared a failover it should lose, or displaced on the wrong basis. Every open in the durable-reconnect tests, and every open a single-client instance holds, sees the same value either way — which is why the suites are green.

marmos91:
Fixed by #2743, which closes half (1) and deliberately leaves half (2) alone.

**Half (1) — the real bug.** `ProcessAppInstanceId` compared `OpenFile.ClientGUID` (the GUID that established the open) where §3.3.5.9.13 condition 4 names `Open.Session.Connection.ClientGuid` (the connection the session is on now). It now resolves the open's client through its session, at all three comparison sites. The origin GUID is still recorded on the open, and a persisted durable row — which has no live session — is still judged by it.

[...]

PR #2743, opened 2026-09-17
fix(smb): judge AppInstanceId failover by the session's current connection (#2660)

marmos91:
## The bug

MS-SMB2 §3.3.5.9.13's fourth AppInstanceId match condition names **`Open.Session.Connection.ClientGuid`** — the GUID of the connection the open's *session is on now*. `ProcessAppInstanceId` compared `OpenFile.ClientGUID`, which is the GUID that **established** the open.

The two diverge for an open that reconnected from a different ClientGuid. This server permits that on a non-lease DHnC/DH2C reconnect: the ClientGuid gate on reconnect is lease-scoped (`leaseReconnectClientGUIDMismatch` requires a non-zero `LeaseKey`), because a lease is scoped per `(ClientGuid, LeaseKey)` and the smbtorture reopen ladders pin that — Samba's own `selftest/knownfail` lists only the `reopen1a-lease` variants, while the non-lease `reopen1a` tests set `options.client_guid = GUID_random()` and assert `NT_STATUS_OK`.

So the failover misjudged such an open **in both directions**: it spared an open it should have lost, and it displaced one on the basis of a connection the open no longer had.

## The fix

Half (1) only. No reconnect gate is added or removed, so **no smbtorture row changes status** — the conformance risk of the earlier, reverted approach is not reintroduced.

- `Session` now records the connection it is served over, with the ClientGuid that connection negotiated (`RecordCurrentConnection`), called after `AddChannel` succeeds on both the NTLM and Kerberos bind paths. `CurrentClientGUID()` reports it. The origin `ClientGUID` field is unchanged and stays the value the session originated on — the field comment now says so, since that distinction is what the bug rested on.
- `openClientGUID` resolves the open's client through its session, and `ProcessAppInstanceId` uses it at all three comparison sites (candidate collection, the re-test under the claim, and the persisted-row pass).
- Fallbacks, each answering a different question:
  - **No session to ask** (session gone, or a persisted durable row, which by definition has no live session): the recorded GUID is the only identity the open has — and is the session the open was on.
  - **A session that never recorded a connection** (a directly constructed session, or state predating the recording): fall back to the recorded GUID, which is what the condition used before.
  - **A session that did record one, but that connection negotiated no ClientGuid**: return the zero GUID, i.e. *unknown*. Returning the recorded GUID there would make the condition evaluable in the wrong direction — a CREATE could displace an open on a client match nobody established. `sameOrUnknownClient` already treats zero as undecidable, so the open survives.

## Tests

`TestProcessAppInstanceId_JudgesByTheSessionsCurrentClientGUID` — three subtests, one per direction plus the fallback:

- a moved open is displaced by a third client (**fails without the fix**: the open survives);
- an open whose session moved to the claimant is spared (**fails without the fix**: the client's own handle is closed);
- a session with no recorded connection falls back to the recorded GUID.

`TestCurrentClientGUID_ReportsTheConnectionTheSessionIsOn` pins the accessor: unknown before any channel, follows a second bind, leaves the origin `ClientGUID` alone, and reports a recorded connection that negotiated no GUID as a zero GUID rather than papering over it with the previous channel's value.

Verified: `go test ./internal/adapter/smb/...`, `go test -race ./internal/adapter/smb/handlers/ ./internal/adapter/smb/session/`, `go vet ./...`. Counterfactual confirmed by reverting only `durable_context.go` and re-running.

Refs #2660


marmos91:
Conformance evidence for the "no reconnect gate is touched" claim — the whole risk this PR carried:

```
smbtorture / badger   pass
smbtorture / memory   pass
WPTS BVT / memory     pass
WPTS BVT / postgres-s3 pass
pynfs v4.0 + v4.1 (memory, postgres-s3) pass
Unit Tests (1.26.x)   pass
```

Both smbtorture variants green means no expected-failure row changed status — in particular `smb2.durable-open.reopen1a` and `smb2.durable-v2-open.reopen1a`, which is what made the superseded #2725 approach a regression. This PR changes only how `ProcessAppInstanceId` *identifies* the claiming client; it adds and removes no reconnect condition.

Merge order: #2729 (now merged as `084376817`) also edits `durable_context.go`. I merged this branch onto develop `83f0b08d4` in a scratch worktree: `durable_context.go` and `durable_context_test.go` auto-merge, `go build ./...` passes, and `TestAppInstanceId_*` / `TestProcessAppInstanceId*` / `TestCurrentClientGUID*` / `TestClose*` all pass together — no semantic conflict.

PR #2744, opened 2026-09-17
test(runtime): release the scheduler tick before the join deadline, not on it

marmos91:
## The flake

`TestDrainStartupWorkers_AWedgedWatcherDoesNotSkipTheSnapshotDrain` failed roughly **1 run in 25** under `-race`, on a drain that had behaved correctly:

```
snapshot_scheduler_shutdown_test.go:380: the startup drain returned with a scheduler
tick still in the control-plane store: the wedged settings-watcher join spent the
snapshot drain's budget too, so the store closes under a running tick
```

It is not flaky timing, it is a **structural tie**. With `startupDrainTimeout = T`:

| | |
| --- | --- |
| watcher join window | `[0, T]` (wedged, so it always spends all of it) |
| `snapCtx` handed to `shutdownSnapshots` | `[T, 2T]` |
| scheduler join budget inside `shutdownSnapshots` | `remaining/2` = `T/2` → deadline `1.5T` |
| **the test released the tick at** | `T + T/2` = `1.5T` |

The release lands exactly on the join's deadline. At that instant the join completing and its timeout firing are both ready, and a two-case `select` picks uniformly — so the test reports a correct drain as a skipped one half the time it is late, which the `-race` scheduler makes frequent. The property under test only requires a release strictly inside `(T, 1.5T)`; the test picked the one value that is a coin flip.

This was red on **develop itself** (`Unit Tests` at `21269462f`, run 35175137998) and on every PR whose diff does not touch this package — including #2728, which was read as a regression from that PR.

## The fix

- Release the tick a **third of the way into** the second window (`budget + budget/3`) instead of at its exact midpoint. Same property, no boundary race.
- Fix the **data race** in the same test: `startupDrainTimeout` was written by the test, read by the release goroutine, and read again by a deferred self-referential restore (`defer func(d time.Duration){...}(startupDrainTimeout)`). The previous value is now captured up front and the budget passed to the goroutine as a constant. The other test in the file that writes the global is corrected the same way.

## Verification

- **0 failures in 40 runs** with `-race`, against **1 in 25** before.
- **The test still has teeth.** Restoring the pre-`ef7143da2` shared-window form (one `stopCtx` covering both joins) makes it fail with the intended message — so the timing change did not defang it:
  ```
  --- FAIL: TestDrainStartupWorkers_AWedgedWatcherDoesNotSkipTheSnapshotDrain (0.50s)
      the startup drain returned with a scheduler tick still in the control-plane store: ...
  ```
- `go test -race ./pkg/controlplane/runtime` passes.

No production code changes — test-only.


marmos91:
Confirms the diagnosis on a live PR: #2743's `Unit Tests (1.26.x)` just failed at run 35199984836 with exactly this test —

```
WARNING: DATA RACE
--- FAIL: TestDrainStartupWorkers_AWedgedWatcherDoesNotSkipTheSnapshotDrain (0.75s)
    the startup drain returned with a scheduler tick still in the control-plane store: ...
```

#2743 touches only `internal/adapter/smb/**`, so it cannot be the cause; the flake is on develop. Both halves of this PR are visible in that one failure — the deadline tie (the assert) and the `startupDrainTimeout` write/read (the race).

PR #2756, opened 2026-09-18
fix(lock): recall an NFSv4 delegation on the SMB lock path instead of denying

### LINKED ISSUE #2754: SMB: Manager.Lock denies a byte-range lock against an NFSv4 delegation instead of recalling it (#2746 regression)
## Summary

`Manager.Lock` (`pkg/metadata/lock/manager.go:852`) consults the cross-protocol
conflict predicate `fileLockConflictsWithUnified` but, unlike `CheckForIO`, it
neither recalls the blocking NFSv4 delegation nor waits for the `DELEGRETURN`.
After #2746 made a delegation participate in that predicate, an SMB byte-range
lock against a file an NFSv4 client has delegated is therefore **denied
outright** instead of recalling the delegation first. The SMB client cannot
acquire the lock at all.

`Manager.TestLock` (the preview at `manager.go:1062`) shares the same shape.

## Regression boundary

`e2f09cc78` (the #2746 squash merge, "recall an NFSv4 delegation before foreign
I/O, and complete the reclaim") is a **single-parent** commit whose parent is
`b8abc5392`, so it contains exactly that PR and nothing else.

`TestNLMAxisInterop` (`test/e2e/nlm_axis_interop_test.go`) result on the boundary:

| develop SHA | E2E run | result |
| --- | --- | --- |
| `b8abc5392` (parent) | [35233569784](https://github.com/marmos91/dittofs/actions/runs/35233569784) | **PASS** |
| `e2f09cc78` (#2746) | [35239826335](https://github.com/marmos91/dittofs/actions/runs/35239826335) | **FAIL** |

The test passed on every one of the 18 earlier runs in which it appears
(`a3001728b`, `dddc64605`, `42e952d9e`, `067babf55`, `555476bb9`, `392ed7eb1`,
`f76168d25`, `1c5032e2f`, `dd9aee6d7`, `7bfb2c286`, `4e51ad587`, `21269462f`,
`bbbeaf47c`, `2f0c62123`, `4e603abe4`, `28b6f1940`, `b8abc5392`, and others), so
this is a regression rather than a flake.

## Observed failure

Only the `SMB_vs_NLM` direction fails; the other three still pass:

```
NLM_vs_SMB   conflict=BLOCKED  afterRelease=ACQUIRED
NLM_vs_NFSv4 conflict=BLOCKED  afterRelease=ACQUIRED
SMB_vs_NLM   conflict=HOLDER_FAIL(err: [Errno 13] Permission denied) afterRelease=NA
NFSv4_vs_NLM conflict=BLOCKED  afterRelease=ACQUIRED
NLM_AXIS_RESULT=FAIL
```

`HOLDER_FAIL` means the *holder's own* lock was refused before any conflict was
tested (`nlm_axis_interop.sh:163`), which is why `afterRelease=NA`. This is not
a conflict being detected; it is the SMB client being unable to take its lock.

## Mechanism

#2746 removed the `IsDelegation()` early-return from
`fileLockConflictsWithUnified`, so a delegation now reports as a conflicting
`UnifiedLock`. It then taught only the I/O path how to handle that:

- `CheckForIO` (`manager.go:1102-1119`) recalls the delegation
  (`lm.breakDelegations`), waits bounded (`lm.waitForDelegationRecall`), and
  re-judges.
- `Manager.Lock` (`manager.go:868`) calls the same predicate and returns
  `NewLockedError` immediately — no recall, no wait.

So a live delegation is now a hard denial on the lock path. `Manager.Lock` holds
`lm.mu` for its whole body via `defer lm.unlock()`, so the recall cannot be
dispatched from inside it for the same reason documented at `manager.go:1109`
(`breakDelegations` takes the write lock); the recall-and-wait has to happen
before the locked region, as it does in `CheckForIO`.

## Reproduction

Deterministic unit test against `NewManager()`, no VM needed:

```go
lm := NewManager()
recalled := false
lm.RegisterBreakCallbacks(&delegReturningCallbacks{onRecall: func() { recalled = true }})

deleg := &UnifiedLock{
    Owner:      LockOwner{OwnerID: "nfs4:c1", ClientID: "nfs4:c1"},
    Type:       LockTypeExclusive,
    Delegation: &Delegation{DelegationID: "d1", DelegType: DelegTypeWrite},
}
lm.AddUnifiedLock(xHandle, deleg)

err := lm.Lock(xHandle, smbExclusive("smb:open-1", 0, 0))
// err      = Locked: resource is locked by another session
// recalled = false
```

`recalled=false` with a non-nil error is the defect: the lock was refused
without ever asking the delegation holder to give it up.

Note the existing `pkg/metadata/lock/cross_protocol_mandatory_probe_test.go`
documents the *pre-#2746* behaviour (`fileLockConflictsWithUnified skips
delegations, so an SMB byte-range lock is granted against it`) and so does not
cover this direction.

## Impact

Cross-protocol byte-range locking is broken for any SMB client taking a lock on
a file an NFSv4 client holds a delegation on: the lock fails with
`STATUS_LOCK_NOT_GRANTED` / `EACCES` rather than waiting for the recall. Because
delegations are granted by default (`ShouldGrantDelegation`,
`internal/adapter/nfs/v4/state/delegation.go:521`), this is reachable on a
default server that serves a share over both protocols.

This is the release blocker: E2E is red on `develop` and the failure is new.

## Fix direction

Route both lock-path callers through the recall-and-wait that `CheckForIO`
already implements, rather than teaching each one separately — the predicate is
the shared seam and it currently has two callers with different handling.
Concretely: hoist the recall-and-wait so `Manager.Lock` performs it before
taking `lm.mu` and then re-judges, and decide whether `TestLock` (a preview that
must agree with acquire) does the same or reports the delegation as a conflict.

Verification should include the real-VM NLM axis, since
`TestNLMAxisInterop` is the test that caught this and it is not covered by the
cross-protocol locking suite #2746 was verified against.

marmos91:
## What was wrong

`Manager.Lock` consults `fileLockConflictsWithUnified`, which since #2746 makes an
NFSv4 delegation participate as the whole-file claim it stands in for. Only
`CheckForIO` was taught to handle that: it recalls the delegation and waits for the
`DELEGRETURN` before re-judging. `Manager.Lock` and `Manager.TestLock` called the
same predicate and returned a conflict immediately.

So an SMB byte-range lock against a file an NFSv4 client had delegated was **denied
outright** — `STATUS_LOCK_NOT_GRANTED` / `EACCES` — instead of recalling first. The
holder's own lock was refused before any conflict was even tested, which is why
`TestNLMAxisInterop`'s `SMB_vs_NLM` direction reported `HOLDER_FAIL` with
`afterRelease=NA`.

## The fix

Route all three lock-path callers through one recall-and-re-judge loop, run before
`lm.mu` is taken (the recall cannot run under the lock, because `breakDelegations`
takes the write lock):

- `Manager.Lock` — recall, judge-and-insert, and if the judgment was blocked by a
  delegation, recall again.
- `Manager.TestLock` — the same shape, so the preview agrees with acquire.
- `CheckForIO` — folded onto the shared helper, removing its inline copy.

Three things this gets right that the obvious version does not:

**The recall predicate is the conflict predicate.** The helper renders the
delegation as the whole-file lock it stands in for (offset 0, length 0,
`delegationToLockType`) and asks `fileLockConflictsWithUnified` — exactly how
`GrantDelegation` already weighs it in the other direction. Deriving the predicate
from the lock's exclusivity instead makes a **zero-byte** SMB lock, which by SMB2
semantics never conflicts, recall a write delegation and park on it for the whole
timeout.

**A terminal conflict is not made to wait on a delegation.** The cross-protocol
scan goes through one helper that lets a definite byte-range conflict win over a
delegation the caller would otherwise recall and wait on, and reports a delegation
blocker only when *every* blocker is a delegation. A read delegation coexists with
a shared NLM lock, so without this an exclusive request meeting the delegation
first paid the full recall timeout only to be denied by the NLM lock behind it.

**The wait covers only the delegations this request selected.** Multiple read
delegations may coexist, so an unrelated one already breaking for another reason
could hold the caller until the deadline after its own delegation had already been
returned, and the stale conflict was then reported. The wait now takes the same
predicate the recall was dispatched with.

**Every path re-judges, not just the acquire path.** The judgment and the insertion
are one critical section, but they cannot share a section with the recall (the lock
ordering forbids it), so a delegation granted by a concurrent OPEN after the recall
decision would otherwise deny the request without ever being recalled — the same
defect through a narrower window. `Lock`, `TestLock` and `CheckForIO` all judge
first, recall only when a delegation is what stands in the way, and loop. A test
drives the window deterministically by granting a second delegation from inside the
first recall's callback; it fails on a single-pass acquire.

**The wait budget belongs to the recall, not to the caller.** The budget runs from
when each recall was dispatched (`BreakStarted`), so a client that never answers
`CB_RECALL` ends in a denial after one `delegationIOWaitTimeout` in total. This
matters because the SMB handler retries a blocking lock every
`BlockingLockRetryInterval` and the fail-immediately attempt comes through the same
call: with a per-call budget, ten retries against one unanswered recall cost ten
timeouts and one SMB request could accumulate waits far past its own 5s/35s
deadline. A timed-out recall denies rather than retrying, so the loop still ends in
a conflict rather than a spin.

## Verification

**Unit.** `pkg/metadata/lock/lock_path_delegation_recall_test.go` reproduces the
defect against `NewManager()` and pins the behaviour: a returned delegation lets the
lock be granted (waking on the break signal, not the timeout); an unanswered recall
denies it; a read delegation does not block a shared lock; a zero-byte lock neither
recalls nor waits; a delegation arriving during acquire is recalled, not denied; a
delegation ahead of a terminal conflict fails fast instead of waiting; and the wait
is scoped to the selected delegations; and repeated attempts against one unanswered
recall share a single budget. The same interleaving test covers the I/O path's
separate re-judge loop. Reverting any of the predicate fix, a re-judge loop, the scan
order, the wait scoping, or the budget anchoring makes the corresponding test fail on
its assertion, not on a build error — measurably so: 2.00s against 0.20s, 3.00s
against immediate, and 4.01s against 0.40s for ten retries.

`go test ./...` and `go test -race ./pkg/metadata/lock/` are green.

**Real hardware (Scaleway, Ubuntu 24.04, dual-netns NLM harness).** Two-build
verification on one VM, same reproduction against both binaries, on the final
commit including the race fix:

| build | `NLM_vs_SMB` | `NLM_vs_NFSv4` | `SMB_vs_NLM` | `NFSv4_vs_NLM` | result |
| --- | --- | --- | --- | --- | --- |
| pre-fix (`develop` 58ae43b3) | BLOCKED/ACQUIRED | BLOCKED/ACQUIRED | **HOLDER_FAIL (EACCES) / NA** | BLOCKED/ACQUIRED | **FAIL** |
| post-fix (`0e4696c0`) | BLOCKED/ACQUIRED | BLOCKED/ACQUIRED | BLOCKED/ACQUIRED | BLOCKED/ACQUIRED | **PASS** |

The pre-fix row reproduces the issue's reported failure exactly. On the same VM the
real Go tests also pass:

```
--- PASS: TestNLMAxisInterop (101.74s)   # the test that caught the regression
--- PASS: TestNLMSystemRpcbindRegistration (1.27s)
--- PASS: TestCrossProtocolLocking (6.77s)
--- PASS: TestCrossProtocolLockingByteRange (2.35s)
```

Closes #2754


marmos91:
Addressed both Copilot findings in `65059f04e`:

**1. Stale preflight / race (real, now fixed).** Confirmed with a probe: a delegation granted between the recall decision and the conflict judgment denied the lock with zero recalls dispatched — the original defect through a narrower window. `Lock`/`TestLock` now judge-and-insert in one critical section and re-judge after each recall, with a deterministic regression test (a second delegation granted from inside the first recall's callback) that fails on a single-pass acquire.

**2. Recall wait must honour timeout/cancellation (partly real, now fixed).** The total wait is now bounded once by a single deadline threaded through the re-judge loop, instead of 10s per iteration. On cancellation: `Manager.Lock` and the `LockManager` interface take no `context.Context`, so there is nothing to observe here — threading a ctx through the interface and every caller is a larger change than this fix, and the bounded wait already guarantees a denial rather than an unbounded hold. Happy to open a follow-up for ctx-aware lock acquisition if that is wanted.

marmos91:
@github-copilot please re-review

marmos91:
Recorded the recall wait's ceiling as a `decision:` marker at the code site (`c48eb1d65`), per the repo convention for a gate that is deliberately narrower than its name: the wait ignores the caller's lock-wait policy because denying without the recall is the defect being fixed, it is bounded by one deadline, and `FileLock` carries no fail-immediately bit and `LockManager.Lock` no context — honouring the policy means threading one through the interface and every caller. The marker names that as the way to overturn it.

marmos91:
@github-copilot please re-review the current head (c48eb1d65). Both findings from your previous review are addressed: the race is closed by the re-judge loop with a deterministic regression test, and the total wait is bounded once rather than once per recall.

marmos91:
Both findings from the second review are fixed in `700e2615`, and the fix changed the lock path enough that I re-ran the real-hardware verification on it.

**1. Scan stopped at the first blocker (fixed).** Confirmed as a measurable defect: a read delegation coexists with a shared NLM lock, so an exclusive request meeting the delegation first recalled and waited out the full 2s timeout only to be denied by the NLM lock behind it. All three scanners now go through one helper that lets a terminal conflict win and reports `delegationBlocked` only when every blocker is a delegation. `TestLock_DelegationAheadOfTerminalConflictDoesNotWait` fails at **2.002s** on the old scan and passes at **0.20s** on the new one.

**2. `CheckForIO` did a single recall and re-judge (fixed).** Correct — a delegation granted between the recall and the check (by the recall callback or a concurrent OPEN) was returned as a conflict without being recalled. It now loops under the same shared deadline, as the acquire paths do. While fixing this I also moved the acquire paths to judge first and recall only when a delegation is what stands in the way, so a definite conflict no longer pays for a recall it does not need.

**3. Fail-immediately / cancellation (deliberate, documented).** Kept as-is and recorded as a `decision:` marker at `Manager.Lock` with the ceiling and the way to overturn it (thread a context through `LockManager` and every caller). Denying without the recall is the defect this path exists to fix, the wait is bounded, and the only client that waits is one that asked the server to reclaim caching rights and never answered.

**Re-verified on hardware after the change** (fresh VM, two builds on one box, now torn down):

| build | `SMB_vs_NLM` | result |
| --- | --- | --- |
| pre-fix `develop` 58ae43b3 | HOLDER_FAIL (EACCES) / NA | **FAIL** |
| post-fix `700e2615` | BLOCKED / ACQUIRED | **PASS** |

`TestNLMAxisInterop` 102.44s, `TestNLMSystemRpcbindRegistration` 1.30s, `TestCrossProtocolLocking` 7.21s, `TestCrossProtocolLockingByteRange` 2.28s — all pass.

marmos91:
@github-copilot please re-review the current head (`700e26150`).

marmos91:
Two of the three findings from the third review are fixed in `52c8b752`; the third is the deliberate behaviour already recorded at the code site. Re-verified on hardware again, since the lock path changed.

**1. Wait covered every breaking delegation, not the selected ones (fixed).** Confirmed as a measurable defect. Multiple read delegations may coexist, so an unrelated one already breaking — its client never answering — held the caller until the deadline *after* its own delegation had been returned, and the stale conflict was reported. `waitForDelegationRecall` now takes the same predicate the recall was dispatched with. `TestLock_WaitIsScopedToTheSelectedDelegations` fails at **3.002s** unscoped and passes immediately when scoped.

**2. No `CheckForIO` interleaving coverage (fixed).** Correct — the I/O path's loop is separate, and every existing `CheckForIO` test would have stayed green through a single-pass regression. `TestCheckForIO_DelegationArrivingDuringRecallIsRecalled` grants a second delegation from the first recall callback and asserts both are recalled before the I/O is allowed; collapsed to a single pass it fails reporting `deleg:nfs4:c2`.

**3. Fail-immediately (kept, and the evidence is why).** I checked this against the harness rather than reasoning about it: the NLM axis `hold` path takes **one** `LOCK_NB` lock and never retries (`test/e2e/testdata/nlm/nlmlock.py:70`). So a non-blocking lock that is denied is never retried, and not waiting for the recall on that path would put the `SMB_vs_NLM` direction straight back to `HOLDER_FAIL`. Denying without the recall *is* the bug. The wait is bounded, and the `decision:` marker at `Manager.Lock` records the ceiling and the way to overturn it (thread a context through `LockManager` and every caller).

**Re-verified on hardware after the change** (fresh VM, two builds on one box, now torn down):

| build | `SMB_vs_NLM` | result |
| --- | --- | --- |
| pre-fix `develop` 58ae43b3 | HOLDER_FAIL (EACCES) / NA | **FAIL** |
| post-fix `52c8b752` | BLOCKED / ACQUIRED | **PASS** |

`TestNLMAxisInterop` 104.29s, `TestNLMSystemRpcbindRegistration` 2.08s, `TestCrossProtocolLocking` 9.45s, `TestCrossProtocolLockingByteRange` 3.05s — all pass.

marmos91:
@github-copilot please re-review the current head (`52c8b7523`).

marmos91:
The amplification finding in the fourth review is real and is fixed in `0e4696c0`. Re-verified on hardware again, since the wait logic changed.

**Fresh 10s budget per `Lock` call (fixed).** You are right that this compounds: the SMB handler retries a blocking lock every `BlockingLockRetryInterval` while the delegation stays `Breaking`, and the fail-immediately attempt comes through the same call, so each retry started a new `delegationIOWaitTimeout` and one SMB request could accumulate waits far past its own 5s/35s deadline.

The budget now runs from when each recall was dispatched (`BreakStarted`) rather than from the call, so repeated attempts against the *same* outstanding recall share it. `waitForDelegationRecall` reports whether the recall completed, and a caller that timed out denies instead of retrying — so the loop still ends in a conflict rather than a spin, and the behaviour is unchanged when the delegation is returned promptly.

`TestLock_RetriesShareOneRecallBudget` runs ten attempts against a client that never answers: **4.01s** with a per-call budget, **0.40s** (one budget) anchored to the recall.

On the context half of the finding — `LockManager.Lock` takes no `context.Context`, so there is nothing here to observe cancellation with. The anchored budget is what bounds the wait, and the `decision:` marker at `Manager.Lock` records that ceiling and names threading a context through the interface as the way to overturn it.

**Re-verified on hardware after the change** (fresh VM, two builds on one box, now torn down):

| build | `SMB_vs_NLM` | result |
| --- | --- | --- |
| pre-fix `develop` 58ae43b3 | HOLDER_FAIL (EACCES) / NA | **FAIL** |
| post-fix `0e4696c0` | BLOCKED / ACQUIRED | **PASS** |

`TestNLMAxisInterop` 101.74s, `TestNLMSystemRpcbindRegistration` 1.27s, `TestCrossProtocolLocking` 6.77s, `TestCrossProtocolLockingByteRange` 2.35s — all pass.

marmos91:
@github-copilot please re-review the current head (`0e4696c0c`).

marmos91:
Two things here — one is the documented decision, the other is a call-graph error I want to correct with evidence.

**`Service.TestLock` / LOCKT blocking — the premise does not hold.** `Service.TestLock` has **no production caller**. The only references in the tree are its definition (`pkg/metadata/service.go:927`) and `pkg/metadata/service_test.go`. `LOCKT` never reaches it:

```
internal/adapter/nfs/v4/handlers/lock.go:402  ->  h.StateManager.TestLock(...)
internal/adapter/nfs/v4/state/lockowner.go:907 ->  lm.TestUnifiedLock(handleKey, testLock)
```

and `NLM TEST` goes `pkg/adapter/nfs/nlm.go:171 -> s.lockMgr.TestUnifiedLock(...)`. `TestUnifiedLock` lives in `pkg/metadata/lock/byterange.go`, is **not touched by this diff**, and contains **no recall or wait at all** (`grep -c recall` → 0). So neither `LOCKT` nor `NLM TEST` can block on a recall. `Manager.TestLock`'s only non-test caller is `Manager.TestLockByParams`, which is reached only from `Service.TestLock` — the dead one. The preview path is not on any live request path.

**Fail-immediately (kept, and the harness is the evidence).** I checked the client side rather than reasoning about it: the NLM axis `hold` path takes **one** `LOCK_NB` lock and never retries (`test/e2e/testdata/nlm/nlmlock.py:70`; `_cmd` always ORs `LOCK_NB`). A non-blocking lock that is denied is never retried, so *not* waiting for the recall on that path puts `SMB_vs_NLM` straight back to `HOLDER_FAIL` — the exact regression this PR fixes, verified on hardware five times. Denying without the recall is the bug. The wait is bounded to one budget per recall, and the `decision:` marker at `Manager.Lock` records the ceiling and names the change that would overturn it (thread the lock-wait policy through `LockManager` and every caller).

If you still see a live path from a request into a recall wait that I have missed, name the handler and the line and I will fix it — but I have traced every caller of `Lock`, `TestLock` and `CheckForIO` in the tree.

PR #2761, opened 2026-09-18
fix(nfs): close the v4.1 callback gaps a shared connection exposes

marmos91:
## What

Reworks the NFSv4.1 delegation/callback path for a shared connection: backchannel sender
selection, reply routing, retired reply tables, probe verdicts, and the revocation statuses
a revoked delegation's stateid returns.

29 commits, none of them present in `develop`. This is *not* the merged #2598 work — it is a
later, separate rework that never got a PR.

## Status

Conflicts with `develop` on rebase. Opening to preserve the work and make it reviewable;
needs a rebase before it can merge.


marmos91:
Closing: this is already on develop.

The PR body claims "29 commits, none of them present in develop. This is *not* the merged #2598 work." Both halves are wrong, and the branch name says so — this *is* the #2598 branch.

Evidence:

- Squashing the 29 commits and rebasing onto develop drops the result as **already applied**: `git rebase origin/develop` reports `skipped previously applied commit`, leaving the branch 0 commits ahead with an empty diff.
- The patch-id of the branch's cumulative diff (`3f9c3593..72ccc9cc6`) is identical to that of `b2b010bee` — `fix(nfs): grant and recall NFSv4.1 delegations over a shared connection (#2598)`, merged 2026-09-16.
- The tip commit's own text is live on develop: the `decision: the generation is sampled before the attempt` marker sits at `internal/adapter/nfs/v4/state/delegation.go:690`.

Forcing the replay with `--reapply-cherry-picks` conflicts, and develop is the **superset** on every hunk — `PendingCBReplies.Register(xid, SessionId4{})` vs the branch's `Register(xid)`, plus `CancelSession`, `queueWaitBound` and `EnsureBackchannelWriterForConn`, none of which the branch has. Merging this would be a regression, not a rework.

Copilot's four findings here are not disposed of by closing — I verified all four against develop and they are live there. Filed separately.

PR #2763, opened 2026-09-18
refactor(engine): require a remote block store

marmos91:
## What

Part of the Step 5 block store simplification
(`.planning/2026-09-14-step5-IMPLEMENTATION-PLAN.md`): remove the local/remote block store
distinction so a share has one journal and one named, remote, durable block store.

Branch tip: `step5/engine-localonly`.

## Status — draft, needs a rebase before it can merge

- Based on `chore/drop-local-blockstore-config`, which no longer exists on any remote.
- Conflicts with `develop` on rebase; `develop` still carries the local/remote split, so
  none of this landed through another route.
- One of four parallel slices of the same plan (`e2e-helpers`, `engine-localonly`,
  `smbtorture-regression`, `task9-harness`); they are independent, not a stack.

Opened to preserve the work and make it reviewable. Not ready to merge.


marmos91:
## Closing — the trunk landed via #2579, and the one unlanded item should not ship

Cut 2026-09-14; #2579 `refactor(shares): one block store, one journal, no kind` merged the same work the next day, with #2607, #2609 and #2613 following. The description above says *"develop still carries the local/remote split, so none of this landed through another route"* — that is not the case. Develop has the `Kind` drop, the single `--block-store` CLI, `JournalRoot`, `CommitAck`/`RelaxedMetadataCommit`/`JournalSize`, and `filepath.Rel` containment.

Unlike the five sibling drafts, this branch did have one genuinely unlanded piece: spec item 15 / D20, *"remove the ~10 dead `HasRemote` branches"*. Having read it, it should not ship, and the spec has been corrected to mark item 15 abandoned rather than open.

**The branches are not dead.** A share is refused unless it names a block store, which is what makes them look unreachable — but `engine.BlockStoreConfig.Remote` is nillable at the package boundary (`engine.go:27` documents nil as local-only mode) and fourteen construction sites build a store that way, including the eviction, warm-read-integrity, cold-read and durability fixtures.

**They are not a fast path either.** Each decides what a read returns when there is nowhere to hydrate from:

- `read_internal.go:56` — `st.Cold || (st.Hole && bs.HasRemoteStore())`: a hole reconciles only against a remote, and otherwise must read as the zeros it is.
- `read_internal.go:79` — `healCorruptWarmRead` fails closed with `ErrIntegrityCheckFailed` when there is no remote, "rather than returning corrupt or zero-filled bytes".
- `flush.go:627` — `SeedColdRefs` skips, because a cold range with no remote "fails its reads closed, where the same range left as a hole reads as the zeros it is".

Collapsing these to the true arm turns each into a silent wrong answer — the same failure shape as the cold-interval and pre-journal-layout defects. `TestWarmReadSelfHeal_LocalOnlyFailsClosed` already pins one of them, so the removal would have gone red in CI.

Spec §8 understated the cost as the evict guard alone losing its reachable condition; the warm-read heal and cold-range seeding guards lose theirs too.

Reviving item 15 means first making `Remote` non-nillable — a constructor that refuses it — and retargeting all fourteen fixtures, so the compiler rather than a grep shows the branches are unreachable. That is a real change with a real benefit to argue, not the cosmetic cleanup the item describes.

Follow-up PR records the ceiling as a `decision:` marker on `engine.Store.HasRemoteStore` and corrects the spec's work-breakdown table (items 1-14 and 16-18 were still marked open).

PR #2765, opened 2026-09-18
test: one block store across the harness and workflows

marmos91:
## What

Part of the Step 5 block store simplification
(`.planning/2026-09-14-step5-IMPLEMENTATION-PLAN.md`): remove the local/remote block store
distinction so a share has one journal and one named, remote, durable block store.

Branch tip: `step5/task9-harness`.

## Status — draft, needs a rebase before it can merge

- Based on `chore/drop-local-blockstore-config`, which no longer exists on any remote.
- Conflicts with `develop` on rebase; `develop` still carries the local/remote split, so
  none of this landed through another route.
- One of four parallel slices of the same plan (`e2e-helpers`, `engine-localonly`,
  `smbtorture-regression`, `task9-harness`); they are independent, not a stack.

Opened to preserve the work and make it reviewable. Not ready to merge.


marmos91:
`step5/task9-harness` is a strict subset of #2762 — its single unique commit has an equivalent there, and is on develop.

## Superseded — the same work landed via #2579

This branch was cut 2026-09-14. The following day #2579 `refactor(shares): one block store, one journal, no kind` merged the same body of work, followed by #2609 (journal-root containment by relative path), #2607 (docs) and #2613 (operator journal path).

The PR description above claims *"develop still carries the local/remote split, so none of this landed through another route."* That premise is false. On develop today:

- the block-store `Kind` discriminator is gone
- `dfsctl store block local` no longer exists; `share create/edit` take `--block-store`
- `JournalRoot` is wired through `start.go` and the adapter test runtimes
- `CommitAck`, `CommitAckJournal`, `RelaxedMetadataCommit` and `JournalSize` are on the share model
- journal-root containment is decided by `filepath.Rel` (`pkg/controlplane/runtime/shares/journal_open.go:62`) — a better version than this branch's prefix check

Verified per-commit rather than per-PR: for every commit unique to this branch tip, distinctive added lines were matched against develop, and deletion-only commits were checked in the opposite direction (the deleted lines being absent). Two further signals confirm staleness: the least-advanced draft in this set differs from develop in more files than the most-advanced one (they are catching up, not ahead), and the files that look branch-only are ones develop deleted later — `internal/logger/context.go` went in #2755.

Closing rather than rebasing: a rebase would resolve ~100 commits of drift to re-apply code already on develop. Nothing is lost. The one genuinely unlanded piece of the step-5 plan is item 15 / D20 (removing the dead `HasRemote` branches), which is being taken forward on a fresh branch off current develop.

